PR #10, opened 2026-04-23
fix(#1152): Array.prototype higher-order methods via __get_builtin path (string receivers)

ttraenkler:
## Summary

- After PR #195 (`__get_builtin` + `__extern_get` for `BuiltinCtor.prototype`), `Array.prototype.map.call(\"abc\", cb)` and siblings fell through to the `__proto_method_call` host bridge, which then handed the Wasm closure callback across the boundary as an uninvokable externref. V8's native Array.prototype.map threw `TypeError: object is not a function` for every such test.
- Fix keeps string primitives in the in-Wasm array-like loop (they compile to externref, so `__extern_length` + `__extern_get_idx` iterate the code units correctly). Only boolean/numeric primitive literals still bail to the legacy bridge — those compile to i32/f64 and `extern.convert_any` can't lift them.
- `__extern_has_idx` now short-circuits for primitive-string receivers (`idx in \"abc\"` throws TypeError).

## Test plan

- [x] `tests/issue-1152.test.ts` — 4/4 pass (`map`, `forEach`, `every`, `some` with StringLiteral + NoSubstitutionTemplateLiteral receivers).
- [x] `tests/equivalence/array-prototype-methods.test.ts` — 13/13 pass (no regression).
- [x] `tests/equivalence/string-methods.test.ts` — 39/42 pass (3 failures pre-existing on main, unrelated to this change).
- [ ] Full CI test262 run — expect the \"object is not a function\" Array-prototype cluster to drop substantially.

## Out of scope for this PR

Remaining failures in the same cluster:
- Boolean/Number primitive-literal receivers (`Array.prototype.map.call(false, cb)`, `.call(2.5, cb)`) — need primitive → wrapper boxing.
- `flatMap` species tests (`this-value-ctor-object-species*.js`) — need Array.@@species lookup.

### DIFF
--- a/src/codegen/array-methods.ts
+++ b/src/codegen/array-methods.ts
@@ -354,15 +354,23 @@ export function compileArrayLikePrototypeCall(
     (ts.isIdentifier(receiverArg) && receiverArg.text === "undefined");
   if (isNullReceiver) return undefined;
 
-  // Bail out on primitive literal receivers (boolean, number, string). Our `extern.convert_any`
-  // coercion only works on ref/anyref values; a primitive compiled to i32/f64 would produce
-  // invalid Wasm. The legacy __proto_method_call path handles ToObject(primitive) correctly.
+  // Bail out on boolean / numeric primitive literal receivers. Those compile
+  // to i32/f64, which `extern.convert_any` can't lift — routing them through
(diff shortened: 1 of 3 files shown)


ttraenkler:
Superseded by PR #247 which already landed this fix on main. #1152 is closed.

PR #73, opened 2026-04-27
feat(ir): slice 7b — non-numeric yields, bare yield, yield* delegation (#1169f)

ttraenkler:
## Summary

Extends slice 7a's generator IR (PR #70) so the IR path covers all yield variants except those gated by spec-correct `.next(arg)` plumbing (still deferred to a future "real coroutine" workstream).

- **New IR node `gen.yieldStar`** — represents `yield* <iterable>` delegation; lowered to `__gen_yield_star(buffer, externref_iterable)`. Operand is pre-coerced externref. Side-effecting (DCE pinned).
- **`gen.push` dispatch widening** — switches on the operand IrType: f64 → `__gen_push_f64`, i32 → `__gen_push_i32` (booleans), everything else → `__gen_push_ref` (caller pre-coerces via `coerce.to_externref`).
- **Null-externref const lowering** — extends `case "null"` to emit `ref.null.extern` for externref result types. Used by bare `yield;` lowering.
- **`lowerYield` widening** — three new arms: `yield*`, bare `yield;`, and `yield <any-Phase-1-type>`. New helper `coerceYieldValueToExternref` skips the coerce when the value's underlying Wasm valtype is already externref (`IrType.val.externref` or `IrType.string` in host-strings mode), avoiding a Wasm validation error from `extern.convert_any` on non-anyref-subtype operands.
- **Selector widening** — accepts bare `yield;`, `yield* <expr>`, and `yield <any-Phase-1-expr>`. Threads `isGenerator` through `isPhase1StatementList` / `isPhase1Tail` so bare `return;` is allowed in generator tails (non-generator bare returns continue to be rejected).

## Test plan

- [x] `npm test -- tests/issue-1169f-7b.test.ts` — **10/10 pass** (5 dual-run cases × 2: selector + equivalence)
- [x] `npm test -- tests/equivalence/` — **0 regressions vs origin/main** (105/1190/1295 identical baseline)
- [x] `npm test -- tests/issue-1169{a,b,c,d,e-bridge}.test.ts tests/issue-1182.test.ts` — **182/182 pass** (all prior IR slices unaffected)
- [ ] CI on the PR — waiting

## Slice 7b scope

**In:**
- `function* g() { yield <any Phase-1>; }` — boolean / string / numeric / object yields
- `function* g() { yield; }` — bare yield (no value)
- `function* g() { yield* <iterable>; }` — delegation to another iterable
- Generator tails with bare `return;` (no expression)
- Mixed-type yield sequences in the same generator

**Out (defers to follow-up slices):**
- try/catch wrapping for deferred-throw semantics (#928) — slice 7-throw
- `.next(arg)` argument passthrough — needs real coroutine transform
- async generators (`async function*`) — separate slice

## Files

- `src/ir/nodes.ts` — `IrInstrGenYieldStar` interface + IrInstr union arm
- `src/ir/builder.ts` — `emitGenYieldStar` method
- `src/ir/from-ast.ts` — widened `lowerYield`, generator-aware `lowerTail` for any return type, `coerceYieldValueToExternref` helper
- `src/ir/lower.ts` — widened `gen.push` dispatch, `gen.yieldStar` emit case, externref null-const lowering, use-collection arm
- `src/ir/select.ts` — accept bare/star/non-numeric yields; thread `isGenerator` through tail recursion
- `src/ir/verify.ts` — verifier use-collection arm
- `src/ir/passes/dead-code.ts` — DCE pin + use-collection arm
- `src/ir/passes/inline-small.ts` — operand-rewrite arm
- `tests/issue-1169f-7b.test.ts` — 5 dual-run cases

### DIFF
--- a/src/ir/builder.ts
+++ b/src/ir/builder.ts
@@ -605,6 +605,20 @@ export class IrFunctionBuilder {
     return result;
   }
 
+  /**
+   * Slice 7b (#1169f): emit a `gen.yieldStar` instr — drain the inner
+   * iterable into the outer generator's buffer via
+   * `__gen_yield_star(buf, inner)`. The caller MUST coerce `inner` to
+   * externref upstream (e.g. via `emitCoerceToExternref`) — the host
+   * import expects an externref in arg position 1.
+   */
+  emitGenYieldStar(inner: IrValueId): void {
+    if (this.funcKind !== "generator") {
+      throw new Error(`IrFunctionBuilder: emitGenYieldStar requires funcKind=generator (${this.name})`);
+    }
+    this.pushInstr({ kind: "gen.yieldStar", inner, result: null, resultType: null });
+  }
+
   private requireBlock(): OpenBlock {
     if (this.current === null) {
       throw new Error(`IrFunctionBuilder: no open block (func ${this.name})`);
--- a/src/ir/from-ast.ts
+++ b/src/ir/from-ast.ts
@@ -314,32 +314,34 @@ function lowerStatementList(stmts: readonly ts.Statement[], cx: LowerCtx): void 
  */
 function lowerTail(stmt: ts.Statement, cx: LowerCtx): void {
   if (ts.isReturnStatement(stmt)) {
-    // Slice 7a (#1169f): generator return. Match the legacy semantics
+    // Slice 7a/7b (#1169f): generator return. Match the legacy semantics
     // (`compileReturnStatement` in `codegen/statements/control-flow.ts`
     // line 89-123): a `return <value>` inside a `function*` pushes
     // `<value>` onto the eager buffer as a final yielded value, then
     // wraps the buffer with `__create_generator` to produce the
     // externref Generator object. This is non-spec — JS spec says the
     // return value lands in `IteratorResult.value` with `done:true` —
-    // but matching legacy is the correctness target for slice 7a so
-    // existing test262 coverage doesn't drift.
+    // but matching legacy is the correctness target so existing
+    // test262 coverage doesn't drift.
     //
-    // Slice 7a only supports numeric (f64) return values — same scope
-    // as `lowerYield`. Non-numeric returns throw and the function
-    // falls back to legacy.
+    // Slice 7b widens the return type: we accept any Phase-1 expression
+    // and route it through the same `lowerYield`-style dispatch
+    // (f64/i32 stay native; ref/string/object/class coerce to
+    // externref → __gen_push_ref). Same dispatch logic as `lowerYield`
+    // except we get a `ts.Expression` already, not a YieldExpression.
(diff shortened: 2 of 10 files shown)


ttraenkler:
## Regression analysis (post-baseline-refresh)

CI feed shows 145 regressions on `5e8ff0a8` against the refreshed baseline. Detailed breakdown by direct comparison of CI artifacts (this run + the immediately-prior scaffolding-only revert run `070d7663` against the same fresh baseline):

| Count | Bucket | Verdict |
|-------|--------|---------|
| 127 | `compile_timeout` | 7b run timeouts ∩ revert-run timeouts = **20**. 7b has 125 unique timeouts; revert has 108 unique timeouts. Each CI run picks a different ~120-tests-each subset to time out. **Non-deterministic w.r.t. 7b code.** |
| 12 | `fail` / `compile_error` | Also failing in the scaffolding-only revert run (which is an effective no-op vs main). Pre-existing drift. |
| 6 | `fail` | Only in the 7b widening. **No test source contains `function*` or `yield`.** Scattered across Promise (×4), TypedArray (×1), block-scope/for-in (×1). |

### The 6 7b-specific failures
```
Promise/resolve/resolve-non-thenable.js
Promise/prototype/then/rxn-handler-fulfilled-invoke-strict.js
Promise/prototype/finally/resolved-observable-then-calls-argument.js
Promise/race/resolve-ignores-late-rejection-deferred.js
TypedArray/prototype/length/invoked-as-accessor.js
language/block-scope/syntax/for-in/acquire-properties-from-array.js
```

`grep -E 'function \*|yield' <these_files>` → zero matches. My selector widening only affects generator/yield shape acceptance; for non-generator code paths it's byte-identical to slice 7a (the new `isGenerator` parameter defaults to `false` and threads only through generator-tail recursion).

The `block-scope/for-in/acquire-properties-from-array.js` example: pure `function props(x) { for (let p in x) array.push(p); }`. Compiled in 65ms / passed on revert; compiles in 109ms / fails on widening. **Same source code path through the compiler in both runs** — slice 7b doesn't touch non-generator function compilation. The 65→109ms slowdown + status flip is consistent with shared-fork accumulated state.

### Locally non-reproducible
- 200-test bulk run via `IncrementalLanguageService` (mimicking CI's worker pattern): 0 timeouts, 0 errors
- 12 specific failing tests run individually with full CI wrapper: all <500ms compile, all instantiate + execute correctly

### Assessment
None of the failing tests touch the slice 7b surface (yields, generators). The 127 compile_timeouts vary across runs (no overlap signal). The 6 non-timeout 7b-specific failures are scattered, locally-irreproducible, and don't contain code my selector can claim.

Documenting per protocol — escalating merge decision to tech lead.

cc @loopdive — see `/dev-self-merge 73` algorithmic gate (failed criterion 2: 10.76% > 10%).

PR #74, opened 2026-04-27
fix(legacy): re-resolve __str_charAt by name in compileForOfString (#1186)

ttraenkler:
## Summary

Implements **#1186** — fixes a pre-existing legacy bug where `compile(source, { nativeStrings: true })` produced invalid Wasm for `for (const c of <string>)` patterns.

**Root cause**: `compileForOfString` captured `__str_charAt`'s funcIdx from `ctx.nativeStrHelpers` at registration time, but late-import additions later in the pipeline shift the function index space. `shiftLateImportIndices` walks `ctx.mod.functions[].body` and `ctx.funcMap` but does NOT update `ctx.nativeStrHelpers`. The captured index goes stale and at runtime resolves to whatever import landed at that position (typically `__is_truthy`), causing:

```
WebAssembly.CompileError: call[0] expected externref, found local.get of type i32
```

**Fix**: replace the captured index with a name walk against `ctx.mod.functions[i].name` at the call site. Mirrors the IR resolver pattern from #1183 (`src/ir/integration.ts:resolveFunc`'s `ctx.mod.functions` fallback). Single-site change, ~25-line comment block explaining why.

## Test plan

- [x] New `tests/issue-1186.test.ts` — 11 cases pass:
  - 5 legacy+nativeStrings runtime checks (count chars, empty string, single-char, `c.length` in body, BMP unicode)
  - 5 legacy↔IR equivalence dual-runs in nativeStrings mode (re-enables what #1183 had to skip)
  - 1 host-strings sanity check
- [x] Prior IR tests (128/128: #1169d / #1169e / #1182 / #1183 / `tests/ir/`) — pass.
- [x] `npx tsc --noEmit` — 0 errors.
- [x] Local equivalence suite (`tests/equivalence/**`) — exit 0.
- [ ] CI (GitHub Actions) — wait for `.claude/ci-status/pr-<N>.json`, self-merge per dev protocol.

## Out of scope (follow-ups)

- Full audit of other `ctx.nativeStrHelpers.get(...)` consumers for similar staleness — only `compileForOfString` is known to capture-and-reuse across late shifts.
- Extending `shiftLateImportIndices` to rewrite `ctx.nativeStrHelpers` entries (option B from spec) — kills the bug class once but is more invasive. Defer until another consumer surfaces.
- Re-enabling the native-mode dual-run cases in `tests/issue-1183.test.ts` that currently assert against JS-computed expected values. The new tests in this PR effectively cover that gap, so the cleanup is optional.

### DIFF
--- a/plan/issues/ready/1186.md
+++ b/plan/issues/ready/1186.md
@@ -139,3 +139,53 @@ inspect each call site for the capture-vs-walk pattern.
 - A full audit of every `ctx.nativeStrHelpers.get(...)` call site for
   similar bugs — spot-fix the known sites, file follow-ups for any
   others surfaced.
+
+---
+
+## Implementation Notes (senior-developer, 2026-04-27)
+
+### Fix applied
+
+Replaced the funcIdx capture in `compileForOfString`
+(`src/codegen/statements/loops.ts:~1487`) with a name walk against
+`ctx.mod.functions[i].name`. Mirrors the IR resolver pattern from
+#1183 (`src/ir/integration.ts:resolveFunc`'s `ctx.mod.functions`
+fallback). The resolved index is `ctx.numImportFuncs + i` — the
+absolute funcIdx in the post-shift index space.
+
+The change is local to one site (option A from the spec). Other
+`ctx.nativeStrHelpers.get(...)` consumers in the legacy codebase
+were left as-is; if they exhibit the same staleness pattern each
+should be filed as its own issue with a reproducer.
+
+### Verification
+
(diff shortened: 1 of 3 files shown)


ttraenkler:
FYI: PR #75 (the parallel branch tracking this issue) just merged as well — it ended up landing six array-perf sprint follow-up issues (#1195–#1200) on top of the equivalent #1186 fix that this PR already shipped. The fix on main from this PR (commit d21b81962) is the canonical one; #75 took its tests/loops.ts via 'git checkout --theirs' rather than fight over duplicate work.

PR #116, opened 2026-04-30
ci(#1217): test262 smoke-canary — fresh-cache double-run flip detector

ttraenkler:
## Summary

New nightly + workflow_dispatch workflow that runs `Test262 Sharded` twice on `main` HEAD with **no `actions/cache` step** (fresh build per run) and reports flip count: the number of tests whose status differs between runs A and B. Purely **observational** — never blocks PRs, never touches the baseline.

Closes the observability gap from #1190 acceptance criterion #5: detect future drift regressions before they start producing false positives at the merge-gate.

## Implementation

- **`.github/workflows/test262-canary.yml`** — two parallel matrix jobs (`test262-shard-run-a` / `test262-shard-run-b`) emit per-shard JSONLs; a third job (`flip-count`) merges and diffs. Mirrors the `test262-sharded.yml` matrix shape but without any cache step (the whole point — without fresh cache the two runs would replay each other).
- **`scripts/test262-canary-flip-count.ts`** — symmetric JSONL diff that bucketises flips into `flips_real` (excluding `compile_timeout` transitions, which are timing noise) and `flips_compile_timeout`. Emits a JSON summary on stdout's last line for downstream parsing. Exits 0 on `flips ≤ threshold`, 1 otherwise, 2 on IO errors.

### Threshold rationale

Initial value: **10** (documented inline). Captures meaningful determinism regressions (e.g. a config change that doubles flap count) without false-firing on the routine 5–8 timeout-flap baseline (PR #104 saw 48 transitions, mostly `compile_timeout`). Override at dispatch time via `flip_threshold` input. Tune up/down after the first few canary runs land their numbers in the workflow summary.

### Triggers

- `cron: 0 4 * * *` UTC (outside peak runner-pool load)
- `workflow_dispatch` with overridable `flip_threshold` and `include_proposals`

## Acceptance criteria

| # | Criterion | Status |
|---|-----------|--------|
| 1 | New workflow runs on schedule + workflow_dispatch | ✅ |
| 2 | Workflow runs main HEAD twice with no `actions/cache` step | ✅ |
| 3 | Workflow output reports flip count distinctly from regression count | ✅ JSON summary fields: `flips`, `flips_real`, `flips_compile_timeout` |
| 4 | Workflow fails with `::error::` if flip count > threshold | ✅ |
| 5 | Threshold value + rationale documented inline | ✅ |
| 6 | Initial flip baseline measured | ⏳ pending first scheduled run |

## Out of scope (per issue notes)

- Statistical merge-gate metric (#1190 Q2 — separate concern, deferred).
- Per-test flake quarantine list (would solve the symptom but not the observability problem).
- Cross-Node-version flap detection (separate concern).

## Test plan

- [x] Local smoke-test of `scripts/test262-canary-flip-count.ts` with synthetic JSONLs:
      - threshold=10, 3 flips → exit 0, PASS reported.
      - threshold=2, 3 flips → exit 1, FAIL reported.
      - JSON summary on stdout's last line for both.
      - Compile_timeout transitions correctly classified out of `flips_real`.
- [ ] First nightly run will land flip count baseline in the issue's comments.

### DIFF
--- a/.github/workflows/test262-canary.yml
+++ b/.github/workflows/test262-canary.yml
@@ -1,0 +1,279 @@
+name: Test262 Canary (engine determinism)
+
+# Smoke-canary for test262 engine determinism (#1217).
+#
+# Runs `Test262 Sharded` twice on `main` HEAD with **fresh cache** (no
+# restore-keys, no incremental cache hits) and reports the flip count —
+# the number of tests whose status differs between the two runs.
+#
+# A non-zero flip count means the test262 result is non-deterministic:
+# could be timing-sensitive `compile_timeout`s under runner load,
+# Promise micro-task ordering nondeterminism, or something more
+# substantive. The threshold is documented inline; tune it from the
+# initial baseline run before flipping the workflow into a hard gate.
+#
+# This workflow is purely OBSERVATIONAL — it never blocks PRs, never
+# touches the baseline, never refreshes any reports. It is the cheapest
+# mechanism to detect future drift regressions; downstream concerns
+# (flake quarantine, runner-pool isolation) are out of scope here.
+#
+# Triggered manually via `workflow_dispatch` and on a nightly schedule.
+# Acceptance criterion checklist for the issue lives in
+# plan/issues/sprints/46/1217.md.
+
+on:
+  schedule:
+    # Nightly at 04:00 UTC — outside peak runner-pool load so the flip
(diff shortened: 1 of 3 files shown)


ttraenkler:
Duplicate of #120 (same issue #1217, #120 is newer). Closing this older PR.

PR #155, opened 2026-05-02
fix(#1245): type-guarded Stage 1 of #1177 — recover Stage 1 wins without surfaced-bug regressions

ttraenkler:
## Summary
Investigates and re-lands #1177 Stage 1 with a **type-guard** that prevents the 81 real regressions from PR#125 while still recovering the targeted TDZ-through-closure fixes.

## Investigation findings

**Finding 1 — the 59 compile_timeouts in PR#125 are CI flakiness, NOT compiler hangs.**
I cherry-picked PR#125's exact Stage 1 commit (\`8b34d909\`) onto a fresh branch and compiled all 66 \`pass→compile_timeout\` tests locally through the same \`wrapTest\` + \`compileMulti\` path the test262 runner uses. **Every test compiled in ≤500 ms** (median ≈ 65 ms). The CI's 16-shard parallelism × 30 s precompile timeout produces a stable flap floor — the absolute \`compile_timeout\` count actually *decreased* on PR#125 (156 → 129, net −27). The "59 compile_timeout regressions" header is just diff direction.

**Finding 2 — the 81 real regressions are pre-existing spec-bugs unmasked by Stage 1:**
- ~50× async-gen yield-star tests — sync-fallback iter capture has an unboxed-ref-cell leak.
- ~20× for-await-of dstr-assign tests — destructure-assign write-back path is not box-aware (architect already noted at \`nested-declarations.ts:240–260\` as a deferred follow-up).
- ~10× destructure-of-null tests — pre-existing TypeError-emission bug.
- 1× function-name-binding TDZ.

The unguarded \`fctx.localMap.get(cap.name) ?? cap.outerLocalIdx\` trips when localMap was re-aimed at a *differently-typed* slot (a boxed ref cell), causing the call-site / closure-emit to wrap or read it as the original value type → illegal-cast / null-deref.

## Approach in this PR
Refine the substitution to a **type-matched preference**: use \`localMap.get(cap.name)\` only when its type matches \`cap.valType\`. Otherwise fall back to legacy \`cap.outerLocalIdx\` to preserve main's behavior on the surfaced-bug clusters.

Sites updated (4):
- \`src/codegen/expressions/calls.ts\` — mutable + non-mutable cap-prepend branches
- \`src/codegen/closures.ts\` — both branches in \`emitFuncRefAsClosure\`

## Local validation
- \`tests/issue-1245.test.ts\` — 5/5 pass (new)
- \`tests/issue-1177.test.ts\` — 7/7 pass (Stage 1 targeted equiv)
- \`tests/issue-1016.test.ts\` — 4/4 pass (param-destructure closures)
- \`tests/issue-1128-dstr-tdz.test.ts\` — 8/8 pass
- \`tests/issue-723-tdz.test.ts\` — 7/7 pass

## Follow-up issues if CI net < +90
- Make \`compileForOfAssignDestructuringExternref\` box-aware (~30-test cluster)
- Fix async-gen yield-star sync-fallback unboxed-ref-cell leak
- Emit spec-mandated TypeError for destructure of null/undefined

## Test plan
- [x] Local TDZ + closure equivalence tests (5 files, 31 tests) all pass
- [ ] CI sharded test262 net ≥ +90 with no bucket > 50


ttraenkler:
Closing per #1245 investigation: CI confirms 81 real regressions are pre-existing spec-bugs unmasked by Stage 1, not addressable by type-guard refinement. See plan/issues/sprints/47/1245.md for the 252-line investigation writeup. Stage 1 will be re-attempted after follow-up issues land (destructure-assign box-awareness, async-gen yield-star sync-fallback, destructure-of-null TypeError).

PR #191, opened 2026-05-03
fix(#1252): correct f64.copysign operand order in SameValue check

ttraenkler:
## Summary

The SameValue f64 formula in `src/codegen/object-ops.ts` (added under #1127) had the `f64.copysign` operands pushed in **reversed** order. Wasm `f64.copysign(z1, z2)` returns z1 with the sign of z2; the codegen pushed `value` then `1`, producing `copysign(value, 1) = |value|` (always positive). So `SameValue(+0, -0)` silently returned `true` and the `Object.defineProperty` frozen-object guarantee was broken for ±0.

## Three changes

1. **Extract a single `emitSameValueF64(fctx, oldValLocal, newValLocal)` helper** that pushes the magnitude (`1`) first and the sign source (`value`) second. Documented why on the helper itself.

2. **Wire it into all three call sites** in `src/codegen/object-ops.ts`:
   - `compileObjectDefineProperty` (defineProperty path).
   - `compileObjectDefineProperties` guarded path.
   - `compileObjectDefineProperties` non-guarded path.
   - The two `defineProperties` sites previously used plain `f64.ne` — no SameValue check at all — so they get upgraded as a side effect.

3. **Regression test** `tests/issue-1252.test.ts` — 6 cases:
   - `SameValue(NaN, NaN) = true` → no throw.
   - `SameValue(+0, -0) = false` → throws (the headline bug).
   - `SameValue(-0, +0) = false` → throws (symmetric).
   - Same value → no throw (sanity).
   - Different value → throws (sanity).
   - Non-writable non-configurable redefine — exercises the same SameValue path without `Object.freeze`.

## Follow-up filed in issue notes

`Object.defineProperties` on a frozen object skips the SameValue check entirely — the `needsValueCompare` flag in that path doesn't consult `frozenVars`. That's a distinct bug from #1252 (it's about *whether* the check runs, not *how* the f64 comparison is performed). The `emitSameValueF64` helper this fix introduces is wired into the right call sites, so when the frozen-vars gate is added later, the ±0/NaN semantics will already be correct.

## Test plan

- [x] `tests/issue-1252.test.ts` — 6/6 pass.
- [x] Equivalence smoke (5 files / 34 tests) — no regression.
- [x] `tests/issue-1247.test.ts` (related, recently merged) — 5/5 still pass.
- [ ] CI test262 conformance.


ttraenkler:
Closing as duplicate of #190 (per team-lead) — same fix (swap f64.copysign operand order so SameValue +0/-0 distinction works). #190 was filed by dev-1283 in parallel; both branches landed the same operand-order swap. Picking #190 to merge.

PR #217, opened 2026-05-03
fix(#1300): inline arrow callee=function-typed param uses closure path

ttraenkler:
## Summary

Fix a soundness null-deref in the `isHostCallbackArgument` predicate
that decides whether an inline arrow argument is compiled as a wasm
GC closure struct or as a host callback (`__make_callback`).

When the callee is a function-typed **parameter or local** (e.g.
`compose1(a)` body does `a(() => \"end\")` where `a: Mw`),
`funcMap.get(\"a\")` returns undefined (it's a runtime closure value, not
a top-level function declaration), so the predicate fell through to
`return true`. The arrow was wrapped via `__make_callback` (host
externref), but the receiving Mw's call-site emits the wasm
closure-struct dispatch (`ref.test → ref.cast → struct.get → call_ref`)
which failed `ref.test` and then null-deref'd on the subsequent
`struct.get`.

**Fix**: extend `isHostCallbackArgument` to consult the TS symbol of
the callee identifier. If any declaration is a `ParameterDeclaration`,
`VariableDeclaration`, `FunctionDeclaration`, `FunctionExpression`, or
`ArrowFunction`, the callee is a user-side callable holding a wasm
closure value — return false so the arrow becomes a wasm GC struct
that the receiver can unwrap.

## Acceptance criteria

1. ✅ The 2-level repro returns `\"<a><b>end</b></a>\"`.
2. (deferred to #1297) Tier 5 hono compose can collapse its
   module-level step workarounds — that cleanup belongs in #1297.

## Test plan

- [x] `npm test -- tests/issue-1300.test.ts --run` → 8/8 pass
- [x] `npm test -- tests/equivalence/{fn-variable-call,array-callback-three-params,illegal-cast-assert-throws}.test.ts --run` → 14/14 pass
- [x] `npx tsc --noEmit -p tsconfig.json` → clean
- [x] Pre-existing `LinkError` failures in `tests/illegal-cast-closures-585` and tests importing the missing `tests/helpers.js` confirmed identical on main HEAD before the patch (environment issues, not regressions)


ttraenkler:
Duplicate of PR #215 (dev-1293). Standing down per tech lead.

PR #222, opened 2026-05-05
Audit diverged branch salvage candidates

ttraenkler:
This PR records the manual review of the useful parts from the remaining diverged branches. No file changes were ported because the valuable source changes are already present on main: scripts/generate-editions.ts already contains the edition/proposal bucket classification fix; regenerating public/benchmarks/results/test262-editions.json from current benchmarks/results/test262-current.jsonl produced no diff; scripts/serve-report.ts and playground/vite-plugin-test262.ts already expose /api/test262-source-status; and the old report timeline branch would replace the newer shared components/t262-charts.js timeline component with older inline report-page code, so it was intentionally not ported. The branch contains an empty audit commit so GitHub can show this review as a PR. Verification: git diff origin/main...HEAD was empty before the audit commit, and regenerated edition data from current main inputs produced no diff.


ttraenkler:
Closing after audit: the useful diverged-branch changes are already present on main, so this no-op review branch is being dropped.

PR #250, opened 2026-05-07
feat(#1323): $IteratorResult struct, eliminate __iterator_done/__iterator_value

ttraenkler:
## Summary
- Implements #1323 — replaces per-iteration-step `__iterator_done` and `__iterator_value` host calls with pure-Wasm `struct.get` on a canonical `$IteratorResult` WasmGC struct.
- The JS host's `__iterator_next` now calls a wasm-exported helper `__make_iterator_result(done, value)` to construct the struct after invoking `iter.next()`. The wasm side recovers the struct via `any.convert_extern + ref.cast` and reads `.done` / `.value` via `struct.get`.
- **Net win**: 3 host calls per iteration step → 1.

## Design

Per the discussion in the issue file, calling `.next()` on a JS iterator inherently requires going through the JS host. The reshape eliminates the **per-field** host calls (`__iterator_done`, `__iterator_value`), not the `next()` call itself. The host bridge constructs the struct once per step and the wasm side reads fields directly.

## Files changed

- `src/codegen/context/types.ts` — add `iteratorResultTypeIdx` field on `CodegenContext`
- `src/codegen/index.ts:addIteratorImports` — register `$IteratorResult` struct, drop `__iterator_done`/`__iterator_value` imports, add `__make_iterator_result` exported wasm helper
- `src/codegen/statements/loops.ts` — rewrite for-of legacy lowering to use `struct.get` instead of host imports
- `src/ir/lower.ts` — rewrite `iter.done`/`iter.value`/`forof.iter` IR lowerings to use `struct.get`
- `src/runtime.ts:__iterator_next` — build the struct via wasm export helper instead of returning the raw result object

## Test plan

- [x] `tests/equivalence/for-of-generator.test.ts` (5 tests) — all pass
- [x] `tests/equivalence/generator-nested.test.ts` (4 tests) — all pass
- [x] `tests/equivalence/symbol-iterator-class.test.ts` (3 tests) — all pass
- [ ] CI — full equivalence + test262 to verify no broader regressions

### DIFF
--- a/plan/issues/sprints/50/1323-iterator-protocol-pure-wasm.md
+++ b/plan/issues/sprints/50/1323-iterator-protocol-pure-wasm.md
@@ -1,12 +1,12 @@
 ---
 id: 1323
 sprint: 50
-title: "Iterator protocol bridging: implement $IteratorResult struct in pure Wasm, eliminate host bridge"
+title: "Iterator protocol: $IteratorResult WasmGC struct, reshape __iterator_next, eliminate per-field host calls"
 status: ready
 created: 2026-05-07
 updated: 2026-05-07
 priority: medium
-feasibility: easy
+feasibility: medium
 reasoning_effort: medium
 task_type: feature
 area: runtime, codegen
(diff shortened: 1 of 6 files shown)


ttraenkler:
Closing: the  struct approach (late user-function registration in addIteratorImports) corrupts shared codegen/TypeScript checker state across sequential compile() calls. Errors are non-deterministic — type.id.toString, identifiers.set, declaredType — all pointing to shared state mutation. The optimization is not a correctness requirement; main is clean with the original 3-host-call approach. Will reopen with a redesign that pre-registers the struct before any user functions (always, unconditionally).

PR #255, opened 2026-05-07
fix(#1311): typed-callable container dispatch (Map.set / arr.push of arrows)

ttraenkler:
## Summary

`isHostCallbackArgument` returned `true` for any method-call (PropertyAccess callee) without inspecting the parameter type. Arrows passed to `routes.push(arrow)` on `Handler[]`, `m.set("k", arrow)` on `Map<string, Handler>`, or `app.get(path, handler)` on a user-defined class therefore went through the host `__make_callback` path and were stored as JS-wrapped externrefs. A later `routes[i](...)` / `m.get(k)(...)` would `extern.convert_any` + `ref.cast` to the expected `__fn_wrap_N_struct`, fail (the JS wrapper isn't a struct), and null-deref at the subsequent `struct.get` — exactly the App.dispatch crash described in #1311 (and surfaced from #1309 Slice A).

## Fix (two-path discriminator)

- **Path A** — non-extern receiver (user class): use the resolved-signature param at the arg index. If it has call signatures, the arrow is being stored in a method that will dispatch via `call_ref` later, so use the GC-struct closure path. Restricting to non-extern receivers excludes Map.forEach / Array.map / etc., whose callback parameter is also callable but must round-trip through the JS host.
- **Path B** — extern receiver (`Map<K, V>`, `Set<T>`, `T[]`): inspect the receiver's element/value type. If it has call signatures, the storage method (`push`, `unshift`, `set`, `add`) is depositing a closure into a callable slot and we need the GC-struct path. Discriminates from HOFs because their callback's parameter type is not assignable from the element/value type — `Map<K, number>.forEach` has V=number which is not callable, so we never switch.

## Test plan

- [x] `tests/issue-1311.test.ts` — 6 cases: sync Map, async Map (full App.dispatch repro), mixed sync+async Map, `Handler[].push` + extract + call, `Handler[].push` + inline call, multi-handler Map dispatch (all pass).
- [x] No regressions in `tests/issue-859.test.ts` (Map.forEach callback mutations), `tests/issue-1306.test.ts` (callable element-access call), `tests/issue-1300.test.ts` (closure capture chain) — 25 tests across the four files all pass.
- [x] `tests/async-await.test.ts`, `tests/map-set*.test.ts`, `tests/illegal-cast-closures-585.test.ts`, etc. show identical pass/fail counts on this branch and `origin/main` (16 failed / 28 passed across 9 files — all pre-existing, none introduced by this fix).
- [ ] CI test262 sharded run + baseline-validate.

### DIFF
--- a/src/codegen/closures.ts
+++ b/src/codegen/closures.ts
@@ -1000,9 +1000,51 @@ export function isHostCallbackArgument(node: ts.Node, ctx: CodegenContext): bool
         // Fall through to host-callback path on any checker error
       }
     }
-    // For method calls (property access), check if the method is known array HOF
-    // (filter, map, etc.) — those have dedicated inline compilation and ARE handled
-    // as closure calls. For other property accesses, treat as host callback.
+    // For method calls (property access), check if the method is on a
+    // user-defined class. User-defined methods receive the closure as the
+    // GC-struct shape (`__fn_wrap_N_struct`) and may store it for later
+    // dispatch (e.g. `app.routes.set(path, handler)`). Routing through
+    // `__make_callback` here produces a JS-wrapped externref that fails the
+    // dispatch-site `ref.cast` and null-derefs at `struct.get`. (#1311)
+    if (ts.isPropertyAccessExpression(parent.expression)) {
+      const propAccess = parent.expression;
+      const methodName = propAccess.name.text;
+      try {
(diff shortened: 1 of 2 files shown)


ttraenkler:
Regressed after PR #264 merged — -18 net, 64 real regressions. The senior-dev fix in #264 already resolved the core #1311 issue. This branch's changes conflict; closing.

PR #257, opened 2026-05-07
fix(#1312): pre-register nested fn so self-reference resolves to its own closure

ttraenkler:
## Summary

Recursive nested function declarations passed as a parameter (the Hono `compose` pattern's `function next() { ... return mw(c, next); }`) failed with `TypeError: Cannot access property on null or undefined`. The recursive `next` self-reference inside `next`'s own body fell through to the graceful `ref.null.extern` fallback in `identifiers.ts` — and the compiled wasm called `call(ref.null extern)` which null-derefed.

Two converging gaps:

1. **`funcMap` registered too late.** `compileNestedFunctionDeclaration` (nested-declarations.ts) registered `funcMap[funcName]` and `nestedFuncCaptures[funcName]` AFTER compiling the body. References to the function inside its own body missed the lookup at `identifiers.ts:478` and emitted `ref.null.extern`.

2. **Self-reference captures sourced from wrong scope.** Even after pre-registering, `emitFuncRefAsClosure` (closures.ts:2647) materialized the closure struct's capture fields by reading `local.get cap.outerLocalIdx` — a local index in the OUTER (declaring) scope. From inside the lifted function's own body that index points to a different local, so recursive calls saw a stale (zero-initialised) ref-cell instead of the live outer one. Probe10 (`let counter = 3; function next() { counter = counter - 1; ...; return 1 + call(next); }`) returned 0 instead of 3.

## Fix

- Reserve a placeholder `mod.functions[]` entry for the nested function BEFORE compiling its body. Set `funcMap` + `nestedFuncCaptures` immediately. Fill in `body`/`locals` after the compile via the saved entry reference (so `addUnionImports`' index-shift doesn't desync the slot). Symmetrical change in both branches of `compileNestedFunctionDeclaration` (with-captures + no-captures).
- In `emitFuncRefAsClosure`, when the lifted fn's own body references itself (`fctx.name === funcName`), source captures from the leading params at indices `[0..numCaptures-1]` for value captures and `[numCaptures..numCaptures+numTdzFlags-1]` for TDZ-flag boxes — matching the lifted fn's parameter layout. The captures arrive as boxed ref-cells / raw values that can be re-pushed unchanged into the new self-closure struct, preserving the live mutation state.

## Test plan

- [x] `tests/issue-1312.test.ts` — 5 cases all pass:
  - sync compose with two arrow middlewares produces `"[A][B]end"`
  - async compose with two arrow middlewares produces `"[A][B]end"`
  - regression guard: direct named async recursion still works
  - sync inner-recursion via param with mutable ref-cell capture returns 3
  - async inner-recursion via param with mutable ref-cell capture returns 3
- [x] Non-regression sweep (30 / 30 pass): `tests/stress/hono-tier5.test.ts`, `tests/stress/hono-tier6.test.ts`, `tests/stress/lodash-tier2.test.ts`, `tests/issue-1283.test.ts`, `tests/issue-1306.test.ts`, `tests/issue-1308.test.ts`.

### DIFF
--- a/src/codegen/closures.ts
+++ b/src/codegen/closures.ts
@@ -2755,7 +2755,22 @@ export function emitFuncRefAsClosure(
 
     // Emit: struct.new with fields: func, cap0, cap1, ..., __tdz_*..., ...
     fctx.body.push({ op: "ref.func", funcIdx: trampolineFuncIdx });
-    for (const cap of nestedCaptures) {
+    // (#1312) Self-reference inside the lifted body of `funcName` itself —
+    // e.g. `function next() { return call(next); }`. The captures are
+    // already in scope as the leading params [0..numCaptures-1] of the
+    // lifted fn (mutable captures arrive as boxed ref cells, immutable as
+    // raw values). We re-push them by param index instead of trying to
+    // dereference `cap.outerLocalIdx`, which points into a different
+    // (outer) scope and yields garbage / null when reused inside the
+    // current lifted body.
+    const isSelfRef = fctx.name === funcName;
+    for (let i = 0; i < nestedCaptures.length; i++) {
+      const cap = nestedCaptures[i]!;
+      if (isSelfRef) {
+        // Captures arrive at param index `i` in the lifted fn (#1312).
+        fctx.body.push({ op: "local.get", index: i });
+        continue;
+      }
       if (cap.mutable && cap.valType) {
         const refCellTypeIdx = getOrRegisterRefCellType(ctx, cap.valType);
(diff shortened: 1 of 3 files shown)


ttraenkler:
Closing per tech-lead decision after triage. The pre-registration in the no-captures branch regressed 38 Function.caller/arguments tests that pass incidentally (their pre-#1312 codegen left self-references as ref.null.extern, satisfying assert.throws(TypeError) via null deref). Replaced by PR with narrower fix that pre-registers only in the has-captures branch — see new PR.

PR #259, opened 2026-05-07
fix(#1311): route user-defined-method arrow callbacks via closure-struct path

ttraenkler:
## Summary

- Fix the `Map<string, AsyncHandler>` dispatch null_deref reported in #1311
- Arrow callbacks passed to user-defined class methods now use the closure-struct path (`compileArrowAsClosure`) instead of `__make_callback`
- Closure can now round-trip through `Map.set` / `Map.get` and dispatch correctly via `call_ref` on the `__fn_wrap_N_struct` shape

## Root cause

`isHostCallbackArgument` returned `true` for ALL property-access call branches, sending arrows passed to `app.get(path, async c => ...)` through `__make_callback`. That host import returns a JS-wrapped externref that is not a Wasm GC struct. When stored in a Map and retrieved at the dispatch site, `any.convert_extern + ref.test (ref __fn_wrap_N) + ref.cast` falls through with `ref.null`, then `struct.get` derefs null in `App_dispatch`.

## Fix

Extend `isHostCallbackArgument` (`src/codegen/closures.ts`) to detect property-access callees whose method is on a user-defined class. We look up `\${ClassName}_\${methodName}` in `funcMap` (also walking `getBaseTypes()` for inheritance). If found with a non-import index, return `false` so the arrow is built via `compileArrowAsClosure`.

Narrow scope: host array HOFs (`forEach`, `map`, …) and other host method callbacks (`Promise.then`) continue through `__make_callback` since their methods aren't in `funcMap` as user-defined.

## Test plan

- [x] `tests/issue-1311.test.ts` (7 tests, all pass): canonical async reproducer, sync variant, 404 fallback, multiple-route dispatch, direct field-access (no Map), Array.forEach regression guard, Map.forEach regression guard
- [x] Adjacent suites pass: `issue-859`, `issue-1283`, `issue-1298`, `issue-1300`, `issue-1306`, `issue-457`, `issue-837`, `issue-1135`
- [x] Pre-existing failures on main (`promise-combinators`, `async-await`, `flatmap-closure`, `illegal-cast-closures-585`, `optional-direct-closure-call`) confirmed unchanged by running same suites on `origin/main`

### DIFF
--- a/plan/issues/sprints/50/1311-map-async-handler-dispatch-null-deref.md
+++ b/plan/issues/sprints/50/1311-map-async-handler-dispatch-null-deref.md
@@ -2,7 +2,7 @@
 id: 1311
 sprint: 50
 title: "Map<string, AsyncHandler> dispatch null_deref in App.dispatch path"
-status: needs-architect-spec
+status: in-progress
 created: 2026-05-07
 updated: 2026-05-07
 priority: medium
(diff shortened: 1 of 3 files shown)


ttraenkler:
Closing in favor of dev-1328's PR #255 (compatible fix, further along in CI). Per team-lead's call.

PR #269, opened 2026-05-08
fix(#1379): drop emitSafeExternrefToF64's typeof guard for spec ToNumeric

ttraenkler:
## Summary

`emitSafeExternrefToF64` (src/codegen/type-coercion.ts) previously gated `__unbox_number` behind a `__typeof_number` check and went to `f64.const NaN` for any non-Number externref. That broke the spec's ToNumeric (§7.1.4) for prefix/postfix `++` / `--` on null / undefined / string / boolean operands:

- `var x = null; ++x` → expected 1, got NaN
- `var x = "1"; x--` → expected 0, got NaN
- `var x = ""; ++x` → expected 1, got NaN
- `var x = " 42 "; ++x` → expected 43, got NaN
- `var x = true; ++x` → expected 2, got NaN

The guard was added when `__unbox_number` would throw "Cannot convert object to primitive value" on plain wasm-structs without valueOf / toString. With #1319's `_hostToPrimitive` fallback now returning `"[object Object]"` for that shape, calling `__unbox_number` directly is safe — V8's `Number(v)` does the full ToNumber chain (null→0, undefined→NaN, "1"→1, "abc"→NaN, true→1, object via ToPrimitive→Number).

## Fix

Removes the typeof check + the NaN-fallback `if/else`, calls `__unbox_number` directly. ~7 fewer Wasm instructions per externref→f64 site, identical observable behaviour for true Number externrefs (V8 short-circuits `Number(v)` when v is already a number), spec-correct behaviour for the rest.

## Test plan

- [x] `tests/issue-1379.test.ts` — 8 cases all pass:
  - `++null` is 1
  - postfix `x--` on `"1"` returns 1, leaves x at 0
  - postfix `x++` on null returns 0, leaves x at 1
  - `++""` is 1
  - `++" 42 "` is 43
  - `++"abc"` is NaN
  - regression guard: `++x` on number-typed local works
  - `++true` is 2, `++false` is 1
- [x] Pre-existing failure in `tests/issue-866.test.ts` ("missing argument triggers default") reproduces on main HEAD — unrelated to this change.
- [ ] CI quantifies the test262 net delta on the language/expressions/{prefix,postfix}-{increment,decrement} cluster (~30 expected).

### DIFF
--- a/src/codegen/type-coercion.ts
+++ b/src/codegen/type-coercion.ts
@@ -2138,25 +2138,18 @@ function emitToStringResultToF64ByKind(ctx: CodegenContext, fctx: FunctionContex
 export function emitSafeExternrefToF64(ctx: CodegenContext, fctx: FunctionContext): void {
   addUnionImports(ctx);
   const unboxIdx = ctx.funcMap.get("__unbox_number")!;
-  const typeofNumIdx = ctx.funcMap.get("__typeof_number")!;
-  const tmpLocal = allocTempLocal(fctx, { kind: "externref" });
-  fctx.body.push({ op: "local.tee", index: tmpLocal } as unknown as Instr);
-  // Check if it's a JS number (typeof === "number")
-  fctx.body.push({ op: "call", funcIdx: typeofNumIdx });
-  fctx.body.push({
-    op: "if",
-    blockType: { kind: "val", type: { kind: "f64" } },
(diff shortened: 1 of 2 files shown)


ttraenkler:
Superseded by PR #270 which has already merged.

PR #272, opened 2026-05-08
fix(#1380) slice 1: strict equality cross-type returns false

ttraenkler:
## Summary

`IsStrictlyEqual` (ECMA-262 §7.2.14) returns false whenever operands have different runtime types. The two-externref strict-eq path in `binary-ops.ts` routed through `__host_eq` (JS `===`) and, when that returned false, fell back to numeric unboxing comparison:

```ts
__host_eq(a, b) ? 1 : Number(a) === Number(b)
```

That fallback (#1065) was meant to handle "boxed numbers that differ in identity but have the same value", but our `__box_number` is identity for Number primitives — two `__box_number(5)` calls return the same primitive 5, and JS `===` handles that correctly via `host_eq`. The fallback never fired for its intended case but did unsoundly turn every cross-type strict-eq into numeric coercion: `null === 0` became `Number(null) === Number(0)` → `0 === 0` → true.

**Fix**: drop the numeric fallback. `__host_eq` is definitive for strict equality.

## Slice 1 of #1380

This PR fixes acceptance criterion 4 (`S11.9.4_A8_T4`) only. The other criteria need deeper work and are noted in the issue file as separate follow-ups:

- **Criterion 1** (ReferenceError eval-order propagation) — needs parser/checker work; cross-cuts the whole binary-expression compilation pipeline, not just equality.
- **Criterion 3** (`0n != Object(1n)`) — root cause is `Object(BigInt)` builtin dropping the BigInt argument before reaching `__new_plain_object`. That's an Object-builtin bug, not an equality bug.
- **Criterion 2** (Symbol.toPrimitive returning Symbol → TypeError) — overlaps with #1343 (Boolean wrapper + Symbol coercion).

## Changes

- `src/codegen/binary-ops.ts`: in the externref-vs-externref strict-eq path, drop the `__host_eq ? 1 : Number(a) === Number(b)` fallback. Trust `__host_eq` directly.
- `tests/issue-1380.test.ts`: 24 cases — all 14 strict-eq comparisons from `S11.9.4_A8_T4`, 2 strict-!== mirrors, 8 regression-coverage cases (object identity, primitive eq, string eq, loose `==` should still respect §7.2.15).

## Test plan

- [x] `tests/issue-1380.test.ts` — 24 / 24 pass on the fix; 6 fail on main HEAD
- [x] No regressions in scoped equality tests: issue-1014, issue-1016, issue-1018, issue-1024, issue-1025, issue-1021-null-vs-undefined
- [x] Pre-existing failure in issue-1015 reproduces unchanged on `origin/main` HEAD — not introduced by this fix
- [ ] CI: Test262 Sharded — `language/expressions/strict-equals/` and `strict-does-not-equals/` net pass delta


ttraenkler:
Closing: net -12 with 56 regressions. The numeric-unboxing fallback was load-bearing. Narrower typeof-gated fix needed before retry.

PR #303, opened 2026-05-08
feat(#1366a): class extends Error/TypeError/… — message + instanceof support

ttraenkler:
## Summary

Closes #1366a. Subclasses of builtin error types (`class MyError extends Error {}`) previously couldn't access `.message` (parent has no Wasm struct fields, super(msg) was a no-op) and `subInstance instanceof Error` always returned false (no struct tag for builtin errors).

This PR takes the **conservative path**: keep the WasmGC-struct instance representation, auto-add a `message` field to subclasses, and fix the static-instanceof analysis. The full externref-backed approach is reserved for #1366b/c/d (Array/Map/Set/Promise) where host structural state can't be inlined.

## What changed

- **`src/codegen/context/types.ts` + `create-context.ts`**: `builtinErrorParentMap: Map<string, string>` on `CodegenContext`.
- **`src/codegen/class-bodies.ts`**: `BUILTIN_ERROR_PARENTS` set, `resolveBuiltinErrorAncestor` walker. When parent is a builtin error: populate the map and auto-add `message: externref` to the subclass struct. `compileSuperCall` writes `this.message = arg[0]` (or `arg[1]` for AggregateError, which takes `(errors, message)`).
- **`src/codegen/expressions/identifiers.ts`**: `tryStaticInstanceOf` resolves `subInstance instanceof <BuiltinError>` based on the chain — `instance instanceof <Ancestor>` and `instance instanceof Error` both return true (NativeError ⊂ Error per spec); orthogonal builtins return false.

## Acceptance criteria

All 5 ACs met by `tests/issue-1366a-extends-error.test.ts` (10 cases):

- [x] AC#1 — `new MyError("x") instanceof MyError`
- [x] AC#2 — `new MyError("x") instanceof Error`
- [x] AC#3 — `new MyError("hello").message` returns "hello"
- [x] AC#4 — regression coverage (regular user classes still NOT instanceof Error; two-level user-class chain still works)
- [x] AC#5 — 7 test262 `subclass-builtins/subclass-<X>.js` files pass locally (Error / TypeError / RangeError / EvalError / ReferenceError / SyntaxError / URIError) — was 0/7 on main

## Out of scope (deferred)

- `extends Array/Map/Set/Promise/Function` — needs externref-backed instance representation (#1366b/c/d)
- `Symbol.species` for typed-container methods
- Throw + catch with subclass `instanceof` recovery — exception integration is broader than this slice
- `Object.prototype.toString.call(e) === "[object Error]"`

## Test plan

- [x] 10/10 unit tests pass
- [x] 7 test262 subclass-builtins tests verified locally
- [x] Regression check: regular user classes + plain inheritance unaffected
- [ ] CI green; +N net-per-test from the 7 confirmed passes (architect estimated +40-60; my analysis suggests +7-15 since the conservative path doesn't enable proto-chain or Object.prototype.toString tests)


ttraenkler:
Closing in favor of PR #307 which uses the externref-backed approach. The struct approach in this PR cannot pass the try/catch instanceof test — the catch handler sees an externref, not a WasmGC struct. PR #307 handles that case correctly.

PR #329, opened 2026-05-09
fix(#1383): typeof-gated strict-equality fallback for cross-type comparisons

ttraenkler:
## Summary

The strict-equality codegen for two externref operands had this shape:

```
if host_eq(a, b)        // JS ===
  return true
else
  return unbox(a) === unbox(b)   // <-- unsound for cross-type
```

The `unbox` fallback was load-bearing for V8's same-value-different-identity case (V8 sometimes returns different externref ids for numerically-equal JS numbers), but it incorrectly produced `true` for cross-type comparisons: `null === 0` became `unbox(null) === unbox(0)` → `0 === 0` → true. Spec §7.2.16 says strict equality between values of different types is always false.

Earlier PR #272 dropped the fallback entirely and caused **-12 net test262** because the fallback was masking unrelated mismatches in non-numeric comparisons. This PR keeps the fallback but **gates it on a runtime typeof check** — the numeric-unbox path only fires when both operands are `typeof === "number"`. Otherwise host_eq's `false` is final.

## Test plan

- [x] `tests/equivalence/issue-1383.test.ts` — 8 cases covering: `null === 0`, `undefined === 0`, `null !== 0`, same-type-pos cases (`null === null`, `undefined === undefined`), same-value-different-identity load-bearing fallback, `null !== undefined` for strict
- [x] Wider equality suite — 60/60 tests across `strict-equality-edge-cases`, `loose-equality`, `equality-mixed-types`, `comparison-coercion`, `issue-1134`, `issue-1134-string-number`, `issue-1383` preserves existing pass rate
- [x] Typecheck clean

### DIFF
--- a/src/codegen/binary-ops.ts
+++ b/src/codegen/binary-ops.ts
@@ -1566,6 +1566,23 @@ export function compileBinaryExpression(
             // Strict equality: __host_eq (JS ===) for reference identity.
             // If that returns false, fall through to numeric unboxing for
             // boxed numbers that differ in identity but have the same value. (#1065)
+            //
+            // (#1383) Gate the numeric-unbox fallback on a runtime typeof
+            // check — only fire it when BOTH operands are typeof === "number".
+            // The fallback was load-bearing for genuinely-different-identity
+            // boxed numbers (V8 sometimes returns different externref ids for
+            // numerically-equal JS numbers), but it incorrectly succeeded for
+            // cross-type strict comparisons too: `null === 0` produced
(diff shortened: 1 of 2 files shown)


ttraenkler:
Closing: 49/51 = 96% regression ratio, too broad. Same cascade pattern as PR #272 — the typeof gate intercepts legitimate paths (f64===f64 in JSON.parse, receiver-TypeError on Array.prototype.every.call(boolean), null receivers in isPrototypeOf). Will reopen with narrower approach. See escalation thread for full bucket breakdown.

PR #347, opened 2026-05-19
feat(#1323): $IteratorResult struct — eliminate __iterator_done/__iterator_value host imports

ttraenkler:
## Summary
- Introduces `$IteratorResult` WasmGC struct to replace `__iterator_done`/`__iterator_value` host imports
- Eliminates two host import round-trips per iterator step
- Changes in `src/ir/lower.ts` (41 lines) and `src/runtime.ts` (58 lines)

## Status
Implementation complete per spec doc. Marked as draft until CI validates the IR change doesn't regress test262.

## Test plan
- [ ] CI passes
- [ ] No test262 regressions
- [ ] Iterator-dependent tests (for-of, destructuring, generators) still pass


ttraenkler:
Closing — not a drift issue, the #1323 feature is incomplete and regresses conformance. Conflict resolution was done cleanly (sendev-432-347, local commit 4b9f14e30, not pushed), but the $IteratorResult mechanism is broken independent of the merge: #1323 rewrote the legacy codegen path (statements/loops.ts) to unconditionally any.convert_extern + ref.cast __iterator_next's result to $IteratorResult, but the runtime (runtime.ts ~L5904) only constructs that struct when callbackState.getExports().__make_iterator_result is reachable. In the default buildImports(imports, undefined, stringPool) usage the tests use, callbackState is absent → the defensive fallback returns a raw JS object → ref.cast traps with 'illegal cast'. Proven: tests/iterators.test.ts (string for-of) + tests/symbol-iterator-protocol.test.ts PASS on main, FAIL with #1323. Also tests/iterators.test.ts:90-91 still assert the __iterator_done/__iterator_value imports #1323 removes. Re-opened as a fresh #1323 task with this diagnosis as the spec.

PR #440, opened 2026-05-20
fix(#1542): class method dstr-param default not applied (scope propagation)

ttraenkler:
## Summary

Closes #1542. Classes nested inside a `try`/block/`if`/loop/`switch`/labeled
statement of a function were being treated as if at module level. Their
bodies were eagerly compiled BEFORE the enclosing function's
`hoistFunctionDeclarations` pass ran, so sibling function declarations
(used by parameter defaults like `method([,] = g())`) were absent from
`funcMap` at class-body compile time. The call-expression compiler
fell back to `ref.null.extern`, and the destructure guard threw
"Cannot destructure 'null' or 'undefined'" at runtime.

This is the actual root cause of the ~134 test262
`Cannot destructure 'null' or 'undefined' [in C_method() ← test]`
failures across `class/dstr/meth-dflt-*`, `gen-meth-*`,
`async-gen-meth-*`, `private-meth-*`, `private-gen-meth-*`. The
architect spec hypothesised a `coerceType` externref→vec gap, but
inspection showed paramType is externref (not a vec ref) and the
externref→vec materialisation path already handles it correctly — the
missing call to `g` was the real bug.

## Fix

`src/codegen/declarations.ts`: propagate `insideFunction` through every
recursive descent in `compileClassesFromStatements`. Classes nested in
any control-flow construct within a function are correctly deferred
until the enclosing function compiles its body, at which point
`hoistFunctionDeclarations` has registered sibling functions in
`funcMap`.

## Test plan

- [x] `tests/issue-1542.test.ts` — 4 cases (try-block generator default,
      bare-block, object-pattern default, top-level no-regress)
- [x] Smoke 4 representative test262 files: all flip from
      "Cannot destructure null/undefined" → either pass or
      "second == 1" generator-eager-materialisation failure
      (separate issue, not in scope for #1542)
- [ ] CI: test262 conformance net delta


ttraenkler:
**REVERTED via #516** (admin-merge 2026-05-23T02:35:32Z) due to confirmed -1219 test262 regression: 691 wasm_compile errors + 279 runtime_error + 275 assertion_fail. Investigation: this PR's class-deferral change broke shape registration that module-level code depended on. Issue #1542 needs respec: propagation must not eager-skip classes whose shapes are referenced from module-level. See revert PR #516 for full diagnosis.

PR #449, opened 2026-05-20
fix(#1557): use AST param count for method struct hash + fork per-literal method bodies

ttraenkler:
## Problem

ESLint's `flat-config-schema.js` has many `{ validate(value) {...} }` schema objects plus one `{ validate() {...} }`. Because these are typed via JSDoc unions (`@property {Function|string} validate`), TypeScript's `getCallSignatures()` returns 0 — `methodSigParts` was empty, so all literals hashed to the same `__anon_N` struct. Both method bodies shared `__anon_N_validate`, and the second body overwrote the first's `typeIdx`, causing Wasm validation failure:

```
not enough arguments on the stack for call (expected [ref __anon_N, externref], got [ref __anon_N])
```

## Fix (three-part)

1. **`ensureStructForType` hash** (`index.ts`): Use AST `MethodDeclaration.parameters.length` instead of TS sig's for `methodSigParts`
2. **`ensureStructForType` placeholder params** (`index.ts`): Iterate `decl.parameters` (AST) rather than `sig.parameters` (TS type)
3. **`compileObjectLiteralForStruct` fork** (`literals.ts`): When a prior literal claimed a placeholder's body slot, allocate a fresh `funcIdx`

Note: `origin/issue-1557-trampoline-arity` fixes the same bug via the literals.ts fork alone (without the upstream index.ts root-cause fix). This PR is a superset.

Closes #1557


ttraenkler:
Superseded by #461 (per-literal funcIdx for trampoline arity — different/safer approach). Closing duplicate per PR pre-review 2026-05-21.

PR #450, opened 2026-05-20
fix(#1558): coerce i32 left operand to f64 in numeric/equality binary ops

ttraenkler:
## Problem

`compileBinary` in `binary-ops.ts` coerced only the *right* operand from i32 to f64 before calling `compileNumericBinaryOp`. When the left operand's TS type was `number` but its lowered Wasm type was i32 (e.g., `messages.length === 1` — `.length` is i32, `1` is f64), the left stayed on the stack as i32 and the emitted `f64.eq` failed validation:

```
f64.eq[0] expected type f64, found call of type i32
```

## Fix

- Route the **both-i32** case through `compileI32BinaryOp` (faster, no needless f64 round-trip)
- For **left-i32, right-f64**: use save-right / convert-left / restore-right pattern before `compileNumericBinaryOp`

## Impact

Unblocks ESLint Tier 1d — `Linter_verifyAndFix` now validates.

Closes #1558


ttraenkler:
Superseded by #462 (widen LEFT i32 to f64 — more consistent with surrounding code). Closing duplicate per PR pre-review 2026-05-21.

PR #478, opened 2026-05-21
ci(#110): harden auto-refresh-prs.yml — per-PR error isolation

ttraenkler:
## Summary

Per task #110. The `Refresh each open PR branch` step in `.github/workflows/auto-refresh-prs.yml` runs under bash's default `-eo pipefail`. When `git push` fails for any single PR (permissions, ref-update conflict, etc.), the entire script exits and the remaining PRs are silently skipped.

Observed in run [26229881489](https://github.com/loopdive/js2wasm/actions/runs/26229881489): the loop aborted mid-iteration after PR #473 first merged, leaving the queue in an inconsistent state. A later run (26237106363) succeeded for PR #475 — confirming the failure is intermittent and depends on which branch trips first.

## Fix

Move the per-PR work into a `refresh_one` shell function:

- `set +e` at the top of the function — local error tolerance.
- Each git operation guarded with `if ! …; then return 1; fi` instead of relying on `-e`.
- `return 1` exits the function but **not** the script, so the outer `while read` keeps iterating.
- Push failures now leave a PR comment ("merged locally but push failed — please refresh manually") so devs see the state.
- Outer loop tracks `ok` / `failed` counters and prints a summary.
- Switched from `echo "\$prs" | while read` (subshell — counters lost) to `done <<< "\$prs"` (here-string — counters persist).
- The step explicitly `exit 0` so per-PR failures are reported in the log without failing the workflow itself.

## Test plan

- [ ] Workflow-only change — no source touched, no equivalence tests needed.
- [ ] YAML syntax validated locally with `js-yaml`.
- [ ] After merge, monitor the next 1–2 runs of `auto-refresh-prs` for the new summary line and confirm per-PR isolation.


ttraenkler:
Superseded by [`8ba69bbee`](https://github.com/loopdive/js2wasm/commit/8ba69bbee) which disabled `auto-refresh-prs.yml` entirely — the merge queue handles base-currency natively now, so the per-PR error-isolation hardening this PR delivered is no longer load-bearing. The function-based error-isolation pattern (`refresh_one`, `return 1` vs `exit 1`) is good and can be reused if we ever re-enable the workflow.

Thanks for the clean fix — keeping it referenced here for future revival.

PR #543, opened 2026-05-23
fix(#1151): broaden isAsyncCallExpression for indirect async callees (Gap A1)

ttraenkler:
## Summary
- #1151 Gap A1: broaden `isAsyncCallExpression` (`src/codegen/expressions.ts:154`) so indirect async callees are recognised and their synchronous throws become rejected Promises per ECMA-262 §27.7.5.2.
- Previously the detector only matched an `async` modifier on a resolvable declaration. Calls through a variable holding an async fn (`const f = asyncFn; f()`) or a `() => Promise<T>` callback carried no such modifier, so the call-site try/catch wrap (#1150) was skipped and the throw trapped at the wasm boundary instead of rejecting.
- Fix: structural fallback — for callees not under a `NewExpression`, iterate the callee TS type's call signatures and treat the call as async if any return type is `Promise<T>` (`isPromiseType`). Async generators stay excluded (asteriskToken short-circuit; their return type is `AsyncGenerator`, not `Promise`).

Scope: Gap A1 only. Gap B's binding-pattern guard already landed on main (closures.ts:1229-1232); Gap C (IR `IrInstrAsyncThrow`) stays deferred until the #1373b gate flip. Body-wrap Option 1 remains declined per the joint async spec.

## Test plan
- [x] New `tests/issue-1151-gap-a1.test.ts` (4 cases): variable-typed async call returns Promise / does not trap on body throw; `() => Promise<T>` callback treated as async; non-async indirect call NOT wrapped (returns raw value). All pass.
- [x] Regression: async-function, async-iteration, binding-null-guard, async-await, issue-1151-gap-b equivalence suites show IDENTICAL pre-existing failures on clean-main baseline vs this branch — zero new regressions. (Remaining fn-expr Gap B `illegal cast` failures are the un-mirrored `compileFunctionExpression` path, out of A1 scope.)
- [ ] CI test262 regression gate (−5 ≤ Δ ≤ +30 expected)

### DIFF
--- a/plan/issues/backlog/1151-async-function-synchronous-throws-bypass.md
+++ b/plan/issues/backlog/1151-async-function-synchronous-throws-bypass.md
@@ -469,7 +469,29 @@ spec §5: ≤ 10 regressions, no single bucket > 50. Net target: **−5 ≤ Δ �
 
 | Gap | Owner | Status |
 |-----|-------|--------|
-| B — `closures.ts:1170` binding-pattern coercion | dev (1-line + mirror) | ready, ship first |
-| A1 — `expressions.ts:154` detector broadening | dev (~15 LOC) | ready, ship second |
+| B — `closures.ts:1170` binding-pattern coercion (arrow/fn-expr) | dev | DONE — guard landed on main at closures.ts:1229-1232 |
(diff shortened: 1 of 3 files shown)


ttraenkler:
Closing as duplicate of #542, which already merged the same #1151 Gap A1 fix (broaden isAsyncCallExpression to Promise-returning callees via getCallSignatures). Parallel-dispatch collision — both PRs implemented the equivalent structural fallback. Not merging this to avoid redoing landed work.

PR #613, opened 2026-05-24
fix(#1130): array-index-exotic length growth on defineProperty (PR-0)

ttraenkler:
## Summary
PR-0 of the #1130 getter-observing array-methods work (per the 2026-05-24 authoritative re-spec). `Object.defineProperty(arr, "n", desc)` with a canonical array index `n >= arr.length` now bumps the WasmGC vec's logical length (struct field 0) to `n + 1`, per ES §10.4.2.1 (array exotic `[[DefineOwnProperty]]` / ArraySetLength).

This is the prerequisite slice: without length growth, the array-method callback loop never reaches a newly accessor-defined index. PR-0 only fixes the loop bound; element-read accessor observation is PR-1 scope.

## Changes
- `src/codegen/object-ops.ts` — new helpers `parseCanonicalArrayIndex`, `resolveVecTypeIdx`, `maybeEmitVecLengthGrowth` (guarded `if (n >= vec.length) vec.length = n+1` on the raw vec ref). Wired into all three array-receiver defineProperty paths:
  - struct-accessor branch (`get`/`set` on a typed vec → `<vec>_get_n` accessor func),
  - `emitExternDefinePropertyNoValue` (accessor + flag-only),
  - `emitExternDefinePropertyValue` (value descriptor; raw vec ref captured via `local.tee` before `extern.convert_any`).
- `tests/issue-1130.test.ts` — 7 tests.
- Issue file: PR-0 results note appended; `status: ready` retained (PR-1/PR-2 follow).

## Notes
- Gated on the receiver statically resolving to a vec struct (length+data) — non-array defineProperty and externref-typed locals are untouched (a type guard prevents an invalid `struct.get`). Only canonical array indices in `[0, 2^32-2]` grow length (`"length"`, `"01"`, `"4294967295"` do not).
- The re-spec's claim that *every* array shape routes accessor defineProperty through `__defineProperty_accessor` holds only for `any`-typed arrays; a statically-typed `number[]` accessor routes to the struct-accessor branch (`<vec>_get_n`), so growth is wired there too. Flagged in the issue file for PR-1.

## Test plan
- [x] `tests/issue-1130.test.ts` — 7/7 pass (accessor/value growth, no-grow-within-length, non-index keys, string-element vec, forEach-unaffected guard)
- [x] Verified pre-existing `array-methods.test.ts` strict-type failures reproduce on clean main (unrelated to this change)
- [ ] CI required checks green

### DIFF
--- a/plan/issues/1130-array-methods-getter-observing-property.md
+++ b/plan/issues/1130-array-methods-getter-observing-property.md
@@ -3,7 +3,7 @@ id: 1130
 title: "Array methods — getter-observing property access on indices and length"
 status: ready
 created: 2026-04-20
-updated: 2026-04-28
+updated: 2026-05-25
 priority: medium
 feasibility: hard
 reasoning_effort: high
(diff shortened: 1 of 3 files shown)


ttraenkler:
Superseded by #614 (same 3-file PR-0 scope for #1130). #614's author is active and its fix also grows the backing $data array — avoiding the iteration-trap the naive 'bump field 0 only' approach hits. dev-c's branch issue-1130-array-getter (b00babe27) is preserved. Closing this orphaned duplicate so the two don't race in the merge queue.

PR #614, opened 2026-05-24
fix(#1130): array-index-exotic length growth on defineProperty (PR-0)

ttraenkler:
## Summary

PR-0 of the multi-PR #1130 sequence (getter-observing array methods),
implementing item A of the 2026-05-24 authoritative re-spec.

`Object.defineProperty(arr, "n", desc)` on an array exotic object where the
numeric index `n >= arr.length` now grows the array length to `n+1` per
ES §10.4.2.1 (ArraySetLength via `[[DefineOwnProperty]]`). Previously the
length stayed unchanged.

## Key correction to the spec

The re-spec proposed "bump struct field 0 (length) only." **That is unsafe.**
Bumping logical length without growing the physical backing `$data` array
leaves the vec inconsistent (`length > backing capacity`), so a subsequent
`for`-loop to `arr.length` or `arr.forEach` **traps** with "array element
access out of bounds" (verified by probe).

PR-0 therefore also grows the backing array (capacity → `n+1`,
`array.new_default` + `array.copy`), mirroring the existing indexed-assignment
grow path in `src/codegen/expressions/assignment.ts`. The vec stays internally
consistent and iteration no longer traps.

## Scope boundary

PR-0 grows length + capacity only. It does **not** store the value-descriptor's
value nor invoke accessor getters during iteration — grown slots read as the
element default. The element/length accessor slow-path read (getter
observation) is **PR-1's scope**, not in this PR. PR-0 is independently safe
and valuable (fixes a real `arr.length`-after-numeric-`defineProperty` bug).

## Changes
- `src/codegen/object-ops.ts` — `parseCanonicalArrayIndex` + `maybeEmitVecLengthGrowth`, called at the top of `compileObjectDefineProperty`.
- `tests/issue-1130.test.ts` — 8 regression cases.
- `plan/issues/1130-...md` — PR-0 implementation note, status → in-review.

## Test plan
- [x] `tests/issue-1130.test.ts` 8/8 pass (index grow value+accessor, `length`-key exclusion, no-shrink in-range, for-loop + forEach no trap, default read, `string[]`).
- [x] `tsc --noEmit` clean.
- [x] Existing defineProperty suites (1460, 846, 846-class-static) pass unchanged.
- [x] Array suites (capacity, bounds-elimination, oob, callback-three-params) show identical pass/fail vs clean main — zero regressions.

### DIFF
--- a/plan/issues/1130-array-methods-getter-observing-property.md
+++ b/plan/issues/1130-array-methods-getter-observing-property.md
@@ -1,9 +1,9 @@
 ---
 id: 1130
 title: "Array methods — getter-observing property access on indices and length"
-status: ready
+status: in-review
 created: 2026-04-20
-updated: 2026-04-28
+updated: 2026-05-25
 priority: medium
 feasibility: hard
 reasoning_effort: high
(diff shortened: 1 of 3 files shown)


ttraenkler:
Closing — duplicate of #613 (dev-c owns the #1130 PR sequence). Tech-lead double-dispatch; standing down.

PR #673, opened 2026-05-27
fix(#1620): multi-value __iterator_next eliminates done/value host imports

ttraenkler:
## Summary
- Changes `__iterator_next` from `(externref)→externref` to `(externref)→(i32 done, externref value)` — a Wasm multi-value result. The runtime reads `done`/`value` off the raw JS result and returns `[done?1:0, value]`; V8 destructures the two-element iterable onto the Wasm stack via the multi-value ABI.
- **Deletes the `__iterator_done` and `__iterator_value` host imports outright** (issue's primary acceptance criterion + `goal: host-independence`). Collapses 3 host calls per for-of step into 1.
- Both returned values are primitives, so there is **no `$IteratorResult` GC struct** to survive a JS import hop — sidesteps the v1 BLOCKED failure (a WasmGC struct externalized through the `__iterator_next` JS import surfaces as `undefined` in V8/Node 25). The BLOCKED failure mode is now structurally impossible.

## Implementation
Follows the `## Implementation Plan (v2)` in `plan/issues/1620-iterator-result-struct-runtime-wiring.md` (Direction 1: multi-value import).
- `src/codegen/index.ts` — `addIteratorImports`: `__iterator_next` registered with the 2-result func type; `__iterator_done`/`__iterator_value` + the `extToI32` type removed.
- `src/codegen/statements/loops.ts` — `compileForOfIterator`: single multi-value call; value (top of stack) → element, done (below) → break check. Doc-comment updated.
- `src/ir/lower.ts` — `forof.iter` lowering consumes the multi-value result inline (value into element slot, done left on stack for `br_if`); the now-incompatible standalone `iter.next`/`iter.done`/`iter.value` IR cases (unused — front-end only emits `forof.iter`) throw a clear error.
- `src/runtime.ts` — `__iterator_next` folds in done/value extraction and returns `[done?1:0, value]`; old split branches deleted.
- `tests/equivalence/helpers.ts` — host helper updated to the multi-value shape.
- `tests/iterators.test.ts` — stale WAT assertions flipped to assert `__iterator_done`/`__iterator_value` are **gone**.

## Test plan
- [x] `tsc --noEmit` clean
- [x] `tests/iterators.test.ts` (5 string for-of) + `tests/symbol-iterator-protocol.test.ts` (custom iterable — the real `__iterator_next` exerciser, was the v1 proven regression) — **10/10 pass**
- [x] IR-path: `issue-1372-ir-destructuring-params`, `issue-1374-ir-string-iter-inline` — 17/17 pass
- [x] Equivalence sweep (for-of-basic, for-of-generator, iterator-protocol-custom, map-set, ir-slice10-map-set, generator-yield-delegation, spread-in-new) — pass
- [x] No new failures: `for await...of` (async-iteration), `tests/helpers.js`-missing files, and map-set type errors all **pre-exist on clean main** (verified via stash). Async-iterator parity is explicitly out of scope per the v2 spec (sync custom-iterable is the acceptance gate).

### DIFF
--- a/plan/issues/1620-iterator-result-struct-runtime-wiring.md
+++ b/plan/issues/1620-iterator-result-struct-runtime-wiring.md
@@ -1,9 +1,10 @@
 ---
 id: 1620
 title: "$IteratorResult struct: eliminate __iterator_done/__iterator_value host imports (runtime wiring gap)"
-status: ready
+status: done
 created: 2026-05-24
 updated: 2026-05-27
+completed: 2026-05-27
 priority: medium
 feasibility: medium
 reasoning_effort: medium
--- a/src/codegen/index.ts
+++ b/src/codegen/index.ts
@@ -6746,23 +6746,14 @@ export function addIteratorImports(ctx: CodegenContext): void {
   const extToExt = addFuncType(ctx, [{ kind: "externref" }], [{ kind: "externref" }]);
   addImport(ctx, "env", "__iterator", { kind: "func", typeIdx: extToExt });
 
-  // __iterator_next: (externref) → externref — calls iter.next()
+  // __iterator_next: (externref) → (i32 done, externref value) — calls iter.next()
+  // Multi-value result avoids the $IteratorResult struct: a WasmGC struct cannot
+  // survive the JS import hop — it surfaces as undefined in V8 (see #1620). The
+  // two primitives (i32 done + externref value) cross the boundary cleanly.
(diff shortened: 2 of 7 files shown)


ttraenkler:
Closing duplicate — PR #672 (issue-1620-multivalue-next) has the same fix plus a dedicated test file. Will proceed with #672 after the #1590 quality gate fix lands on main.

PR #686, opened 2026-05-27
fix(#1613): accept binding-pattern for-in head

ttraenkler:
## Summary
- `compileForInStatement` rejected for-in heads with an array/object binding pattern (`for (var [x, x] in obj)`) at compile time ("for-in variable must be an identifier").
- Now destructures the enumerated key string into the pattern each iteration via `compileForOfDestructuring` — array-pattern destructuring of a string iterates its characters per spec (`for (var [x, x] in { ab: 1 })` binds `x = 'b'`).

## Test results (test262 for-in dir, 20 binding-pattern/scope candidates)
- Baseline (main): 10 pass / 10 fail
- This branch: 12 pass / 8 fail — net +2, no regressions
- Newly passing: `head-var-bound-names-dup.js`, `head-let-destructuring.js`
- Remaining fails are unrelated (fresh-binding-per-iteration, TDZ closure capture, default-on-exhausted-iterator).

## Test plan
- [x] `tests/issue-1613.test.ts` (new) passes
- [x] `tests/issue-forin.test.ts` (existing) still passes
- [ ] CI sharded test262 + quality green

### DIFF
--- a/plan/issues/1613-for-in-variable-must-be-identifier.md
+++ b/plan/issues/1613-for-in-variable-must-be-identifier.md
@@ -1,9 +1,10 @@
 ---
 id: 1613
 title: "codegen: for-in head with binding pattern / non-identifier rejected ('for-in variable must be an identifier')"
-status: ready
+status: done
 created: 2026-05-24
-updated: 2026-05-24
+updated: 2026-05-27
+completed: 2026-05-27
 priority: low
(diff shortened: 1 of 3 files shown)


ttraenkler:
Superseded by PR #683 which covers the same binding-pattern case plus member-expression targets. Closing as duplicate.

PR #688, opened 2026-05-27
fix(#1318): bring sharded CI runner up to parity for assert diagnostics

ttraenkler:
## Summary

The first #1318 fix (commit 39fa6ef3f) only improved the `runTest262File` smoke-test path. The **sharded CI runner** (`tests/test262-shared.ts`, driven by the `test262-chunk*.test.ts` files that generate the live conformance JSONL) and the legacy `tests/test262-vitest.test.ts` still used the old `findNthAssert`: a 120-char cap, a narrow `assert|verify\w+` regex, and a bare `returned N (found M asserts in source)` fallback. That path is what produces the ~8,900 vague `assertion_fail` entries.

Extracted the locator into `tests/test262-assert-locator.ts` (side-effect-free, unit-testable) and consume it from both runners:

- `extractFullAssert` balances parens across lines, capturing the full `assert.sameValue(actual, expected, "message")` call — including the message argument `wrapTest` strips from the compiled body but is still in the source. Cap raised 120 → 500.
- Broadened the assert-detection regex to also match `assert.*`, `compareArray`, `$DONOTEVALUATE`, and `throw new Test262Error`.
- When the executed-assert counter outruns the static source scan (loops, helper-internal asserts), anchor on the **last** assertion instead of a bare `returned N`. The `returned N —` prefix is preserved so `classifyError` bucketing is unchanged (still `assertion_fail`).

This is a test-harness diagnostics change only — no compiler/runtime code touched, so test262 pass/fail counts are unaffected; only the `error` text of existing failures gets richer.

## Test plan

- [x] `tests/issue-1318-locator.test.ts` — 10 new cases (findNthAssert + extractFullAssert)
- [x] Existing `tests/issue-1318.test.ts` (Part 1 / runTest262File path) stays green
- [x] `tsc --noEmit` clean on touched files
- [ ] CI sharded test262 run shows no regression in pass/fail counts

### DIFF
--- a/plan/issues/1318-returned-n-bare-exit-code-context.md
+++ b/plan/issues/1318-returned-n-bare-exit-code-context.md
@@ -1,9 +1,10 @@
 ---
 id: 1318
 title: "test harness: 'returned N' bare exit code — capture last assertion detail (~8,900 vague failures)"
-status: ready
+status: done
 created: 2026-05-07
-updated: 2026-05-07
+updated: 2026-05-27
+completed: 2026-05-27
 priority: high
 feasibility: medium
 reasoning_effort: medium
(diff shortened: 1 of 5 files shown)


ttraenkler:
Superseded by #720 (clean cherry-pick on current main — resolves DIRTY merge state)

PR #694, opened 2026-05-27
fix(#1658): apply parameter default in call-site inlining path

ttraenkler:
## Summary
- Fixes #1658 — a function-parameter default was not applied when the callee was inlined, so `process(5)` returned 5 (not 15) and the `tests/equivalence/destructuring-extended.test.ts` "destructured function parameters with defaults" case returned **30** instead of **40**.

## Root cause
A small expression-shaped function with a **constant** parameter default (`function f(x, y = 10)`) is marked inlinable by `registerInlinableFunction`. For constant defaults the callee-side default guard is elided (#869), so the caller must materialize the default. The inline path in `src/codegen/expressions/calls.ts` padded missing args with a plain `pushDefaultValue` (`f64.const 0`), ignoring the declared default → `f(5)` inlined to `y = 0`.

## Fix
In the inline path, consult `ctx.funcOptionalParams` for the callee. When an argument is **missing** OR an explicit **`undefined`-like literal**, emit the recorded default via `pushParamSentinel(..., optEntry)` (which yields the constant value for constant defaults). Expression defaults can't reach this path — their callee guard emits an `if`/`local.set`, which `INLINE_DISALLOWED_OPS` already excludes from inlining.

## Test plan
- [x] `tests/issue-1658.test.ts` — scalar omitted, explicit `undefined`, default-of-0, multiple trailing defaults (4 pass)
- [x] `tests/equivalence/destructuring-extended.test.ts` now returns 40 (4 pass)
- [x] No new regressions: pre-existing failures in `arrow-call-apply`, `null-destructure-param-object`, tagged-template, `destructuring-initializer` are identical with and without this change (verified via stash A/B)

### DIFF
--- a/plan/issues/1658-destructured-function-param-default-not-applied.md
+++ b/plan/issues/1658-destructured-function-param-default-not-applied.md
@@ -1,10 +1,11 @@
 ---
 id: 1658
 title: "Destructured/scalar function-parameter default not applied (returns wrong value)"
-status: ready
+status: done
 sprint: Backlog
 created: 2026-05-24
-updated: 2026-05-24
+updated: 2026-05-27
+completed: 2026-05-27
 priority: high
 feasibility: medium
 reasoning_effort: medium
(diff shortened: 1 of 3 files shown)


ttraenkler:
Duplicate of #692 — same root cause and fix. Closing in favor of the teammate's PR.

PR #771, opened 2026-05-28
fix(#1648): Object.create descriptor ToBoolean fidelity (+32 test262)

ttraenkler:
## Summary
- Object.create static-expansion fast path in `calls.ts` only treated literal `true` as truthy when reading the writable/enumerable/configurable flags, silently degrading every other shape (`123`, `'x'`, `{}`, `[]`) to `false`. Per §6.2.6 ToPropertyDescriptor these flags must run through ToBoolean.
- Adds `staticToBoolean(expr)` that resolves ToBoolean for literal shapes (numeric/string/object/array/function/regex/null/true/false; identifier `undefined`/`NaN`/`Infinity`; `void`; `!` prefix). Unwraps `as`/`<T>`/`()`/satisfies/non-null wrappers so `123 as any` still resolves.
- Outer fast-path gate tightened: every flag's value must be statically ToBoolean-resolvable. Non-resolvable flags (`configurable: someVar`) fall through to the runtime `__defineProperties` path which honors ToBoolean natively.
- Result: `built-ins/Object/create` 173 → 205 / 320 (+32, 54.1% → 64.1%). `tests/issue-1648.test.ts` 8/8 pass.

## Test plan
- [x] 8/8 focused tests in `tests/issue-1648.test.ts` pass
- [x] `built-ins/Object/create` 173 → 205 / 320 via real test262 runner
- [x] No source impact on `built-ins/Object/defineProperty` or `built-ins/Object/defineProperties` buckets (this fix lives in the Object.create call-site, not the shared descriptor runtime)

## Remaining out of scope (see issue Resolution section)
- ~12 "Cannot convert object to primitive value" — ToPrimitive coercion
- ~12 "Getter must be a function: [object Object]" — accessor-descriptor unwrap
- ~10 missing TypeError on invalid descriptors
- Misc data-vs-accessor transitions, Proxy `[[GetPrototypeOf]]` counting

These overlap #1630 (ESCALATED-NEEDS-SPEC) and #1334 edge cases — not regressions from this fix.

### DIFF
--- a/plan/issues/1648-spec-gap-object-create-properties-map.md
+++ b/plan/issues/1648-spec-gap-object-create-properties-map.md
@@ -1,9 +1,10 @@
 ---
 id: 1648
 title: "spec gap: Object.create(proto, descriptors) ignores descriptor map (162 test262 fails)"
-status: ready
+status: done
 created: 2026-05-08
-updated: 2026-05-24
+updated: 2026-05-28
+completed: 2026-05-28
 priority: medium
 feasibility: medium
 reasoning_effort: medium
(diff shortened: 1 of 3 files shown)


ttraenkler:
Closing in favor of #NEW — re-cut on fresh origin/main base. PR #771 was based on stale main and the merge brought conflicts that would have reverted #1644 slice-A / #1685 native ToNumber work landed since. The replacement contains only the surgical staticToBoolean fix (+102/-10 in calls.ts) + tests.

PR #794, opened 2026-05-28
fix(#1528a): dynamic new via __reflect_construct (architect spec)

ttraenkler:
## Summary

Route `new <runtime-value>()` — where the callee is a parameter, local binding, member access on an `any`-typed receiver, or any other value the compiler cannot resolve to a known class at compile time — through the host's `__reflect_construct`. The host wrapper delegates to `Reflect.construct`, which performs spec `IsConstructor(F)` and throws the canonical `TypeError("X is not a constructor")` when the check fails.

Implements the architect plan in `plan/issues/1528-non-constructor-typeerror-promise-and-species.md` (L114). #1528b (the static unwrap subset, PR #756) merged 2026-05-27; this PR completes #1528a.

## Changes

- **`src/codegen/expressions/new-super.ts`** — new `compileDynamicConstruct` helper lowers `new <expr>(...)` to `__js_array_new` / `__js_array_push` / `__reflect_construct`. Wired into the legacy `__new_<ctorName>` unknown-constructor fallback: when no host import is registered and the callee is an identifier, route through the dynamic path under JS-host mode; standalone (no JS host) falls back to `emitThrowTypeError("is not a constructor") + ref.null.extern` (matches the existing static-guard behaviour).
- **`src/codegen/declarations.ts`** — pre-pass `collectUnknownConstructorImports` no longer registers stale `env.__new_<name>` host imports for callees that resolve to runtime locals (parameters, non-class let/const/var, function declarations). Top-level FunctionDeclarations still take the in-module function-style class path that precedes the unknown-constructor fallback.

Spec: §13.3.5.1.1 EvaluateNew, §7.3.15 Construct, §7.2.4 IsConstructor.

## Test plan

- [x] `tests/issue-1528a.test.ts` (4 cases) — all pass:
  - arrow-valued param: `new <arrow>()` throws TypeError
  - member-of-any: `new obj.fn()` throws TypeError
  - `new (Math.abs as any)()` throws TypeError
  - `new null()` throws TypeError
- [x] 2 confirmed net new test262 passes for the canonical spec pattern (`assert.throws(TypeError, () => { new resolve(); })` where `resolve` is a parameter):
  - `test/built-ins/Promise/create-resolving-functions-resolve.js`
  - `test/built-ins/Promise/create-resolving-functions-reject.js`
- [x] Related-issue suites green (1605, 1605-cpn, 1594, 1682, 1679): 17/17 pass.
- [x] `class*.test.ts` pass/fail counts identical between this branch and `origin/main` (the 27 helpers-setup failures are pre-existing and unrelated to this change).
- [ ] CI test262-sharded regression gate

## Remaining (out of scope)

The 5 originally-named test262 files in the issue still fail, but in distinct upstream paths — `Promise.resolve.call(NotPromise)` host delegation still throws the legacy `[object Object] is not a constructor` host-string. That is the next layer of #1528, tracked separately.

- **Spread in dynamic `new`** — bridged to #1609.
- **Standalone/WASI dynamic construct** — would need a Wasm-native `IsConstructor`; currently throws real TypeError statically.

### DIFF
--- a/plan/issues/1528-non-constructor-typeerror-promise-and-species.md
+++ b/plan/issues/1528-non-constructor-typeerror-promise-and-species.md
@@ -1,9 +1,10 @@
 ---
 id: 1528
 title: "spec gap: non-constructor TypeError — Promise.all / allSettled species and executor paths"
-status: in-progress
+status: done
 created: 2026-05-20
-updated: 2026-05-27
+updated: 2026-05-28
+completed: 2026-05-28
 priority: medium
 feasibility: medium
 reasoning_effort: medium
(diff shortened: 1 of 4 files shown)


ttraenkler:
Catastrophic regression confirmed 3x: 945 pass→fail, net −816. compileDynamicConstruct calls flushLateImportShifts 3x and fires as broad catch-all at line 3262 (old reportError site), corrupting already-emitted call indices across the entire function body — same #618 pattern. Closing; the architect spec needs a rework before re-implementation.

PR #817, opened 2026-05-28
docs(#820m, #1694): NamedEvaluation findings + file Promise subclass issue

ttraenkler:
## Summary

Two related docs-only commits:

- **#820m** — Investigation confirms NamedEvaluation `fn-name-class` /
  `__proto__-fn-name` is **NOT a localized fix**. The class value in
  `{ prop: class {} }` and `const c = class {}` returns `null` (not a class
  with the wrong `.name`) — the class itself is missing because
  `compileObjectLiteralForStruct` doesn't call `compileClassExpression` for
  property values, and `collectAnonymousClassesInNewExpr` doesn't walk
  PropertyAssignment / BindingElement initializers. Reaches the same
  conclusion as #1513/#1632/#1674: needs architect spec for class-as-value
  representation. Independently probed against current main and confirmed
  ALL function/class `.name` reads fail (even `function bar(){}.name` and
  `class Foo {}.name` declarations) — the representation gap is deeper
  than the original ~12 NamedEvaluation cases.

- **#1694** — New backlog issue filed for Promise.any/all/allSettled/race
  subclass/capability bugs (~50 fails total). Two distinct fingerprints:
  non-Promise capability `this` (~40 fails, combinators hard-wire
  `new Promise(...)`) and `class X extends Promise` derived static-method
  codegen (~28 fails, `extern.convert_any[0]` invalid Wasm). Sub-cluster B
  overlaps with #1632b / #1364b representation work.

## Test plan

- [x] Docs-only — no source changes
- [x] Probe artifacts retained in `/home/node/.claude/jobs/8d9a5e7c/820m-probe*.mts`
- [ ] CI required checks (cheap gate + test262 shards + quality) should pass
      as a no-op for docs changes

### DIFF
--- a/plan/issues/1694-promise-subclass-capability.md
+++ b/plan/issues/1694-promise-subclass-capability.md
@@ -1,0 +1,115 @@
+---
+id: 1694
+title: "Promise.any/all/allSettled/race: non-Promise capability `this` + extends-Promise codegen (~50 fails)"
+status: backlog
+created: 2026-05-28
+priority: medium
+feasibility: hard
+reasoning_effort: high
+task_type: bugfix
+area: codegen, runtime
+language_feature: promises, subclassing
+goal: spec-completeness
+sprint: Backlog
+related: [1368, 1465, 1528, 1116, 1644, 1682]
+---
+# #1694 — Promise combinators: non-Promise capability `this` + extends-Promise codegen
+
+## Problem
+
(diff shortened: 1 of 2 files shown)


ttraenkler:
Superseded by PR #826 which cleanly cherry-picks the 1694 and 820m issue files. This PR has merge conflicts and is no longer needed.

PR #830, opened 2026-05-28
docs(#1694): investigate Promise subclass capability — escalate to architect

ttraenkler:
## Summary

Senior-developer investigation pass for #1694 (Promise.any/all/allSettled/race subclass/capability bugs, ~50 fails).

**Verdict: NEEDS-ARCHITECT-SPEC, blocking on #1596 + #1632b + #1682-Half-B. No Promise-layer patch moves the needle.**

## Findings

Probed each fingerprint against current main (JS-host mode, real `compile()` + `WebAssembly.instantiate()`). The issue's original two-sub-cluster decomposition is partially incorrect; failures span three distinct root causes, only one of which is in any way Promise-layer.

| Original framing | What I actually found |
|---|---|
| Sub-cluster B: ~28 fails — `class X extends Promise; X.METHOD(…)` emits `extern.convert_any[0]` invalid Wasm | **Already works on main.** `resolvePromiseSubclassThisArg` (calls.ts:717) routes via `__promise_subclass_ctor`. Probe `MyPromise.{any,all,allSettled,race}([…])` compiles + instantiates 5/5. The "invalid Wasm" claim was a probe artifact (`as any` TS cast breaking identifier detection). |
| Sub-cluster A: ~40 fails — non-Promise capability `this` | **Decomposes into two unrelated gaps:** |
| A.i: `Promise.X.call(fn, [])` ~24 | wasm-fn `_wrapForHost` not [[Construct]]-able by V8 — **same root as #1596 / #1632b** |
| A.ii: `Promise.X.call(declaredSubclass, [])` ~20 | `__promise_subclass_ctor` returns a *bodyless* `class extends Promise {}`; user constructor never invoked (callCount stays 0, executor not captured) — **same root as #1682 Half B / #1632b** |

Both A.i and A.ii require the same upstream architect decision: how does a compiled-Wasm function/class get exposed to the host as a callable AND [[Construct]]-able value whose `[[Call]]`/`[[Construct]]` invoke the user's compiled body?

## Changes

- `plan/issues/1694-promise-subclass-capability.md` — appended investigation section with corrected decomposition, root-cause analysis (with concrete code refs to `src/codegen/expressions/calls.ts:4916-4932`, `src/runtime.ts:6386-6405`), and the architect-spec recommendation. Updated frontmatter: `needs_architect_spec: true`, `updated: 2026-05-28`, `investigation_done: 2026-05-28`, added related: 1596, 1632b.

## Test plan

- [x] Probes reproduce all three failure modes on current main (gitignored as `tests/probe-1694*.test.ts`; shapes documented in the issue file for verbatim reproduction)
- [x] Sub-cluster B negative — `MyPromise.METHOD([…])` direct-subclass form passes (no `as any`)
- [x] No source code changes; docs-only PR

### DIFF
--- a/plan/issues/1694-promise-subclass-capability.md
+++ b/plan/issues/1694-promise-subclass-capability.md
@@ -1,0 +1,251 @@
+---
+id: 1694
+title: "Promise.any/all/allSettled/race: non-Promise capability `this` + extends-Promise codegen (~50 fails)"
+status: backlog
+created: 2026-05-28
+updated: 2026-05-28
+investigation_done: 2026-05-28
+priority: medium
+feasibility: hard
+reasoning_effort: high
+task_type: bugfix
+area: codegen, runtime
+language_feature: promises, subclassing
+goal: spec-completeness
+sprint: Backlog
+related: [1368, 1465, 1528, 1116, 1644, 1682, 1596, 1632b]
+needs_architect_spec: true
+---
+# #1694 — Promise combinators: non-Promise capability `this` + extends-Promise codegen
+
+## Problem
+
+Across `Promise.any`, `Promise.all`, `Promise.allSettled`, `Promise.race`, ~50
+test262 cases fail with two distinct error fingerprints that share the same
+underlying gap: the **NewPromiseCapability(C)** step of each combinator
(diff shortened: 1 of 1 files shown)


ttraenkler:
Superseded — plan/issues/1694-promise-subclass-capability.md already committed to main in batch docs commit c58338348.

PR #834, opened 2026-05-28
fix(#1698): native ArrayBuffer.prototype.slice in --target wasi

ttraenkler:
## Summary

Under `--target wasi`, `ab.slice(begin, end)` compiled but produced an
`illegal cast` trap at runtime: the extern-class dispatch routed the
call to a missing host import, leaving a degraded path that dropped the
args and emitted `ref.null extern` for the result. Any subsequent
struct-field read (e.g. `new Uint8Array(sliced)`) then trapped on the
`ref.cast`.

Mirrors the #1654 dual-mode pattern: a Wasm-native emitter in
`src/codegen/dataview-native.ts` copies bytes between i32_byte vec
structs, applying ECMAScript §25.1.5.3 normalisation
(negative-offset → clamp from end, default-end → byteLength,
out-of-bounds → clamp). Hook in `src/codegen/expressions/calls.ts`
under the existing `noJsHost(ctx)` DataView block, gated on receiver
symbol === \"ArrayBuffer\" and method === \"slice\". JS-host mode
remains untouched.

## Test plan

- [x] `tests/issue-1698.test.ts` — 8/8 pass: compile-no-trap, basic
      slice, default-end, negative-begin/end, OOB-clamp, empty-when-
      begin>=end, source-independence.
- [x] `tests/wasi.test.ts` — 24/24 unchanged green.
- [ ] CI: required checks (cheap gate, test262 shards, equivalence gate, quality).
- [ ] \`/dev-self-merge\` on CI completion.

Closes #1698.

### DIFF
--- a/plan/issues/1629-spec-gap-object-defineproperty-descriptor-attributes.md
+++ b/plan/issues/1629-spec-gap-object-defineproperty-descriptor-attributes.md
@@ -158,3 +158,23 @@ Recommend splitting into sub-issues:
   fails) — likely overlaps #1130.
 
 No code change landed under this task; needs architect spec before implementation.
+
+## Partial fix #1629b (2026-05-28)
+
+Sub-cluster fixed: `Object.getOwnPropertyDescriptor` attribute readback
+for plain-object struct fields that were redefined via
+`Object.defineProperty`. Root cause: the GOPD fast path in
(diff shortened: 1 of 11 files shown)


ttraenkler:
Duplicate of #827 (already merged at db1503444). Branch was re-pushed after the merge, producing this empty PR.

PR #844, opened 2026-05-28
fix(#1594A): closure-aware __typeof for hoisted block fns (Slice A)

ttraenkler:
## Summary

Slice A of #1594A (AnnexB legacy block-function hoisting). The runtime
`__typeof` host helper at `src/runtime.ts:7038` returned `"object"` for
Wasm closure structs because they are externref-typed in JS and lack a
`[[Call]]` internal method. Per the architect spec in
`plan/issues/1594-annexb-strict-function-code-tdz-referenceerror.md`
(*typeof defect* section), this is the dominant root cause of the ~202
annexB function-code failures — most use `assert.sameValue(typeof f,
\"function\")` as their oracle.

Fix: when the operand is a JS object, consult the `__is_closure` export
(emitted alongside every closure type, same authoritative discriminator
used by `_maybeWrapCallableUnknownArity` at `runtime.ts:1024`) and report
`\"function\"` when it answers 1. When `__is_closure` is absent (no
closures were emitted in this module) we fall through to the legacy
`typeof v` — strict superset of prior semantics.

## What this does NOT cover

- **Slice B** (the §B.3.3.1 legacy var-binding write at
  `nested-declarations.ts:150`) is a separate follow-up PR. Without
  Slice A landing first, Slice B is a no-op for the test oracle.
- **`__any_typeof`** (wasm-generated, in `any-helpers.ts`) is only
  reached under `ctx.fast` mode and operates on Any-tagged structs; its
  gcref arm (tag 6) returns `\"object\"` without closure detection.
  Adding `__is_closure` plumbing there needs new codegen wiring and is
  deferred — the default-mode test262 oracle goes through the host
  `__typeof` path patched here.

## Test plan

- [x] Reproduced the spec's minimal repro: `function outer() { { function f(){}; } return typeof f; }` returned `\"object\"` pre-fix.
- [x] `tests/issue-1594a.test.ts` — 5 cases (block-hoisted fn, externref-erased fn, plus regression guards for plain object / null / array). Pre-fix: 2 failures. Post-fix: 5/5 green.
- [x] Typeof-related equivalence suites pass: `typeof-comparison.test.ts`, `typeof-narrowing.test.ts`, `typeof-member-expression.test.ts`, `symbol-typeof.test.ts` — 14/14 green.
- [x] Typecheck clean (`tsc --noEmit`).
- [x] Pre-push hook (typecheck + lint) passed.
- [ ] CI will run full equivalence-gate, lint, and test262 — measure annexB/language/function-code lift from `pass=108, fail=202, ce=2` baseline (spec target: `pass≥290`).

### DIFF
--- a/src/runtime.ts
+++ b/src/runtime.ts
@@ -7035,7 +7035,25 @@ assert._isSameValue = isSameValue;
       if (name === "__call_2_f64") return (fn: Function, a: number, b: number) => fn(a, b);
       if (name === "__call_1_i32") return (fn: Function, a: number) => fn(a);
       if (name === "__call_2_i32") return (fn: Function, a: number, b: number) => fn(a, b);
-      if (name === "__typeof") return (v: any) => typeof v;
+      if (name === "__typeof")
+        return (v: any) => {
+          // (#1594A Slice A) Wasm closure structs are externref-typed in JS
+          // and have no `[[Call]]`, so the bare `typeof v` returns "object".
+          // Per spec they should report "function". `__is_closure` (emitted
+          // alongside any closure type) is the authoritative discriminator,
+          // identical to the one used by `_maybeWrapCallableUnknownArity`.
+          if (v != null && typeof v === "object") {
+            const isClosureFn = callbackState?.getExports()?.__is_closure as ((x: any) => number) | undefined;
+            if (typeof isClosureFn === "function") {
+              try {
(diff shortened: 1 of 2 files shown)


ttraenkler:
Closing duplicate — PR #843 (same fix, same runtime.ts patch) is already fully green on CI and will merge first. The typeof closure fix will land via #843.

PR #847, opened 2026-05-28
fix(#1594A) Slice B: §B.3.3.1 legacy var-binding write for block-hoisted fns

ttraenkler:
## Summary

Implements AnnexB §B.3.3.1 legacy var-binding semantics for block-nested
sloppy-mode `function` declarations. After the block executes, outer
reads of the name resolve to the (now-bound) closure value rather than
the funcMap funcref wrapper.

Companion to Slice A (PR #844, closure-aware `__typeof`).

## What changes

- **`src/codegen/context/types.ts`** — adds `fctx.annexBLegacyLocals?: Map<string, number>`.
- **`src/codegen/statements/nested-declarations.ts`** — Phase A: in
  `hoistFunctionDeclarations`, for every block-nested FunctionDecl that
  hoists successfully, allocate an outer-fctx externref local and record
  it. Eligibility gates per §B.3.3 NOTE 1/2: skip generator/async/strict,
  skip when name is already in localMap (param or earlier hoist), skip
  when hoist failed. Also fix recursion to descend into brace-less
  `if (cond) function f(){}` (§B.3.2 form).
- **`src/codegen/statements.ts`** — Phase B: in compileStatement's
  FunctionDeclaration branch, when the decl matches a registered legacy
  local, emit `emitFuncRefAsClosure → extern.convert_any → local.set`
  so each pass through the declaration rebinds the legacy local (spec
  §B.3.3.1 step 2.c).
- **`src/codegen/expressions/calls.ts`** — `isLocallyShadowed` gate
  OR-s in `annexBLegacyLocals?.has(name)` so call sites for legacy-bound
  names route through the local rather than the funcMap direct path.
  The no-capture case is handled by the existing callable-fallback
  path, which recovers the closure via `any.convert_extern +
  emitGuardedRefCast` (null-safe — calls before the block ran produce
  null → guarded cast → null → emitNullCheckThrow → TypeError).

## Tests

`tests/issue-1594a-slice-b.test.ts` (3 tests, all pass):
- function-init: outer call after block executes (canonical pattern)
- function-update: loop re-binding observes latest closure
- brace-less if-then: `if (cond) function h(){}` hoists correctly

## Test plan

- [x] typecheck + lint
- [x] tests/issue-1594a-slice-b.test.ts: 3/3 pass
- [x] tests/issue-1594a.test.ts (Slice A): pass
- [x] tests/ts-wasm-equivalence.test.ts: pass
- [x] tests/tdz-reference-error.test.ts: pass
- [ ] CI test262 — gauges the actual AnnexB function-code/global-code bucket movement

## Notes

- Pre-existing `helpers.js` import errors in some test files are unrelated to this PR (they fail on `origin/main` HEAD too — `tests/helpers.ts` doesn't exist).
- Three stale `issue-1592-*` worktrees noted in tech-lead msg — not touched.

### DIFF
--- a/src/codegen/context/types.ts
+++ b/src/codegen/context/types.ts
@@ -295,6 +295,18 @@ export interface FunctionContext {
       materializedLocalIdx: number; // ref_null $AnyString — reserved for future cache
     }
   >;
+  /**
+   * #1594A Slice B — §B.3.3.1 AnnexB legacy var-binding writes for
+   * block-hoisted function declarations. Maps the function name to an
+   * outer-fctx externref local index allocated during hoisting. When the
+   * statement is later visited in textual order, the FunctionDeclaration
+   * compile path emits a runtime `localSet(idx, closureExternref)` so that
+   * outer reads of the name (after the block has executed) see the bound
+   * value rather than the funcMap funcref wrapper. Sloppy-mode only;
+   * strict-mode declarations and would-be-early-error cases are NOT
+   * registered here (see `hoistFunctionDeclarations` Phase A gates).
+   */
+  annexBLegacyLocals?: Map<string, number>;
 }
 
 /** Context shared across all codegen. */
--- a/src/codegen/expressions/calls.ts
+++ b/src/codegen/expressions/calls.ts
@@ -7622,16 +7622,7 @@ function compileCallExpression(ctx: CodegenContext, fctx: FunctionContext, expr:
     // emit cap-prepend logic are already correct, even if a coincidental
     // local with the same name exists in the current scope.
     let isLocallyShadowed = false;
-    if (
-      fctx.localMap.has(funcName) &&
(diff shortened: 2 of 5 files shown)


ttraenkler:
Closing per team-lead direction (2026-05-28). PR introduced 674 wasm-change regressions; the issue (`localMap` > `funcMap` priority making any same-named externref local shadow the funcref slot) is architectural, not a local fix.

Suspended Work doc (key finding, two viable fix directions, secondary writeback-direction defect, resume steps) appended to `plan/issues/1594-annexb-strict-function-code-tdz-referenceerror.md`. Branch `issue-1594a-slice-b` is abandoned; re-attempts should branch fresh from `origin/main` after an architect picks fix direction (a) funcref-first when local is null, or (b) sidecar shadow outside `localMap`.

Slice A (PR #844) is unaffected and stays merged.

PR #856, opened 2026-05-28
docs(#1690): root-cause acorn isInAstralSet validator failure — two defects

ttraenkler:
## Summary

Investigation only — no code change. Documents the root cause of acorn.mjs validator failure for the team to fix.

Reproduced on current main: `WebAssembly.compile()` on the acorn.mjs binary fails with `function isInAstralSet failed: f64.lt[0] expected type f64, found global.get of type (ref null 57)`.

## Root cause

Two compounding defects, both required:

1. **`var i` inside a nested function is wrongly hoisted to the module-level `$__mod_i` global.** `walkModuleStmtForVars` in `src/codegen/declarations.ts:2873` correctly does NOT recurse into FunctionDeclarations / methods / arrows — but identifier resolution then consults `moduleGlobals` and short-circuits the inner-function `var i`, instead of allocating a fresh function-local.
2. **The for-loop CONDITION read of `i` is not shifted by `fixupModuleGlobalIndices`** even though the init (`global.set 2474`) and increment (`global.get/set 2474`) writes ARE shifted. The loop-condition read stays at the stale index 2472 → it points at `$__mod_unicodeScriptValues` of type `(ref null 57)` → `f64.lt` rejects.

## WAT evidence (in issue file)

```wat
(func $isInAstralSet ...
  f64.const 0
  global.set 2474              ;; i = 0 — correct
  (block (loop
    global.get 2472            ;; STALE — should be 2474
    local.get 1  struct.get 3 0  f64.convert_i32_s
    f64.lt                     ;; FAIL
    ...
    global.set 2474            ;; i += 2 — correct
```

## Test plan

- [ ] Localize defect #1 in identifier-resolution path; confirm a fix where inner-function `var i` shadows module-level `$__mod_i`
- [ ] Validate defect #2 independently with a focused repro (module var + nested for-loop without name collision)
- [ ] Re-run probe at `.tmp/acorn/probe.mts` after each fix
- [ ] Ensure test262 doesn't regress (esp. #1618 / #1314 / #1677 / #1601 shift-class issues)

Issue file: `plan/issues/1690-acorn-isinastralset-global-array-f64-mismatch.md` (see `## Investigation 2026-05-28 (sendev-1542)` section).

### DIFF
--- a/plan/issues/1690-acorn-isinastralset-global-array-f64-mismatch.md
+++ b/plan/issues/1690-acorn-isinastralset-global-array-f64-mismatch.md
@@ -133,3 +133,67 @@ codegen (also a finding — see Notes).
 - Builds on #1679: that issue's `new this(...)` blocker is resolved on
   `e622751f7`, exposing this validation failure as the next gate to a clean
   acorn compile.
+
+## Investigation 2026-05-28 (sendev-1542)
+
+**Repro confirmed on current main** (after PR #845 merge): `WebAssembly.compile(acorn.mjs binary)` fails with `function #58:"isInAstralSet" failed: f64.lt[0] expected type f64, found global.get of type (ref null 57) @+185628`. The function index and global index moved (`#58` vs the issue's `#56`, `2472` vs `56`) but the symptom shape is identical.
+
+### WAT-level findings (`.tmp/acorn/acorn.wat`)
+
+`isInAstralSet` body (L2840-2939):
+
+```wat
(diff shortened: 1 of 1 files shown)


ttraenkler:
Superseded by PR #862 (fix lands the actual code change). This was a docs-only investigation that documented the root cause — the impl PR captures the same context in its commit message + issue file. Closing to avoid duplicate landing of the issue-file edits.

PR #867, opened 2026-05-28
fix(#1340): function-decl closure-singleton cache for first-class function values

ttraenkler:
## Summary

Mirrors the per-class-method singleton cache from #1394 (`emitCachedMethodClosureAccess`) for the top-level function-declaration case:

- **Without caching**, every textual occurrence of `foo` in value position emitted a fresh `struct.new $closure_struct`, so `foo === foo` was false and sidecar writes on `(foo as any).prototype = X` did not round-trip — struct A from the write site never matched struct B from the read site.
- **With caching**, one lazy-init externref global per captureless top-level function declaration; all later reads return the same externref.

Implements the architect spec at `plan/issues/1340-spec-gap-iterator-helpers-wasm-compile.md`.

## Test262 impact

The runner's `Iterator` shim at `tests/test262-runner.ts:1705-1708` writes `(Iterator as any).prototype = …` and later reads `Iterator.prototype.drop`. Before this PR, the read yielded `undefined` and threw "not a function", which `classifyError` bucketed as **wasm_compile** (regex match on `/not a function/i`). That accounted for ~89 of 245 failures under `built-ins/Iterator/prototype/*`.

After the fix the prototype write round-trips, unblocking those tests for the actual helper-invocation paths (whether each helper succeeds is now a downstream concern, not a misclassified compile error).

## Changes

- `src/codegen/context/types.ts` — new `funcClosureGlobals` map; optional `noThisParam` flag on `pendingMethodTrampolines` entries.
- `src/codegen/context/create-context.ts` — initialize `funcClosureGlobals`.
- `src/codegen/registry/imports.ts` — include `funcClosureGlobals` in the `addUnionImports` shift block (parallels `methodClosureGlobals`).
- `src/codegen/closures.ts`:
  - new `emitCachedFuncClosureAccess` exported helper (mirrors `emitCachedMethodClosureAccess`, no `this`-drop trampoline);
  - `finalizeMethodTrampolines` honors `noThisParam` (skip `sig.params.slice(1)` and the `ref.null` prologue).
- `src/codegen/expressions/identifiers.ts` — call `emitCachedFuncClosureAccess` in the `funcRefIdx` branch for captureless functions; fall through to the legacy per-site `emitFuncRefAsClosure` for captures.
- `tests/issue-1340.test.ts` — 3 focused unit tests.

## Why the existing `closureMap` path isn't enough

`compileIdentifier` already has a `closureMap`/`moduleGlobals` branch (identifiers.ts:661-675), but it is populated only for **functions registered as closures** (those with captures, or hoisted lambdas). Plain top-level `function foo(){}` declarations have no `closureMap` entry, so they fell past that branch directly into the per-site allocation. The cached-singleton pattern from #1394 solves the identity problem precisely and keeps the *call* hot path (resolved via `funcMap` + direct `call funcIdx`) unchanged.

## Edge cases

- **Captures**: skipped — caller falls back to capture-aware `emitFuncRefAsClosure`. Identity is not required for these (they aren't typical `.prototype = X` targets).
- **`__` internal funcs, class constructors**: already excluded by existing prefix / `classSet` checks.
- **Late `funcIdx` re-resolution** (generators, default params, async): trampoline enrolled in `pendingMethodTrampolines` with `noThisParam: true`, so `finalizeMethodTrampolines` rebuilds the body against the function's final signature without trying to drop a non-existent `this`.
- **Global index shifting** under late-import insertion: `funcClosureGlobals` added to `shiftMap` block; mirrors `methodClosureGlobals`.

## Test plan

- [x] `npm test -- tests/issue-1340.test.ts` — 3/3 pass (identity, `.prototype` round-trip, Iterator-shim prototype assignment).
- [x] Closure-adjacent equivalence tests: `tests/equivalence/{generator-methods,symbol-iterator-class,iterator-protocol-custom,issue-799-prototype-chain,array-prototype-methods}.test.ts` — 33/33 pass.
- [x] `pnpm run typecheck` — clean.
- [x] Verified the 2 pre-existing failures in `tests/equivalence/optional-direct-closure-call.test.ts` exist on the baseline before this change (unrelated to #1340).
- [ ] CI test262: expected delta ≥ +60 (the misclassified ~88 minus modest 2nd-order losses).

### DIFF
--- a/plan/issues/1340-spec-gap-iterator-helpers-wasm-compile.md
+++ b/plan/issues/1340-spec-gap-iterator-helpers-wasm-compile.md
@@ -1,9 +1,10 @@
 ---
 id: 1340
 title: "spec gap: Iterator.prototype helpers wasm_compile errors (89 of 245 fails)"
-status: blocked
+status: done
 created: 2026-05-08
 updated: 2026-05-28
+completed: 2026-05-28
 priority: high
 feasibility: hard
 reasoning_effort: high
@@ -14,7 +15,7 @@ goal: spec-completeness
 sprint: 50
 parent: 1328
 related: 1320, 1323
-blocked_on: architect-spec
+resolution: "Architect spec landed and was implemented in PR #867 — function-decl closure-singleton cache (`emitCachedFuncClosureAccess`). Each captureless top-level `function foo(){}` now resolves to a single lazy-init externref global at every value-context read, so `(foo as any).prototype = X` round-trips and the test262 Iterator-shim no longer misclassifies as wasm_compile."
 ---
 # #1340 — Iterator.prototype helper methods: wasm_compile failures
 
--- a/src/codegen/closures.ts
+++ b/src/codegen/closures.ts
@@ -3103,8 +3103,12 @@ export function emitObjectMethodAsClosure(
 export function finalizeMethodTrampolines(ctx: CodegenContext): void {
   for (const t of ctx.pendingMethodTrampolines) {
     const sig = getFuncSignature(ctx, t.methodFuncIdx);
-    if (!sig || sig.params.length === 0) continue;
-    const methodUserParams = sig.params.slice(1);
+    if (!sig) continue;
+    // (#1340) Plain function decls have no hidden `this`; method sigs lead
+    // with `this` at param 0 and need it dropped. The legacy method path
+    // requires `sig.params.length >= 1` because it slices off `this`.
+    if (!t.noThisParam && sig.params.length === 0) continue;
+    const methodUserParams = t.noThisParam ? sig.params : sig.params.slice(1);
     // Only rebuild when the user-param arity is unchanged. The trampoline's
     // OWN func type (its wrapper type) was fixed at registration with
     // `userParamCount` params and is shared/cached, so it cannot change here;
@@ -3166,7 +3170,10 @@ export function finalizeMethodTrampolines(ctx: CodegenContext): void {
       savedBodies: [],
     };
 
-    const newBody: Instr[] = [{ op: "ref.null", typeIdx: t.objStructTypeIdx } as Instr];
+    // (#1340) Function-decl trampolines have no `this` prologue; method
+    // trampolines emit `ref.null <objStruct>` as the receiver before
+    // forwarding user params.
+    const newBody: Instr[] = t.noThisParam ? [] : [{ op: "ref.null", typeIdx: t.objStructTypeIdx } as Instr];
     for (let i = 0; i < methodUserParams.length; i++) {
       newBody.push({ op: "local.get", index: i + 1 } as Instr);
       const from = wrapperUserParams[i];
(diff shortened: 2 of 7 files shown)


ttraenkler:
## ESCALATE — net regression (-23 tests, criteria 4 & 5 violated)

CI on `d5f924b59` ran clean (no failing checks, catastrophic guard 53/200), but the actual delta vs the js2wasm-baselines main baseline is **net-negative**:

- **Regressions (pass → fail/other): 53** — 38 with wasm-hash change, all real
- **Improvements (other → pass): 15** — from sibling main-merges (#1345 Reflect, etc), not from #1340
- **Net per test: ~-23** (acceptance criteria #4 forbids ANY `built-ins/Iterator/prototype` regression — but the regressions are in **Array.prototype**, not Iterator)
- **Spec target: ≥+60 net** — missed by ~80 tests

### Two distinct regression patterns

1. **`Array.prototype.every.call(obj, fn)` → returned 3** (18 tests) — the `.call()` dispatch path doesn't recognise the new struct-ref-returning cached closure, so the arity/coercion of the callback drops. The fast-path fix (`any.convert_extern + ref.cast` returning struct ref) helps the direct `[arr].every(fn)` path but the `.prototype.every.call` host-bridge dispatch still goes through `__call_fn_N` which can't consume the new shape.
2. **Maximum call stack size exceeded** (~25 tests) — annexB `if-decl-else-decl-*-func-existing-fn-no-init.js` + `Array/prototype/filter/15.4.4.20-9-c-ii-9.js`. The cached trampoline `__fn_tramp_${name}_cached` collides somewhere with `funcMap` lookup or `nestedFuncCaptures`, producing an infinite mutual recursion.

### What works

- Local scoped tests pass: `tests/issue-1340.test.ts` (3/3), `tests/issue-820m.test.ts` (6/6), `tests/issue-1394.test.ts`, `tests/issue-1693.test.ts`, `tests/issue-1690.test.ts`, `tests/equivalence/array-filter-obj-length.test.ts` (4/4).
- The 89 misclassified `wasm_compile` errors for `Iterator.prototype` helpers ARE the intended fix target. Spot-checking the report shows those probably did flip — but they're swamped by the new `.every.call` regressions because Iterator.prototype tests are a much smaller bucket than Array.prototype tests.

### Why this needs a respec, not a patch

The architect spec assumed the cached closure's externref read could substitute everywhere the per-site struct.new used to flow. That holds for the direct array-method call_ref path, but **fails for the `Function.prototype.call` host bridge** which extracts the callable through a different routing (per `setupArrayCallback` line 5077: `__call_2_f64` / `__call_2_i32`). The fix needs:

1. A consistent representation that the bridge AND call_ref both recognise.
2. Resolution of the trampoline-name vs funcMap conflict that causes infinite recursion in some hoisted-function scenarios.

Both are codegen-architecture-level decisions, not localized patches. Recommend:
- **DO NOT merge** PR #867 in its current shape.
- Architect respec needed: either (a) the cached closure only fires when *all* downstream consumers can take struct-ref (i.e. only `[arr].method(fn)` direct dispatch, not `.method.call`), or (b) we extend the host-bridge to consume the new shape.

I'm setting the issue back to needs-spec with the diff sample attached.

PR #921, opened 2026-05-29
arch(#1727): internal async-call result-extraction spec

ttraenkler:
Docs-only. Adds the `## Implementation Plan` to #1727 (the equivalence-shard-4 async-function-returns-NaN drift that fails every PR's CI).

## Root cause (pinned via WAT)
`instance.exports.f()` (async, `Promise<number>`) returns raw `42` — the export path calls `$f` directly with its real `() -> f64` signature. But an **internal** `f()` consumed as a primitive (`return f() as unknown as number`) hits the async-call wrap at `expressions.ts:946` → `wrapAsyncReturn`, which boxes the f64 and wraps it in `Promise_resolve(...)` producing a **JS Promise object**. The numeric `return` sink then `__unbox_number`s that Promise object → `Number(Promise{42})` === **NaN**. WAT for `$main`: `f64.const 42; __box_number; Promise_resolve; ... __unbox_number; return`.

## Fix
Extend the existing await-parent skip (`expressions.ts:964-978`, which already returns the raw `T` for `await f()`) to also skip the wrap when the consumer is a non-Promise cast / primitive sink (`f() as any`, `as unknown as number`, etc.). Minimal-diff variant: ~15 lines in the async-call block + a small helper. Forward-compatible with the CPS work (#1042/#1373b) — does not touch `wrapAsyncReturn`/`wrapAsyncCallInTryCatch`.

Feasibility: easy–medium. Mandatory gates: full-CI net ≥ 0, equivalence-shard-4 flips green, zero new async/generator regressions.

### DIFF
--- a/plan/issues/1727-async-function-call-result-nan.md
+++ b/plan/issues/1727-async-function-call-result-nan.md
@@ -1,0 +1,292 @@
+---
+id: 1727
+title: "internal async-function call result wrapped in Promise then unboxed → NaN"
+status: ready
+created: 2026-05-29
+updated: 2026-05-29
+priority: high
+task_type: bugfix
+area: codegen
+language_feature: async-functions, type-coercion
+goal: test262-conformance
+related: [1042, 1373, 1151]
+---
+
+# #1727 — internal async-function call result reads as NaN
+
+## Problem
+
+In the synchronous-wasm async model, calling an async function **internally**
+(a wasm `call`/`call_ref`, not the JS export boundary) and consuming its value
+as a primitive returns **NaN** instead of the resolved value. Fails the 7
+`tests/equivalence/async-function.test.ts` cases (+1 math-pow that depends on
+it) — the pre-existing `equivalence-shard (4)` drift failing every PR.
+
+```ts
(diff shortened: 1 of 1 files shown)


ttraenkler:
Closing as superseded — #921's design-spec content is already fully present on `main`.

The #1727 implementation merged via #924, which also (a) set `plan/issues/1727-async-function-call-result-nan.md` to `status: done` (+`completed`), (b) fixed the `plan/issues/1042-*.md` glob to the real filename `1042-async-await-state-machine-lowering.md` (the #1616 link gate), and (c) recorded the `## Resolution (dev, 2026-05-29)` section with the implemented minimal-diff fix and the #1730 arrow-closure carve-out.

This PR is an older snapshot of the same file from before #924. Merging it would *regress* main: it reverts `status: done` → `ready`, drops `completed`, re-introduces the broken `1042-*.md` glob (fails the link gate), and deletes the recorded resolution. The `## Implementation Plan (architect — 2026-05-29)` section it adds is identical to what main already carries.

Nothing to salvage — no union merge needed. (Resolving TaskList #258.)

PR #937, opened 2026-05-29
docs(#1719): architect spec — intactness-gate for overridden Array @@iterator

ttraenkler:
Docs-only. Adds the `## Implementation Plan` section to `plan/issues/1719-…` specifying the **intactness-gate** approach (option a) for honoring a monkeypatched `Array.prototype[Symbol.iterator]` / `Array.prototype.values` in array destructuring (71 test262 fails).

Spec summary:
- New module-scoped `ctx.arrayIteratorMaybeOverridden` flag (`context/types.ts` + `create-context.ts`).
- Module pre-scan in `index.ts` (modeled on `sourceContainsClass`) sets the flag on any write to `Array.prototype[...]`/`.values` (assignment or `Object.defineProperty`) — conservative, keys on the LHS object not the key expression.
- Gate inserted in `compileArrayDestructuring` (destructuring.ts) and `destructureParamArray` (destructuring-params.ts): when set, coerce the vec/tuple struct RHS to externref and delegate to the existing GetIterator lane (§7.4.2 / §8.5.2). Strings + already-boxed numerics are skipped.
- When the flag is clear (every non-monkeypatching program) codegen is byte-identical to today, so #1016/#1021/#1024/#1025/#1320 do not regress.
- for-of *source* iteration is explicitly scoped out (follow-up), gated on the same flag.

Includes edge cases (override-after-site, conditional override, per-realm, `.values`-only, aliased symbol, string RHS, empty pattern), the 71-file test set + the 5 regression guards, and the exact files/line refs for the dev.

Flips issue `status: blocked` → `ready` (kept `feasibility: hard`).

No compiler source touched.

### DIFF
--- a/plan/issues/1719-array-destructuring-ignores-overridden-array-prototype-iterator.md
+++ b/plan/issues/1719-array-destructuring-ignores-overridden-array-prototype-iterator.md
@@ -1,8 +1,7 @@
 ---
 id: 1719
 title: "Array destructuring ignores overridden Array.prototype[Symbol.iterator] ('items[Symbol.iterator] must be a function', 71 fails)"
-status: blocked
-blocked_on: needs-architect-spec
+status: ready
 created: 2026-05-29
 updated: 2026-05-29
 priority: high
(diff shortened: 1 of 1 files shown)


ttraenkler:
Superseded by #940. The intactness-gate approach in this spec was disproven during implementation (#1719): the externref GetIterator lane can't reach a host Array.prototype override because compiled WasmGC arrays aren't host JS Arrays. #940 replaces it with the $ArrayObj object-value representation (the array sibling of #1732's $FuncObj). Closing to unblock the merge queue; #940 lands the correct spec into the same issue file.

PR #944, opened 2026-05-29
fix+refactor(#389): native-messaging 1 MiB byte-corruption + 3-symbol API + host.imports doc

ttraenkler:
Addresses guest271314's #389 test feedback. Three coherent changes to `examples/native-messaging/`.

## (a) BUG fix — large-payload byte corruption (the core change)

A 1 MiB framed message round-tripped as an array of NULLs (guest's report); under a non-zeroing runtime the same path traps `memory access out of bounds`.

**Root cause:** `process.stdout.write(Uint8Array | ArrayBuffer)` under `--target wasi` stages the bytes into a fixed scratch region (`WASI_WRITE_SCRATCH_START` = 128 KiB, page 2) before a single `fd_write`. The string writer grows linear memory to fit the staging buffer (#1723), but `ensureWasiWriteUint8ArrayHelper` / `ensureWasiWriteArrayBufferHelper` in `src/codegen/index.ts` **lacked that guard** — so any body past the default 3-page reserve overflowed.

**Fix:** add the identical `neededPages = ceil((SCRATCH_START + len) / 65536)` `memory.grow` guard to both byte-array writers before they stage. `≥1 MiB` writes are now byte-exact.

Verified locally: pre-fix a 1 MiB Uint8Array write traps OOB; post-fix the shipped host round-trips 1 MiB byte-for-byte with **zero nulls**.

## (b) REFACTOR — guest's cross-language 3-symbol pattern

`host.ts` now uses `getMessage()` (read 4-byte LE header + exactly N body bytes, return raw `Uint8Array`) / `sendMessage(message)` (LE-frame + stdout write) / `main()` (the `while(1)` getMessage→sendMessage loop). The body flows as **raw bytes** end to end — never lossily stringified — so it's binary-safe at any size and forward-compatible with Chromium's in-progress Uint8Array-over-Native-Messaging ([issues.chromium.org/issues/497341241](https://issues.chromium.org/issues/497341241)). Keeps the strict verbatim echo (#930).

## (c) DOC — `host.imports.js`

New README section: it's the generated JS host-imports glue, **not** needed for the WASI/wasmtime path (`host.wasm` imports only `wasi_snapshot_preview1`); it's for embedding the module in a JS host instead.

## Tests

- `tests/issue-1530.test.ts` gains a **1 MiB byte-exact round-trip regression** driving the shipped host's Uint8Array path (asserts zero nulls/garbage). All 6 issue-1530 tests pass.
- `tests/wasi.test.ts` 24/24 pass; typecheck clean.
- `smoke-test.sh` unchanged (runs under wasmtime in CI).

### DIFF
--- a/examples/native-messaging/README.md
+++ b/examples/native-messaging/README.md
@@ -76,6 +76,26 @@ runs on any standards-compliant WASI preview1 runtime.
 > The `-o` flag is an **output directory**, not a filename. js2wasm names the
 > output after the input basename (`host.wasm`).
 
+### What is `host.imports.js`?
+
+The build also emits `host.imports.js` (plus `host.d.ts` and `host.wat`). It is
+the **generated JS host-imports glue** for the module — a small ES module that
+re-exports `createImports()` / `instantiateBytes()` / `instantiateFromUrl()`
+helpers wired to the right import manifest and string pool for this `.wasm`.
+
+You do **not** need it to run the Native Messaging host: this build targets
+`--target wasi`, so `host.wasm` imports only `wasi_snapshot_preview1` and runs
+directly under wasmtime / wasmer / wazero (or Chrome's native-host launcher)
+with no JS at all — that is the whole point of the standalone WASI target.
+
+`host.imports.js` is for the **other** way to run the module — **embedding it in
+a JavaScript host** (a browser tab, a Node service, a Worker) instead of a WASI
+CLI runtime. There it supplies the `env` / `string_constants` imports the
+js2wasm runtime expects, so you can `import { instantiateFromUrl } from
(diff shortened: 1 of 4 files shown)


ttraenkler:
Superseded by #945 (same #389 native-messaging 1 MiB fix + 3-symbol API + host.imports doc, more complete with issue-file hygiene). This PR reworded the host's stderr debug line ('chars, declared body length' → 'bytes, body length') and dropped its trailing newline WITHOUT updating smoke-test.sh's EXPECTED_STDERR, so the smoke test failed (the 'stray byte? off-by-one' message is the smoke guard's generic wording — it's actually a stderr text mismatch, not a real stray byte). #945 keeps the original console.error wording, so smoke passes. Closing this duplicate.

PR #951, opened 2026-05-29
feat(#1718 S1): Iterator.prototype.flatMap (type-check + runtime polyfill)

ttraenkler:
## Summary

Sprint 57 #1718 **Slice 1** — `Iterator.prototype.flatMap` (the most localized of the iterator-helper failures, mirrors the #1340 map/filter/take/drop infra).

`built-ins/Iterator/prototype/flatMap/*` failed two ways:
1. **Type-check CE** — `it.flatMap(...)` → "Property 'flatMap' does not exist on type 'ArrayIterator'". The checker's lib set omitted `lib.esnext.iterator.d.ts` (which declares the whole helper family).
2. **Runtime** — `flatMap is not a function`: `_installIteratorHelperPolyfills` installed `Iterator.zip/zipKeyed/concat` but not `flatMap`.

## Fix (two parts)

- `src/checker/index.ts`: add `lib.esnext.iterator.d.ts` — unblocks the whole Iterator-helper family's type-checking (map/filter/take/drop/flatMap + Iterator.concat/zip), not just flatMap.
- `src/runtime.ts`: `Iterator.prototype.flatMap` polyfill mirroring the existing zip/concat helpers (`_makeHelperIterator` + `_getFlattenable`), per §27.1.4.x — `mapper(value, counter)` → GetIteratorFlattenable → yield each inner value before advancing the outer; close the outer on abrupt completion.

## Verification

`tests/issue-1718-flatmap.test.ts` — 5 cases (flatten arrays + strings, skip empty inner, + both type-check assertions). flatMap test262 CE bucket → 0 (was 4); runtime pass 0 → 13/44 locally.

## Guardrail note

The residual `flatMap is not a function` failures are **prototype-chain identity** — a *compiled* iterator's proto must resolve to the polyfilled `%Iterator.prototype%`. That is the #1320 iterator-bridge foundation (`related: [1320]`), intentionally NOT forced into this slice. S2 (`zip`/`zipKeyed`) and S3 (`concat`) remain; their polyfills are already installed and their residuals are dominated by the same #1320 chain.

Additive only — no codegen change, no regression risk to existing iterator helpers.

### DIFF
--- a/plan/issues/1718-iterator-sequencing-helpers-concat-zip-flatmap.md
+++ b/plan/issues/1718-iterator-sequencing-helpers-concat-zip-flatmap.md
@@ -81,3 +81,40 @@ dynamic-import/TLA). `flatMap` is shipped ES; prioritise it first. Feasibility
 
 Filed by product-owner test262 triage 2026-05-29 against main baseline
 (`.test262-cache/test262-current.jsonl`, 48,117 records).
+
+
+## S1 implementation (landed) — Iterator.prototype.flatMap
+
+Two-part fix:
+
+1. **Type-check** (`src/checker/index.ts`): added `lib.esnext.iterator.d.ts`
+   to the checker's lib set. Without it, every Iterator-helper call
+   (`it.flatMap(...)`, and also `map`/`filter`/`take`/`drop`,
+   `Iterator.concat`/`zip`) CE'd with "Property does not exist on type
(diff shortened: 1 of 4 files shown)


ttraenkler:
Superseded by a clean re-cut (the lib.esnext.iterator.d.ts change regressed the equivalence-gate; dropped it). New PR off current main with the runtime polyfill only.

Issue #986, opened 2026-05-30
Bundling and compiling fails due to require() calls in optimize.ts - move to top-level import declarations


guest271314:
**Describe the bug**

Bundling and compiling fails due to `require()` calls in `optimize.ts` - move to top-level import declarations.

**To Reproduce**
Install

```shell
bun install --trust --no-cache --cache-dir=/media/user/x/.bun-cache https://github.com/loopdive/js2 
```

Try to bundle to single, standalone script with `bun build`

```shell
bun build --trust --verbose --no-cache --cache-dir=/media/user/x/.bun-cache --target=node --compile --outfile=js2wasm ./node_modules/@loopdive/js2/src/cli.ts
250 |     binaryen = require("binaryen");
                             ^
error: This require call is not allowed because the transitive dependency "node_modules/binaryen/index.js" contains a top-level await
    at /home/user/native-messaging-js2wasm/node_modules/@loopdive/js2/src/optimize.ts:250:24
```
**Expected behavior**
When `require()` calls are substituted for top-level import decarations and assignments

```typescript
const binaryen = await import("binaryen"); 
const fs = await import("node:fs");
```

```shell
bun build --cache-dir=/media/user/x.bun-cache --target=node --outfile=js2wasm.js ./node_modules/@loopdive/js2/src/cli.ts
Bundled 129 modules in 1197ms

  js2wasm.js  33.67 MB  (entry point)

```

```shell
DENO_COMPAT=1 DENO_CACHE_DIR=/media/user/x/.deno-cache deno compile -A --no-remote --exclude=node_modules --exclude=package.json --cached-only js2wasm.js -o js2wasm
Compile js2wasm.js to js2wasm

Embedded Files

js2wasm
└── js2wasm.js (32.11MB)

Files: 32.11MB
Metadata: 1.73KB
Remote modules: 12B


```

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

ttraenkler:
Thanks for the detailed report — the root cause is exactly what you identified, and it's fixed on `main` (#992).

The optimizer no longer `require()`s binaryen. The problem was the mismatch you hit: binaryen ships a top-level `await`, so it can only be evaluated asynchronously, and a synchronous `require()` can't load an async module — which is why `bun build --compile` rejected it. The fix loads it with `await import("binaryen")` instead, which a bundler *can* follow and embed. As part of that, the compile pipeline became async.

[...]

PR #988, opened 2026-05-31
fix(#1756): route optimize.ts bare require() through createRequire shim (GH #986)

ttraenkler:
Resolves GitHub #986 (guest271314): `bun build --compile` / `deno compile` of js2wasm fail on `src/optimize.ts:250 require("binaryen")` because binaryen's transitive deps use top-level await.

**Root cause:** two bare `require()`s in optimize.ts (binaryen + node:fs). binaryen is an *optional* dep in the *synchronous* compile path, so it can't be a static import and shouldn't force the public compile API async.

**Fix:** the file already loads node: builtins via a bundler-opaque `process.getBuiltinModule("node:module").createRequire(...)` shim (bundlers can't statically follow it). Extracted that as `getNodeRequireSync()` and routed both bare requires through it. Sync API preserved, dependency stays optional.

**Verified:** no bare `require()` left, `tsc` clean, `--optimize` still runs binaryen (sample `-O3` 272 B vs 412 B unoptimized → wasm-opt active).

**Note:** unblocks the *build*; a standalone `--compile` binary resolves binaryen at runtime (graceful skip if absent). Embedding binaryen in a single-file binary needs the async-API migration — separate enhancement, noted in #1756.


ttraenkler:
Superseding: closing in favor of #992 (the #1757 async-compile migration), which merged to main at 07:59 UTC and **includes the same `src/optimize.ts` createRequire-shim fix** this PR carries (routing the synchronous binaryen load through `process.getBuiltinModule("node:module").createRequire` so bundlers don't statically follow it). #992 additionally makes the optimizer load via `await import("binaryen")` on the async path, which is what lets a `bun build --compile` / `deno compile` standalone binary actually *embed* binaryen — the full #986 end-state. So #1756's fix is delivered on main via #992 and this PR is redundant. Closing without merge to avoid a duplicate/conflicting change to optimize.ts.

ttraenkler:
Closed as superseded by #992 (the #1757 async-compile migration, merged to main 07:59 UTC). #992 includes the same `src/optimize.ts` createRequire-shim fix this PR carried, plus the async `await import("binaryen")` path that lets a standalone `bun build --compile` / `deno compile` binary embed binaryen — the full #986 end-state. #1756's fix is therefore on main via #992; merging this too would be a duplicate/conflicting optimize.ts change.

PR #999, opened 2026-05-31
feat(#1761): self-healing merge-queue watchdog (auto-recover from merge_group dispatch wedge)

ttraenkler:
## What

A self-healing watchdog for the GitHub merge queue. The queue periodically wedges — PRs sit in `AWAITING_CHECKS` with **zero `merge_group` runs dispatched**, no githubstatus.com incident (~4× in 24h). The one proven recovery is a ~10-min disable of the `merge_queue` ruleset rule, then restore. This automates detect + recovery so the wedge self-heals without human intervention.

Closes #1761.

## Files

- **`scripts/merge-queue-watchdog.mjs`** — detect + recover.
  - **DETECT** (conservative, 3 ANDed clauses): oldest queue entry is `AWAITING_CHECKS`, enqueued > `WEDGE_THRESHOLD_MIN` (default 20) min ago, **and** no `merge_group` Actions run was *created* since that `enqueuedAt`. The last clause distinguishes a provably-dead dispatcher from a slow-but-live one — we never reset a healthy queue.
  - **RECOVER** — prefers the **ruleset reset** (back up ruleset 16700772 → PUT with `merge_queue` rule removed → `sleep 600s` → restore the captured original verbatim → re-enqueue green PRs). The script restores the *captured backup*, never reconstructs params, so recovery can't drift the queue config. **Falls back** to draining the stuck green head via `gh pr merge <N> --admin --merge` (logged gate bypass) when no admin token is available.
  - **Idempotent / rate-limited**: aborts if the `merge_queue` rule is already absent (a reset is mid-flight) or if the ruleset `updated_at` is within the last 30 min (cooldown). `DRY_RUN=1` supported.
- **`.github/workflows/merge-queue-watchdog.yml`** — schedule every 15 min + `workflow_dispatch` (with a `dry_run` input); `concurrency: { group: mq-watchdog, cancel-in-progress: false }` (single-flight across the 10-min sleep); `timeout-minutes: 25`.
- **`docs/runbooks/merge-queue-wedge.md`** — symptom, auto-watchdog, manual fallback, and the GitHub Support recommendation.
- **`plan/issues/1761-...md`** — issue with full rationale + implementation notes.

## Auth (the crux)

Editing ruleset 16700772 needs **`Administration: write`**, which the default `GITHUB_TOKEN` lacks. The workflow uses `secrets.MQ_ADMIN_TOKEN || secrets.AUTO_ENQUEUE_TOKEN || secrets.GITHUB_TOKEN`, and the script **auto-detects** capability at runtime (a no-op re-PUT of the unchanged ruleset is the probe; 403 → fall back to the admin-merge drain). **Recommendation:** wire a fine-grained PAT/App token with `Administration: write` as `MQ_ADMIN_TOKEN` so the clean reset path is always available — the fallback only drains one head and doesn't fix the processor.

## Validation

- `DRY_RUN=1 node scripts/merge-queue-watchdog.mjs` against the live queue prints correct decisions (verified against both an empty queue and a 3-entry recovering queue: `DETECT: healthy — oldest entry only N.Nm old (< 20m threshold)`).
- 6/6 detect decision branches asserted (empty / building / young / slow-but-alive / dead-dispatcher / no-runs-ever).
- Prettier-clean; node `--check` passes.

## Root cause — maintainer action

The watchdog is **recovery, not a fix**. The recurring `merge_group` non-dispatch with clean config and no incident is GitHub-side; the issue recommends filing a **GitHub Support ticket** (needs the org's support portal — I cannot file it).

## Do not enqueue

Per the task, this PR is **not** enqueued — and the irony is the queue may currently be wedged. The maintainer is resetting in parallel and will land this.

### DIFF
--- a/.github/workflows/merge-queue-watchdog.yml
+++ b/.github/workflows/merge-queue-watchdog.yml
@@ -1,0 +1,59 @@
+name: Merge-queue watchdog
+
+# WHY: GitHub's merge queue periodically wedges — PRs sit in AWAITING_CHECKS and
+# ZERO merge_group runs dispatch, with no githubstatus.com incident (#1761, ~4x
+# in 24h). The one proven recovery is a ~10-minute disable of the merge_queue
+# rule in the branch ruleset, then restore (see
+# .claude/memory/feedback_wedged_merge_queue_reset). This watchdog detects the
+# wedge and performs that reset automatically, idempotently and rate-limited.
+#
+# The recover path needs Administration:write to edit the ruleset. If MQ_ADMIN_TOKEN
+# is wired it takes the clean reset path; otherwise it falls back to draining the
+# stuck green head with `gh pr merge --admin --merge` (a logged gate bypass).
+
+on:
+  schedule:
+    # every 15 min — must be > the script's 600s sleep + API time, single-flight
+    # concurrency below prevents overlap if a recovery run is still sleeping.
+    - cron: "*/15 * * * *"
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: "Detect + log decision only, no mutation"
+        type: boolean
+        default: false
+
+permissions:
+  contents: write
+  pull-requests: write
+  actions: read
+  checks: read
+
+concurrency:
+  group: mq-watchdog
+  cancel-in-progress: false
+
+jobs:
+  watchdog:
+    runs-on: ubuntu-latest
+    timeout-minutes: 25 # recover path sleeps 600s; leave margin for API + restore
+    steps:
+      - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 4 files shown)


ttraenkler:
Closing — the maintainer opted to disable auto-enqueue (the idle-PR sweeper) entirely rather than build self-healing around the merge queue. Auto-enqueue churn was the suspected wedge *trigger*, so removing it is the empirical test; if wedges still recur, the manual 10-min ruleset reset is documented (feedback_wedged_merge_queue_reset memory). This also frees issue #1761 for the string-hash presize carve (PR #1000). Watchdog code/runbook remain on branch issue-1761-merge-queue-watchdog if ever needed.

PR #1088, opened 2026-06-03
feat(#1574): §2.4 sound monomorphize clone re-inference

ttraenkler:
## #1574 §2.4 — Re-infer resultType in monomorphize clones (sound subset)

`src/ir/passes/monomorphize.ts`. Relaxes the pass's "body must not consume any parameter as an operand" gate so param-consuming single-block callees become eligible for cloning. `cloneWithParamTypes` now re-types the clone body under the substituted params via `reinferCloneBody` → `reinferInstr` → `reinferBinary`/`reinferUnary`, and **abandons** the clone (the call-site group stays on the original callee) whenever any param-tainted instruction can't be soundly re-typed.

### Soundness model (this narrows the spec's §2.4 estimate)
Wasm arithmetic/comparison ops consume their operands' **exact** ValType off the stack — there's no implicit numeric coercion at the IR→Wasm boundary. So cloning `add(a,b){return a+b}` for an i32 tuple would leave `f64.add` consuming i32 operands → invalid Wasm, unless an `f64.convert_i32_s` is inserted (a separate operand-coercion layer, **future work**). The re-inferer therefore requires exact operand-storage match and abandons any clone that would change an arithmetic operand's storage type. **Strictly additive: never emits invalid Wasm.**

### What the sound subset unlocks
- identity-like bodies (param flows only to return/call) — already cloneable; re-inference is a no-op;
- **reference-polymorphic ops** (`ref.is_null`, `select` over a shared ref supertype) — a param-consuming body using those can now be cloned per call-site *reference* type. This is the genuine new capability.

### Tests
`tests/ir/phase3c.test.ts`:
- negative: `double(n)=n+n` called with f64+i32 tuples → i32 clone abandoned (would retype an f64 operand), call site stays on original;
- positive: `isNull(x)=ref.is_null(x)` called with externref + struct-ref tuples → one valid clone, result stays i32, body verifies.

IR fallback gate unchanged (this pass affects emitted-Wasm quality, not selection). Typecheck clean. The pre-existing `__box_number` LinkErrors in the local inline-small/passes/ir-*-equivalence suites are environment-related (no host import object supplied) and fail identically on the base — confirmed via `git stash`; this change adds zero new failures.

Full implementation notes + follow-ups (operand-coercion layer; closure-typed-param straggler; §2.5-already-done) recorded in `plan/issues/1574-ir-improvement-spec.md`. The issue stays `status: ready` (it's a multi-item research backlog, not a single deliverable).

### DIFF
--- a/plan/issues/1574-ir-improvement-spec.md
+++ b/plan/issues/1574-ir-improvement-spec.md
@@ -9,6 +9,7 @@ sprint: Backlog
 owner: architect
 related: 1131, 1167a, 1167b, 1167c, 1126, 1231, 1169, 1370, 1373
 ---
+
 # IR Type Analysis & Optimization Pass Improvements
 
 A research document scoping concrete, file/line-anchored improvements to the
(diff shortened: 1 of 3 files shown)


ttraenkler:
Parking #1574 per tech-lead decision (Option 1). The §2.4 monomorphize re-inference here is correct, tested, and strictly additive, but its claim-back is ~0 until two blockers land: (1) an operand-coercion insertion layer in clones (to unlock the numeric-helper-across-f64/i32 case the spec estimated), and (2) closure-typed-param support in the IR selector. The implementation + soundness analysis is preserved on branch issue-1574-ir-type-analysis and documented in plan/issues/1574-ir-improvement-spec.md. Re-open / re-cut when a blocker is resolved.

PR #1208, opened 2026-06-05
fix(#1890): re-resolve dstr-rest-param fallback funcIdx after late-import shift

ttraenkler:
## Summary

Fixes #1890 — the largest non-owned standalone-lane compile bucket in the
2026-06-05 harvest (**~1,142** failures):

```
invalid Wasm binary … i32.trunc_sat_f64_s[0] expected type f64, found call of type externref
```

### Root cause (#1839 defect class)

The standalone array-destructuring iterator fallback in
`src/codegen/destructuring-params.ts` captured `fbLenFn` / `fbGetIdxFn`
funcIdx integers **before** registering `__array_from_iter_n`. Under
`--target standalone` that registration adds a **new** `env::` import (the
helper has no native object-runtime impl and is not in
`OBJECT_RUNTIME_HELPER_NAMES`), which **shifts every later function index**.
`flushLateImportShifts` rewrites `fctx.body`, but not these escaped integer
captures — so the emitted `call fbLenFn` targeted the wrong
(externref-returning) function and `i32.trunc_sat_f64_s` received externref
instead of f64, producing invalid Wasm.

### Fix

Re-read the three fallback helper funcIdxs (`__extern_length`,
`__extern_get_idx`, `__array_from_iter_n`) from `ctx.funcMap` after the late
imports are registered, so the `call` ops target the post-shift indices.
Pure dev-lane destructuring-param codegen — does not touch
`object-runtime.ts` or the #1472 family.

## Scope — 1-of-3 in the dstr-rest-param standalone cluster

This PR ships only the **trunc_sat** slice. The cluster has two siblings,
routed separately:
- **#1** sibling `__str_flatten` late-import-shift → routed to sd-1886.
- **#3** gen-method `array.set` dstr coercion gap → follow-up (mine, next).

So this is a **partial** delta on the 1,142 bucket; CI's full standalone run
measures the actual landed gain.

## Test

`tests/issue-1890-dstr-trunc-sat.test.ts` (2 tests, both pass): a dstr param
over an `any`-typed value compiles to a valid standalone module, and the
`trunc_sat_f64_s expected f64` signature does not appear.

### DIFF
--- a/plan/issues/1890-standalone-dstr-rest-param-stale-funcidx-trunc-sat.md
+++ b/plan/issues/1890-standalone-dstr-rest-param-stale-funcidx-trunc-sat.md
@@ -1,0 +1,77 @@
+---
+id: 1890
+title: "standalone: dstr-rest-param fallback emits invalid Wasm (i32.trunc_sat_f64_s expected f64, found externref) — stale funcIdx after late-import shift (~1,142 fails)"
+status: done
+created: 2026-06-05
+updated: 2026-06-05
+completed: 2026-06-05
+priority: high
+feasibility: medium
+task_type: bugfix
+area: codegen
+language_feature: destructuring-params, late-imports
+goal: standalone-mode
+sprint: 59
+related: [1839, 1602, 1592, 1219]
+---
+# #1890 — dstr-rest-param iterator-fallback uses a stale funcIdx → invalid Wasm
+
+## Symptom
+
+**~1,142 official standalone-lane compile failures** (largest non-owned bucket in
+the 2026-06-05 standalone JSONL harvest):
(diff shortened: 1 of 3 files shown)


ttraenkler:
Closing — superseded by sd-1886's #329 (the general re-resolve-funcIdx-after-deferred-flush mechanism + serializer assertion), which covers this fbLenFn/fbGetIdxFn site, the sibling __str_flatten shift, and the @@toPrimitive late-shift captures in one place. Per tech-lead direction, dropping this narrow per-site version to avoid colliding with #329. The #1890 issue file flips back to ready/blocked-on-#329.

PR #1212, opened 2026-06-05
feat(#1886 Slice B): intraprocedural linear-backed Uint8Array codegen

ttraenkler:
## #1886 Slice B — intraprocedural linear-backed `Uint8Array` codegen

Slice A (escape analysis) is merged. This is **Slice B**: the codegen that backs
a qualifying `new Uint8Array(...)` **local** by a `(ptr,len)` pair in linear
memory instead of a WasmGC vec — `b[i]` → `i32.load8_u`, `b[i]=v` →
`i32.store8`, `b.length` → a `len` local, and `process.std*.{read,write}(b)` →
zero-copy `fd_read`/`fd_write`. GC fallback is intact for every other
`Uint8Array`.

### Root cause of the WIP's wasm-validation failure (fixed)

The WIP failed validation with a void `main` ending in a bare `ref.null extern`
(`expected externref, found i32`). Two distinct bugs:

1. **Allocator funcIdx desync.** The WIP pre-emitted `__lin_u8_alloc` *before*
   `collectAllSourceImports`. `addImport` does **not** shift already-registered
   defined functions, so later imports shifted the import/defined boundary and
   left `call __lin_u8_alloc` pointing at an *import*. The wrong arity made the
   function's net stack delta `-1`; `stackBalance`'s function-level
   `fixBranch(expected=0)` then pushed a `ref.null extern` into the
   empty-block-typed void body to "fill" the perceived missing value → a value
   on the stack at function exit → invalid. **Fix:** emit `__lin_u8_alloc` in
   the deferred-helper zone (after the import scan, before the final
   native-string reconcile + user functions), matching
   `emitToUint32Helper`/`emitDeferredWasiHelpers`.
2. **Predicate mismatch.** The hoist pre-pass skipped the GC `$buf` local on the
   raw Slice-A `safeBindings` set, while the declaration site used the tighter
   Slice-B predicate. A buffer Slice A admitted but Slice B declined was left
   with no storage and pulled in GC extern helpers, desyncing `__str_flatten`.
   Both sites now share `isLocalLinearNewBinding`.

### Slice B scope (strictly intraprocedural + leak-free)

A buffer is linear-backed only if it is (1) a `new Uint8Array(...)` local,
(2) used **only intraprocedurally** (no user-function call argument — that needs
Slice C's signature rewrite), and (3) **allocated at most once per run** (not in
a loop; enclosing fn called ≤1× and never from a loop) — the bump arena has no
reset until Slice D. The arena global is registered on demand, so a WASI module
with no linear buffer is byte-identical to today.

### Validation

- `probe_u8` (single buffer in `main`) is linear-backed and round-trips
  correctly under wasmtime **v44** and **v45** (stdin `0x41` →
  `buf[0]=(0x41+1)&255=0x42` → stdout, rest zero-filled).
- `examples/native-messaging/nm_js2wasm.ts` compiles **byte-identical** to main
  (`cmp` passes). All its buffers escape via helper params (Slice C) or are
  per-frame-in-loop (Slice D), so Slice B correctly declines them; h2h on v44/v45
  is unchanged (`validJSON=true match=true`, 65 frames, flat ~24 MB on v45).
  **The host speedup is Slice C/D, not Slice B.**
- New `tests/issue-1886-slice-b.test.ts` (9 tests): validation, `buf[i]` r/w +
  I/O round-trip, `.length`, loop writes, and the escape/param/loop GC-fallback
  guards. `tests/issue-1886.test.ts` + `tests/wasi.test.ts` green.

Issue file: corrected the overstated "hold the body to beat AS" framing (linear
is the lever, not retention) and added `## Slice C` (interprocedural signature
rewrite — the host win) + `## Slice D` (zero-copy subarray view + arena reset)
sections. #1886 stays `in-progress` (Slices C/D remain).

### DIFF
--- a/plan/issues/1886-linear-backed-uint8array-wasi-io.md
+++ b/plan/issues/1886-linear-backed-uint8array-wasi-io.md
@@ -4,7 +4,7 @@ title: "Linear-backed Uint8Array for WASI I/O buffers (escape analysis) — avoi
 status: in-progress
 sprint: Backlog
 created: 2026-06-04
-updated: 2026-06-04
+updated: 2026-06-05
 priority: medium
 feasibility: hard
 reasoning_effort: high
@@ -21,6 +21,16 @@ related: [1863, 1527, 389]
 the `.wat` shows why, and it points at a concrete, *targeted* optimization.
 
 ## Why AssemblyScript is faster (measured + confirmed from its `.wat`)
+
+> **Framing correction (esch 2026-06-05):** the original issue title/body framed
+> this as "hold the body in memory to beat AS." That is **wrong** and the title
+> is kept only for issue continuity. The team-lead's held-GC-body measurement
+> confirmed that holding the 64 MiB body in a **GC array** is both *slower* and
+> *fatter* than streaming a 1 MiB window — retention is not the lever. **The
+> lever is getting the bytes out of the GC heap and into linear memory so
+> `fd_read`/`fd_write` touch them with zero copy** (and so element ops are
+> `i32.load8_u`/`i32.store8` against that same linear region). We *beat* AS on
+> memory precisely by NOT holding the body — we stream a small linear window.
 
 AssemblyScript uses **linear memory exclusively** — no WasmGC, no `array.new`/
 `array.copy`/`struct.new`. Its message buffer *is* linear memory, so `fd_read`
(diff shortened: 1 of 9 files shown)


ttraenkler:
Closing as superseded by #1202, which landed the same intraprocedural linear-backed Uint8Array (Slice B) implementation on main at 03:31 UTC (impl 9622c53ee, merge c791d7434). This branch was an independent reimplementation authored ~36 min after #1202 had already merged — a cross-session race, not drift (the DIRTY state was an add/add collision on linear-uint8-codegen.ts). main's version passes all the same gates (probe_u8 round-trips on wasmtime v44 and v45; nm-host h2h validJSON/match true; 20 tests green) and carries a more complete test suite and issue doc, so there's nothing unique here to land.

PR #1218, opened 2026-06-05
fix(#1898): __object_create struct.new omits nextSeq — P0 standalone regression (-1,805)

ttraenkler:
## P0 — restores ~1,805 standalone test262 passes

Standalone test262 regressed **28.94% → 24.76% (pass 12,481 → 10,676, −1,805)** with
`compile_error +5,582` immediately after #1196 (native prototype-chain ops) merged.
Default (gc) lane unaffected; standalone isn't a merge-gating check, so it slipped through.

### Root cause
#1837 grew `$Object` to 6 fields (added `nextSeq`). #1196 (commit `c696b1ffc`) added
`__object_create` whose `struct.new $Object` pushed only **5** operands
(`proto, props, 0, 0, 0`) — missing the `nextSeq` push the sibling `__new_plain_object`
already had. Since `ensureObjectRuntime` emits `__object_create` **unconditionally**
(standalone-gated), **every** standalone program touching the open-object runtime — even
`const o:any={}; o["y"]=2;` which never calls `Object.create` — emitted an invalid module:

```
Compiling function #N:"__object_create" failed:
  not enough arguments on the stack for struct.new (need 6, got 5)
```

### Fix
Add the missing `nextSeq: 0` push so `__object_create`'s `struct.new` supplies all 6
fields (matching `__new_plain_object`). One-line fix-forward.

Verified: only two `struct.new $Object` sites exist (`object-runtime.ts:332`
`__new_plain_object`, `:1205` `__object_create`); both now 6-operand. #1211
(`__extern_get` re-route) touches no `struct.new` — ruled out.

### Relationship to #1195
PR #1195 (in/hasOwn, in CI) carries this identical fix bundled with its feature. This
minimal fix-forward restores the −1,805 passes **immediately** rather than waiting on
#1195's larger payload. The two changes are identical/idempotent and merge cleanly.

### Test
`tests/issue-1898.test.ts` (new, 3 tests): bare open-object + `Object.create` both
instantiate (empty imports) without a `struct.new` arity error; default-gc path
unaffected. A stale operand count fails `WebAssembly.compile`, guarding every
`struct.new $Object` site. tsc clean.

### DIFF
--- a/plan/issues/1898-standalone-object-create-nextseq-regression.md
+++ b/plan/issues/1898-standalone-object-create-nextseq-regression.md
@@ -1,0 +1,78 @@
+---
+id: 1898
+title: "standalone REGRESSION: __object_create struct.new omits nextSeq → invalid Wasm for all open-object programs"
+status: done
+created: 2026-06-05
+completed: 2026-06-05
+priority: critical
+feasibility: easy
+reasoning_effort: low
+task_type: bugfix
+area: codegen, runtime
+language_feature: objects, standalone
+goal: standalone-mode, host-independence
+sprint: 59
+related: [1472, 1196, 1837]
+---
+# #1898 — Standalone __object_create struct.new omits nextSeq (P0 regression)
+
+## Symptom
+
+Standalone test262 regressed hard immediately after #1196 (native
+prototype-chain ops) merged: **28.94% → 24.76% (pass 12,481 → 10,676,
+−1,805)**, with `compile_error` jumping **+5,582 (21,854 → 27,436)** and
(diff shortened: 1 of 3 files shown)


ttraenkler:
Closing as redundant: the P0 regression is already fixed on main. Slice 7 PR #1216 (merged at 6c9f3767b) carried the same __object_create nextSeq(#1837) one-liner alongside Object.setPrototypeOf. Verified on current main aaf762478: standalone open-object + Object.create programs now produce valid Wasm. The tests/issue-1898.test.ts regression guard from this PR can be re-landed separately if a dedicated guard is wanted (the #1897 standalone CI gate is the broader safety net).

PR #1274, opened 2026-06-07
feat(temporal): add minimal native Temporal subset

ttraenkler:
## Summary

Adds a minimal native Temporal implementation for issue #661. The compiler now lowers selected `Temporal.PlainDate`, `Temporal.PlainTime`, `Temporal.Duration`, and `Temporal.Now.plainDateISO()` construction, parsing, getters, formatting, equality, and arithmetic directly into Wasm/runtime helpers.

The implementation follows the TC39 Temporal proposal shape for the supported surface while keeping the subset explicit instead of trying to compile the full polyfill in this pass.

## Validation

- `pnpm exec tsc --noEmit --pretty false`
- `pnpm exec vitest run tests/issue-661.test.ts`
- `pnpm exec prettier --check src/codegen/temporal-native.ts src/codegen/expressions/calls.ts src/codegen/expressions/new-super.ts src/codegen/property-access.ts src/runtime.ts tests/issue-661.test.ts plan/issues/661-temporal-api-via-compiled-polyfill.md`
- pre-push hook: typecheck, lint, format check, issue integrity


ttraenkler:
## Shepherd analysis — codegen corruption fixed; regression gate red is the feature's inherent tradeoff, NOT drift

**The bisected defect is fixed** (commit 7507924d6, head e795e6a00). Mechanism: `temporalKindForExpression` probed the local table with `getLocalType(fctx, fctx.localMap.get(name) ?? -1)`. For identifiers with no local slot (builtin namespaces: `Number`, `Math`, `Symbol`, `Object`), index `-1` made `getLocalType` evaluate `fctx.params[-1].type` → TypeError. The exception escaped the Temporal property-access hook into the generic codegen fallback, which compiled the whole member access to `f64.const 0` — corrupting `Number.POSITIVE_INFINITY`, Math constants, `Symbol.iterator`, `Object.create`, etc. in non-Temporal programs. Fix: only probe `getLocalType` when `localMap` actually has the identifier, so the hook declines cheaply and side-effect-free.

**Result**: equivalence-gate green (was 48 regressions), all 8 equivalence shards green, all 114 test262 shards green, all 3 required checks green (cheap gate / merge shard reports / quality). The catastrophic 7,512 pass→fail explosion (CE 1,095 → 14,783) is gone.

**Remaining red: `check for test262 regressions` — net -18 (77 ↓ / 59 ↑). I verified this against the FRESH baselines-repo JSONL (post-fbe31c3a9): identical 77/59, so it is not stale-baseline drift.**

Breakdown of the 77:
- **71 are `test/built-ins/Temporal/**` negative-validation tests** — unique to this PR. On main, `Temporal` is undefined, so any expected-throw test passes spuriously; the minimal subset evaluates instead of throwing (the issue file explicitly defers option bags / validation / edge ordering to follow-up). ~40 are deep ISO-8601 grammar validation (`argument-string-*`, UTC designators, calendar annotations, sub-minute offsets); ~31 are option-bag / range / non-integer / mixed-sign validation.
- 2 `pass → compile_timeout` (TypedArray) — runner-load flake per project rule.
- 4 misc (eval/annexB, RegExp escape, RegExp legacy-accessors, String split) — overlap with PR #1262's simultaneous 4-test failure → shared drift/flake.

The 59 improvements are Temporal tests that now genuinely pass.

**Escalation per dev-self-merge criteria** (net_per_test < 0, single bucket 71 > 50): NOT enqueueing on my own authority. Options for the lead:
1. **Accept the tradeoff** and merge — the PR trades 71 spurious passes for 59 real ones plus a working Temporal subset; the binary regression check is not in the required set.
2. **Require follow-up validation first** — a spec-first ISO-grammar validator in the `__temporal_*_from_string_field` runtime helpers (throw RangeError) would flip ~40 back, turning net to roughly +22 with bucket ~31 (<50), satisfying the self-merge criteria. That is feature work beyond this shepherding fix.
3. Hold the PR.

PR #1303, opened 2026-06-10
fix(scripts): raise maxBuffer on update-issues.mjs git-log spawn

ttraenkler:
`loadIssueHistory` in `scripts/update-issues.mjs` spawns `git log --find-renames --name-status` over all of `plan/issues`. On a full-history checkout that output now exceeds Node's default 1 MiB `spawnSync` buffer, so the script dies with `ENOBUFS` — which crashes the pre-push #1616 issue-integrity gate with an opaque 1 MB rename dump instead of a verdict. (Shallow CI clones stay under the limit, which is why CI never caught it.)

I hit this on a fork with full history; the fix is a `maxBuffer: 256 MiB` on the spawn — ~2 orders of magnitude headroom over today's output.

Verified: `node scripts/update-issues.mjs --check` exits 0 on a full-history clone after the change.


ttraenkler:

PR #1312, opened 2026-06-10
fix(#1919): re-base nativeStrHelperImportBase when late-import shifts repair the native-string helpers

ttraenkler:
Slice 2 of #1919 — the `__str_flatten` sub-bucket (~165 standalone invalid-Wasm test262 binaries: `call[0] expected (ref null N), found i32.const`).

## Root cause
Two shift regimes overlap. When an `ensureLateImport` batch lands, `shiftLateImportIndices` repairs the native-string helper map AND the helper bodies — but did not advance `nativeStrHelperImportBase`. The next `reconcileNativeStrFinalizeShift` computed `added = numImportFuncs - base` over the SAME imports and re-applied the delta: `__str_flatten`'s internal `call __str_copy_tree` ended one slot high (calling itself); dead-import elimination preserved the skew into the final binary.

## Fix
- `shiftLateImportIndices` and `addStringImports`' inline shift re-base `nativeStrHelperImportBase = numImportFuncs` after repairing the helpers (mirrors `addUnionImports`' inline shift since #1677-fast-path). Base stays -1 on the default GC path — host mode is a hard no-op (#618 hazard).
- `ensureNativeStringHelpers` settles any pending late-import batch before baking funcIdx values (same #1919 slice-1 guard as `ensureObjectRuntime`).

## Validation (fork has no CI)
- `tests/issue-1919-strflatten.test.ts` — 3/3 pass (standalone, wasi, host-guard)
- Representative `set-access-of-missing-private-setter.js` now compiles + instantiates + returns 1 (was: invalid Wasm)
- `node scripts/equivalence-gate.mjs`: 3 flagged entries are load flakes — none reproduce in solo runs on the branch; "basic arithmetic" (38s timeout) also fails on main solo. Solo failure sets: main 12, branch 11 (strict subset).
- `tests/issue-1923.test.ts`: end-to-end Uint8Array timeout is pre-existing on main (49s there too) — #1915 territory, not this PR.

Issue table re-measured on main 3b8013d37 and updated in the issue file.


ttraenkler:
Closing as superseded: this branch's fix landed on the ttraenkler/js2 fork (PR ttraenkler/js2#6) and reaches this repo through the consolidated sync-back PR #1313 — keeping both open would double-apply the same commits. The PR itself was opened against this repo by accident (a stale `gh-resolved` default in the fork's git config, since fixed). Follow-up #1919 slices will arrive as individual PRs here per the new workflow.

PR #1325, opened 2026-06-11
fix(#1998,#1997): Array.prototype.join element coercion + toString

ttraenkler:
## Summary

Fixes two related `Array.prototype` string-conversion bugs.

**#1998 — `join` traps "illegal cast" on non-numeric elements.** Raw array
elements flowed straight into the `wasm:js-string` `concat` builtin, which
requires string operands, so any `externref`/any-typed element trapped. Each
element is now stringified per §23.1.3.18 step 7.c/d before `concat`:

- **externref/ref elements** route through a new `__extern_join_str` host
  import: `undefined`/`null` → `""`, else `ToString`, recursing into nested
  WasmGC vecs (so `[[1,2],[3]]` → `"1,2,3"`).
- **f64 vecs** substitute `""` for the undefined/hole sNaN sentinel
  (`0x7FF00000DEADC0DE`), while a *genuine* `NaN` (distinct bit pattern) still
  renders `"NaN"`.
- `number_toString` is ensured on demand for `any[]` receivers (the up-front
  import collection only fired for statically-numeric receivers), and the
  default `","` separator + empty-string constant are registered so no-arg
  `join()` and empty arrays produce a real string instead of a `null` that
  traps in `concat`.

**#1997 — `Array.prototype.toString()` returned `"[object Array]"`.**
§23.1.3.36 delegates to `join` with the default `","` separator. Added
`toString` to `ARRAY_METHODS` and routed it to the same lowering as `join`.

Late-import index shifts are settled (`ensureLateImport` + `flushLateImportShifts`)
before any `funcIdx` is captured, so `concatIdx`/`toStrIdx` stay correct.

## Repros (all now match Node)

| Expression | Before | After |
|---|---|---|
| `([10,9] as any[]).join(",")` | trap | `"10,9"` |
| `[1,undefined,2].join("-")` | trap | `"1--2"` |
| `[1,null,2].join("-")` | trap | `"1--2"` |
| `[1,,3].join(",")` | trap | `"1,,3"` |
| `[1,NaN,2].join("-")` | — | `"1-NaN-2"` |
| `([[1,2],[3]] as any[]).toString()` | `"[object Array]"` | `"1,2,3"` |
| `[1,2,3].toString()` | `"[object Array]"` | `"1,2,3"` |
| `[].toString()` | — | `""` |

## Tests

`tests/issue-1998.test.ts` — 14 equivalence cases (any / undefined / null /
hole / NaN / empty join, nested + empty toString, numeric & string
non-regression). Related suites green: #1215, #1286, array-prototype-methods.

## Out of scope (separate issues)

- Boolean elements still render `1`/`0` (pre-existing i32 storage — confirmed
  on clean upstream).
- `Array(3).join(",")` → `"0,0,0"` instead of `",,"` (the constructor
  materialises dense zeros — #2000).


ttraenkler:

I re-ran this PR against the freshly-promoted baseline (`bcf6ea3e`, 31075/43135 host) and reproduced the diff locally with `scripts/diff-test262.ts`. The regression counts are **identical before and after the baseline refresh**, so this is a genuine regression from this PR's own changes, **net -2 pass**. The gate is correctly blocking the merge until it's fixed.

**What broke:** Array.prototype.toString / Array `[object Array]` string conversion

```
test/built-ins/Array/length/S15.4.2.2_A1.1_T2.js → fail (returned 2 — assert #1 at L16: assert.sameValue(x.toString(), "[object Array]...)
test/built-ins/Array/S15.4.1_A1.1_T2.js → fail (returned 2 — assert #1 at L15: assert.sameValue(x.toString(), "[object Array]...)
test/built-ins/Array/S15.4.2.1_A1.1_T2.js → fail (returned 2 — assert #1 at L15: assert.sameValue(x.toString(), "[object Array]...)
```

PR #1313, opened 2026-06-11
sync: fork work-back — index-shift validation, flush guards, deep-audit issues, ratified specs

ttraenkler:
Syncs back everything developed on the ttraenkler/js2 fork (which merged loopdive/js2 main at cae704dc1, so this diff is the fork's net contribution):

**Compiler fixes (validated locally; full suites green on the fork):**
- `#1923`-class structural work: always-on emit-time index validation inline at encoder write sites (`src/emit/binary.ts`) — converts the recurring `u32 out of range: -1` late-import index-shift class into named, located codegen errors. Measured overhead statistically zero (inline checks; the separate-walk variant cost ~15% and was rejected). 23 new tests.
- `#1919` slice 1: flush pending late-import batches before native defined-function registration (`ensureObjectRuntime` + `addUnionImports` guards) — fixes the `__obj_find` invalid-Wasm bucket (146 standalone test262 rows) whose root cause was an over-shift of freshly-baked helper indices, not a key-type bug.
- `fix(scripts)`: 256 MiB maxBuffer on `update-issues.mjs`'s git-log spawn (ENOBUFS on full-history clones — supersedes #1303 if this merges first).
- `fix(hooks)`: resolve hook scripts via `$CLAUDE_PROJECT_DIR`.

**Ratified architecture specs (plan-only; supersedes #1304 if this merges first):**
- BigInt i64-brand ValType decision (option (a), with the #1919 i64-bucket attribution: it's the index-shift class, not BigInt)
- Per-backend dynamic-value representation (slices G1–G5)
- BackendEmitter legalization boundary (slices L1–L5)

**Issue filings:** standalone-gap review issues, sprint-61 merged-PR code-review follow-ups, a 55-bug deep-audit sweep (#1929–#1983), and the index-shift child issues — renumbered twice to stay collision-free against this repo's namespace (fork-local #1915 → #2029; #1955/#1956 → #1984/#1985). The large deletion count is a truncated `test262-run.log` and regenerated standalone report.

I kept the fork's history linear over your main; this PR is the reconciliation point. Happy to split it if you prefer reviewing the src/ changes separately from the plan/ batch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1330, opened 2026-06-11
feat(#2043): index-shift class hardening — always-on emit validation, flush guards, strflatten rebase

ttraenkler:
Granular sync from the ttraenkler/js2 fork — first of the per-topic splits of #1313.

**Code:**
- Always-on, per-index-space validation inline at the encoder write sites (`src/emit/binary.ts`): every function/type/global/local/tag/table index write is range-checked, turning the recurring late-import index-shift class (#1809/#1839/#1602/#1886/#1666/#1677 history) from an opaque `u32 out of range: -1` encoder crash — or worse, silently-wrong indices — into a named, located codegen error. Inline placement measured statistically-zero overhead (a separate pre-emit walk cost ~15% and was rejected; details in the #2043 issue file).
- Flush pending late-import batches before native defined-function registration (`ensureObjectRuntime`, `addUnionImports`): fixes the over-shift that produced the standalone `__obj_find` (146 test262 rows) and arguments-arity (93 rows) invalid-Wasm buckets.
- `nativeStrHelperImportBase` re-base when shifts repair native-string helpers: fixes the `__str_flatten` bucket (~165 rows).

**Tests:** 23 validation tests + flush-guard regression + strflatten regression (all green on the fork; the fork also ran the standalone repro acceptance: `class A extends Uint8Array {}` now reports `global index out of range … at function 'MyArr_new'` instead of the raw RangeError).

**Plan:** the issue family for the class — #2043 ratified design, #2039 signature-split bucket, #2029 producer diagnosis, #1984/#1985 follow-up slices. Numbers are post-renumber and collision-free with this repo's #1916–#1950.


ttraenkler:

PR #1335, opened 2026-06-11
feat(#1712): host-side vec mutators — __vec_push/__vec_pop exports + __is_vec discriminator

ttraenkler:
## Summary

**Stacked on #1327** (`issue-1712-dyn-dispatch`), which is stacked on #1307 — draft until both land; the diff then collapses to this slice's single commit.

Fixes acorn's `push is not a function` blocker (`enterScope`: `this.scopeStack.push(new Scope(flags))` through dynamic `this` dispatch — the receiver is an opaque WasmGC vec struct the host cannot grow). Three stacked defects:

1. **No host-side vec mutation**: new Wasm-side exports in `_emitVecAccessExportsInner` (`src/codegen/index.ts`): `__is_vec`, `__vec_mut_supported`, `__vec_push`, `__vec_pop` — per-vec-type ref.test dispatch chains mirroring `__vec_len`/`__vec_get`, grow on the Wasm side with the `compileArrayPush` discipline (newCap = max((len+1)*2, 4), array.new_default + array.copy + struct.set). Element coverage: externref always, f64/i32 when box/unbox imports exist; unsupported kinds return -1/0 sentinels so the runtime stays fail-loud (no silent no-op mutations).
2. **`closureBridge` wrapped DATA fields**: `_wrapForHost`'s get trap bridged ANY struct field into a callable (#1090 heuristic), so the `scopeStack` field READ returned a JS function. Vetoed via `__is_vec` — the positive discriminator (`__vec_len` can't serve: its not-a-vec default 0 is ambiguous with an empty vec; `__is_closure` can FALSE-POSITIVE on a vec whose canonicalized layout collides with a closure capture struct — observed in the 850KB acorn module).
3. **Routing**: `__extern_method_call`'s not-a-function arm routes push/pop on vec receivers through the new exports, unwrapping the receiver through both the host-proxy map and `_wasmClosureWrapperTargets`, and pushing RAW args (not host proxies) so Wasm-side element reads see the structs.

## Validation

- acorn now executes `enterScope`/`exitScope`; the parse chain advances to the next root-caused blocker (the two-shape instance member call in static `Parser.parse` — analysis in the issue file).
- `tests/issue-1712-dynamic-dispatch.test.ts` + `issue-1712-capture-closure-dispatch` pins green; typecheck clean.
- Known pre-existing, NOT addressed here (documented in the issue file): TS-typed receivers take the STATIC vec path with an unguarded checker-shape cast (`.tmp` L1/L2 probes) — separate slice.


ttraenkler:
Closing as superseded/already-merged. Both substantive commits on this branch — `3b1659bc` (acorn dynamic-dispatch chain) and `3e5d83a9` (host-side vec mutators: `__vec_push`/`__vec_pop`/`__is_vec`/`__vec_mut_supported`) — were already landed on `main` via the merged PR #1327 (merged 2026-06-11). After merging current `origin/main` into this branch, `git diff origin/main HEAD` is empty and there are no non-merge commits ahead of main: the branch is fully contained in main.

Verification on the merged tree:
- `__is_vec` / `__vec_push` / `__vec_pop` / `__vec_mut_supported` exports + runtime `__extern_method_call` push/pop dispatch are present in `src/codegen/index.ts` and `src/runtime.ts` on `origin/main`.
- `tests/issue-1712-dynamic-dispatch.test.ts` (8 tests) and `tests/issue-1712-ifelse-global-shift.test.ts` pass on the merged tree.

Nothing unique remains here to merge; closing to keep the queue clean.

PR #1339, opened 2026-06-11
fix(#2031,#2032): array dstr default+rest+short trap; computed-key object pattern binds 0

ttraenkler:
## Summary

Two destructuring spec-conformance fixes from the 2026-06-10 sweep (issue files #2031/#2032 live on PR #1315's branch, not yet on `main`). Both bugs live in `src/codegen/destructuring-params.ts`.

### #2031 — array destructuring default + rest + short source traps
`const [p, q = 9, ...rest] = [1]` threw `RuntimeError: array element access out of bounds` (Node: `p=1, q=9, rest=[]`). The rest lowering clamped `restLen = max(0, len - i)` but kept the **unclamped source offset `i`** in `array.copy(restArr, 0, srcData, i, restLen)`; WasmGC `array.copy` traps when `srcOffset > src.len` even for a zero-length copy (here `i=2 > len=1`). Default-only (`[p, q=9]`) and rest-only (`[p, ...rest]`) short sources already worked — only **default + rest + short** tripped it.

Fix: clamp the source offset to `min(i, srcLen)` so `restLen=0` copies validate; longer sources (`i <= len`) are unaffected.

### #2032 — computed-key object destructuring silently binds 0
`const k = "dyn"; const { [k]: v } = { dyn: 6 }` bound `0` (Node `6`). The struct fast path cast `element.propertyName ?? element.name` to `Identifier`; a `ComputedPropertyName` has no `.text`, so `fields.findIndex` returned `-1` and the binding was skipped, leaving the local zero-initialized.

Fix: `resolveStaticPropKey` recovers the constant string from the checker for string/number-literal-typed keys (covers `const k = "..."`, string-literal keys `["dyn"]`, and param patterns), routing them to the correct struct field. A **non-constant** computed key now reports a clear diagnostic instead of silently binding 0. Static-key destructuring is unchanged.

## Testing
- `tests/issue-2031.test.ts` — 11 cases: short/exact/longer/empty array sources with default+rest; computed keys via `const`, string-literal, function-param, and mixed-with-static bindings; loud-error path for a widened (non-constant) key. All green.
- IR fallback budget gate: OK (no unintended increases).
- `tsc --noEmit`: clean.

Issue files are on PR #1315's branch, so status is noted here rather than flipped in-tree: #2031/#2032 are **done** with this PR.


ttraenkler:

I re-ran this PR against the freshly-promoted baseline (`bcf6ea3e`, 31075/43135 host) and reproduced the diff locally with `scripts/diff-test262.ts`. The regression counts are **identical before and after the baseline refresh**, so this is a genuine regression from this PR's own changes, **net -4 pass**. The gate is correctly blocking the merge until it's fixed.

**What broke:** for-await-of and `for(;;)` destructuring on the error path (obj-ptrn-prop-eval-err) now emits compile errors

```
test/language/statements/for-await-of/async-func-dstr-const-obj-ptrn-prop-eval-err.js → compile_error
test/language/statements/for-await-of/async-func-dstr-let-obj-ptrn-prop-eval-err.js → compile_error
test/language/statements/for-await-of/async-func-dstr-var-obj-ptrn-prop-eval-err.js → compile_error
test/language/statements/for-await-of/async-gen-dstr-const-obj-ptrn-prop-eval-err.js → compile_error
test/language/statements/for-await-of/async-gen-dstr-let-obj-ptrn-prop-eval-err.js → compile_error
test/language/statements/for-await-of/async-gen-dstr-var-obj-ptrn-prop-eval-err.js → compile_error
test/language/statements/for/dstr/var-obj-ptrn-prop-eval-err.js → compile_error
```

PR #1340, opened 2026-06-11
fix(#1986,#1987): any-operand strict equality uses JS === semantics

ttraenkler:
## Problem

Strict equality (`===` / `!==`) with an `any`-typed operand was unboxed via
ToNumber and lowered to `f64.eq`, so `===` behaved **looser** than `==`:

| expr | before | node |
|---|---|---|
| `(false as any) === 0` | true | false |
| `("1" as any) === 1` | true | false |
| `(null as any) === 0` | true | false (wronger than even `==`) |
| `(true as any) === 1` | true | false |
| `(0 as any) === (-0 as any)` | false | true |

The `0 === -0` case (#1987) failed because the AnyValue numeric path split
numbers across the i32 box (tag 2) and f64 box (tag 3).

## Root cause

`compileBinaryExpression`'s externref-equality path (`src/codegen/binary-ops.ts`)
had no strict-equality branch for an `any` externref vs a typed primitive — it
fell to a numeric fallback that ToNumber-unboxes the `any` side (`null→0`,
`false→0`, `"1"→1`) and emits `f64.eq`.

## Fix

Route a strict comparison whose operands are an `any` externref and/or a known
number / null / string primitive through the existing `__host_eq` import
(JS `===`, #1065). `__host_eq` is spec-exact for §7.2.16: a type mismatch →
false, `+0 === -0` → true, `NaN !== NaN`.

**Booleans are excluded** — `coerceType(i32 → externref)` boxes a boolean as a
JS *number* (`__box_number`), so `__host_eq(true, 1)` would be false; boolean
operands keep their existing correct lowering (and a boolean `any` compared to a
boolean falls through to it). **JS-host only** — the standalone / WASI
`noJsHost` tag-dispatch path above is untouched (`#1776` still passes).

## Tests

`tests/issue-1986.test.ts` (16 tests, all pass): the four #1986 repros,
matching number / boolean / string equality, `!==` negation, the #1987 `+0`/`-0`
and `NaN` cases, plus object-identity, both-any string, and loose-`==`
regression guards. `#1065` / `#1134` / `#1776` equality suites still pass; the
`any`-heavy `#1296` analytics tests now run and pass where they were skipped.

> Two adjacent pre-existing bugs are out of scope and left untouched:
> `(null as any) == 0` (loose `==` returns true) and `(5 as any) === "5"`
> (number-any vs string traps "illegal cast") — both reproduce on `main`.

## Issue files

Issues #1986 and #1987 are part of an unmerged spec-conformance batch not yet on
any branch, so their `status: done` flip is noted here rather than committed.


ttraenkler:

I re-ran this PR against the freshly-promoted baseline (`bcf6ea3e`, 31075/43135 host) and reproduced the diff locally with `scripts/diff-test262.ts`. The regression counts are **identical before and after the baseline refresh**, so this is a genuine regression from this PR's own changes, **net -19 pass**. The gate is correctly blocking the merge until it's fixed.

**What broke:** invalid Wasm emitted for Array iteration callbacks (every/filter/find/map/reduce/some), nested destructuring, and for-in/of/await-of `let` bindings

```
test/built-ins/Array/length/S15.4.5.1_A1.3_T1.js → compile_error
test/built-ins/Array/prototype/every/15.4.4.16-4-7.js → compile_error
test/built-ins/Array/prototype/filter/15.4.4.20-4-7.js → compile_error
test/built-ins/Array/prototype/find/predicate-called-for-each-array-property.js → compile_error
test/built-ins/Array/prototype/find/return-undefined-if-predicate-returns-false-value.js → compile_error
test/built-ins/Array/prototype/findLast/predicate-called-for-each-array-property.js → compile_error
test/built-ins/Array/prototype/findLast/return-undefined-if-predicate-returns-false-value.js → compile_error
test/built-ins/Array/prototype/forEach/15.4.4.18-4-7.js → compile_error
test/built-ins/Array/prototype/map/15.4.4.19-4-7.js → compile_error
test/built-ins/Array/prototype/reduce/15.4.4.21-4-7.js → compile_error
test/built-ins/Array/prototype/reduce/15.4.4.21-5-1.js → compile_error
test/built-ins/Array/prototype/reduceRight/15.4.4.22-4-7.js → compile_error
test/built-ins/Array/prototype/reduceRight/15.4.4.22-5-1.js → compile_error
test/built-ins/Array/prototype/some/15.4.4.17-4-7.js → compile_error
test/built-ins/Array/S15.4.5.2_A3_T2.js → compile_error
test/language/expressions/assignment/dstr/array-elem-nested-array.js → compile_error
test/language/expressions/assignment/dstr/obj-prop-nested-array.js → compile_error
test/language/statements/for-await-of/let-identifier-with-newline.js → compile_error
test/language/statements/for-in/let-identifier-with-newline.js → compile_error
test/language/statements/for-of/let-identifier-with-newline.js → compile_error
```

PR #1347, opened 2026-06-11
fix(hooks): resolve hook scripts via $CLAUDE_PROJECT_DIR instead of cwd-relative paths

ttraenkler:
Granular sync from ttraenkler/js2 — splits #1313. Hook commands break when an agent's shell sits in a worktree or a deleted directory; anchoring on $CLAUDE_PROJECT_DIR makes them cwd-independent. (We hit both failure modes in practice this week.)


ttraenkler:

PR #1374, opened 2026-06-11
fix(#2059): any < any performs §7.2.13 string comparison

ttraenkler:
## Problem

Relational operators (`<`/`<=`/`>`/`>=`) on two `any` operands always returned `false` for strings:

```ts
function lt(a: any, b: any): boolean { return a < b; }
lt("a", "b");   // wasm: false — node: true
lt("10", "9");  // wasm: false — node: true (lexicographic)
```

## Root cause

In the default JS-host path `any` compiles to **externref** (`ctx.anyValueTypeIdx` is `-1`, so the `__any_lt` machinery the issue pointed at never registers). The numeric fast path then unboxes both externrefs to f64 (`Number("a") = NaN`), so `f64.lt(NaN, NaN)` is false for every string relational.

## Fix

- New `__host_relational` host import + `host_relational` intent (`import-manifest.ts`, `index.ts`, `runtime.ts`): `(externref, externref, i32 opcode)` where opcode 0/1/2/3 = `<`/`<=`/`>`/`>=`, delegating to the native JS operator which already implements §7.2.13 (ToPrimitive + lexicographic string compare).
- In `compileBinaryExpression` (`binary-ops.ts`), before the numeric coercion, route relationals through `__host_relational` when a JS host is available **and neither operand is statically a primitive we already handle numerically** (number / boolean / bigint / string). That guard leaves the provably-numeric and provably-string fast paths untouched (no perf regression). Standalone/WASI falls through to the numeric path — no unsatisfiable host import.

## Tests

New `tests/equivalence/any-relational-string.test.ts` (16 cases: string lexicographic, mixed string/number, pure numeric, NaN, null/undefined, all four operators). All pass. No regressions in string-relational-operators / comparison-coercion / boolean-relational-comparison / i32-loop-compare / string-arithmetic-coercion / logical-operators (48 tests green).

Closes #2059.

### DIFF
--- a/plan/issues/2059-any-relational-no-string-comparison.md
+++ b/plan/issues/2059-any-relational-no-string-comparison.md
@@ -1,10 +1,11 @@
 ---
 id: 2059
 title: "relational operators on two any/externref operands never perform string comparison (\"a\" < \"b\" → false)"
-status: ready
+status: done
 sprint: 61
 created: 2026-06-10
-updated: 2026-06-10
+updated: 2026-06-11
+completed: 2026-06-11
 priority: high
 feasibility: medium
 reasoning_effort: medium
@@ -66,3 +67,35 @@ policy). Likely shares plumbing with #2058.
 Grepped `relational`, `string comparison`, `__any_lt` — #117 (harness string
 compare, done), #295 (comparison ops, bigint-focused, done), #1563/#1577
 (broad spec audits, item absent). Not covered.
+
+## Resolution (2026-06-11)
+
+The root-cause note about `__any_lt` was a red herring for the **default**
+(JS-host) compile path: there `ctx.anyValueTypeIdx` is `-1`, so the AnyValue
+helpers are never registered and `any` operands compile to **externref**, not
+`$AnyValue`. The numeric fast path then coerces both externrefs to f64
+(`__unbox_number` → `Number("a") = NaN`), so every string relational was false.
+
+Fix (JS-host path, dual-mode safe):
+- Added a `__host_relational` host import + `host_relational` intent
+  (`src/compiler/import-manifest.ts`, `src/index.ts`, `src/runtime.ts`). It
+  takes `(externref, externref, i32 opcode)` where opcode 0/1/2/3 =
+  `<`/`<=`/`>`/`>=` and delegates to the native JS operator, which already
+  implements §7.2.13 (ToPrimitive + lexicographic string compare).
+- In `compileBinaryExpression` (`src/codegen/binary-ops.ts`), before the numeric
+  coercion, when the op is relational, a JS host is available, **and neither
+  operand is statically typed as a primitive we already handle numerically**
+  (number / boolean / bigint / string), compile both operands as externref and
+  call `__host_relational`. The "both non-primitive" guard keeps the
+  provably-numeric and provably-string fast paths untouched (no perf
+  regression). Standalone/WASI (`ctx.standalone`/`ctx.wasi`) falls through to
+  the existing numeric path — no new unsatisfiable host import is emitted.
+
+### Test Results
+
+New `tests/equivalence/any-relational-string.test.ts` — 16 cases covering
+string lexicographic (`"a"<"b"`, `"10"<"9"`, `"abc"<"abd"`, `"abc"<"ab"`),
+mixed string/number (numeric), pure numeric, NaN, and null/undefined, across
+all four operators. All pass. No regressions in string-relational-operators /
+comparison-coercion / boolean-relational-comparison / i32-loop-compare /
+string-arithmetic-coercion / logical-operators (48 tests green).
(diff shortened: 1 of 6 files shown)


ttraenkler:
Closing in favor of a reimplementation per the merged cluster spec in `plan/issues/2058-any-plus-runtime-string-numeric-add.md` / `2059-any-relational-no-string-comparison.md` (written after this PR).

Why I'm not salvaging this branch:

1. **The gate is keyed on TS checker types, not lowered Wasm types** — it fires for *any* relational where neither side is a primitive (class instances, arrays, plain objects), coerces those WasmGC refs to externref, and hands them to the host where `a < b` on an opaque Wasm object throws (ToPrimitive has nothing to call). That matches the CI signature: 14 regressions, all `runtime_error`, all wasm-hash-changed, net −11. The spec's gate is `leftType.kind === "externref" || rightType.kind === "externref"` on the lowered types, placed just before the externref-numeric fallback (after every dedicated static path), so it only fires where the bug actually lives.
2. **Import shape diverges from the cluster design**: the spec shares one `__host_compare(externref, externref) -> i32` (4-way −1/0/1/2 result) with #2058's plumbing instead of a per-operator opcode import.
3. **No standalone path** — the spec requires the inline §7.2.13 nativeStrings dispatch (dual-mode policy); falling through to the numeric path leaves standalone wrong.
4. The branch is conflicting with main, and #2059 is step 3 in the staged landing order (#2063 → #2058 → #2059), so it would have to be held and re-merged twice anyway.

I'll carry over `tests/equivalence/any-relational-string.test.ts` (the 16 cases are good) into the reimplementation. Tracking: issue #2059, lands after #2058.

ttraenkler:

I re-ran this PR against the freshly-promoted baseline (`bcf6ea3e`, 31075/43135 host) and reproduced the diff locally with `scripts/diff-test262.ts`. The regression counts are **identical before and after the baseline refresh**, so this is a genuine regression from this PR's own changes, **net -36 pass**. The gate is correctly blocking the merge until it's fixed.

**What broke:** String.prototype endsWith/includes/startsWith/slice/charCodeAt position-argument handling, Array.sort comparison, Object.defineProperty

```
test/built-ins/Array/prototype/sort/S15.4.4.11_A2.1_T1.js → fail
test/built-ins/Array/prototype/sort/S15.4.4.11_A2.1_T2.js → fail
test/built-ins/Object/defineProperty/15.2.3.6-4-416.js → fail (returned 3 — assert #2 at L37: var verifyNameProp = false;)
test/built-ins/Object/defineProperty/15.2.3.6-4-592.js → fail (returned 3 — assert #2 at L43: var verifyNameProp = false;)
test/built-ins/String/prototype/charCodeAt/S15.5.4.5_A3.js → fail (returned 2 — assert #1 at L14: assert.sameValue(__instance.charCodeAt(3), NaN);)
test/built-ins/String/prototype/endsWith/coerced-values-of-position.js → fail (returned 8 — assert #7 at L34: assert(str.endsWith('The future', 10.4), '10.4...)
test/built-ins/String/prototype/endsWith/return-abrupt-from-position.js → fail (returned 2 — assert #1 at L20: throw new Test262Error();)
test/built-ins/String/prototype/endsWith/searchstring-found-with-position.js → fail (returned 2 — assert #1 at L27: assert( str.endsWith('The future', 10), 'str.e...)
test/built-ins/String/prototype/endsWith/searchstring-not-found-with-position.js → fail (returned 2 — assert #1 at L27: assert.sameValue( str.endsWith('is cool!', str...)
test/built-ins/String/prototype/endsWith/String.prototype.endsWith_Fail_2.js → fail (returned 2 — assert #1 at L12: assert.sameValue('word'.endsWith('d', 3), fals...)
test/built-ins/String/prototype/endsWith/String.prototype.endsWith_Success_4.js → fail (returned 2 — assert #1 at L12: assert.sameValue('word'.endsWith('r', 3), true...)
test/built-ins/String/prototype/includes/coerced-values-of-position.js → fail (returned 10 — assert #9 at L38: assert.sameValue(str.includes('The future', t...)
test/built-ins/String/prototype/includes/return-abrupt-from-position.js → fail (returned 2 — assert #1 at L20: throw new Test262Error();)
test/built-ins/String/prototype/includes/return-false-with-out-of-bounds-position.js → fail (returned 2 — assert #1 at L25: assert.sameValue( str.includes('!', str.length...)
test/built-ins/String/prototype/includes/searchstring-not-found-with-position.js → fail (returned 2 — assert #1 at L24: assert.sameValue( str.includes('The future', 1...)
test/built-ins/String/prototype/includes/String.prototype.includes_FailBadLocation.js → fail (returned 2 — assert #1 at L12: assert.sameValue('word'.includes('w', 5), fals...)
test/built-ins/String/prototype/includes/String.prototype.includes_FailLocation.js → fail (returned 2 — assert #1 at L12: assert.sameValue('word'.includes('o', 3), fals...)
test/built-ins/String/prototype/slice/S15.5.4.13_A1_T7.js → fail
test/built-ins/String/prototype/startsWith/coerced-values-of-position.js → fail (returned 9 — assert #8 at L34: assert.sameValue( str.startsWith('The future',...)
test/built-ins/String/prototype/startsWith/return-abrupt-from-position.js → fail (returned 2 — assert #1 at L19: throw new Test262Error();)
```

PR #1396, opened 2026-06-12
fix(#1959): RepeatMatcher empty-iteration progress guard for native RegExp VM

ttraenkler:
## Problem

The standalone (pure-WasmGC) RegExp VM had no §22.2.2.3.1 RepeatMatcher progress guard. A nullable-bodied quantifier (`(?:a?)*`, `(a*)+`) ran a zero-width iteration that looped pushing backtrack frames until the 1,000,000-step cap, which `runAt` reports as "no match" — a silent wrong answer plus a multi-second perf cliff at every scan position. `/(?:a?)*/.test('b')` returned `false` (node: `true`) in ~3s.

## Fix

New `ReOp.EMPTYCHECK = 13` opcode + per-loop scratch slot:
- **compile.ts** — `nodeMatchesEmpty()` nullability analysis. A nullable `star` records sp at iteration start (`SAVE scratch`) and verifies progress before looping (`EMPTYCHECK scratch` → fail ⇒ take the loop exit). A nullable `plus` lowers to `body·star(body)` so the mandatory first rep may legally match empty while 2nd+ reps are guarded. `repeat` inherits the guard via its star/opt expansion.
- **bytecode.ts** — `EMPTYCHECK` opcode + `CompiledRegex.nSlots` (`2*nGroups` capture slots + one scratch slot per EMPTYCHECK).
- **vm.ts** (reference VM) + **native-regex.ts** (hand-authored Wasm VM) — `EMPTYCHECK` handler; `nSlots` threaded for caps allocation through `__regex_search/run/replace/split/match_all` while match-result array shape still uses `nGroups`.
- **regexp-standalone.ts** — `$NativeRegExp` struct gains field 6 `nSlots`; caps allocation + every helper call pass it.

## Tests

`tests/issue-1959.test.ts` — 14 standalone-Wasm cases via `String.prototype.search` vs native (headline `(?:a?)*`, nested `(?:a*)*`, `(a*)+`, `(?:|a)*`, anchored-fail `(a?)*x`, plus non-nullable controls). All match native, all <1ms (was 300ms–3s, silent wrong answer). `tests/regex-bytecode.test.ts` (TS VM) + the #1539 suite: 454/455 pass — the lone failure (`refuses unicode flag (u)`) is **pre-existing on main** (the u-flag refusal was lifted earlier and that test was never updated; confirmed on a clean checkout), tracked separately.

### DIFF
--- a/plan/issues/1959-regex-vm-empty-quantifier-no-progress-guard.md
+++ b/plan/issues/1959-regex-vm-empty-quantifier-no-progress-guard.md
@@ -1,10 +1,11 @@
 ---
 id: 1959
 title: "native RegExp VM: empty-body quantifier loops burn the 1M-step cap and silently report no-match (/(?:a?)*/ fails)"
-status: ready
+status: done
 sprint: 61
 created: 2026-06-10
-updated: 2026-06-10
+updated: 2026-06-12
+completed: 2026-06-12
 priority: high
 feasibility: medium
 reasoning_effort: high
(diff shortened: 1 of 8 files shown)


ttraenkler:
Closing as duplicate — #1395 already merged the same #1959 fix (RepeatMatcher empty-iteration progress guard). I opened this in parallel without seeing #1395 first. No action needed.

PR #1425, opened 2026-06-12
docs(#1983): scope class-member name-collision + flag extern-key overload trap

ttraenkler:
## What this is

Scoping + analysis for #1983 (synthetic class-method names collide with user functions). **Doc-only — no code change.**

## Findings

- **Bug reproduced** on main: `class A { m(){} }` + `function A_m(){}` compiles to an **invalid Wasm binary** (both `new A().m()` and `A_m()` resolve to one overwritten `funcMap` entry with mismatched signatures).
- **Scoped** the collision to `ctx.funcMap` (the only map a user `function A_m()` also writes to). Accessor/static/closure maps use the same `${name}_${member}` string convention but key disjoint namespaces — they must NOT flip.
- **CRITICAL trap found mid-implementation:** funcMap keys are **overloaded** between user-class members and **extern/host-class** members — `Array_push`, `Map_get`, `Map_new`, `RegExp_new` are real `env` host-import names registered under the same `${name}_${member}` convention (index.ts:10184/10192/10271). A blind separator flip would break extern-class construction and method dispatch — **strictly worse than the original bug**.

## Why this needs arch input

A correct fix must thread a **user-vs-extern discriminator** through all ~20 funcMap sites (only mangle when the receiver is a user class; leave extern host-import names as `_`). I documented that requirement plus a narrower alternative (uniquify only on detected collision at the single registration site + a shared lookup wrapper, avoiding the 20-site flip). Set `status: blocked` pending an `[ARCH input first]` pass.

The `#`-separator helper primitive I prototyped is described in the spec but intentionally **not** committed as code — the discriminator design should land with it so it isn't misused as a blind flip.


ttraenkler:
Closing — against the fresh test262 baseline this regresses **3,829 tests (net −3,678)**, i.e. it broadly breaks compilation for a whole category, and it's been abandoned for ~2 days. It can't be merged in this state without destroying conformance. The underlying issue #1983 stays open for a clean reimplementation; I'll reopen if it's actively reworked.

PR #1430, opened 2026-06-12
fix(#1967): widen HOF gates to struct (ref) element arrays

ttraenkler:
## Problem

`map`/`filter`/`reduce`/`reduceRight`/`forEach`/`find`/`findIndex`/`findLast`/`findLastIndex`/`some`/`every` silently returned empty/garbage on struct (`ref`) element arrays:

```ts
[{v:"a"},{v:"b"}].map(o => o.v);   // wasm: "null|0"   node: "a|2"
[{k:1},{k:2},{k:3}].filter(o => o.k > 1).length;  // wasm: 0   node: 2
[{k:1},{k:2},{k:3}].reduce((a,o) => a+o.k, 0);    // wasm: 0   node: 6
```

The dispatch gates in `compileArrayMethodCall` only passed `f64|i32|externref` element kinds, so `ref`-element receivers fell into the generic no-op fallback. Companion to #1390, which widened the **sort** gate the same way; this finishes #1967.

## Fix

The native WasmGC loop machinery (`setupArrayLoop` + `buildClosureCallInstrs`) is **already generic over `elemType`** — it reads each element via `array.get` and coerces it to the closure param type via `coercionInstrs(elemType, paramTypes[0])` (a no-op when the closure param is the same struct ref). So, as with the #1390 sort gate, widening the gate is the whole fix for the iterating methods. Added an `isHofElemKind()` helper and applied it to all HOF gates (now also accepts `ref`/`ref_null`).

`find`/`findLast` additionally hard-coded an `f64` result local (init to NaN) in non-fast mode — storing a struct ref into it produced an **invalid binary**. `ref`/`externref` elements now use a **nullable element-typed** result local with a `null` not-found sentinel (new `nullableValType` / `nullRefInstrs` helpers), matching JS `find` returning `undefined`.

## Tests

`tests/equivalence/hof-struct-elements.test.ts` (15, match Node): map (string field / numeric / struct-returning), filter (length + chained field access), reduce/reduceRight (numeric accumulator over structs), find/findLast (+ null not-found sentinel), findIndex/findLastIndex, some/every, forEach. Regression-clean across array-prototype-methods / sort-nonnumeric / array-of-structs / reverse-struct-map / array-callback-three-params / array-filter-obj-length (31 green); tsc + biome + prettier clean.

## Scope

Out of scope (distinct issues, **not** regressed by this change):
- `some`/`every` return an i32 truthy flag not boxed to a JS boolean at the export boundary — pre-existing for **numeric** arrays too (verified on main).
- A **struct accumulator** in `reduce` with no initial value still hits the numeric `accTmp` typing generalised under #1994.


ttraenkler:
Closing — regresses **300 tests (net −299)** for only +1 improvement against the fresh baseline, and abandoned for ~1 day. Not mergeable as-is. Issue #1967 stays open for a proper redo; reopen if you pick it back up.

PR #1439, opened 2026-06-13
fix(#2036): standalone Array.prototype generics over array-like refuse loud (stage 1)

ttraenkler:
## Problem (#2036)

In **standalone/WASI** mode, `Array.prototype.<m>.call(arrayLike, …)` over a
NON-array receiver (open `$Object`, `arguments`, `any`) produced three broken
outcomes — two of which violate the #1888 dual-mode invariant ("any uncertainty
⇒ fail loud, never invalid Wasm"):

1. **Invalid Wasm** — `local.set expected f64, found externref` (indexOf), `call
   expected externref, found f64.convert_i32_s` (filter)
2. **Null deref** at runtime
3. **Silently wrong** — `Array.prototype.indexOf.call({0:5,5:'length',length:6}, 'length')` → `-1`

Meanwhile `map`/`reduce`/`lastIndexOf` already refuse such receivers loudly — so
the refusal gate exists, but `indexOf`/`filter`/`forEach`/… had arms slipping
past it into broken codegen.

## Fix — stage 1 ("stop the bleeding")

Per the issue's own staged plan. In `src/codegen/array-methods.ts`
`compileArrayPrototypeCall`: when `ctx.standalone || ctx.wasi` AND the receiver
is not a genuine native-array vec (`resolveArrayInfo` returns null), return
`undefined` so the borrowed-method dispatch in `expressions/calls.ts` emits the
loud `#1888 Slice 3/4` refusal — exactly like `map`/`reduce`.

Why: the typed shape-inferred fast paths emit f64/i32 element loads on an
externref (invalid Wasm), and `compileArrayLikePrototypeCall` depends on the
`__extern_length`/`__extern_get_idx` JS-host `env` imports that **don't exist
standalone**. Genuine native-array receivers still take the fast path. The gate
is `standalone/wasi` only, so **host mode is byte-for-byte unchanged**.

This converts the ~430 invalid-Wasm/null-deref/wrong-result standalone rows into
honest refusals, protecting conformance from silent wrongness.

## Tests

`tests/issue-2036.test.ts` (5): standalone indexOf/filter/forEach.call on an
array-like → loud refusal; standalone `indexOf.call([10,20,30], 20)` → `1`; host
array-like call still compiles (refusal didn't leak into host mode).

Pre-existing failures in `#1461` (concat-spreadable), `#1131` (fib), `#1888`
(wasi-roundtrip) confirmed **identical on clean main** — not caused by this
change. `tsc --noEmit`, `biome lint`, `prettier --check` clean.

## Stage 2 (follow-up, NOT in this PR)

The generic `$Object` arm (read `length` via `__extern_get`, keyed element get,
externref comparison per §23.1.3.17) is the larger Slice-4 work. Issue stays
`in-progress`. The host-mode array-like wrong-results are a separate, pre-existing
concern outside this standalone issue's scope.

Refs #2036 (stage 1).

### DIFF
--- a/plan/issues/2036-standalone-array-generics-arraylike-invalid-wasm.md
+++ b/plan/issues/2036-standalone-array-generics-arraylike-invalid-wasm.md
@@ -1,10 +1,10 @@
 ---
 id: 2036
 title: "standalone: Array.prototype generics over array-like receivers emit invalid Wasm / null-deref / wrong results instead of refusing loud (~500+ tests)"
-status: ready
+status: in-progress
 sprint: Backlog
 created: 2026-06-10
-updated: 2026-06-10
+updated: 2026-06-13
 priority: high
 feasibility: medium
 reasoning_effort: high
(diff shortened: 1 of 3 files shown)


ttraenkler:
Closing per tech-lead direction: the stage-1 'refuse the array-like path' approach is structurally unable to go net-positive (persistent -30 standalone regression — the compileArrayLikePrototypeCall path is NOT universally broken standalone; ~30 struct/array-like receivers with __sget_* getters legitimately pass through it). #2036 is reassigned to the architect for the stage-2 generic-arm spec (read length via __extern_get, keyed element get, externref comparison per §23.1.3.17). Issue file documents both stage findings.

PR #1459, opened 2026-06-14
fix(#2085): array-HOF predicate truthiness matches §7.1.2 ToBoolean

ttraenkler:
## Problem

`buildTruthyCheck`/`buildFalsyCheck` (the truthiness test for array HOF callback RESULTS — filter/find/some/every) were a second hand-rolled ToBoolean that drifted from the canonical `ensureI32Condition`:
- f64 arm used `f64.ne 0` → **NaN wrongly truthy**;
- ref/externref arm only did `ref.is_null` → **boxed `0`/`""`/`false`/`NaN` wrongly truthy**.

```ts
[1, 2, 3].filter((x) => NaN as any)   // wasm kept all; node keeps none
[0, 1].find((x) => (x as any))        // boxed-0 predicate wrongly truthy
```

## Fix

Route both through a shared `buildToBooleanInstrs` mirroring `ensureI32Condition` (§7.1.2):
- **f64** → `|x| > 0` (NaN / ±0 falsy);
- **any-boxed ref** (`isAnyValue`) → `__any_unbox_bool`;
- **externref** → `__is_truthy`;
- **native-string ref** → flatten + `len > 0`;
- **i32/i64** → as-is / `i64.eqz; i32.eqz`.

`buildFalsyCheck` is now `!buildToBoolean` (reuse + `i32.eqz`).

## Tests

`tests/issue-2085.test.ts` — 7/7: filter NaN/boxed-0/`""`/false → keeps none, truthy → keeps all, `find([0,1], x=>(x as any))` → 1, some/every boxed-falsy, and normal boolean predicates unaffected.

No regressions: `issue-2074` green. The `functional-array-methods.test`/`array-methods.test` fixture failures (TS "number | undefined not assignable" errors) reproduce **identically on baseline main** — pre-existing, unrelated.

## Residual (noted in issue)

`find((x) => (0 as any))` with an INLINE `(literal as any)` boxes the closure RESULT to externref → `__is_truthy` on the box wrapper (host `v?1:0` sees the wrapper as truthy). That's a separate closure-return-boxing / `__is_truthy`-unwrap concern; the `buildTruthyCheck` drift this issue names is fixed (element-typed and variable-boxed `any` predicates — the common shapes — work via `__any_unbox_bool`).

Closes #2085.


ttraenkler:
## Root-cause of the −6 regression gate (senior-dev review)

I traced all 18 of the regressing test262 tests with runtime proof. **The ToBoolean fix in this PR is correct** — the regressions are NOT a `buildTruthyCheck` defect. They are a pre-existing `this`-modeling bug that this PR *unmasks*:

- The 18 regressions are `every/filter/some/reduceRight` callbacks that read `this` (15 use a passed `thisArg`; 3 return top-level `this`/global).
- Array-HOF callbacks compile to the closure (`call_ref`) path with **no `this` parameter** — `this` inside the callback is a literal `__get_undefined()`. So `this.PROP` is `undefined`/falsy and `thisArg` is never forwarded.
- On `main` those wrong-`undefined` results were rendered *truthy* by the two truthiness bugs this PR removes (`f64.ne 0` → NaN truthy; `ref.is_null` → non-null `undefined` sentinel truthy), so the tests passed **by accident**.
- Proven: `main` gives `filter(cb,{res:true}).length === 3` (Node=3) via the missing-thisArg path; this PR correctly gives the value as falsy → `0` → assertion fails.

The 12 improvements and 18 regressions flow through the **same (correct) ToBoolean arms** and are inseparable there — masking the `this` bug in `buildTruthyCheck` would re-break the `""`/boxed-empty wins (#2080/#2085) and is rejected. So **no change to `buildTruthyCheck` can make this PR net-positive**; the fix must be upstream.

I filed **#2152** (thisArg + top-level-`this` forwarding to array HOF callbacks — architect-spec sized: `.call`-aware host bridge vs threading a `__this` param through the closure ABI). Full runtime-proven breakdown is in `plan/issues/2085-...md` and `plan/issues/2152-...md` (pushed to this branch).

**Recommendation:** do not merge this PR alone (net −6). Land #2152 first so this becomes net-positive, or land #1459 + #2152 together. The 8 genuinely `this`-independent ToBoolean wins (literal `NaN`/`""`/`false` predicates, `findIndex`/`findLastIndex return-negative-one`) all pass and are real.

PR #1509, opened 2026-06-16
fix(#2160): standalone Number.parseInt / Number.parseFloat namespaced aliases

ttraenkler:
## Summary

One independent slice of the **#2160** standalone String/Number residual.

`Number.parseInt` / `Number.parseFloat` (§21.1.2.12-13) are the same functions
as the global `parseInt` / `parseFloat` and lower through the same call-site
routing (`calls.ts` reads `funcMap.get("parseInt"/"parseFloat")`). The **bare**
forms worked standalone because the parse import-collector
(`src/codegen/declarations.ts`) registers a native WasmGC scanner under those
names — but the collector only recognized the **bare-identifier** call form, so
the `Number.`-prefixed property-access form never registered the scanner and
standalone compilation failed with a `__get_builtin` codegen error.

**Fix:** detect the `Number.parseInt` / `Number.parseFloat` call shape in the
collector and add the same helper to `parseNeeded`. Bare forms unchanged; no
call-site change needed (routing already reads the same funcMap names).

## Verified (host + standalone)

| program | result |
|---|---|
| `Number.parseFloat("1.5") === 1.5` | true |
| `Number.parseInt("42", 10) === 42` | true |
| `Number.parseInt("42") === 42` (no radix) | true |
| `Number.parseInt("ff", 16) === 255` | true |
| `Number.parseInt("42px", 10) === 42` | true |
| `Number.parseFloat("abc")` is NaN | true |
| namespaced === bare form | true |
| bare `parseInt`/`parseFloat` (regression guard) | true |

`tests/issue-2160-number-parse.test.ts` — 16 cases (8 × {host, standalone}),
all green. `tsc --noEmit` clean.

## Scope

Issue **stays `ready`**. Re-measured against latest main: the common
String/Number methods + coercion already pass standalone (many closed by the
value-rep P0/P1 work that just landed). The remaining bulk of the 635 is the
**wrapper objects** `new String(...)`/`new Number(...)` (gated on value-rep
boxing #2072/#2104 — the `__new_String`/`__new_Number` leak in the acceptance
criteria) plus harder coercion edges overlapping the coercion engine (#1917).


ttraenkler:
Dequeued from the merge queue by the queue shepherd: this PR's merge-group run (head `cc413c9`) failed two required checks against the live baseline, and the failure is a **real regression** (not the known async-args standalone-guard false-positive — the wasm hash changed):

- **`Differential test` (Delta gate)**: `string/13-template-literal.js: match → mismatch` — a program that matched V8 on the baseline no longer does. Net delta +13 improvements / **−1 regression**; the delta gate is zero-tolerance on regressions.
- **`merge shard reports` / "check for test262 regressions"**: **Net −1 pass (32734 → 32733)**, category `assertion_fail: 1`, **with a wasm-hash change** (a genuine codegen delta from this PR, not baseline drift — baseline age was only 31m).

All 96 individual test262 shards passed; the failure is purely the −1 regression surfacing in the aggregate gate. The Number.parseInt / Number.parseFloat namespaced-alias change appears to affect template-literal string handling.

Applied the `hold` label so the auto-enqueue backstop won't re-sweep it. **Action for the author:** fix the `string/13-template-literal.js` regression (or, if intentional, update the differential + test262 baselines with justification), then remove `hold` and re-enqueue.

PR #1514, opened 2026-06-16
fix(#2106): find/findLast miss observable through ?? (value-rep P3 slice S3.1)

ttraenkler:
## #2106 value-rep P3 — slice S3.1: `find`/`findLast` miss observable through `??`

First bounded increment of S3 (zero arithmetic blast radius — keeps the f64 result type).

`[1,2,3].find(x=>x>5) ?? -1` returned **`NaN`** instead of `-1`: a numeric-array `find`/`findLast` miss is lowered to f64, and `??` treats any f64 as never-nullish, so `?? rhs` never fired.

### Fix (two spots, composing with #2104's `value-tags.ts`)
1. **`array-methods.ts`** (find, findLast): the f64 miss now carries the **distinguished `UNDEF_F64` sentinel** (`0x7FF00000DEADC0DE` via `pushUndefF64`) instead of a plain `0/0` NaN — so a genuine NaN **element** (`[NaN].find(Number.isNaN)` is a real hit, never nullish) stays distinct from a miss.
2. **`expressions/logical-ops.ts`** `compileNullishCoalescing`: added `isFindFamilyCall` (mirrors the existing #2004 `isCodePointAtCall` case); for an f64 find/findLast LHS, branch on `emitIsUndefF64` (the exact sentinel, NOT plain NaN) so `find(miss) ?? rhs` yields `rhs`.

### Scope / honest limits
- Covers the **inline** `find(...) ?? rhs` form (syntactic recognition, exactly like the merged #2004 codePointAt approach).
- The **bound-variable** form (`const v = find(...); v ?? rhs`), `=== undefined`, and `typeof` are **not** covered — an f64 can never be `=== undefined`; those need the f64→externref result-type widening (**S3 increment 2**, measure-first). Bound-variable `??` is unchanged from before (still NaN), so no regression.
- **No regression**: the sentinel is still a NaN bit pattern, so `Number.isNaN`, truthiness, and arithmetic propagation on a miss are byte-identical to the old plain-NaN behavior; only inline `??` improved.

### Validation
- `tests/issue-2106-find-nullish.test.ts` — 7/7 (find/findLast miss→rhs, hit, genuine-NaN-element-is-a-hit, `0` not coalesced, inline arithmetic).
- `tsc`, `ir-nullish-coalesce` (4/4), `issue-1136-findlast` (6/6), `check:any-box-sites` all clean.

S1/S2/S4 + S3 increment 2 (optional-param + externref widening) remain open under #2106.

### DIFF
--- a/plan/issues/2106-valuerep-p3-undefined-observability.md
+++ b/plan/issues/2106-valuerep-p3-undefined-observability.md
@@ -1,10 +1,11 @@
 ---
 id: 2106
 title: "value-rep P3: undefined observability — UNDEF_F64 sentinel, union-collapse reversal (flagged), standalone $undefined singleton"
-status: ready
+status: in-progress
+assignee: ttraenkler/sdev7
 sprint: 62
 created: 2026-06-11
-updated: 2026-06-15
+updated: 2026-06-16
 priority: high
 feasibility: hard
 reasoning_effort: max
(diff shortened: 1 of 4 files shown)


ttraenkler:
Parking this as a draft. The original author's session ended and the local worktree for this branch got contaminated with unrelated cross-issue changes, so a clean conflict resolution isn't safe to do in place. The conflict itself is trivial (one planning-doc section), and #2106 S0 already landed via #1511 — I'll redo this S3.1 slice cleanly from origin/main in a fresh branch when a senior slot frees up, rather than salvage a polluted worktree. Un-drafting when that happens.

ttraenkler:
Closing this parked draft to stop the merge-queue monitor from re-flagging it every cycle. Non-destructive: the branch `issue-2106-s3-num-undef-externref` and its commits are preserved (closing a PR doesn't delete the branch), so this is fully reopen-able. Context: the original author's session ended and this branch's local worktree got contaminated with unrelated cross-issue changes, so an in-place conflict resolution wasn't safe. #2106 S0 already landed via #1511, so the value-rep P3 lane isn't blocked — I'll redo this S3.1 slice (find/findLast through ??) cleanly from origin/main in a fresh branch when a senior slot frees up.

PR #1518, opened 2026-06-16
fix(#2130): runtime presence model for delete/in — shared _wasmStructHasOwn predicate

ttraenkler:
## Summary

`in` / `hasOwnProperty` / `Object.keys` answered against the **static struct shape**, never the runtime delete tombstone — so `delete o.a; "a" in o` stayed true, object-rest `"e" in rest` reported the source-struct field present, and `Object.keys` kept deleted keys. Stage A + Stage B of the #2130/#1991 presence-predicate plan (runtime-side, JS-host).

## What changed (`src/runtime.ts`)
- **Stage A** — new shared `_wasmStructHasOwn(obj, key, exports)`: tombstone + key-based sidecar (§7.3.12 value-independent) + descriptors + proto/static methods + per-receiver struct fields. `__hasOwnProperty` and `__extern_has` (`in`) both route through it. **Deleted the `__sget_<key>` getter-doesn't-throw probe** — it was a module-global field-name test (any receiver "has" a key if ANY struct declares it) and the root of the `in` false positives. Object-rest fixed with zero codegen (rest is a plain JS object; no `__sget_` fallthrough).
- **Stage B** — read-path tombstone gate in `__extern_get` (before `__sget_`) + `_safeGet` (top of the WasmGC-struct branch); `_safeSet` clears the tombstone on re-add (single choke point for sidecar + `__sset_` arms); `__delete_property` coerces a WasmGC/native-string key to a primitive before the tombstone key.
- **A4** — `_enumerableOwnStructKeys` (tombstone filter + enumerable sidecar union) backs `Object.keys`/`values`/`entries`.

## Tests
`tests/issue-2130-delete-in-presence.test.ts` — 10 cases (JS-host, `buildImports`+`setExports`): `delete o.a; "a" in o`→false, `"b" in o` stays true, dynamic-key delete, object-rest `"e"/"f" in rest`, delete-then-re-add `in`, `Object.keys` drops the key, `Object.keys(rest)===["f"]`, present-prop control, **opaque (param `any`) `o.a` reads undefined after delete**. All green. No regression: hasOwnProperty / delete-fastpath / #2066 / class-method-delete suites stay green.

## Known residual (documented in the issue; spec §A5/A6)
Reading `o.a` after `delete o.a` on a **statically-struct-shaped LOCAL `any` var** still struct.gets the stale field — the compiler's `ref.test`/`struct.get` fast path bypasses the runtime gate (the deleted field is physically still in the struct). Needs codegen read/delete struct-type symmetry (the delete sentinel and the read must agree on the type), distinct from this runtime presence model. The **opaque-receiver (param/cast `any`) read IS fixed**. #2130 kept `in-progress` for that codegen follow-up.

### DIFF
--- a/plan/issues/2130-delete-and-in-resolved-against-static-struct-shape.md
+++ b/plan/issues/2130-delete-and-in-resolved-against-static-struct-shape.md
@@ -1,10 +1,10 @@
 ---
 id: 2130
 title: "delete o.prop is a no-op and `in` answers against the static struct shape — post-delete / dynamic-key / object-rest all wrong"
-status: ready
+status: in-progress
 sprint: 62
 created: 2026-06-12
-updated: 2026-06-12
+updated: 2026-06-16
 priority: high
 feasibility: hard
 reasoning_effort: high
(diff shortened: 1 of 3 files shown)


ttraenkler:
Closing as superseded by #1525 (`issue-2130-delete-in-presence-predicate`), which is the newer, more complete Stage-A implementation of the same #2130 tombstone-aware `in`/hasOwnProperty predicate (both authored by d1 during a brief duplicate-worktree overlap). #1525 is the keeper; this branch was UNMERGEABLE in the queue. Avoids a double-merge of #2130.

PR #1552, opened 2026-06-16
fix(#2017): getter-only accessor write throws catchable TypeError

ttraenkler:
## #2017 — assignment to getter-only accessor must throw strict-mode TypeError

```ts
const o: any = { get x() { return 1; } };
o.x = 99;  // was: silent no-op   node: TypeError (only a getter)
```

### Root cause
The "illegal cast" trap was already gone (s62 accessor work). Residual: `o.x = 99` on a getter-only accessor **silently no-op'd** instead of throwing. Object literals `{ get x(){…} }` install a real host accessor on a `__new_plain_object` JS object; `o.x = …` lowers to `__extern_set` → `_safeSet`, whose strict `obj[key] = val` correctly threw V8's getter-only TypeError — but `_safeSet`'s catch swallowed it and diverted to the sidecar.

### Fix
In `_safeSet`'s write catch (`src/runtime.ts`), detect a getter-only accessor via the new `_isGetterOnlyAccessor(obj, key)` helper (walks the proto chain; a data property anywhere shadows it) and re-throw a spec-worded TypeError instead of falling through to the sidecar. Mirrors the existing `_isRevokedProxyError` rethrow (#2180). get/set pairs and plain dynamic writes are unaffected.

### Tests
`tests/issue-2017.test.ts` — 4 cases: throws catchable TypeError; getter value unchanged; get/set pairs still accept writes; new dynamic property still works.

The 3 pre-existing `accessor-side-effects.test.ts` failures are unchanged by this fix (confirmed failing on unmodified main).

Closes #2017.

### DIFF
--- a/plan/issues/2017-getter-only-assignment-trap-not-typeerror.md
+++ b/plan/issues/2017-getter-only-assignment-trap-not-typeerror.md
@@ -1,10 +1,12 @@
 ---
 id: 2017
 title: "assignment to a getter-only object-literal property traps 'illegal cast' instead of throwing strict-mode TypeError"
-status: ready
+status: done
+completed: 2026-06-16
+assignee: ttraenkler/dev-a
 sprint: 63
 created: 2026-06-10
-updated: 2026-06-12
+updated: 2026-06-16
 priority: low
 feasibility: medium
 reasoning_effort: medium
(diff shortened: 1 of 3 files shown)


ttraenkler:
Closing: the runtime-only fix in _safeSet was too broad. It conflated user assignment ([[Set]], must throw on getter-only) with object-literal data-property installation (DefineOwnProperty, must replace). Regressed test/language/expressions/object/prop-dup-get-data.js (`{ get foo(){}, foo: 1 }` threw during construction). A correct fix needs to distinguish [[Set]] from DefineOwnProperty at the codegen layer — routing to senior-dev with the error-model family (#581/#1536). Details in plan/issues/2017.

PR #1564, opened 2026-06-16
fix(host-interop): dispatch Promise-executor resolve/reject through __call_function (#2028)

ttraenkler:
## Summary

`new Promise(executor)` invokes the compiled executor body with **real host functions** for the `resolve`/`reject` params. The in-body call `resolve("ok")` was dispatched down the closure-struct `ref.test`/`ref.cast`/`struct.get`/`call_ref` path, which fails on a foreign callable: the cast nulls and `struct.get` traps `"dereferencing a null pointer"`, so the promise **rejected with a RuntimeError instead of fulfilling**. (Verified with a live repro now that the async CPS gate is ON post-#1796.)

## Fix

New predicate **`calleeIsPromiseExecutorParam(ctx, expr)`** (`src/codegen/expressions/calls.ts`), OR'd into the existing `hostCallFallback` gate. It fires only when the callee is a parameter of the arrow/function-expression passed **directly as arg 0 to `new Promise(...)`** and the param type is callable, routing the call through the already-wired `__call_function(fn, undefined, argsArray)` host arm (JS-host only — gated `!ctx.standalone && !ctx.wasi`). No `__make_callback` bridge surgery, no new host import.

## Scope decision (why executor-param, not "any externref callable param")

arch1's spec suggested "any externref-typed callable param." I implemented that first and it **regressed the #1941 dual-mode guarantee**: an ordinary local-closure callback (`apply(cb,v){return cb(v)}`) ALSO lowers its `cb` param to a bare `externref` local, so the broad predicate fired on it and pulled `__call_function`/`__js_array_new` into a self-contained module. The discriminator isn't the lowered type — it's **provenance**: executor params receive host fns; closure callbacks receive wasm closures (`ref.test $closure` succeeds, fast path taken). Scoping to the executor-param shape restores the #1941 guard.

## Validation

- `tests/issue-2028.test.ts` (6 cases): sync resolve → "ok"; reject → reason; resolve-twice → first wins; resolve(number); `.then` chain → 11; **#1941 dual-mode guard** (no host imports for a pure local-closure callback param).
- tsc clean; IR fallback budget unchanged.
- **Net new regressions: 0** — pre-existing-on-main failures unaffected (`promise-combinators` ×2, `optional-direct-closure-call` ×2, `accessor-side-effects` ×3, all verified identical on the main checkout).

## Out of scope

The `promise-combinators` ×2 failures (`await Promise.all(src.getPromises())`) are a **separate** defect — the host `declare`-class **method** `getPromises()` returns `__get_undefined`, unrelated to the resolve/reject dispatch. That host-method-return marshaling gap is what blocks dropping the `awaitedExprIsPromiseCombinator` exclusion in #1796; it remains a follow-up.

Closes #2028.

### DIFF
--- a/plan/issues/2028-promise-executor-resolve-reject-trap.md
+++ b/plan/issues/2028-promise-executor-resolve-reject-trap.md
@@ -1,10 +1,12 @@
 ---
 id: 2028
 title: "new Promise(executor): invoking the host-provided resolve/reject from wasm traps null deref — executor pattern fully broken in JS-host mode"
-status: ready
-sprint: 62
+status: done
+assignee: ttraenkler/sen-b
+completed: 2026-06-16
+sprint: 63
 created: 2026-06-10
-updated: 2026-06-12
+updated: 2026-06-16
 priority: high
 feasibility: hard
 reasoning_effort: high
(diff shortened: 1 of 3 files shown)


ttraenkler:
Closing as duplicate of #1563 (same #2028 fix — dispatch new Promise(executor) resolve/reject via __call_function, same calls.ts change, opened 1 min apart). #1563 is the keeper.

PR #1571, opened 2026-06-16
fix(#2025): extracted-method null-this throws catchable TypeError, not a trap

ttraenkler:
## #2025 — extracted-method null-`this` traps uncatchably

```ts
class A { x = 42; m(): number { return this.x; } }
const a = new A(); const f = a.m;
try { return "got:" + f(); } catch (e) { return "threw"; }
// before: wasm trap "dereferencing a null pointer" ESCAPES the try/catch
// node:   "threw" (TypeError: this is undefined)
```

The method-extraction trampoline (`buildTrampolineThisSlot`, `closures.ts` —
shared by the object-literal, cached class-method, and per-call-site paths, all
rebuilt by `finalizeMethodTrampolines`) forwarded `ref.null` for the unbound
`this`. The method body's `this.x` then did `struct.get` on a null ref → an
**uncatchable Wasm trap** that escaped the user's `try/catch`.

### Fix
- When the method body **actually reads `this`** (`local.get 0`, detected by
  `methodReadsThisParam` walking the now-final compiled body via
  `walkInstructions`), the trampoline's null-receiver arm throws a **catchable
  JS TypeError**. When the method never touches `this`, it keeps forwarding the
  harmless `ref.null`, so extracting a `this`-free method still works (the
  issue's required non-regression case).
- `buildTypeErrorThrow(ctx, msg, fctx?)` generalises `buildDestructureNullThrow`
  (`destructuring-params.ts`) — a JS-host + standalone (`__new_TypeError` /
  `__throw_type_error`) catchable throw as `Instr[]`.
- `ensureTypeErrorThrowImports` pre-registers the throw's late imports **once**
  before the `finalizeMethodTrampolines` loop, so adding them mid-loop can't
  shift the already-built `call methodFuncIdx` forwarders (invalid-module
  hazard I hit during development).

### Acceptance criteria — verified
- ✅ Repro returns "threw" with a real `TypeError` (`e instanceof TypeError`)
- ✅ bound/direct calls unchanged (`a.m()` → 42)
- ✅ extracting a `this`-free method still works (no spurious throw)
- ✅ value-receiver method dispatch unchanged; standalone mode works

`tests/issue-2025.test.ts` — 7/7. No regressions vs. baseline across the
trampoline / `this` / generator-method suites (#1118, #1394, #1602, #1636,
#1702, #1672, generator-methods). typecheck / lint / format clean.

Sets issue #2025 `status: done` (self-merge path).

### DIFF
--- a/plan/issues/2025-extracted-method-trap-not-catchable-typeerror.md
+++ b/plan/issues/2025-extracted-method-trap-not-catchable-typeerror.md
@@ -1,10 +1,12 @@
 ---
 id: 2025
 title: "calling an extracted method (const f = a.m; f()) traps uncatchably instead of throwing catchable TypeError"
-status: ready
+status: done
 sprint: 63
 created: 2026-06-10
-updated: 2026-06-12
+updated: 2026-06-16
+completed: 2026-06-16
+assignee: ttraenkler/tld-2108
 priority: low
 feasibility: medium
 reasoning_effort: medium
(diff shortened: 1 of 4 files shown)


ttraenkler:
Closing without merging: this fix introduced 75 test262 regressions (net -75 pass, 0 improvements, 71/75 with wasm-hash change — so real, not drift; buckets: wasm_compile 40, runtime_error 35). The method-extraction trampoline (`buildTrampolineThisSlot`) is a shared, fragile path also reached via #2015 `__call_fn_method_N` dispatch and the #1602/#1636 paths; throwing on its null-`this` arm fires in far more cases than just unbound extraction (receiver set but not `ref.test`-matching the exact struct, subclass/boxed receivers), breaking valid method calls and emitting malformed Wasm. Given #2025 is priority:low, the right fix is at the `this`-deref site, not the trampoline. Reverting #2025 to `ready` with findings; the branch is preserved for a future attempt.

PR #1595, opened 2026-06-16
fix(#2119): module code gets an unmapped arguments object

ttraenkler:
## #2119 — module code (always strict) got a *mapped* arguments object

ES modules are always strict ([§11.2.2](https://tc39.es/ecma262/#sec-strict-mode-code)), and strict functions get an **unmapped** arguments object ([§10.4.4](https://tc39.es/ecma262/#sec-arguments-exotic-objects)): a parameter write must NOT be visible through `arguments[i]`. `isStrictFunction` didn't infer module strictness, so every directive-less function got the mapped (sloppy) arguments — miscompiling genuine module input.

```ts
function f(a: number): string { a = 99; return "" + arguments[0] + "," + arguments.length; }
export function test(): string { return f(5); }
// wasm: "99,1"   node (.mjs): "5,1"
```

### Change

`isStrictFunction` (`src/codegen/helpers/is-strict-function.ts`) now ORs in a new `isModuleSourceFile(sf)` at the SourceFile-scope branch: true iff the file carries a top-level `import`/`export` (TypeScript's internal `externalModuleIndicator`) or an ESM `impliedNodeFormat`.

It deliberately does **NOT** key on `scriptKind` — the comment's original concern. The test262 harness compiles sloppy `.js` cases with `fileName: "test.ts"` (→ `scriptKind: TS`), so keying on that would wrongly unmap all of them. Verified empirically: a sloppy `.js` source compiled as `test.ts` has `externalModuleIndicator === undefined` → stays sloppy/mapped; a source with a top-level `export` has it set → unmaps. This is the genuine-signal distinction the issue's Fix-direction note asked for (without needing a flag).

`isStrictFunction` has exactly 4 consumers, all the `emitArgumentsObject` mapped/unmapped decision (function-body.ts, literals.ts, nested-declarations.ts ×2), so the change is scoped to the arguments object — it doesn't alter function-declaration scoping or other strict-mode behaviours.

### Acceptance criteria

- [x] Module-input repro matches Node — numeric variant returns `51` (unmapped) not `991` (mapped), verified via `assertEquivalent`.
- [x] test262 sloppy-mode arguments mapping does not regress — sloppy source (no import/export) stays mapped; unit test pins it. Broad conformance delta validated by sharded CI.
- [x] `"use strict"`-directive path unchanged (unit test).

### Validation

- `tests/equivalence/issue-2119-module-strict-arguments.test.ts` — 5/5: 1- and 2-param module-unmapped-arguments equivalence (vs Node) + unit `isStrictFunction` checks for sloppy(false)/module(true)/`"use strict"`(true).
- No new failures across arguments / strict / closures / function equivalence suites. (One pre-existing failure in `arguments-nested-and-loops` — a function-decl-in-block hoisting bug — fails identically on `main`, unrelated.)
- `npm run typecheck` + `npm run lint` (Biome) clean.


ttraenkler:
## Do NOT merge — needs the pre-wrap approach (handoff spec for #2119)

This PR's post-wrap `externalModuleIndicator` approach is **defeated by the test262 harness** and fails the conformance gate. Captured here as the spec for whoever picks it up (senior/architect lane).

### What CI showed
`check for test262 regressions` FAILED: **net −8** (14 regressions, 6 improvements). All 14 are `test/language/arguments-object/mapped/*` (`flags: [noStrict]`, sloppy) tests that must stay **mapped**. Cross-checked against other recent PRs (all pass the gate) → genuinely caused by this change, not baseline drift.

### Root cause
`tests/test262-runner.ts` `wrapTest` (~line 1792, see also ~346) wraps every test in a synthetic `export function test()`. That `export` sets `externalModuleIndicator` on the **wrapped** SourceFile, so `isModuleSourceFile` here fires for *every* test262 case — including sloppy `noStrict` ones — and unmaps their `arguments`, breaking the `mapped/*` suite. This is exactly the over-fire the original `is-strict-function.ts` comment warned about.

(My earlier empirical probe looked clean because it used the **unwrapped** `doCompile` path in `scripts/test262-worker.mjs`; the conformance runner uses the wrapping `test262-runner.ts` path. So post-wrap detection cannot distinguish "user wrote `export`" from "harness added `export`".)

### Correct fix direction (matches the issue's own note)
Detect genuine top-level `import`/`export` in the **ORIGINAL user source, before wrapping**, and thread an `unmapped`/`moduleStrict` flag through the compile pipeline into `isStrictFunction` / `emitArgumentsObject` — rather than reading `externalModuleIndicator` off the post-wrap SourceFile. `wrapTest` must record/forward the pre-wrap module-ness so the harness path stays sloppy/mapped while genuine module input unmaps.

### Reusable from this branch
- `isModuleSourceFile(sf)` (the genuine-signal check: `externalModuleIndicator` / ESM `impliedNodeFormat`, deliberately NOT `scriptKind`) — correct for the **pre-wrap** source.
- `tests/equivalence/issue-2119-module-strict-arguments.test.ts` — module-unmapped equivalence cases + unit `isStrictFunction` checks (sloppy=false / module=true / `"use strict"`=true).

Leaving this PR open as the reference; #2119 released back to `ready` for the senior/architect lane.

PR #1630, opened 2026-06-17
fix(#2029): retire __new_<Builtin> + __tag_user_class host-import leaks for standalone extends-Error

ttraenkler:
## What

Under `--target standalone`/`wasi`, a `class X extends Error/TypeError/RangeError/…` subclass leaked the `env.__new_<Parent>` and `env.__tag_user_class` HOST imports, making the module non-instantiable without a JS runtime. This is the runtime residual of #2029 after PR-1 fixed the emit crash.

## Fix (3 coordinated changes)

**`src/codegen/class-bodies.ts`**
1. Both subclass ctor sites (implicit-ctor `isExternrefBacked` ~1506, explicit `super(msg)` ~2272) register the Wasm-native `__new_<Parent>` via `emitWasiErrorConstructor` (builds a `$Error_struct`) before the `ensureLateImport` lookup, gated on `(wasi||standalone) && isWasiErrorName(parent)`. The lookup resolves to the native func → no host import. Mirrors `new-super.ts:1981` for direct `new Error(...)`.
2. `__tag_user_class` registration skipped under no-JS-host — it only feeds the `__instanceof` HOST import (unavailable standalone); standalone instanceof resolves at compile time / via the native `$tag` field. Mirrors the PR-1 `emitSetSubclassProto` skip.

**`src/codegen/property-access.ts`**
3. `.message`/`.name`/`.stack` on a user error subclass resolves the receiver to its builtin error ancestor (`classBuiltinParentMap`→`classParentMap` walk), reading the native struct field instead of the null `__extern_get` path.

## Validation

`extends Error` subclasses now construct, satisfy `instanceof`, and read `.message` natively standalone with **ZERO `env.*` imports** — verified by instantiating with a truly empty `{}` import object. Host mode unchanged (all gates on no-JS-host flags). `tsc --noEmit` clean; prettier clean. Test: `tests/issue-2029-subclass-builtin-standalone-emit.test.ts` (new 'runs standalone with zero host imports' block, 13/13 pass).

## Scope

Residual (pre-existing, NOT introduced): a multi-level `B extends A extends Error` returns `instanceof Error === false` (the `tryStaticInstanceOf` transitive-ancestor walk gap) — confirmed 0 on main pre-fix; this PR removes the leak, not that gap. Direct (1-level) `instanceof Error` is correct. Issue stays `in-progress`.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -339,7 +339,7 @@ The issue frontmatter `status:` field tracks where an issue is, set by whichever
 3. Update `plan/issues/backlog/backlog.md` if the issue was listed there
 
 <!-- AUTO:conformance-start -->
-**test262 conformance**: 31,353 / 43,135 (72.7 %) — baseline a3fd74c7, 2026-06-16T10:35:16Z
+**test262 conformance**: 31,357 / 43,135 (72.7 %) — baseline unknown, 2026-06-17T03:16:20.635Z
 <!-- AUTO:conformance-end -->
 
 ### Sprint History
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ STATUS.md rather than duplicating numbers that go stale. Standalone
 README until the current standalone regression is fixed.
 
 <!-- AUTO:conformance-start -->
-**test262 conformance**: 31,353 / 43,135 (72.7 %) — baseline a3fd74c7, 2026-06-16T10:35:16Z
+**test262 conformance**: 31,357 / 43,135 (72.7 %) — baseline unknown, 2026-06-17T03:16:20.635Z
 <!-- AUTO:conformance-end -->
 
 ## Current Status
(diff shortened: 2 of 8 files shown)


ttraenkler:
Closing as **superseded by #1536c** (merged via #1604). 

I verified on current main that the #2029 runtime residual is fully resolved: `class MyErr extends Error {}` constructs with **zero `env.*` host imports** and `.message`/`.name`/`.stack` read the native `$Error_struct` field. #1536c independently implemented all three fixes this PR contained — the native `__new_<Parent>` ctor (`emitWasiErrorConstructor`), the standalone `__tag_user_class` skip, and the user-error-subclass field read. Merging this PR would only duplicate them (double import + double registration).

The one nuance this PR added — transitive multi-level `B extends A extends Error` resolution — is a partial mitigation of the same class that #2188 (filed by #1536c) tracks properly via per-user-class brands. No unique value remains here.

PR #1634, opened 2026-06-17
fix(#2162): native Set.prototype.forEach in standalone mode

ttraenkler:
## What

`Set.prototype.forEach(cb)` was the one remaining broken collection method in standalone: the #1510 native Set runtime served add/has/delete/clear/size but **not iteration**, so `s.forEach(...)` fell through to the generic path and produced **invalid Wasm**.

## Fix

The shared collection-forEach machinery (`tryCompileNativeCollectionForEach`, added for Map.forEach in #1527) already supports a Set via its `isSet` flag — it passes the element as BOTH `value` and `key` per §24.2.3.6. This wires Set's method dispatch (`tryCompileNativeSetMethodCall`, `set-runtime.ts`) to call it: a `forEach` branch intercepting **before** the add/has/delete/clear gate and before the receiver is compiled (the shared helper compiles its own receiver). One import line + one dispatch branch — no new runtime helper. `nativeStrings`-gated; host/gc unchanged.

## Validation

`tests/issue-2162-set-foreach.test.ts` (7 cases, `--target wasi`, asserts valid Wasm + ZERO `Set_*`/`Map_*` imports + values): value sum, value===key, insertion order, tombstone-skip after delete, SameValueZero dedupe, empty set, string elements. Map.forEach + standalone-set suites still green (19/19 total). `tsc --noEmit` + prettier clean.

## Scope

Slice 3 of #2162; issue stays `in-progress`. Remaining: Map/Set keys/values/entries + for-of iterators, `new Map/Set(iterable)`, ES2025 set-algebra.


ttraenkler:
Superseding-commit check: closing as redundant. The native `Set.prototype.forEach` standalone implementation this PR adds already landed on main via fc30dcc5c ("fix(#2162): native Set.prototype.forEach in standalone mode") — same `tryCompileNativeCollectionForEach(..., /* isSet */ true)` dispatch and an equivalent `tests/issue-2162-set-foreach.test.ts`. After merging main, this branch's functional `src/` diff is empty (comment-only). Closing to keep the queue clean; the keys()/values()/entries() iterator slice continues in #1642.

PR #1693, opened 2026-06-18
feat(#2026 PR-3): dynamic-new spread + new.target edges

ttraenkler:
## #2026 PR-3 — dynamic-`new` argument & meta edges

Follow-up to PR-1/PR-1b (the dynamic-`new K()` uniform ctor ABI, already on main via #1647/#1656/#1672). Two argument/meta edges of `emitDynamicNewFallback` were broken; this PR fixes both. Additive, `new-super.ts` only, no struct-shape change, helpers by name. The static `new C()` path is untouched.

### PR-3a — spread args (fixes an INVALID-Wasm crash)
`new K(...[a,b])` **and** `new K(...x)` previously emitted invalid Wasm (`extern.convert_any[0] expected anyref, found call of type i32`) — the whole module failed to instantiate. The per-arg eval loop compiled a `SpreadElement` verbatim, yielding an i32 (array length) that the downstream box rejected. `new K(4, ...[5])` silently returned `null`.

Fix: before emitting any code, flatten an array-literal spread via the existing shared `flattenCallArgs` (same compile-time flatten the static class-`new` path uses). A non-flattenable (variable) spread returns `false` so the caller falls through to the legacy path — a runtime miss, **not** a broken module. The bail-before-emit ordering is load-bearing.

Result: `new K(...[4,5])` → 9, `new K(4,...[5])` → 9, `new K(...someVar)` instantiates cleanly (runtime variable-spread drive is a noted follow-up, matching how even the static path defers non-literal spread to `compileSpreadCallArgs`).

### PR-3b — new.target
`new.target === K` read `0` inside a dynamically-constructed ctor (the dynamic path never set the new-target global). Fix: call the shared `emitSetNewTargetBeforeCall(ctx, fctx.body, className)` before the ctor call, mirroring the static path. No-op unless `ctx.usesNewTarget`, so zero cost otherwise.

### Validation
- New `tests/issue-2026-dynamic-new-edges.test.ts` (6 cases): array-lit spread, mixed spread, non-flattenable-spread no-crash, new.target true, new.target discrimination, plain-arg PR-1 regression guard.
- All 13 existing #2026 tests still green; `issue-2023` (new.target) + `class-expressions` regression-clean.
- tsc + prettier + biome-lint clean.

### Still open under #2026 (separate slice, not this PR)
PR-2: `.constructor === A` for an externref/`any`-typed receiver (`make(A).constructor` still reads 0; the static receiver already works). Needs a `__tag`→`__class_<Name>`-singleton dispatch in `property-access.ts`.

### DIFF
--- a/plan/issues/2026-class-as-first-class-value.md
+++ b/plan/issues/2026-class-as-first-class-value.md
@@ -2,7 +2,7 @@
 id: 2026
 title: "classes are not first-class values: new K() on a parameter throws 'No dependency provided for extern class', .constructor identity broken"
 status: in-progress
-assignee: ttraenkler/sdev-async2
+assignee: ttraenkler/sdev-ctor
 sprint: 63
 created: 2026-06-10
 updated: 2026-06-18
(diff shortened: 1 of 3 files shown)


ttraenkler:
Superseded by the split per tech-lead: this bundled PR-3a (spread) + PR-3b (new.target). PR-3a — the invalid-Wasm spread crash fix — now lands alone as #1699 (with the variable-spread case hardened to a loud compile-time refusal). PR-3b (new.target) and PR-2 (.constructor) follow as their own PRs.

PR #1734, opened 2026-06-19
fix(#2371): stop standalone for-in from leaking __for_in_* host imports

ttraenkler:
## #2371 — standalone for-in leaks `env::__for_in_*` host imports

Found in a standalone host-import-leak hunt (object/class/property lane).
`for (const k in obj)` in `--target standalone --nativeStrings` registered four
unsatisfiable JS-host imports — `env::__for_in_keys/_len/_get/_has` — so the
module **failed to instantiate** against an empty host, for every receiver shape
(object literal / `any` / array / class instance).

**Root cause:** `declarations.ts` calls `addForInImports()` whenever
`state.forInFound`, **without** the `!ctx.standalone && !ctx.wasi` host-mode
guard its sibling iterator-import finalizers carry.

### Slice 1 — stop the leak, refuse cleanly
- `declarations.ts`: gate `addForInImports` on `!ctx.standalone && !ctx.wasi`.
- `statements/loops.ts`: in the no-host-import fallback, refuse standalone for-in
  with a clear compile-time diagnostic instead of the prior naive
  static-property unroll.

**Why not a native for-in here?** A correct one needs two broad capabilities a
static unroll can't provide, so an unroll would *silently miscompute*:
1. **Native key enumeration over the runtime value** — array keys are the integer
   indices `"0".."length-1"`, NOT the static array type's `.getProperties()`
   (Array.prototype methods + `length`); an `any` receiver has no static key set.
2. **Native dynamic property read by a runtime string key** — the dominant
   `for (k in o) … o[k] …` body reads by the runtime key, and `o[k]` with `k` a
   runtime string returns `0`/default in standalone today (verified independently
   of for-in).

Both are the #2371 phase-2 follow-up. Refusing is **strictly better** than the
prior un-instantiable module, and does not regress any test262 standalone pass
count (those tests could never run standalone — the CE just re-buckets a
never-passing test while removing the host-import leak).

**Host mode is unchanged** — the `__for_in_*` imports are still registered and
the JS-host enumeration path is byte-identical.

### Tests
`tests/issue-2371-standalone-forin.test.ts` (6): object/array/any/class
standalone for-in emits ZERO `__for_in_*` imports and a clear CE; a no-for-in
program stays clean; host mode still registers all four imports. The #2066
deleted-property + #forin host suites (11) stay green. tsc + prettier +
biome(lint) + stack-balance + coercion-sites + any-box gates clean.

### DIFF
--- a/plan/issues/2371-standalone-for-in-host-import-leak.md
+++ b/plan/issues/2371-standalone-for-in-host-import-leak.md
@@ -1,0 +1,108 @@
+---
+id: 2371
+title: "standalone: for-in enumeration leaks __for_in_* host imports (un-runnable in --target standalone)"
+status: in-progress
+assignee: ttraenkler/sd1
+sprint: 64
+created: 2026-06-19
+updated: 2026-06-19
+priority: high
+feasibility: medium
+reasoning_effort: medium
+task_type: bugfix
+area: codegen, standalone
+language_feature: for-in, enumeration
+goal: standalone-mode
+origin: "2026-06-19 sd1 standalone host-import-leak hunt (object/class/property lane)"
+---
+
+# #2371 — standalone for-in leaks `__for_in_*` host imports
+
+## Problem
+
+`for (const k in obj)` in `--target standalone --nativeStrings` registers four
+unsatisfiable JS-host imports — `env::__for_in_keys`, `env::__for_in_len`,
+`env::__for_in_get`, `env::__for_in_has` — so the module **fails to instantiate**
+against an empty import object. for-in is effectively un-runnable in standalone
+mode for EVERY receiver shape (object literal, `any`, array, class instance).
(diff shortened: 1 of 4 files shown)


ttraenkler:
Closing — net -89 on the standalone shard, 0 improvements. The refusal-CE approach is wrong: the test262 standalone harness DOES provide the `__for_in_*` host imports, so for-in actually works in standalone for real tests. The hard CE turned 89 passing tests (built-ins/{decodeURI,encodeURI,JSON,global,...}) into compile_errors. The for-in import 'leak' is benign in the harness; the only correct fix is a working native for-in (native key enumeration + dynamic property read by runtime key — #2371 phase 2), not a refusal/gate. Downgrading #2371 to phase-2-only.

PR #1740, opened 2026-06-19
fix(#2376): route Uint8ClampedArray methods natively (add to BUILTIN_TYPES)

ttraenkler:
## #2376 — `Uint8ClampedArray` methods mis-dispatch to host extern-class imports

`new Uint8ClampedArray([1,2,3,4]).reduce((a,x)=>a+x,0)` (and every other
`Uint8ClampedArray` method) routed through the **host extern-class dispatch**
instead of the **native typed-array array-method path** all the other typed
arrays use:
- **GC mode:** invalid Wasm — the synthesized `env.Uint8ClampedArray_reduce`
  import's externref `self` param mismatched the GC vec receiver
  (`call[0] expected externref, found … (ref null N)`).
- **standalone/WASI:** leaked an unsatisfiable `env.Uint8ClampedArray_*` import.

Every sibling typed array (`Int8/Uint8/Int16/Uint16/Int32/Uint32/Float32/
Float64Array`) works; only `Uint8ClampedArray` was broken.

### Root cause (one-line omission)
`isExternalDeclaredClass()` (`src/checker/type-mapper.ts`) treats a
`declare var X: XConstructor` lib type as an extern class **unless** the name is
in `BUILTIN_TYPES`. That set listed every typed array **except
`Uint8ClampedArray`** — so `isExternalDeclaredClass(Uint8ClampedArray)` returned
`true`, and the method dispatch fired `compileExternMethodCall` before the native
`compileArrayMethodCall` path.

### Fix
Add `"Uint8ClampedArray"` to `BUILTIN_TYPES`, next to its `Uint8Array` sibling.
It already lowers to the same `(ref null $Vec[f64])` vec representation
(`TYPED_ARRAY_NAMES` / `typedArrayVecStorage` include it), so the native path
handles all its methods once `isExternalDeclaredClass` stops claiming it.

### Tests (`tests/issue-2376.test.ts`)
- `.reduce` value (was invalid Wasm), `forEach`/`map`/`indexOf`/`length` native
  dispatch, standalone **no `env.Uint8ClampedArray_*` leak**, and a
  sibling-typed-array (Uint8/Int8/Float64) no-regression case.
- `tsc` + prettier clean. The pre-existing `functional-array-methods` /
  `externref` test failures (a missing `tests/helpers.js`) fail **identically on
  `main`** with the change stashed — verified, not a regression.

### Out of scope
`Uint8ClampedArray` value-clamping on store (`a[0]=300→255`) is the f64-vec
representation gap shared with the other non-`Uint8Array` integer typed arrays
(#2159 packed-storage family), unaffected by this dispatch fix.

Closes #2376.


ttraenkler:
Note: the issue was renumbered #2376 → **#2379** (`2376` collided with the concurrently-merged `issue-2376-date-proto-value-read`, PR #1733). The issue file is `plan/issues/2379-uint8clampedarray-method-misdispatch-builtin-types.md` and the test is `tests/issue-2379.test.ts`. The BUILTIN_TYPES fix itself is unchanged. (PR title edit is blocked by a gh/GraphQL projectCards deprecation warning in this repo — title still shows #2376 but the content is #2379.)

PR #1741, opened 2026-06-19
feat(#2400): Wasm-native encodeURI/encodeURIComponent (standalone) — S1+S2

ttraenkler:
## #2400 — Wasm-native URI percent-encoding (S1 + S2: encode)

Under `--target wasi`/`--target standalone` there is no JS host, so
`encodeURI` / `encodeURIComponent` silently fell through to a `ref.test`/
`ref.cast` of the argument and returned `undefined` (~133
`built-ins/{encodeURI,encodeURIComponent}` test262 fail; the `env.*` import
was not even satisfiable).

This implements **S1 + S2** of sd3's 4-slice plan together — `encodeURI` and
`encodeURIComponent` share a single native helper and differ only by a mask.

### What changed
- **`src/codegen/uri-encoding-native.ts`** (new) — `__uri_encode(s, preservedMask)`,
  a pure-Wasm implementation of ECMAScript §19.2.6.5 Encode (hand-built
  `Instr[]`, #679/#682 dual-backend pattern): UTF-8 transcode (RFC 3629, 1–4
  octets), surrogate-pair handling, per-variant preserved-set mask
  (`encodeURIComponent = uriUnescaped`; `encodeURI = + uriReserved ∪ #`),
  uppercase `%XX` output, catchable **URIError** on unpaired surrogates.
- **`declarations.ts`** — register the helper at the URI finalize under
  `ctx.wasi || ctx.standalone`; host mode keeps the `env.*` imports unchanged.
- **`calls.ts`** — route the encode names to `__uri_encode` with the
  per-function mask (decode names still use the host import — that is S3/S4).

The helper's `funcIdx` is claimed only *after* `emitWasiErrorConstructor`
appends `__new_URIError`, so its module slot is never shifted out from under it.

### Validation
`tests/issue-2400-uri-encoding.test.ts` — 2 tests green. The standalone module
is compiled with `target: "wasi"` and instantiated with an **empty import
object** (proving no host); exports return numbers (`.length` / `.charCodeAt`)
so results read back without a string-marshaling host. Covers ASCII
passthrough + percent-encoding, all `uriUnescaped` marks, reserved-set
behavior (escaped by component, preserved by encodeURI), 2/3/4-byte UTF-8
(`©`/`€`/`😀`), URIError on lone/unpaired surrogates, empty string, and a
host-mode env-import regression check.

### Remaining (follow-on)
S3/S4 = native `__uri_decode` for `decodeURI`/`decodeURIComponent` (the
`%XX` → UTF-8 → code-point reassembly + URIError on malformed). The issue
stays `in-progress` until those land.

### DIFF
--- a/plan/issues/2400-native-uri-percent-encoding.md
+++ b/plan/issues/2400-native-uri-percent-encoding.md
@@ -1,0 +1,131 @@
+---
+id: 2400
+title: "Wasm-native decodeURI / encodeURI / decodeURIComponent / encodeURIComponent (percent-encoding, ~133 test262)"
+status: in-progress
+assignee: ttraenkler/sd5
+sprint: 64
+created: 2026-06-19
+updated: 2026-06-19
+priority: medium
+feasibility: medium
+reasoning_effort: high
+task_type: feature
+area: codegen
+language_feature: global-functions
+goal: standalone-mode
+test262_bucket: uri-encoding
+test262_count: 133
+---
+
+# #2400 — Wasm-native URI percent-encoding
+
+## Problem
+
+`decodeURI` / `decodeURIComponent` / `encodeURI` / `encodeURIComponent` are
+dispatched to **host imports** (`env.decodeURI` etc., `calls.ts:8892` +
+`declarations.ts:504`). Under `--target standalone`/`wasi` there is no JS host,
+so each leaks an unsatisfiable `env.*` import → instantiation failure. ~133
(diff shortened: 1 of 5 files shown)


ttraenkler:
Closing — renaming the URI work to #2500 to avoid a cross-session issue-id collision (#2378 collides with the parallel proto-read session's PR #1739). Reopening under branch issue-2500-uri-percent-encoding. Same encode (S1+S2) implementation.

PR #1753, opened 2026-06-19
fix(#2506): standalone any[] join/toString — boxed-any element via __extern_toString

ttraenkler:
## #2506 — standalone `any[].join()` / `any[].toString()` emitted invalid Wasm

```ts
const a: any[] = [1, 2, 3];
a.join(",");   // INVALID Wasm (standalone)
a.toString();  // INVALID Wasm
```

```
local.set[0] expected type (ref null 6), found ref.as_non_null of type (ref extern)
```

`number[]` / `string[]` / `new Array(N)` joins were all VALID — the defect is specific to a **boxed-any (`externref`) element array** (`any[]`, mixed-type literals).

### Root cause
`compileArrayJoinNative` (`src/codegen/array-methods.ts`) builds the element→string fold; its non-numeric/non-boolean branch assumed a `$NativeString` ref element and emitted `ref.as_non_null` to lift it to `(ref $AnyString)`. But an `any[]` element array is `$__arr_externref` — each element is a raw `externref` (`__box_number`/`__box_boolean`-boxed), so the non-null'd externref mistypes into the `(ref $AnyString)` result local. Same #2505-class boxed-any element-rep divergence, in the join fold.

### Fix
New `elemType.kind === "externref"` arm routes each boxed-any element through **`__extern_toString`** (externref → externref native string) — the *same* runtime ToString that `String(x)` and template literals (`` `${x}` ``) use for an `any` value — then `any.convert_extern` + `ref.cast $AnyString` for the concat fold. The late-import shift is flushed up front so the baked call funcIdx is stable.

**Not `__any_to_string`** (the `$AnyValue`-tag dispatcher): an `any[]` element is a `__box_number` externref, not a `$AnyValue` struct, so that path mis-stringifies it to `"[object Object]"`. A first attempt used it and produced **VALID-but-WRONG** output (`[1,2,3].join(",")` → `"[object Object],…"`); routing through `__extern_toString` gives the correct `"1,2,3"`.

### Tests
`tests/issue-2505-anyarray-join.test.ts` (9): `any[]`/mixed/boolean/string/empty `join` + `toString` produce **correct content** (asserted via `.length`/`.charCodeAt`, instantiated with empty imports `{}`); `number[]`/`string[]`/`boolean[]` regressions green. `#2074`/`#2502`/`#1993`/`#2106` suites pass; `tsc` clean.

(This branch also carries the `#2504` doc — file-verified diagnosis of the separate standalone `console.log(string)` host-bridge bug surfaced during this investigation; doc-only.)


ttraenkler:
I shipped a duplicate of this (#1755, now closing) — your fix landed first. One spec edge my version additionally covered that I don't see here: **§23.1.3.18 step 7d — a `null`/`undefined` array element joins as the empty string** (not `"null"`/`"undefined"`). Routing the element through `__extern_toString` alone renders `null`→`"null"`/`undefined`→`"undefined"`, so:

```ts
const a: any[] = [1, null, undefined, 2];
a.join(",");   // spec: "1,,,2"  — __extern_toString alone gives "1,null,undefined,2"
```

I guarded it by teeing the boxed element and testing `ref.is_null` OR `__extern_is_undefined` before the `__extern_toString` call, emitting `""` for the nullish case. Worth a follow-up test/guard on this PR if you want full join spec-correctness — here's the focused case I used:

```ts
// null/undefined elements join as empty string (§23.1.3.18 step 7d)
const a: any[] = [1, null, undefined, 2];
// a.join(",") === "1,,,2"
```

Closing #1755 as a dup of this (join) + #1756 (find).

PR #1754, opened 2026-06-19
fix(#2074,#66,#70): standalone join/toString of externref-element (any[]/empty/boxed-any) array

ttraenkler:
## Problem

In `--target standalone`, `Array.prototype.join()` / `toString()` over an **externref-element** vec (untyped `[]`, `any[]`, `(string|number)[]`, object arrays) emitted an **invalid Wasm module**:
```
local.set[0] expected type (ref null 6), found ref.as_non_null of type (ref extern)
```
This tripped even for an **empty** such array (`[].join()`), since the element-conversion's static type must match the fold accumulator's `(ref null $AnyString)` even when the loop never runs.

## Root cause

`compileArrayJoinNative`'s catch-all `else` arm `ref.as_non_null`'d the element assuming a `(ref null $NativeString)` string element. For an externref-element array that yields a `(ref extern)`, which mistypes into the `(ref null $AnyString)` accumulator.

## Fix (additive — one new element arm)

Add `else if (elemType.kind === "externref")` that stringifies each element through the native `__extern_toString` (§7.1.17, `ensureObjectRuntime`) — the **same** ToString `String(x)` uses — then casts its externref result to `(ref $AnyString)`. Critically `__extern_toString` (not the $AnyValue tag-dispatcher `__any_to_string`) recovers the boxed-number element a `new Array(N)`/`any[]` array stores (`$__box_number_struct`), so numeric `any[]` joins to `"1,2,3"` not `"[object Object]"`. `Array.prototype.toString` delegates to join via the same fold, inheriting the fix.

## Measured (standalone, upstream/main ed8c4f6e6)

- empty `any[]`/`[]` join → valid, `""` (was invalid module)
- `any[] = [1,2,3]`.join(",") → `"1,2,3"` (was invalid; `__any_to_string` gave `"[object Object]"`)
- `any[]` string join + `any[]` toString correct
- all 33 pre-existing join tests (#2074/#2088/#2105) pass — **0 regressions**
- 400-file `built-ins/Array/prototype` sweep: 1 invalid (no regression cluster)

Folds in **#66** (join invalid Wasm) + **#70** (any[]/boxed-any join & toString). Array-like-object join (`obj.join=Array.prototype.join`) and empty `new Array()` construction are separate lanes (#54 / #2379), excluded.

## Test

`tests/issue-2074.test.ts` — new "externref-element ... join is valid + correct" block (+11 cases incl. toString delegation), all green.

### DIFF
--- a/plan/issues/2074-standalone-externref-element-join-residual.md
+++ b/plan/issues/2074-standalone-externref-element-join-residual.md
@@ -1,0 +1,103 @@
+---
+id: 2074-residual
+title: "standalone: Array.prototype.join()/toString() of an externref-element (any[]/empty/boxed-any) array emits invalid Wasm (folds #66 + #70)"
+status: done
+assignee: ttraenkler/sdev-protoglue
+sprint: Backlog
+created: 2026-06-19
+updated: 2026-06-19
+completed: 2026-06-19
+priority: high
+feasibility: medium
+reasoning_effort: max
+task_type: conformance
+area: codegen
+language_feature: builtins, arrays, strings
+goal: standalone-mode
+related: [2074, 2075, 2088, 2105, 2379, 1917, 1461]
+origin: "2026-06-19 — sdev-arrayrep flagged #66/#70: standalone join of any[]/empty array emits invalid Wasm"
+---
+
+## Problem
+
+In `--target standalone` (native strings), `Array.prototype.join()` /
+`Array.prototype.toString()` over an **externref-element** vec — an untyped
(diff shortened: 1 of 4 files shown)


ttraenkler:
Duplicate of #1753 (#2506, sdev-arrayrep's element-rep lane) — same compileArrayJoinNative __extern_toString fix, created ~8 min before mine. #1753's 9 tests already cover empty any[], any[] numeric/string/mixed/boolean, and toString delegation (a superset of mine). Closing in favor of #1753.

PR #1755, opened 2026-06-19
fix(#2505): standalone any[]/boxed-any join() & toString() invalid Wasm

ttraenkler:
## Summary

`any[].join(sep)` and `any[].toString()` emitted **invalid Wasm** under `--target standalone` (`local.set[0] expected (ref null \$AnyString), found ...`), failing instantiation.

**Root cause:** a boxed-any element array stores each element as a generic boxed value (externref), not a `(ref null \$NativeString)`. `compileArrayJoinNative`'s string-element arm took a bare `ref.as_non_null`, leaving the value typed externref — which can't `local.set` into the `(ref \$AnyString)` fold result local. `number[]`/`string[]`/`boolean[]`/`new Array(N)` were unaffected (#2505-class divergence, join lane).

**Fix:** route a boxed-any element through `__extern_toString` (the same ToString path `String(any)` uses, which correctly recovers a boxed number/boolean/string element — a bare `__any_to_string` over `any.convert_extern` misses the host-boxed-number externref shape and mis-renders `1` as `[object Object]`), with the §23.1.3.18 step-7d `null`/`undefined` → empty-string guard. `Array.prototype.toString` delegates to join, so it's fixed by the same change.

## Verification

`tests/issue-2505.test.ts` (6 new) — valid Wasm + spec-correct values:
- mixed `[1,"x",true]` → `"1,x,true"`
- numeric `[1,2,3]` → `"1,2,3"` (not `[object Object]`)
- `[1,null,undefined,2]` → `"1,,,2"` (§23.1.3.18 step 7d)
- object `[1,{a:1},"z"]` → `"1,[object Object],z"`
- explicit separator; `toString()` delegation

Controls unchanged (number[]/string[]/boolean[]); existing #2074 join tests (12) still green; tsc clean.

### DIFF
--- a/scripts/coercion-sites-baseline.json
+++ b/scripts/coercion-sites-baseline.json
@@ -1,5 +1,5 @@
 {
-  "codegen/array-methods.ts": 19,
+  "codegen/array-methods.ts": 20,
   "codegen/async-scheduler.ts": 3,
   "codegen/binary-ops.ts": 37,
   "codegen/closed-method-dispatch.ts": 1,
--- a/src/codegen/array-methods.ts
+++ b/src/codegen/array-methods.ts
@@ -4903,6 +4903,33 @@ function compileArrayJoinNative(
     }
   }
 
+  // #2505 — boxed-any element (`any[]`): each element is a generic boxed value
+  // (externref/anyref), stringified through `__extern_toString` (the same path
+  // `String(any)` uses) in the per-element fold below. Resolve the import index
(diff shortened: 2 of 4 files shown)


ttraenkler:
Dup of arrayrep #1753 (any[].join/toString) + #1756 (any[].find/findLast) — same root cause + __extern_toString fix, theirs landed first. Posted my additional null/undefined→empty-string (§23.1.3.18 step 7d) join edge + test as a comment on #1753.

PR #1767, opened 2026-06-19
fix(#2508): nested dstr-param default struct.new arity via liveBodies patch

ttraenkler:
## #2508 — nested destructuring-param default object: short struct.new (invalid Wasm)

> Re-filed from the colliding `#2503` (that id is the unrelated ToPrimitive
> residual on main). This PR is the **invalid-Wasm / arity** half; the
> destructured-VALUE-FLOW half is filed as **#2509**.

### Root cause (NOT a `$shape` gap)
The colliding structs here carry no `$shape` field — this is the field-pad
hazard of a struct-type **field unification**, the same bug CLASS as #2158 for a
different patch.

A nested destructuring-param default object materializes into a **detached
`if.then` buffer** swapped onto `fctx.body` by a plain JS-local swap (not
`pushBody`, so absent from `fctx.savedBodies`), registered in `ctx.liveBodies`
for the late-import-shift coverage window (#2158). When a later same-shape object
literal grows the anon struct's field set (the inner `{x:4,y:5,z:6}` default
appends `y` to the 2-field `{x,z}` shape), `patchStructNewForAddedField`
retro-pads every existing `struct.new` — but it walked ONLY `mod.functions` +
`fctx.body` + `fctx.savedBodies`, **never `ctx.liveBodies`**. So the earlier
`struct.new` in the orphaned buffer stayed one operand short of the grown
3-field type → invalid Wasm ("struct.new need 3, got 2").

### Fix
Add a `ctx.liveBodies` traversal to `patchStructNewForAddedField`
(`src/codegen/expressions/late-imports.ts`), mirroring the late-import-shift
walk that already covers these detached buffers.

### Validation
- `tests/issue-2508-dstr-param-default-arity.test.ts` (5 cases): class method /
  class-expr static / plain fn / generator method / value-undef variant all
  compile to VALID Wasm.
- No regression: `basic-destructuring` + `destructuring-extended` +
  `destructuring-initializer` + `issue-2158`. `tsc --noEmit` clean.

### Scope — net-0 in the pass-gate (CE→fail, not →pass)
This clears the **24 invalid-Wasm compile_errors** (class/expr
`meth-…-dflt-obj-ptrn-prop-obj` + async/gen/private variants) — a real
robustness + bucket-clearing improvement — but those tests flip
`compile_error` → **fail** (assertion), not → pass: the destructured
**value-flow** when the outer default object fires is a separate pre-existing
bug, filed as **#2509**. So **no passes are banked** (net-0). Requesting a
tech-lead net-0 correctness-fix approval per the agreed plan.


ttraenkler:
Renumber note: this issue is now **#2512** (the arity fix) with the value-flow follow-up at **#2513** — per tech-lead, the interim #2508/#2509 and the original #2503 collided with a parallel session's ToPrimitive issue. The commits + issue files in this PR use #2512/#2513; the PR title still shows #2508 only because `gh pr edit` is erroring on a Projects-classic GraphQL deprecation in this gh version. Canonical id: **#2512**.

PR #1770, opened 2026-06-19
fix(#2042 R2): standalone object-key ToPropertyKey — obj[{toString}] illegal-cast

ttraenkler:
## #2042 R2 — standalone `obj[{toString}]` traps illegal-cast (ToPropertyKey)

```ts
const o: any = {};
const k: any = { toString: () => "abc" };
o[k];   // TRAP: "illegal cast"  (standalone)
```

A computed member access whose key is an **object** trapped: `__to_property_key` (#2042 S1, PR #1629) ToPropertyKey'd AnyString and boxed-number keys but returned a `$Object` key **unchanged**, so it reached the `ref.cast $AnyString` in the `$Object` runtime's `__obj_find`/`__obj_hash` and trapped.

### Fix
`__to_property_key` (`src/codegen/object-runtime.ts`) now routes a `$Object` key through **`__extern_toString`** (§7.1.1 ToPrimitive(string) → ToString — the canonical ToString that `String(x)` / template literals use), yielding the canonical string key before the cast. `__extern_toString` is registered **later** in the same `ensureObjectRuntime` pass (forward dependency), so the object-key arm is **spliced into** `__to_property_key`'s body (held by reference in `mod.functions`) once its funcIdx is known — inserted before the unchanged-fallthrough so Symbol/opaque keys still pass through. Reuses the built ToPrimitive engine (arch-dynshape #1472 spec R2); does NOT rebuild it. standalone-only; number/string/variable keys unchanged.

### Measured (standalone, instantiate)
`obj[{toString}]` read/write/`in`/`delete` all work (11/22/1/0); a `{toString}` key yielding a numeric string matches the numeric slot (`o[2]=42; o[{toString:()=>"2"}]`→42). string/integer/variable keys unchanged.

### Tests / regression
`tests/issue-2042-r2-topropkey-object.test.ts` (8) green. `tsc` clean; `coercion-sites` baseline +2 (sanctioned engine routing). `#2042` S1/S3 + Reflect + object-keys suites unchanged — the one pre-existing `non-integer numeric key` failure in `issue-2042.test.ts` (`o[1.5]=4; return o[1.5]`→0) is **broken on main without this change** (a separate float-key READ-path gap), out of scope.

### Out of scope (#1917 engine gap)
A `valueOf`-**only** object whose `valueOf` returns a **number** stringifies to `"[object Object]"` (`__to_primitive` valueOf-number recovery), so `obj[{valueOf:()=>2}]` doesn't match `o[2]`. This R2 fix removes the illegal-cast **trap** and fixes the common `toString`-keyed shape; the valueOf-number recovery is the #1917 engine owner's concern.


ttraenkler:
Updated to also land the **#2042 S3 `Object.is` residual** (#80): native `__object_is` (SameValue §7.2.10) — number bit-compare (NaN==NaN, +0!=-0), boolean/bigint unbox, string value-equality, else ref identity; added to `OBJECT_RUNTIME_HELPER_NAMES`. Verify-before-claim found the S3 read-natives already shipped (PR #1639); `Object.fromEntries` deferred (needs native iteration). `tests/issue-2042-s3-object-is.test.ts` (7) green, wasm-opt -O3 passes, 0 reg.

PR #1806, opened 2026-06-21
fix(#2160): standalone Number.prototype.toLocaleString() via number_toString

ttraenkler:
## Summary

`Number.prototype.toLocaleString()` with **no arguments** compiled in host (gc) mode but was a hard **standalone/WASI `compile_error`**: the call routed to the host-only `__extern_toLocaleString` import, which has no native fallback, producing the `'__extern_toLocaleString' (dynamic-shape …)` refusal. Every numeric `toLocaleString()` was a standalone CE — part of the #2160 host-vs-standalone String/Number gap.

## Spec

ECMA-262 §21.1.3.4: with no ECMA-402 (Intl) implementation, the result of `Number.prototype.toLocaleString()` equals `ToString(value)` base 10. The implementation-defined locale formatting reduces to plain ToString.

## Fix

Add a `toLocaleString` arm gated on `(ctx.standalone || ctx.wasi)` + 0 args, in:
- `src/codegen/declarations.ts` (number-method import collector) — registers the native `number_toString` helper
- `src/codegen/expressions/calls.ts` (call-site router) — routes to the **existing** `number_toString` lowering (identical to the 0-arg `.toString()` arm), unwrapping the native-string externref once for native-strings mode

**Host (gc) mode is intentionally excluded** — it keeps `__extern_toLocaleString` so real Intl grouping is preserved (`(1234).toLocaleString() === "1,234"`). A call WITH a locale argument also stays on the host path (real Intl out of scope).

No new coercion matrix — reuses the sealed engine helper `number_toString`. The #2108 coercion-sites drift baseline is refreshed for the two new `number_toString` references (declarations 17→18, calls 18→19).

## Validation

- `tests/issue-2160-number-tolocalestring-standalone.test.ts` (8/8): standalone compiles + no `__extern_toLocaleString` leak; exact char content for int/negative/fractional/zero/variable receivers; host-mode no-regression guard (host still imports `__extern_toLocaleString`).
- Existing `tests/issue-2160-number-parse.test.ts` (16/16) + native-strings (91) green.
- tsc + prettier + format:check + coercion-sites gates clean. Standalone-only; host untouched.

## Still open (unchanged)

Wrapper objects `new String`/`new Number` construction still leaks `__new_String`/`__new_Number` (no native wrapper-object representation — gated on value-rep #2072/#2104); `str.replace(fn)` standalone (RegExp engine limitation).


ttraenkler:
Closing: superseded. Branched from the fork origin/main which was 1188 commits behind upstream, so this never triggered CI. Probed the bug against upstream/main: toLocaleString re-done from upstream/main as #1824.

PR #1808, opened 2026-06-21
fix(#2203): capture-aware native-generator candidacy fixes standalone funcidx desync

ttraenkler:
## #2203 — Standalone closure-capturing native generator emitted invalid funcidx

A nested `function*` that **both** captures an enclosing-function variable **and** yields, compiled `--target standalone`, failed at binary-emit:

```
Codegen error: function index out of range — undefined at function 'g'
```

Minimal repro:
```ts
export function test(): number { let f = 0; function* g() { yield f; } return f; }
```
Dropping either the capture or the yield compiled cleanly.

### Root cause
Same funcidx-desync *class* as #2043/#1461, but it was a **predicate disagreement**, not a `flushLateImportShifts` miss:

- The host-import predicate `sourceNeedsGeneratorHostImports` decides whether to register the JS-host generator buffer imports (`__gen_create_buffer`, `__create_generator`, …) in standalone, via `isNativeGeneratorCandidate`.
- `isNativeGeneratorCandidate` **ignored captures**, so a capturing generator was deemed *native* ⇒ host imports **skipped**.
- But the emission gate in `nested-declarations.ts` only takes the native path when `captures.length === 0`, so the capturing generator **fell through to the host buffer path**, which bakes `funcMap.get("__gen_create_buffer")!`. With the import never registered, that resolved to `undefined` → baked into a `call` → the funcidx error.

### Fix (one root-cause change)
Make `isNativeGeneratorCandidate` capture-aware via a new `generatorCapturesEnclosingScope` helper (TS-checker based; counts **only enclosing-function** bindings, so a top-level generator reading a module global stays native). A capturing generator is now non-native everywhere, so the import predicate and the emission gate agree and the host-fallback imports are registered. No cast/guard over a stale index.

### Verified (standalone)
- minimal repro compiles (no funcidx error);
- `.next()`-driven capturing generators return correct captured values (incl. post-construction mutation through the boxed ref-cell);
- no-capture native generators still emit **zero** host imports (#2172 invariant intact);
- host-mode generators unaffected.

tsc + prettier clean. Pre-existing failures `tests/issue-1169f-7{a,b}.test.ts` (IR/legacy yield-sequence equivalence) and the broken `tests/generator-nested.test.ts` import were confirmed failing identically on `origin/main` — not regressions.

### Follow-up (escalated to architect, separate issue)
`for-of` / array-destructuring over a **host-buffer** generator in standalone still traps (`null deref`) / returns `NaN` — a **pre-existing, capture-independent** iteration gap (fails the same for a no-capture *native* generator on main). The standalone iterator runtime only drives the canonical `$Vec`, not a host JS `Generator` externref. So the `*ary-ptrn-elision*` destructuring cluster flips CE → **fail** (no longer compile_error) rather than CE → pass until that gap is closed. Details in the issue file.

### Tests
- `tests/issue-2203-funcidx-gen-capture.test.ts` (6 cases)


ttraenkler:
Closing as duplicate: #2203 was already fixed and merged on `main` by another agent with an equivalent fix (`generatorCapturesOuterScope` wired into `sourceNeedsGeneratorHostImports` — the same root cause and mechanism this PR implements as `generatorCapturesEnclosingScope`). Independent convergence on the identical diagnosis. The merged fix also split the eager-generator over-consumption value bug to #2566. No need to land this duplicate.

PR #1809, opened 2026-06-21
fix(#2160): standalone Number.prototype.toLocaleString via native number_toString

ttraenkler:
## Summary

`(n).toLocaleString()` on a **number receiver** CE'd in standalone / WASI with `'__extern_toLocaleString' (dynamic-shape object)`: the generic `toLocaleString` fallback routes the receiver to the host `__extern_toLocaleString` import, but a bare number is not an extern object so the standalone codegen refuses it. (Host mode worked via real Intl, e.g. `"1,234"`.)

## Fix (no-JS-host only)

§21.1.3.4 — there is no Intl in standalone/WASI and the no-Intl default is implementation-defined. For a number receiver in no-JS-host targets, delegate `toLocaleString` to the base-10 `Number.prototype.toString` (native `number_toString`), matching the `toLocaleString → toString` delegation already used elsewhere. The import-collector pre-registers `number_toString` for the `toLocaleString` call shape (gated `standalone || wasi`) so the call lowers without a late module-function shift; the call-site emits the native call and unwraps to a native string. Locale/options args are ignored.

**JS-host mode is untouched** — it keeps the Intl-backed `__extern_toLocaleString` (grouping preserved), so no host regression. Non-number receivers (Array/TypedArray/Date/object) also keep the host fallback unchanged.

## Scope

One independent slice of #2160. The `new String`/`new Number` wrapper method-dispatch residual is BLOCKED on a native wrapper constructor (value-rep #2072/#2104/#1910-S2, senior-dev) and is NOT in this PR; `String.raw` invalid-standalone-binary is a separate slice. #2160 stays `ready`.

## Test

`tests/issue-2160-number-tolocalestring.test.ts` (11/11): integer/single-digit/fractional/negative number toLocaleString + string concat across host & standalone, plus a standalone no-`__extern_toLocaleString`-leak assertion. tsc + prettier + biome clean.

### DIFF
--- a/plan/issues/2160-standalone-string-number-coercion-residual.md
+++ b/plan/issues/2160-standalone-string-number-coercion-residual.md
@@ -79,3 +79,46 @@ same helper to `parseNeeded`. Regression test:
 acceptance criteria's `__new_String`/`__new_Number` leak) plus the harder
 String/Number coercion edges that overlap the coercion engine (#1917). Those
 remain the value-rep / #1917 territory called out in the original notes.
+
+---
+
+## Slice (2026-06-21, dev-agent) — `Number.prototype.toLocaleString` for standalone
+
+**Status stays `ready`** — one more independent slice of the 635-bucket,
+unblocked (no wrapper substrate / #1917 dependency).
+
+**Bug:** `(n).toLocaleString()` on a number receiver CE'd in standalone / WASI
+with `'__extern_toLocaleString' (dynamic-shape object)`. The generic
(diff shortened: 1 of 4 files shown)


ttraenkler:
Closing as a duplicate of #1806 (bruno), which fixes the same #2160 slice (standalone Number.prototype.toLocaleString → number_toString) with the same approach. #1806 was the assigned/first PR for this slice; I opened #1809 before seeing the task note that toLocaleString was already covered. Deferring to #1806. My distinct String.raw slice is in #1812.

PR #1810, opened 2026-06-21
fix(#2029): standalone Object.create(proto, descriptors) emit crash

ttraenkler:
## Summary

`Object.create(proto, descriptors)` with a **non-empty** properties object died at emit time under `--target standalone` with `Binary emit error: Codegen error: global index out of range — -1` — a hard `compile_error` that loses the whole file. (The #2029 PR-1 audit marked `built-ins/Object/create` as compiling, but that was the **1-arg** form; the 2-arg descriptor form still crashed.)

## Producer

The compile-time descriptor expansion in the `Object.create` handler (`src/codegen/expressions/calls.ts`) materialized each property **key** with a raw `{ op: "global.get", index: stringGlobalMap.get(key) }`, guarding only `=== undefined`. Under standalone / nativeStrings, `addStringConstantGlobal` stores the documented **-1 sentinel** (#1174 — "no host import, materialize inline"), so `global.get -1` reached the binary encoder and crashed. Two sites: the static `__defineProperty_value` arm and the dynamic-flag `__defineProperty_desc` arm.

## Fix

Emit the key via the existing dual-mode `stringConstantExternrefInstrs` helper (native-string externref standalone; `global.get` for a valid host global; **already guards the -1 sentinel**) at both sites. Mirrors the `emitSetSubclassProto` fix shape from PR-1. gc/host mode is unchanged (real globals, never the sentinel).

## Measured impact

60-file `built-ins/Object/create` standalone sample:

| | pass | fail | CE |
|---|---:|---:|---:|
| main | 3 | 1 | 49 |
| this PR | 8 | 13 | 32 |

**+5 pass, 17 tests moved out of `compile_error`.** Zero host regressions (host `Object.create(null, {a:{value:1},b:{value:2}})` returns correctly before and after).

## Validation

- `tests/issue-2029-object-create-descriptors-standalone-emit.test.ts` (9/9): single / multi / string-literal-key / static-flag descriptors compile standalone; value-descriptor read-back (42, 5); no `-1` global emitted; host no-regression guard.
- Existing `tests/issue-2029-subclass-builtin-standalone-emit.test.ts` (8/8) green.
- tsc + prettier + coercion-sites gates clean.

## Still open (separate, not this slice)

- 4 `Object/create` files still emit `-1` via a **different** producer (`Object.create({}, undefined)` + `instanceof Object` — the assert/instanceof string-global path).
- `o.a + o.b` arithmetic over two `any`-typed props read off an `Object.create(null, …)` result traps standalone with `illegal cast` (native dynamic-read/unbox gap; reading `o.b` alone works; plain `{}` two-prop add works).
- `__defineProperty_value`/`__defineProperty_desc`/`__getOwnPropertyDescriptor` remain host-only imports refused standalone (`dynamic-shape`).

### DIFF
--- a/plan/issues/2029-standalone-u32-out-of-range-binary-emit.md
+++ b/plan/issues/2029-standalone-u32-out-of-range-binary-emit.md
@@ -170,3 +170,52 @@ standalone still leaks the `__new_<Builtin>` HOST import (`class-bodies.ts:1423/
 — a host-import-retirement concern, not the emit crash. Kept #2029 `in-progress`:
 the emit-crash cluster (the headline) is fixed; the `__new_<Builtin>` standalone
 runtime path is the residual. Reassess closing once that lands.
+
+## Slice (2026-06-21, dev-agent) — `Object.create(proto, descriptors)` emit crash
+
+**Status stays `in-progress`** — one independent emit-crash slice of the bucket.
+
+The PR-1 audit said `built-ins/Object/create` "all COMPILE in standalone", but
+that was the **1-arg** form. `Object.create(proto, descriptors)` with a
+**non-empty** properties object still died at emit with
+`Binary emit error: global index out of range — -1` (the documented #1174
+string-global -1 sentinel).
+
+**Producer (confirmed):** the compile-time descriptor expansion in the
+`Object.create` handler (`src/codegen/expressions/calls.ts`) materialized each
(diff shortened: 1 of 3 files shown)


ttraenkler:
Closing: superseded. Branched from the fork origin/main which was 1188 commits behind upstream, so this never triggered CI. Probed the bug against upstream/main: Object.create(proto,descriptors) already fixed upstream.

PR #1812, opened 2026-06-21
fix(#2160): String.raw valid standalone/WASI binary via native-strings concat

ttraenkler:
## Summary

`String.raw` emitted an **INVALID standalone / WASI binary** (even the no-substitution form). `compileTaggedTemplateExpression` built the `externref`-element template-strings vec (`array.new_fixed`) **before** dispatching the `String.raw` builtin short-circuit; in native-strings mode the literals lower to `ref $AnyString` (not externref), so that (unused-for-`String.raw`) vec failed validation: `array.new_fixed expected type externref, found struct.new of type (ref 6)`. With substitutions the `compileStringRaw` concat path additionally mis-fed an f64 into `any.convert_extern`.

## Fix (two parts)

1. **Dispatch `isStringRawTag` at the top** of `compileTaggedTemplateExpression`, before any template-vec scaffolding (cache global + `array.new_fixed`) is emitted — so `String.raw` never builds the dead, mistyped vec.
2. **`compileStringRaw` native-strings branch** (`noJsHost && nativeStrings`): coerce every operand to `ref $AnyString` via the existing `compileNativeConcatOperand` helper (number_toString + ref-from-extern, bool→literal, string passthrough, any→ToString) and concatenate with the native `__str_concat`.

JS-host mode is unchanged (wasm:js-string concat). Host and standalone now return identical results.

## Test

`tests/issue-2160-string-raw-standalone.test.ts` (13/13): no-subst / single / multiple / boolean / string substitutions + raw-escape length, across host & standalone, plus a standalone-validates regression guard. Existing tagged-template suites (issue-2008 incl. its `String.raw` case, issue-2176) green (20/20). tsc + prettier + biome clean.

One independent slice of #2160 (issue stays `ready`; wrapper-object method-dispatch remains value-rep/senior-dev territory).


ttraenkler:
Closing: superseded. Branched from the fork origin/main which was 1188 commits behind upstream, so this never triggered CI. Probed the bug against upstream/main: tagged-template already fixed upstream.

PR #1815, opened 2026-06-21
fix(#2029): object-literal data-key + accessor standalone emit crash

ttraenkler:
## Summary

An object literal mixing a **data property** with a get/set **accessor** (e.g. `{ a: 1, get v() {} }`) is routed through the host-object externref path. There the DATA-property key was materialized with a raw `global.get <stringGlobalMap.get(key)>`, guarding only `=== undefined`. Under `--target standalone` / nativeStrings the key string is the **-1 sentinel** (#1174 — no host string-constant global), so `global.get -1` crashed the binary encoder (`Binary emit error: global index out of range — -1`) — a hard `compile_error` losing the whole file.

This drove the bulk of the standalone `built-ins/Iterator/prototype` failures: their throwing-iterator fixtures return `{ done: true, get value() { … } }` — a data key + accessor. The trigger is NOT `extends Iterator`; the minimal repro is just `{ a: 1, get v() { return 5; } }` (getter-only or data-only literals already worked — only the **mix** crashed).

## Fix

In `compileObjectLiteral`'s host-object path (`src/codegen/literals.ts`), materialize the data-property key via the dual-mode `stringConstantExternrefInstrs` helper — mirroring the accessor-key path immediately below it (which already had this fix). gc/host mode unchanged (real globals, never the sentinel).

## Measured impact

60-file `built-ins/Iterator/prototype` standalone sample:

| | pass | CE | (emit-CE) |
|---|---:|---:|---:|
| main | 4 | 28 | 6 |
| this PR | 6 | 23 | 1 |

+2 pass, 5 emit-CEs cleared in the sample; the full-cluster scan found ~49 emit-CEs of this exact shape. Zero host regressions (the fix is -1-sentinel-gated, never reached in gc mode; host `{ a:7, get v(){return 5} }` → `o.a+o.v === 12` before and after).

## Validation

- `tests/issue-2029-objlit-data-accessor-standalone-emit.test.ts` (7/7): data-before/after-getter, data+setter, multi-data+accessor, the Iterator throwing-iterator `next()` shape, no-`-1`-global assertion, host no-regression.
- tsc + prettier + coercion-sites gates clean.

## Still open (separate)

- The `o.a + o.v` arithmetic read-back over two `any` props of an accessor-routed object traps standalone (`illegal cast`) — the same native dynamic-read/unbox gap noted for the Object.create slice (#1810), not the emit crash.
- 1 residual emit-CE in the sample is a different producer.

### DIFF
--- a/plan/issues/2029-standalone-u32-out-of-range-binary-emit.md
+++ b/plan/issues/2029-standalone-u32-out-of-range-binary-emit.md
@@ -170,3 +170,41 @@ standalone still leaks the `__new_<Builtin>` HOST import (`class-bodies.ts:1423/
 — a host-import-retirement concern, not the emit crash. Kept #2029 `in-progress`:
 the emit-crash cluster (the headline) is fixed; the `__new_<Builtin>` standalone
 runtime path is the residual. Reassess closing once that lands.
+
+## Slice (2026-06-21, dev-agent) — object-literal data-key + accessor emit crash
+
+**Status stays `in-progress`** — another emit-crash slice (the `Iterator/prototype`
+cluster, ~49 standalone CE).
+
+The `built-ins/Iterator/prototype` failures are NOT about `extends Iterator` —
+they're driven by the throwing-iterator fixtures returning
+`{ done: true, get value() { … } }`: an object literal mixing a **data property**
+with an **accessor**. Such a literal is routed through the host-object (externref)
+path, and the DATA-property key was materialized with a raw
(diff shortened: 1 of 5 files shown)


ttraenkler:
Closing: superseded. Branched from fork origin/main (1188 behind upstream) so never triggered CI. Both slices probed against upstream/main: objlit data+accessor key already fixed upstream; the DisposableStack.prototype slice is re-done from upstream/main as #1821.

PR #1819, opened 2026-06-21
fix(#2029): builtin-subclass standalone host-import leaks (__get_undefined + __tag_user_class)

ttraenkler:
## Summary

A builtin subclass (`class X extends Uint8Array {}` / `extends Error` …) + `new X()` leaked **three** host imports standalone, so the module fails to instantiate with an empty import object. This slice clears two of them (`__new_<Builtin>` is a separate, harder slice — see below).

### 1. `__get_undefined`

`compileNewExpression` → `pushDefaultValue` → `emitUndefinedValue` (`src/codegen/type-coercion.ts`) called `ensureLateImport` **directly**, bypassing the canonical `ensureGetUndefined` `ctx.nativeStrings` guard (`ensureLateImport` does not refuse `__get_undefined`). So the host import leaked into the implicit derived-ctor forwarder's externref default. Fix: gate on `ctx.nativeStrings` → emit `ref.null.extern` (undefined ≡ null standalone). Same defect class as the #1702 Error-subclass slice, different site (the `new`-expression default path).

### 2. `__tag_user_class`

The externref-backed-class instanceof tag (`src/codegen/class-bodies.ts`, #1455) was emitted unconditionally. Its **only** consumer is the host `__instanceof` import, which does not exist standalone (a plain class uses native WasmGC type checks and needs no tag). So the tag was dead weight that leaked. Fix: skip it under `ctx.nativeStrings` — mirrors the #1500 `emitSetSubclassProto` host-only-adjustment skip. gc/host keeps it so the host tag-chain still resolves `instanceof`.

## Validation

`tests/issue-2029-ta-subclass-host-leak-standalone.test.ts` (5/5): TypedArray/Error subclass `new X()` assert neither `__get_undefined` nor `__tag_user_class` is imported; empty-declaration guard; host (gc) no-regression (tag still present). Existing `issue-2029-subclass-builtin-standalone-emit` (8/8) green. tsc + prettier + coercion-sites clean. Zero host regressions (both fixes `nativeStrings`-gated).

Verified by direct import inspection — the test262 runner masks the leak because it doesn't instantiate the standalone lane with a strictly empty import object.

## Still open (separate slice)

`__new_<Builtin>` (`__new_Uint8Array`, `__new_Error`, …) — the builtin base constructor. Needs a native WasmGC construction path for the externref-backed base (overlaps #2159). Even an empty `class X extends Uint8Array {}` declaration emits it.

### DIFF
--- a/plan/issues/2029-standalone-u32-out-of-range-binary-emit.md
+++ b/plan/issues/2029-standalone-u32-out-of-range-binary-emit.md
@@ -170,3 +170,43 @@ standalone still leaks the `__new_<Builtin>` HOST import (`class-bodies.ts:1423/
 — a host-import-retirement concern, not the emit crash. Kept #2029 `in-progress`:
 the emit-crash cluster (the headline) is fixed; the `__new_<Builtin>` standalone
 runtime path is the residual. Reassess closing once that lands.
+
+## Slice (2026-06-21, dev-agent) — builtin-subclass host-import leak (__get_undefined + __tag_user_class)
+
+**Status stays `in-progress`.** Two of the three builtin-subclass standalone
+host-import leaks fixed; `__new_<Builtin>` construction remains a separate slice.
+
+A builtin subclass (`class X extends Uint8Array {}` / `extends Error` etc.) +
+`new X()` leaked THREE host imports standalone (module fails to instantiate with
+an empty import object): `__get_undefined`, `__tag_user_class`, `__new_<Builtin>`.
+This slice clears the first two:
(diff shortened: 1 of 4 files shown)


ttraenkler:
Closing: superseded. Branched from the fork origin/main which was 1188 commits behind upstream, so this never triggered CI. Probed the bug against upstream/main: __get_undefined/__tag_user_class already fixed upstream; only __new_<Builtin> remains as a separate slice.

PR #1826, opened 2026-06-21
fix(#2160): String.raw with substitution valid standalone binary

ttraenkler:
## Summary

On pristine upstream HEAD, `String.raw` **with a substitution** (`String.raw`a${1}b``) still emits an **INVALID standalone binary**: a numeric substitution drives an f64 straight into `any.convert_extern` ("expected externref, found f64.const"). The no-substitution case was already fixed upstream by the generic tagged-template `pushStringElem` externref-array bridge, but `compileStringRaw`'s substitution-concat path (`src/codegen/string-ops.ts`) still mixed representations in native-strings mode.

## Fix

In `noJsHost` / native-strings mode, `compileStringRaw` now coerces every operand to `ref $AnyString` via the existing `compileNativeConcatOperand` helper (number_toString + ref-from-extern, bool→literal, string passthrough, any→ToString) and concatenates with the native `__str_concat`. JS-host mode keeps the wasm:js-string concat path unchanged.

## Context

Re-files the still-needed substitution half of the **closed PR #1812**, which was branched off the stale fork ( ~1188 commits behind upstream) and got superseded after only the no-substitution case was probed. I verified the substitution crash on pristine  () before and after the fix, and **branched this PR from ** so it's not stale-based. Claimed at slice granularity (`2160:string-raw-subst`).

## Test

`tests/issue-2160-string-raw-subst-standalone.test.ts` (13/13): no-subst / single / multiple numeric substitutions, raw-escape length, boolean + string substitutions — host & standalone — plus a standalone-validates regression guard. `issue-2008` tagged-template suite 7/7 green. tsc + prettier clean.

### DIFF
--- a/src/codegen/string-ops.ts
+++ b/src/codegen/string-ops.ts
@@ -633,6 +633,37 @@ function compileStringRaw(
   substitutions: readonly ts.Expression[],
 ): ValType | null {
   addStringImports(ctx);
+
+  // (#2160 standalone slice) Native-strings (standalone / WASI) path. Every
+  // operand fed to the native `__str_concat` must be a `ref $AnyString`.
+  // `compileStringLiteral` already yields one here, but a numeric / any
+  // substitution leaves an f64 / externref that the host externref-concat path
+  // below mishandles — a numeric substitution (`String.raw\`a${1}b\``) drove an
+  // f64 straight into `any.convert_extern` → an INVALID standalone binary. Reuse
+  // the proven `compileNativeConcatOperand` coercion (number_toString +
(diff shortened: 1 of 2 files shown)


ttraenkler:
Closing as a duplicate of #1825 (dev-bruno), which is the same fix (String.raw with substitutions → compileStringRaw native-concat branch, same two files) created ~4s earlier. We both independently rebuilt this slice from upstream/main (shared ttraenkler/dev-agent claim handle didn't lock us out of each other — flagged separately). Dedup rule = keep the first; deferring to #1825. Removing this from the merge queue to avoid a conflicting second land.

PR #1844, opened 2026-06-21
fix(#983d): dual-path dynamic dispatch for obj.<field>() host-method calls

ttraenkler:
## #983d — dual-path dynamic dispatch for `obj.<field>()` host-method calls

Fixes the generic-Array-method-on-plain-object cluster (`var o = {}; o.pop =
Array.prototype.pop; o.pop()`): **0 → 11/19** test262 entries flip to pass.

### Root cause (revised from the issue's original framing)

The issue was filed as a "missing `__sset_` host write-back" build, but that
framing is stale — the `__sset_<field>` setters already exist and `_safeSet`
already wires them. The actual bug is a **codegen dispatch gap**:

`obj.method(args)` where `method` is a **host function value** stored in `obj`
(no wasm impl) falls past every static / struct-method handler and hits the
**graceful-null fallback** in `compileCallExpression`, which compiles the callee
property-access, **drops** the resulting method value, and pushes
`ref.null.extern`. So the call is never made — `o.pop()` yields `null` instead
of `undefined`, and any mutation the method would perform on `o` never happens.

### Fix

Before the graceful-null fallback, when the callee is a non-optional property
access `obj.method` and we are in JS-host mode, route to
`emitWrapperDynamicMethodCall` → `__extern_method_call(receiver, "method",
argsArray)`. The host runs the generic method with the live-mirror
`_wrapForHost` proxy as `this`, so BOTH the return value AND mutation
observability round-trip (the proxy's `set` trap → `_safeSet` → `__sset_`).

- Syntactic gate only — never static-classify the receiver (a possibly-proxy
  value carries no TS-type brand, #2501).
- JS-host mode only; standalone falls through to the graceful null (a
  Wasm-native dynamic dispatch is a separate follow-up).
- ~20 LOC, additive, behind the fallback — nothing that already resolved is
  affected.

### Validation

- Cluster **0 → 11/19** pass.
- **0 regressions**: focused proxy/object-method sweep 7/7 (Proxy get/set, Array
  HOF, Object.keys, eval-call); 180-file isolated regression sample all pass
  (the batch-run `new Ctor` crash is a known multi-test-in-one-process
  state-pollution artifact present on base too, not a per-test regression — CI
  shards isolate). Hard-error gate OK. Typecheck clean.
- New test: `tests/issue-983d-generic-method-call.test.ts`.

### Residual carved to #2573

The remaining 8 fail at a *later* assertion `obj.length === undefined` — a
**separate property-read bug** (reading a missing property on a plain `{}`
object returns `null`, not `undefined`), orthogonal to this method-dispatch fix.

### DIFF
--- a/plan/issues/2573-missing-property-read-returns-null-not-undefined-plain-object.md
+++ b/plan/issues/2573-missing-property-read-returns-null-not-undefined-plain-object.md
@@ -1,0 +1,66 @@
+---
+id: 2573
+title: "Reading a missing property on a plain `{}` object returns null, not undefined"
+status: ready
+created: 2026-06-21
+priority: medium
+feasibility: medium
+goal: test262-conformance
+parent: 983d
+test262_fail: 8
+---
+# #2573 — Missing-property read on a plain object yields `null` not `undefined`
+
+## Problem
+
+Reading an own property that does not exist on a plain object literal
+(`var obj = {}; obj.length`) returns JS `null` (`typeof === "object"`) where the
+spec requires `undefined` (§10.1.8 OrdinaryGet → returns `undefined` for a
+missing property).
+
+```js
+var obj = {};
+obj.length;   // expected: undefined ; actual: null
+```
+
+## How it surfaced (#983d residual)
+
+After #983d landed the dual-path dispatch for `obj.<field>()` host-method calls
+(`var o = {}; o.pop = Array.prototype.pop; o.pop()` now actually runs), the
+generic-Array-method-on-plain-object test262 cluster went 0 → 11/19. The
+remaining 8 fail at a **later** assertion — `obj.length === undefined` — because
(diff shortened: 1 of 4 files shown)


ttraenkler:
Closing — the dual-path `obj.<field>()` dispatch fix is net-negative on the full merge_group test262 gate (-200 net / 323 regressions, bucket 65f0b2857a074130), even though it fixed the 19-file generic-method-on-plain-object cluster locally (+11). The graceful-null fallback's undefined-return is load-bearing for far more call shapes than the targeted cluster, and a tighter gate can't be validated without the full sharded gate (scoped ≠ full). Root-cause + a 'retry must gate surgically and validate via full local-ci' note recorded in #983d. Will re-attempt #983d with a narrow, fully-validated gate later.

PR #1851, opened 2026-06-21
fix(#2187): runtime-guarded native-string dispatch for any-typed receivers

ttraenkler:
## #2187 — string `.length` / methods on an `any`-typed native-string receiver (standalone)

String `.length` and string methods on an `any`/`unknown` receiver whose runtime value is a native `$AnyString` took the generic externref path and returned **0** in standalone, because dispatch gates on `isStringType(<TS static type>)` — false for an `any` receiver even though the value IS a real string (`String(o.v)`-style concat already works because it keys off the operand ValType).

Generalizes the #2077/#2192 caught-Error precedent with a runtime `ref.test $AnyString` guard at the any-receiver dispatch sites. **Native-string mode (standalone/WASI) only — host/gc mode is untouched.**

### Fixes (all verified)
- `o.v.length` → 2, `o.v.charCodeAt(0)` → 104
- string-yield generator: `for (const v of g()) n += v.length` → 2; `v.charCodeAt(0)` → 97
- `catch (e:any) { e.message.length / .charCodeAt(0) }` → 4 / 111
- nested `o.a.b.length` → 2
- indexed read-back: `Object.values(o)[0].length`, `Object.entries(o)[0][1].length` → 2
- native methods on `any`: `o.v.slice(1).length` → 4, `o.v.indexOf("l")` → 2

### Approach (3 receiver shapes)
1. **`.length` on an externref `any`** — guard the existing #1472 Slice-2 `__extern_length` arm: string hit → `$AnyString.len`; miss → unchanged array/$ObjVec reader.
2. **`.length` on a typed `$AnyString` local with TS type `any`** (generator loop var) — read field 0 directly via `isNativeStringFamilyTypeIdx`.
3. **native string methods on an externref `any`** — `compileGuardedNativeStringMethodCall` + a `receiverOverride` on `compileNativeStringMethodCall` (single receiver eval, no re-compile), wired through a new `receiverMayBeNativeStringAtRuntime` predicate. `collectStringMethodImports` registers the helpers.

### Non-regression
Boxed `String` wrapper keeps its #1910-R4 path; `any` holding an array keeps array-length; null/number `any` → benign default. `any` (not `any[]`) holding an **array** calling a string-named array method stays at 0 (pre-existing on `main`, out of scope). Full string-method + generator + property-access equivalence suites pass; the 9 pre-existing #1472 slice failures are identical on `main`.

Tests: `tests/issue-2187.test.ts` (12 cases incl. all ACs, non-regression guards, host-mode parity).

Changed: `src/codegen/property-access.ts`, `src/codegen/expressions/calls.ts`, `src/codegen/string-ops.ts`, `src/codegen/index.ts`.

### DIFF
--- a/plan/issues/2187-any-typed-local-string-method-dispatch-by-valtype.md
+++ b/plan/issues/2187-any-typed-local-string-method-dispatch-by-valtype.md
@@ -1,10 +1,12 @@
 ---
 id: 2187
 title: "standalone: string methods on an any-typed local with a native-string ValType take the generic externref path (v.length → 0)"
-status: ready
+status: done
 sprint: 64
 created: 2026-06-17
-updated: 2026-06-17
+updated: 2026-06-21
+completed: 2026-06-21
+assignee: sdev-strdispatch
 priority: low
 feasibility: hard
 reasoning_effort: high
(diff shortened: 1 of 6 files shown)


ttraenkler:
Superseded: #2187 was independently fixed by PR #1845 (sd-3) and merged. That fix covers only the bare-identifier-with-$AnyString-local case; the broader value-rep cluster (opaque-externref any values: o.v.length, o.v.charCodeAt, catch(e:any).message, nested, Object.values[0]) is reopened as #2576, built on top of #2187. New PR incoming from issue-2187-anystr-value-rep.

PR #1863, opened 2026-06-21
fix(#2040): standalone AnyValue tag-5 equality — 3-way field-4 cascade

ttraenkler:
## #2040 — standalone `_isSameValue` / AnyValue tag-5 equality (top-tier standalone unlock)

Implements arch-2040's spec. Cluster A's `assert.notSameValue`/`_isSameValue` failures were **not** a destructuring bug (the rest array IS fresh, `Object.is` correct in pure standalone) — the root cause is the **standalone AnyValue tag-5 equality arm**.

### Root cause
The tag-5 (string) arm of `__any_eq`/`__any_strict_eq` was `strEqualsIdx >= 0 ? [host __str_equals] : [i32.const 0]`. `strEqualsIdx` (the JS-host `equals` import) is **-1 in standalone**, so the arm was a **dead `i32.const 0`** → EVERY tag-5 `===`/`==` returned false. A NUMBER `any` is boxed as tag-5 by the #1888 `boxToAny` policy, so `5 === 5` → false → the harness `assert._isSameValue` (`if(a===b){…} return a!==a && b!==b`) broke, failing every numeric-`any` sameValue/notSameValue once any prior `any`-op set `anyValueTypeIdx`.

### Fix — consumer-side ONLY, NO #1888 boxing change (preserves the −794 baseline gain)
REPLACE the dead ternary with a canonical **3-way field-4 cascade** disambiguating by the runtime externref:
- both `$__box_number` → `__any_to_f64` + `f64.eq` (`5===5` true, NaN-safe)
- both `$AnyString` (BASE type) → flatten + native `__str_equals`
- **else** (object/array/host externref) → REFERENCE IDENTITY (`ref.eq` via the eq heap type) so `a===a` over an array/object `any` is true; mixed primitive → 0

`__any_eq` cross-tag str⇄num, `__any_typeof`, `__any_to_string`, `__any_to_f64` audited — already correct (the `__any_to_f64` #1888 `$BoxedNumber` recovery is the contract precedent).

### Traps avoided (from 3 reverted attempts)
1. `ref.test` the `$AnyString` **BASE** (`ctx.anyStrTypeIdx`), NOT `$NativeString` — cons/utf8 are subtypes (a Flat-only test → indexOf −1).
2. A true **3-way**, not a boolean — objects fall to ELSE.
3. **REPLACE** the whole dead ternary, not nest inside it.
4. ELSE must be **ref-identity**, not `0` — else `a!==a` over arrays mis-returns.

### Validation
- 9 scoped tests (`tests/issue-2040-anyvalue-tag5-eq.test.ts`) + comprehensive probe pass.
- Broad standalone sweep (equals/strict-equals/addition/Number/indexOf/Object.is): **444 vs 440 baseline (+4, indexOf flat — no regression)**.
- Wider sweep (includes/lastIndexOf/find/Math/Boolean): **0 regressions**.
- Gates: hard-error, any-box-sites, coercion-sites OK. #1461 array-search + equality/AnyValue suites green (the one `#2081 null==undefined` failure is pre-existing on `origin/main`, tag-0, untouched).

**⚠️ FULL-GATE mandate:** `_isSameValue` gates a huge fraction of asserts, so a scoped sweep CANNOT capture the magnitude or a subtle broad regression. The authoritative −794-catching gate is the **FULL standalone merge_group (#2097 floor)** — this PR is enqueued for that. Expect a large +delta.


ttraenkler:
Closing as superseded/net-negative per the #2040 shelve verdict (full evidence in #1884). The mixed-tag assert.sameValue(tag-5-number, tag-3-literal) case this targeted already passes on current main — the +311 was realized by intervening merges (#2503/#2358/#2187/#2574), and both fix layers now net-negative (tag-5 cascade −151, numeric-gate broaden 14reg/0imp). Equality helpers correctly left untouched.

PR #1360, opened 2026-06-11
fix(#2039): pushBody/popBody in generator lowering — truthiness + async-gen i64 invalid-Wasm buckets

ttraenkler:
Draft from a recovered fork branch. Root cause (instrumented): compileNestedFunctionDeclaration and its copies in literals.ts/closures.ts swap fctx.body raw instead of pushBody, so the already-baked prologue misses exactly one late-import flush (+1 index skew) — one bug behind both the ~150-row truthiness if[0] and ~230-row async-gen i64 buckets. Fix: pushBody/popBody at 4 sites. Typecheck green; full validation was in progress when the session died — needs merge-with-main + the issue-2039 test suites before review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1864, opened 2026-06-21
fix(#2579): standalone boxed tag-5 string equality → native __str_equals

ttraenkler:
## #2579 — `__any_strict_eq` / `__any_eq` tag-5 string arm always-false in standalone

The `$AnyValue`-boxed equality helpers compared same-tag string operands (tag 5) via the `wasm:js-string` `equals` builtin, which is **absent in standalone** (`strEqualsIdx === -1`), so the arm emitted `i32.const 0` — two equal boxed strings never matched.

Surfaced via `Array.prototype.{indexOf,lastIndexOf,includes}.call(stringArrayLike, str)`, which composes `__any_from_extern` + `__any_strict_eq` / `__any_eq`:

```ts
const al: any = { 0: "x", 1: "y", length: 2 };
Array.prototype.indexOf.call(al, "y");    // -1   → 1
Array.prototype.includes.call(al, "y");   // false → true
```

### Fix (`any-helpers.ts` only)
Route the standalone tag-5 arm to the native `__str_equals` (the same helper `binary-ops.ts` uses for a static string `===`): recover each operand from field 4 (`externval`) via `any.convert_extern`, a `ref.test $AnyString` guard, `ref.cast`, `__str_flatten` (cons-string safe), then `__str_equals`. Applied to **both** `__any_strict_eq` (`===`) and `__any_eq` (`==`). The guard keeps a non-string field-4 carrier on `0` rather than trapping. Host/gc mode keeps the `wasm:js-string` path. Two new anyref temps per helper.

### Non-regression
Numeric/null array-like `.call` search unchanged; static string `===` unchanged; host/gc unchanged. The pre-existing #2036 S6 "refuse loud" and #1360 null-field failures are **identical on `main`** (verified) — untouched by this fix.

### Out of scope (pre-existing on `main`, separate gaps)
- Plain `const a:any=[...]; a.indexOf("y")` (NOT the `.call` form) → 0: routes through `__extern_method_call`'s non-`$Object` `$Vec` brand arm (#1888 "Slice 4" deferral), never reaches this helper.
- Boolean-element array-like `.call` search returns 0 on main too (tag-4 search-value boxing).

Tests: `tests/issue-2579.test.ts` (8 cases: indexOf/includes/lastIndexOf, cons-string, len early-out, miss, non-regression).

### DIFF
--- a/plan/issues/2579-standalone-any-strict-eq-tag5-string-content.md
+++ b/plan/issues/2579-standalone-any-strict-eq-tag5-string-content.md
@@ -1,0 +1,91 @@
+---
+id: 2579
+title: "standalone: __any_strict_eq / __any_eq tag-5 string arm returns always-false (boxed string === and Array.prototype.indexOf/includes.call on string array-likes)"
+status: done
+sprint: 64
+created: 2026-06-21
+updated: 2026-06-21
+completed: 2026-06-21
+assignee: sdev-strdispatch
+priority: low
+feasibility: medium
+reasoning_effort: high
+task_type: bugfix
+area: codegen
+language_feature: strings
+goal: standalone-mode
+related: [2036, 1461, 1360, 296]
+origin: "2026-06-21 — pinpointed by sdev-reflect while closing #2036 array-search"
+---
+
+# #2579 — `__any_strict_eq` / `__any_eq` tag-5 string arm is always-false in standalone
+
+## Problem
+
+The `$AnyValue`-boxed equality helpers `__any_strict_eq` (`===`) and `__any_eq`
+(`==`) compare same-tag string operands (tag 5) by content via the
+`wasm:js-string` `equals` builtin. In standalone/WASI there is no
+`wasm:js-string` host, so `strEqualsIdx = ctx.jsStringImports.get("equals") ?? -1`
+is `-1`, and the tag-5 arm emitted `i32.const 0` (**always false**) — two equal
+strings boxed in `$AnyValue` never matched.
+
+The headline standalone repro (still failed on `origin/main` `93e53919f`):
+
+```ts
+const al: any = { 0: "x", 1: "y", length: 2 };
+Array.prototype.indexOf.call(al, "y"); // -1   expected 1
+Array.prototype.includes.call(al, "y"); // false expected true
+```
+
+`Array.prototype.{indexOf,lastIndexOf,includes}.call(arrayLike, str)` routes
+through `compileArrayLikePrototypeSearch` → `__extern_strict_eq` /
+`__extern_same_value_zero` (composed from `__any_from_extern` + `__any_strict_eq`),
+so the always-false tag-5 arm made every string-element search miss.
+
+## Root cause
+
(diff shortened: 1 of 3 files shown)


ttraenkler:
Heads-up on a collision with #1863 (#2040), which rewrites the same tag-5 arm in `src/codegen/any-helpers.ts`. #1863 implements the full 3-way field-4 cascade:

- **number** → `__any_to_f64` + `f64.eq` (fixes `5===5` / the `_isSameValue` / `assert.sameValue` headline — the bulk of the cluster)
- **native string** → `ref.test $AnyString` (BASE) + `__str_flatten` + `__str_equals` — functionally identical to this PR's string arm, same BASE-type insight
- **else** → `ref.eq` reference identity (so `a===a` over an array/object `any` is true)

So #2040 subsumes #2579's string-only fix and additionally closes the number headline this PR leaves out of scope. I'm landing #1863 (complete and ahead in the queue). To avoid a merge_group conflict on the shared arm, I'd recommend closing #2579 as superseded by #2040 once #1863 lands. Thanks for the string-arm work — the `ref.test $AnyString` BASE-type routing matches exactly what #2040 landed.

ttraenkler:
Update: I've dequeued #1864 from the merge queue and applied `hold` to prevent the auto-enqueue backstop from re-adding it (which would re-create the collision on the shared `any-helpers.ts` tag-5 arm). #1863 (#2040) is landing the complete 3-way cascade — number + string + object-identity — which includes this PR's exact string-arm fix. Once #1863 merges, please close #2579 as superseded by #2040 (or remove `hold` and reopen the discussion if I've missed something). The string-arm work and the `ref.test $AnyString` BASE-type routing here were correct — they're preserved in #1863.

ttraenkler:
Coordination note (tech-lead, s65 substrate wave): I'm landing #1883 (#2583 any-array dispatch) as the tag-5 **content-equality base** — it carries the same native `__str_flatten`+`__str_equals` fix this PR (#2579) implements, because #2583's array `indexOf`/`includes` is a hard dependency of it. Once #1883 merges, I propose **closing #1864 as superseded** (its content-eq is covered). 

The **broader tag-5 field-4 representation** work — #1863 (#2040 numeric) + #2585 (proto-identity ref.eq) — is going to an architect spec, because (confirmed by #2585 prototyping) no local discriminator can separate an identity-bearing object from a boxed number/NaN in field-4 in pure standalone (`nativeBoxNumberTypeIdx` is −1 there); the fix needs a distinct boxed-number tag/carrier. So #1863 should rebase onto the post-#1883 base and fold into that spec rather than re-landing the content-eq. Flagging so we don't double-land or collide. Happy to adjust if you'd rather #1864 be the base instead.

ttraenkler:
Superseded by #1888 (merged 21031ed6284). The native string-eq and the `ref.test $AnyString` guard from this PR are folded into #1888's guarded tag-5 string arm — and that guard's restoration is precisely what fixed the −162 merge_group destructuring regression the first #1888 draft hit (its refactor had dropped the guard). #2579's 8 test cases are preserved verbatim at tests/issue-2579.test.ts, so closing loses zero coverage. The fix landed; closing as superseded.

PR #1868, opened 2026-06-21
fix(#2573): plain-object .length is undefined (fail-safe static gate)

ttraenkler:
## #2573 — `.length` on a plain object is `undefined`, not numeric 0

`.length` is a "reserved accessor" handed to the array-length lowering, which
returns numeric **0** (vec field 0 / `__extern_length`). Correct for a real
array / string / arguments / function, but WRONG for a plain `{}` object where
`length` is an absent property — `({}).length === undefined` (§10.1.8
OrdinaryGet). The #983d generic-Array-method-on-plain-object residual.

### Fail-safe static gate
New `isPlainObjectWithoutLength` (`property-access.ts`), a `.length` arm before
the Function/vec arms: divert `.length` → `undefined` **only on POSITIVE
certainty** that the receiver's TS type is a concrete object type (**NOT**
`any`/`unknown`) with no own/inherited `length` member, no numeric index
signature, and no call/construct signature. An ambiguous `any` receiver **keeps**
the existing numeric vec-field-0 lowering (arrays dominate there) — divert on
certainty, never on "not provably array", so **`any[].length` arithmetic is never
touched**. Safe-partial ≫ regress `array.length` (lead-directed).

### Validation (exhaustive array-like pre-check, box-free, both gc + standalone)
array literal / arithmetic / loop-bound, `any[]`, string, string-arith, function
arity, typed array, empty array, post-push — **all unchanged**. Plain-object
cases (`const obj = {}; obj.length === undefined`) now correct in gc + standalone.
- `tests/issue-2573.test.ts` — 26 cases green.
- `tsc` + `prettier` + `#2108 coercion-drift` + `hard-error` gates all clean locally.

### Scope note — substrate split to #2580
The `any`/dynamically-mutated-receiver case (the test262 `S15.4.4.*_A2` cluster —
`var obj = {}; obj.length`) needs a **runtime** property-presence check + a
uniform-externref `.length` representation = substrate (a return-type change on
the hot `any[].length` path). Split out as **#2580** (value-rep lane + #983d
retry, task #20). So this PR banks the **user-code plain-object correctness with
zero array regression**; the test262 cluster is the #2580 follow-up. (Filed via
`--allocate`, not hand-picked.)

### DIFF
--- a/plan/issues/2573-missing-property-read-returns-null-not-undefined-plain-object.md
+++ b/plan/issues/2573-missing-property-read-returns-null-not-undefined-plain-object.md
@@ -1,7 +1,9 @@
 ---
 id: 2573
 title: "Reading a missing property on a plain `{}` object returns null, not undefined"
-status: ready
+status: done
+completed: 2026-06-21
+assignee: ttraenkler/sd-6
 created: 2026-06-21
 priority: medium
 feasibility: medium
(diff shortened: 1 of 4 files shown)


ttraenkler:
Closing per tech-lead decision: the fail-safe static .length gate ejected from the merge_group on a hidden test262 .length regression that the targeted array-like pre-checks missed. It moves 0 test262 rows, and the .length path is too central to risk a diagnose-fix-revalidate cycle for zero conformance gain. The real conformance value — `any`/dynamic-receiver .length needing a runtime property-presence check + uniform-externref representation — is tracked as the parked substrate issue #2580 (value-rep / #983d lane).

PR #1871, opened 2026-06-21
fix(#2001 S2): sparse-array HOF visit-skip on the dense vec

ttraenkler:
The headline #2001 slice (builds on S1 / PR #1838, now merged).

## What S2 does
S1 made a *visited* hole read as `undefined`; S2 adds the §23.1.3.* visit-SKIP
semantics the read-map alone cannot give — the callback is NOT invoked for a
hole, and a hole never matches `indexOf` / contributes to a `reduce` fold.

Two reusable gate helpers in `src/codegen/array-holes.ts` (both gated on
`usesArrayHoles && elemType.kind === "externref"` → typed `number[]`/`boolean[]`
kernels byte-identical, no `ref.test`):
- **`holeSkipGate`** — wraps branch-free per-iteration `work` in `if($Hole) onHole
  else work`. `forEach`/`filter`/`map`. For `map`, `onHole` writes `$Hole` into the
  result slot (result vec forced to externref-element) so a source hole → a
  **result hole** (`[1,,3].map(x=>x*10)` → `"10,,30"`).
- **`holeContinueGate`** — for work that branches into the loop (`some`/`every`
  early-exit `br 2`, `indexOf`/`lastIndexOf` match-break, `reduce`/`reduceRight`
  fold): tests the hole at the SAME control depth and on a hole does `i±1; br 1`
  (continue), leaving the present-index work unguarded so its `br` depths are
  unchanged (avoids the depth-off-by-one of nesting work in an extra `if`).
  `reverse` flips `i++`→`i--` for reduceRight / lastIndexOf.

## Per-method
- `forEach`/`some`/`every` skip (callback not called, scan continues)
- `map` skips + result-hole; `filter` omits
- `indexOf`/`lastIndexOf` skip — a hole never matches (supersedes S1's
  hole→undefined read-map for these two)
- `reduce`/`reduceRight` skip the fold AND seek the first/last **present** element
  for the no-initial-value seed (empty/all-holes ⇒ TypeError)
- `includes` UNCHANGED (Get → `[,].includes(undefined) === true`)
- `find`/`findIndex`/`findLast`/`findLastIndex` UNCHANGED (Get — visit a hole as
  `undefined`, not HasProperty-skip)

## Validation
- `tests/issue-2001-s2-hof-skip.test.ts` (21 cases — host + standalone)
- `tests/issue-2001-s1-hole-literal.test.ts` updated from S1's interim
  visit-as-undefined to S2 visit-skip; 36 cases
- array-methods (22), array-prototype-methods (13), functional-array-methods (24)
  all green; typecheck clean; typed-kernel byte-identity + deterministic-bytes
  guards
- Broad-impact (shared array-HOF loop driver): the merge_group full-Test262 gate
  is the authoritative validator (per memory `project_broad_impact_validate_full_ci`).

### DIFF
--- a/plan/issues/2001-sparse-holes-materialize-defaults.md
+++ b/plan/issues/2001-sparse-holes-materialize-defaults.md
@@ -308,6 +308,45 @@ destructuring / all HOFs / at / pop / slice, typed no-regression + deterministic
 bytes). Gates green: tsc, prettier, biome lint, `check-test262-hard-errors` (0,
 no growth), `check:ir-fallbacks` (no increase).
 
+### S2 landed (2026-06-21, sd-1838)
+
+**Done.** HOF visit-SKIP semantics on the dense WasmGC vec, the headline fix
+(`*-ary-forEach-15.4.4.18-2-9` family + the `does-not-visit-deleted/absent`
+suite). Two reusable gate helpers in `src/codegen/array-holes.ts`:
+
+- `holeSkipGate(ctx, data, i, arrTy, work, onHole?)` — wraps a branch-free
+  per-iteration `work` in `if ($Hole) onHole else work`; used by
+  `forEach`/`filter`/`map`. For `map`, `onHole` writes `$Hole` into the result
+  slot (and the result vec is forced to externref-element so it can hold the
+  sentinel) → a source hole yields a **result hole** (`[1,,3].map(x=>x*10)` →
+  `"10,,30"`).
+- `holeContinueGate(ctx, data, i, arrTy, onHole?, reverse?)` — for loop bodies
+  whose `work` contains its OWN `br` into the loop/block (`some`/`every`
+  early-exit `br 2`, `indexOf`/`lastIndexOf` match-break, `reduce`/`reduceRight`
+  fold). Emits the hole test at the SAME control depth and, on a hole, does
+  `i±1; br 1` (continue) directly — so the present-index work that follows is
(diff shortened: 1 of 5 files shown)


ttraenkler:
Closing per the #2001-line decision: S2's HOF visit-skip is spec-correct for OWN-property holes (`[1,,3].forEach`→2 is correct JS) but is conformance-net-NEGATIVE on test262 (-6 net, 0 improvements). The merge_group full-Test262 gate ejected it; the 6 regressed files are all `Array/prototype/{every,filter,indexOf,lastIndexOf,map,some}` "inherited accessor property without a get function on Array.prototype".

Root cause: HOFs use HasProperty, which walks the PROTOTYPE CHAIN. These tests do `Object.defineProperty(Array.prototype, "N", {set})`, making a hole "present via inheritance" → the HOF must visit index N and `Get` → the no-getter accessor → undefined → the callback passes. S2's own-property visit-skip treats a $Hole as unconditionally absent, so the callback is never called → the assert fails. js2wasm's dense WasmGC vec cannot model `Array.prototype[N]` accessor inheritance; S1's interim 'visit the hole as undefined' coincidentally matches these tests, so keeping S1 keeps all 6 passing.

Since the reachable test262 hole-HOF surface is exactly this inherited-accessor family, the own-property visit-skip has ~0 conformance payoff while breaking the 6. Keeping S1's read-map (the better-matching approximation) is correct. Deferring S2 — spec-correct but blocked on a 'prototype-chain index-accessor modeling' substrate that does not exist (same value-rep family as #2580).

PR #1875, opened 2026-06-21
docs(#2582): file 3rd compiled-acorn blocker — null receiver into wordsRegexp

ttraenkler:
Files the **third** independent compiled-acorn dogfood blocker, surfaced once #1712 (PR #1874) fixed the tokenizer identity loop.

With the identity loop gone, compiled acorn now executes far past the loop into module-init `buildUnicodeData`, where `wordsRegexp(unicodeBinaryPropertiesOfStrings[ecmaVersion])` throws `Cannot read properties of null (reading 'replace')` — a computed numeric-key read of a module-level object literal (`{9:"",…,14:ident}`) returns `null` where real JS reads `""`.

This is a **planning artifact only** (one new issue file, no code) — kept off the #1874 PR so it doesn't drift the enqueued branch. The issue documents the failure, the list of minimal probes that did NOT reproduce (so the next investigator knows the trigger is the full acorn shape), and the suggested full-shape repro + fix approach (cross-check #2542, the standalone computed-key analog).

#1712 stays open until the full acorn parse + AST-match acceptance is met.

### DIFF
--- a/plan/issues/2582-acorn-wordsregexp-null-receiver.md
+++ b/plan/issues/2582-acorn-wordsregexp-null-receiver.md
@@ -1,0 +1,127 @@
+---
+id: 2582
+title: "compiled-acorn module-init: null receiver into wordsRegexp from a numeric-keyed module-object read"
+status: ready
+created: 2026-06-21
+updated: 2026-06-21
+priority: high
+feasibility: hard
+reasoning_effort: high
+task_type: fix
+area: codegen, runtime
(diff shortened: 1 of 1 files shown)


ttraenkler:
Batch-committed the #2582 issue file directly to main (tech-lead doc-commit) — a one-file planning artifact shouldn't go through the merge queue's full standalone-shard CI. Closing this doc-only PR as superseded; #2582 is tracked on main + task #38.

PR #1903, opened 2026-06-22
fix(#2588): register native __object_is under WASI, not standalone-only

ttraenkler:
## Problem

Compiling an esbuild-bundled Native Messaging host with `--target wasi --wit` fails with a **hard** emit error (external report loopdive/js2#389):

```
Binary emit error: Codegen error: function index out of range — undefined
(valid: [0, 164)) at function '__defineProperty_value'.
```

A minimal source needs no esbuild and no `Object.defineProperty` — a framed `process.stdin.read(buf, offset)` loop + `process.stdout.write` under `--target wasi` is enough to pull in the native `$Object` runtime and reproduce.

## Root cause

`ensureObjectRuntime` registers the native `__defineProperty_value` helper **unconditionally**. Its #2042-S4 ValidateAndApplyPropertyDescriptor preflight bakes a direct `call __object_is` (SameValue value-change check). But `__object_is` was gated on `if (ctx.standalone)` only.

WASI is host-free too — `--target wasi` sets `ctx.wasi = true` but leaves `ctx.standalone = false` (the canonical host-free predicate across codegen is `ctx.standalone || ctx.wasi`). So under WASI the `__object_is` block was skipped, `funcMap.get("__object_is")` returned `undefined`, and the define helper baked an **undefined** funcIdx → the emitter's index guard fired.

This is a dependency-gate mismatch (an unconditionally-registered native helper depending on a more-narrowly-gated one), introduced by #2042 S4 (`b03ac8ca7`) — **not** a late-import index shift despite the guard message.

## Fix

`src/codegen/object-runtime.ts`: gate the `__object_is` registration on `ctx.standalone || ctx.wasi`. WASI runs the native define body (no JS host), so it must register the helper that body calls. Host-only mode (`!standalone && !wasi`) still owns `__object_is` via its JS import → host output byte-identical.

Patching the call site to skip `__object_is` when undefined was rejected: it would silently drop the spec-correct SameValue redefine check in WASI. Registering the dependency keeps WASI semantics aligned with standalone.

## Verification

- Original esbuild bundle + minimal repro both compile to a binary that passes `WebAssembly.validate`.
- `--target standalone` defineProperty data descriptor still executes (`o.x === 42`); host mode compiles + validates (unchanged).
- New `tests/issue-2588-wasi-defineproperty-funcidx.test.ts` (4 cases: framed stdin loop, esbuild-prelude-like bundle, standalone exec, host).
- No regressions in `define-property-patterns`, `issue-2042-s3-object-is`, `issue-1629-S6`, `issue-1127-samevalue`, `issue-2042-r2-topropkey-object`, and the WASI stdin/stdout/dataview suites (52 tests pass).


ttraenkler:
Superseded by #1924 (origin). Re-allocated the double id-collision (#2588→#2602) to #2609 and re-opened on origin (fork unpushable from the dev env). Work preserved + fixed in #1924.

PR #1904, opened 2026-06-22
feat(#2589): auto-resolve ambient `process` typing under --target wasi

ttraenkler:
## Problem

Compiling a `process`-using host with `--target wasi` emits a TS2580 **"Cannot find name 'process'"** warning on *every* use (5× for the example Native Messaging host) — because we serve only the bundled TS lib files, not `@types/node`. Yet the compiler **already lowers** `process.std{in,out,err}`, `argv`, `env`, `exit` for WASI (`node-process-api.ts`). So the global is supported; only its *type* was missing. Reported in #389.

## Fix

On the WASI / node-emulation path, serve a synthetic ambient `process` `.d.ts` (a global script) to the checker so it resolves `process` and stops warning — declaring exactly the lowered surface, not more.

- `src/checker/index.ts` — `AnalyzeOptions.wasi`; `analyzeSource` injects `__js2wasm_node_env.d.ts` as an extra root when set. **Dup-safe**: if the user already declares `process`, it detects the duplicate-identifier diagnostic and rebuilds without injection — never turns a benign warning into a hard error.
- `src/compiler.ts` / `src/compiler/output.ts` — thread `wasi: options.target === "wasi"`.

**Type-level only** — emitted wasm is **byte-identical** (md5-verified: the example host produces the same `nm_js2wasm.wasm` with and without the change). Codegen lowers `process.*` syntactically regardless.

## Verification

- Example host `--target wasi`: **5 → 0** `process` warnings; wasm md5 unchanged.
- `tests/issue-2589-process-ambient-typing.test.ts` (4 tests, all pass): resolves under wasi; still warns *without* wasi; genuinely-undefined names still warn (no blanket suppression); user-declared `process` → no dup-error.

Gate chosen: **wasi-implies-node-emulation** (lower friction than a separate `--emulate-node` flag, since `--target wasi` already lowers `process.*`). Pairs with #2523 / #2524.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


ttraenkler:
Superseded by #1925 (origin). Removed the accidentally-committed node_modules symlink and re-opened on origin (fork unpushable from the dev env). Work preserved in #1925.

PR #1961, opened 2026-06-22
fix(#2106): standalone strict ===/!== over type-erased nullish any

ttraenkler:
## Summary

Fixes a standalone-mode strict-equality observability bug (part of #2106's value-rep P3 work): `undefined === undefined` and `null === null` over two fully type-erased `any` operands returned **`false`** in standalone/WASI mode.

This is the exact shape the test262 `isSameValue` harness hits constantly (`isSameValue(a: any, b: any)` compiles both params to `externref`).

## Root cause

`binary-ops.ts` — the `noJsHost` externref-equality cascade (the #1776 isSameValue path). The **LOOSE** `==` branch already had a both-nullish guard (#2081), but the **STRICT** `===` branch went straight into the number → boolean → bigint → eqref-identity cascade. Two `ref.null extern` operands are none of number/boolean/bigint, so they reached the identity arm where `any.convert_extern(null)` fails `ref.test $eq` → returns `0`. There was **no "both nullish → equal" arm for strict**.

## Fix

Refactor the loose-only nullish guard into a shared `bothNullishGuard(inner)` applied to **both** modes:
`(lNull || rNull) ? (lNull && rNull) : <core cascade>`.

- Gated on `noJsHost` (standalone/wasi); **host mode byte-identical**.
- Loose behaviour unchanged (it always wrapped in the identical guard).
- `nullish vs non-nullish` stays `false` (`undefined === 5` → false).

## Known boundary (deferred to the tag-1 `$undefined` singleton, S1)

A **fully type-erased** `null === undefined` now returns `true` (want `false`): both operands are the identical `ref.null extern` bit pattern, so no inline test can split them. Distinguishing them requires the tag-1 `$undefined` singleton (this issue's S1 slice). The **type-aware** path (statically-known null/undefined types) is unaffected and still keeps `null === undefined` false — `tests/issue-1021-null-vs-undefined.test.ts` stays green. This is a strict net improvement over the prior state where *all* nullish strict-eq through erased `any` was `false`.

## Validation

- New: `tests/issue-2106-standalone-nullish-strict-eq.test.ts` (6 cases, host-free / zero `env` imports).
- No local regression: `tests/equivalence/{strict-equality-edge-cases,loose-equality,equality-mixed-types,typeof-comparison,comparison-coercion,null-narrowing,struct-null-comparison}` (50), `issue-1776` (6), `issue-1021-null-vs-undefined` (5), `issue-1127-samevalue`, `issue-2029` (3).
- Full test262 blast radius (incl. the erased `null===undefined` edge) measured in `merge_group` per the value-rep broad-impact protocol.

### DIFF
--- a/plan/issues/2106-valuerep-p3-undefined-observability.md
+++ b/plan/issues/2106-valuerep-p3-undefined-observability.md
@@ -2,10 +2,10 @@
 id: 2106
 title: "value-rep P3: undefined observability — UNDEF_F64 sentinel, union-collapse reversal (flagged), standalone $undefined singleton"
 status: in-progress
-assignee: ttraenkler/sdev7
+assignee: ttraenkler/sdev-async
 sprint: 65
 created: 2026-06-11
-updated: 2026-06-18
+updated: 2026-06-23
 priority: high
 feasibility: hard
 reasoning_effort: max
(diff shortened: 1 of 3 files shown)


ttraenkler:
**HOLD — labelled `hold` (do not auto-enqueue).**

On deeper probing the `null === undefined → true` edge is NOT rare — it reaches the most common nullish guard in the language. In standalone mode the RHS literal `undefined` and `null` BOTH compile to `ref.null extern` (indistinguishable by construction), so the shared `bothNullishGuard` fires for every `anyNullish === <literal nullish>` comparison:

- `x === undefined` where `x` is an `any` holding `null` → `true` (WRONG, want `false`)
- `x === null` where `x` is an `any` holding `undefined` → `true` (WRONG, want `false`)

This is a genuine representation trade with no inline resolution: while undefined and null share the same `ref.null extern` bits I can make `undefined === undefined`/`null === null` correct OR `null === undefined` correct, never both. The proper fix is the tag-1 `$undefined` singleton (this issue's S1 slice), which gives undefined distinct bits so all four cases resolve correctly.

Holding this PR in favour of S1. Keeping it open as the documented diagnosis + repro harness; the `bothNullishGuard` refactor is reusable once S1 lands (the guard then keys on the distinct representation, not bare `ref.is_null`).

ttraenkler:
Closing — superseded by the #2106 substrate path; fix-forward.

On probing, the standalone `null === undefined → true` edge is pervasive (it reaches the most common nullish guard in the language), and this manual rep-block can't merge: distinguishing `undefined` from `null` in standalone strict-eq requires the `$undefined` tag-1 singleton from #2106 S1, which must land as the architect-re-spec'd full producer+consumer sweep — a partial S1 subset breaches the standalone high-water floor (see PR #2025, NET −1245 test262 rows).

The follow-up is tracked in #2106 (being rescheduled to sprint 67). Retiring this 4-day-stale DIRTY partial in favor of that path.

PR #1968, opened 2026-06-23
fix(#2631): node:fs fd-based readSync/writeSync via a per-module shim

ttraenkler:
## Summary

Switches the Native Messaging host example from the js2wasm-specific `process.stdin.read(buf, offset)` / `process.stdout.write` shapes to **real Node fd-based `fs.readSync(fd, …)` / `fs.writeSync(fd, …)`** from `node:fs`, so the **same source** (a) compiles via js2wasm to standalone WASI **and** (b) runs **unmodified** under real `node`.

Fixes loopdive/js2#389: the reporter correctly noted `process.stdin.read(buffer, offset)` matches **no** real Node API (`process.stdin` is an async Duplex stream with no synchronous buffer-filling `read`). `fs.readSync`/`fs.writeSync` are the faithful synchronous primitives — fd-based (integer fd 0/1/2, no path, no filesystem), mapping 1:1 to WASI `fd_read`/`fd_write`. This is also what Javy uses (`Javy.IO.readSync`).

## Design

`node:fs` is a **per-module shim** (`examples/native-messaging/node-fs.wat`, `scripts/build-node-fs-shim.mjs`), mirroring `js2wasm:node-process` (#2524/#2625). Codegen only wires `import { readSync, writeSync } from "node:fs"` to calls of imported shim functions; the syscall sequence lives in the `.wat` shim, **not** in codegen.

The wasm import **module is `node:fs`** (the declared interface) with the real Node member names `readSync`/`writeSync` — the shim implementation name never leaks into the module's declared dependency. Whether the import is satisfied by `node-fs.wat`, a native WASI host, or the real `node:fs` under a JS host is a link-time concern.

**ABI note:** the narrow ABI keeps the `fd` parameter — `readSync(fd, ptr, len) -> i32` / `writeSync(fd, ptr, len) -> i32`. Dropping `fd` would break stderr telemetry routing (`writeSync(2, …)` must go to fd 2, distinct from stdout fd 1 — exactly the #389 stderr gap). The ABI can be widened to more of `node:fs` later.

## Changes

[...]

PR #1990, opened 2026-06-24
feat(#2632): Phase 2 — WASI fd-readiness reactor (multi-sub poll_oneoff + non-blocking fd0)

ttraenkler:
## #2632 Phase 2 — the fd-readiness reactor (substrate for process.stdin)

Turns the Phase-1 event loop's single-clock blocking sleep into a real reactor that waits on **fd0-readable OR the nearest timer** (multi-subscription `poll_oneoff`) and drains fd0 into an **internal stdin buffer** via a **non-blocking** `fd_read`. This is the substrate Phase 3's `process.stdin` Readable will pull from — Phase 2 does **NOT** yet expose `process.stdin`.

### What landed
- **`async-scheduler.ts`** — a SEPARATE fd-reactor run-loop body gated on `state.stdinReactor`:
  - `__rl_poll_fd0_or_clock(deadlineNs, nowNs)` generalises the #1484 single-clock marshaller to **2 subscriptions** (sub[0]=FD_READ fd0, sub[1]=CLOCK only when a timer is pending), scans the returned `event_t[]` and returns 1 if fd0 readable else 0.
  - `__rl_stdin_drain()` sets fd0 non-blocking once (`fd_fdstat_set_flags(0, NONBLOCK)`), non-blocking `fd_read`s available bytes into the internal buffer (reuses the page-1 `WASI_STDIN_BUF_START`); EAGAIN≠EOF, 0-byte read = EOF (drops the fd0 subscription).
  - The reactor exits only when no timer is pending AND fd0 has hit EOF.
  - All Phase-2 globals/helpers register **only** when the reactor is active, so timer-only / non-timer programs keep Phase 1's exact layout.
- **`index.ts`** — detect `__wasiStdinReadByte()`, register `fd_read` + `fd_fdstat_set_flags`, activate the reactor before `ensureTimerHeap`.
- **`calls.ts`** — lower `__wasiStdinReadByte()` (the internal-buffer test primitive Phase 3's `.read()` builds on) to the next buffered byte (0..255) or -1.
- **`late-imports.ts`** — lockstep-shift the async-scheduler stored func indices on a late import. **Root cause fix** for an "expected i64 but nothing on stack" invalid-Wasm bug: a string-concat import landing before the timer now-reader call was baked shifted defined funcs but not the stored `runLoopNowFuncIdx`. Mirrors the `nativeStrHelpers`/`mapHelpers` lockstep (#1677/#2162). Latent in Phase 1.
- **`runtime.ts`** — `poll_oneoff` polyfill decodes all subs and reports FD_READ when fd0 has stdin remaining, else CLOCK; `fd_fdstat_set_flags` no-op ack.

### Validation (#1968 batch-neutrality lesson)
- **125/125 programs compile byte-identical (SHA256) vs `origin/main` IN BATCH** (120 test262 files + explicit timer-only / interval-only / microtask-only / promise / plain programs). Verified both before and after merging current main (which carried the #2633 node-fs refactor).
- `issue-1326` / `issue-1326c` / `issue-2632-event-loop` / `wasi-timers` suites all green (the async-scheduler path the lockstep touches) — 49 tests green incl. the new suite.
- **New `tests/issue-2632-phase2-fd-reactor.test.ts`** (9 tests): compile-time wiring, polyfill-driven fd reactor (2/0/3-byte stdin, ordering), and end-to-end under **real wasmtime** with piped stdin — the reactor wakes on stdin readiness before a far timer, and on the timer when stdin is idle.

### Scope / bookkeeping
Phase 2 of a phased **GOAL**. `status` stays `in-progress` — Phase 3 (`process.stdin` Readable / #2635) and Phase 4 (Preview-2 `wasi:io/poll`) remain. **Does NOT close #2632.** Phase 2 requires the inline (non-`--link-node-shims`) `fd_read` path; shim integration is Phase 3.


ttraenkler:
Park diagnosed (run 28067978802, `test262 js-host shard 24`): **CI network flake**, not a regression. The shard crashed on `undici SocketError: other side closed` / `TypeError: terminated` (a failed fetch mid-shard) before producing any test results — `merge shard reports` failed only transitively. This is distinct from a real regression (which shows as completed-shard `assertion_fail` with wasm-hash change). The change is verified 125/125 byte-identical in batch vs main + real-wasmtime stdin/timer behavior. Confirmed flake; removing hold and re-enqueuing once.

PR #2145, opened 2026-06-26
fix(#2722): nested optional object-field destructuring default now fires

ttraenkler:
## What

```ts
function f({ a: { b = 3 } = {} }: { a?: { b?: number } } = {}): number { return b; }
```

| Call | Before | After (spec) |
|------|--------|------|
| `f()`              | 0 | **3** |
| `f({ a: {} })`     | 0 | **3** |
| `f({ a: { c: 1 }})`| 1 | **3** |
| `f({ a: { b: 5 }})`| 5 | **5** ✅ |

## Root cause + fix (Path A — two coordinated edits, ~10 LOC, no `function-body.ts` threading)

`a?` is the union `{ b? } | undefined`.

- **Change 1** — `src/codegen/index.ts`, the #1589A empty-object widening guard. It used `propType.getProperties().length === 0` to detect a "genuinely empty `{}`" field, but `getProperties()` on a **union** returns only the *common* properties — intersected with `undefined`'s empty set that is **always `[]`**. So the guard mistook *every* optional object field for an empty `{}` and clobbered the correct `ref_null structB` back to externref, routing the value through the host `__extern_get`/`__sget_b` f64-`0` else-branch so the nested `b = 3` default never fired. Now gated on the **resolved struct actually being empty** (`ctx.structFields`), preserving #1589A's genuinely-empty-`{}` intent (HasProperty / `indexOf.call` unchanged).
- **Change 2** — `src/codegen/literals.ts`, `compileObjectLiteral` contextual-type resolution. Strip a 2-member `T | undefined` union to `T` before `resolveStructName`, so an optional-typed inner literal `{}` / `{c:1}` builds as the inner `structB` (absent fields → sentinel) instead of falling through to a host externref that stores `ref.null` and makes the nested destructure throw *"Cannot destructure null"*. Purely additive to the struct branch — any `T` that maps to externref still falls through unchanged.

No `function-body.ts` / `runtime.ts` / `__sget_*` / destructure-reader changes (the field-type change flows through `ctx.structFields` and the existing struct-read destructure path already handles a nullable-struct field + nested `= {}` default).

## Validation

- `tests/issue-2722.test.ts` — **9/9** (gc + standalone; args built in-Wasm + called via no-arg exported wrappers, since a WasmGC struct param can't be passed from JS): four core repros 3/3/3/5; controls `g`/`h`/`m` green; 3-level deep (5/5/5/9); mixed optional+required (7/15); optional-no-default (1/2); optional-primitive (9/4); genuinely-empty optional (0/1).
- Regression basket diffed against `origin/main` — **0 new failures** (the reds in `issue-1542-repro` (2) + `null-destructure-param-object` (3) are pre-existing on main too). `issue-1589a` (the modified guard), `issue-2158`/`2545`/`2512`/`2568-standalone`, `issue-1372`, `object-literals` all green.
- `tsc --noEmit` clean; `prettier --check` clean.

## Floor

Broad-impact representation change → real validation is the **#2097 merge_group standalone shard**. Change 2 sits *after* the standalone `$Object` diversion, so standalone non-empty literals still route to `$Object` first; the union-strip only affects the closed-struct fallthrough.

Carved from #1556 (its #1543/#1544 cluster + single-level defaults are done); this is the narrow nested-optional residual.

https://claude.ai/code/session_01BuD4FvrUx5imJ9GqRgE2JY

### DIFF
--- a/plan/issues/2722-nested-optional-object-field-default-not-firing.md
+++ b/plan/issues/2722-nested-optional-object-field-default-not-firing.md
@@ -1,7 +1,9 @@
 ---
 id: 2722
 title: "Nested OPTIONAL object-field binding default not firing — externref field + f64 struct-getter can't signal field-absent"
-status: ready
+status: done
+completed: 2026-06-26
+assignee: ttraenkler/sd-accessor
 created: 2026-06-26
 updated: 2026-06-26
 priority: medium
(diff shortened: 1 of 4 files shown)


ttraenkler:
Closing — Path A (the 2-edit shortcut) is NOT floor-safe.

sd-accessor ran a bounded 1,660-file object-pattern dstr sweep (for-of / for-await-of / variable / class / function / arrow families) vs the merged baseline:
- Both edits (this PR): **NET -38** (44 regr / 6 impr) — cluster `*obj-ptrn-prop-obj-value-null` + `*dflt-obj-ptrn-prop-obj-init` ("Cannot destructure null").
- Change 1 alone (#1589A resolvedIsEmpty gating): forces every optional-object field to `ref_null struct` → nested-pattern fields deref `ref.null`.
- Change 2 alone (literals.ts T|undefined→T strip): **NET -18** — `*dflt-obj-ptrn-prop-obj` absent fields get a sentinel/0 instead of `undefined`.

No configuration is net≥0; neither edit is salvageable by narrowing (both perturb the shared nested-pattern representation). This confirms #1556's original assessment: #2722 needs the genuine substrate fix — **nullable-struct-ref threading through function-body.ts param resolution (~150-200 LOC), routed to architect** — not the 2-edit prototype. Re-routing #2722 to architect rework; the validation evidence is on the issue.

PR #2198, opened 2026-06-27
fix(soundness): #2750 S1+S2 — full sound umbrella for .js + externref-array OOB→undefined

ttraenkler:
Implements the two cheap, byte-neutral correctness wins from **#2750** (slices **S1 + S2**), consistent with the JS-semantics-first direction (#2751). #2750 stays **in-progress** (S3/S4/S5 remain).

## S1 — full sound umbrella for single-file `.js` (`src/checker/index.ts`)
`analyzeSource` used `strict: !isJs`, so a single-file `.js` got ONLY the pinned `strictNullChecks` (#2748 C) while BOTH multi-file blocks (`analyzeMultipleFiles`/`analyzeFiles`) use `strict: true` unconditionally. This unifies them: `.js` now gets `strict: true` (→ `strictFunctionTypes`, `strictPropertyInitialization`, `useUnknownInCatchVariables`, …) with an explicit `noImplicitAny: false` override (rejecting untyped JS is NOT the goal — the dynamic/externref path handles it). The `strictNullChecks: true` guard is kept explicit + commented (soundness-critical for the #2748 null-collapse miscompile).

## S2 — externref-element array OOB read → `undefined`, not `null` (`src/codegen/property-access.ts`)
Out-of-bounds index read on an externref-element array (`any[]`, `string[]`) returned `null` (`ref.null.extern`) only because the two plain-read call sites passed `useUndefinedSentinel=false`. Flipped to `true` (the #1396 machinery already existed for destructuring callers). Now `any[]`/`string[]` OOB → JS `undefined` (`__get_undefined`). Packed `number[]`/`boolean[]` are untouched — the sentinel only applies to externref/ref_extern element kinds; the full close (packed `T|undefined` rep) is the deferred S5 `noUncheckedIndexedAccess` epic.

## Validation / byte-neutrality evidence
- **S1 byte-neutral on the `.ts` corpus**: test262/equivalence compile via `test.ts` filenames → `isJs=false` → `strict` was ALREADY `true`. The change only affects the `.js` branch. Confirmed `.js` boundary holds: untyped `.js` still compiles (noImplicitAny off), null-guard still works (strictNullChecks on), `.ts` control unchanged.
- **S2 scoped to externref OOB**: `useUndefinedSentinel=true` is a no-op for packed f64/i32 elements (the sentinel path is gated on externref/ref_extern element kind), so `number[]`/`boolean[]` reads are byte-identical.
- **Regression test** `tests/issue-2750.test.ts` (7 cases): `any[]`/`string[]` OOB → undefined, OOB !== null, negative index, in-bounds unchanged; `number[]`/`boolean[]` OOB still sentinel (locks in the deferred S5 gap).
- Existing related suites green: `array-oob-bounds-check`, `issue-2001-s1-hole-literal`, `issue-1105-at-oob`, `array-methods` (66 tests).
- `tsc --noEmit` clean; `biome lint` clean; `prettier --write` applied.

Note: the #2750 issue file ("sound TS settings") lives on open PR #2195, not on main; this PR adds no issue file. (Heads-up: main currently has a *different* `id: 2750`, "Consolidate process docs" — an id-collision with PR #2195 to reconcile.)

### DIFF
--- a/src/checker/index.ts
+++ b/src/checker/index.ts
@@ -676,8 +676,22 @@ export function analyzeSource(source: string, fileName = "input.ts", analyzeOpti
   const compilerOptions: ts.CompilerOptions = {
     target: ts.ScriptTarget.ES2022,
     module: ts.ModuleKind.ESNext,
-    strict: !isJs,
+    // #2750 S1 — single-file `.js` now gets the FULL sound `strict` umbrella,
+    // matching both multi-file blocks (`analyzeMultipleFiles`/`analyzeFiles`,
+    // which set `strict: true` unconditionally). Previously `strict: !isJs`
+    // gave a single-file `.js` ONLY the pinned `strictNullChecks` (#2748 C),
+    // leaving `strictFunctionTypes`/`strictPropertyInitialization`/
+    // `useUnknownInCatchVariables`/… OFF — a latent single-file vs multi-file
+    // inconsistency. Sound flags keep type-directed codegen accurate.
+    strict: true,
+    // `strictNullChecks` is already implied by `strict: true`; kept explicit as
+    // the soundness-critical guard the #2748 C fix established (a `T|null`
+    // collapse changes the Wasm value-representation and folds null/undefined
+    // guards — the #2748 infinite-loop miscompile).
     strictNullChecks: true,
+    // BOUNDARY (#2750 Prong 1): `noImplicitAny` stays OFF for `.js` — rejecting
+    // untyped JS is NOT the goal; the dynamic/`any`/externref path handles it.
+    // This override MUST come after `strict: true` (which would enable it).
     noImplicitAny: false,
     noEmit: true,
     // Enable JSX parsing for .tsx/.jsx files. ReactJSX desugars JSX to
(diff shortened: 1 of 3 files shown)


ttraenkler:
Closing as **superseded by #2233** (merged 2026-06-28, folds #2760).

Verified on `main`:
- **S1** (single-file `.js` → `strict: true`): already on main (`src/checker/index.ts:712`, `strict: !isJs`). The only merge conflict here was comment-wording on that line.
- **S2** (externref/plain-array OOB → undefined): #2198 used a **shared bounds-check helper flip** (`emitBoundsCheckedArrayGet false→true`), which regressed the S2 canary (the cause of this PR's `check for test262 regressions` failure in the merge_group). #2760/#2233 reworked it the correct way via a **call-site** helper for primitive-element arrays (now on main: `src/codegen/property-access.ts`, `src/ir/from-ast.ts`), deliberately avoiding the shared-helper flip.

Net delta of this branch vs current main is effectively zero. No unique value remains; fixing it would mean redoing #2233's work on the rejected approach.

PR #2213, opened 2026-06-28
fix(#2745): bound-function partial-application args, [[Construct]] args, restricted-property poison

ttraenkler:
## #2745 — Function.prototype.bind residual

Fixes 3 of the 4 concrete bind bugs (groups **a/b/d**). Net test262
`built-ins/Function/prototype/bind`: **66 → 85 of 100 (+19, 0 regressions** in
the directory).

### Root causes fixed
- **(a) partial-application + over-arity `arguments`** — `__bind_function`
  (`src/runtime.ts`) wrapped the wasm-closure target with a *fixed-arity*
  `_wrapWasmClosure(target, lengthHint)` bridge that truncated call-time args to
  the target's declared formals, so a target reading `arguments[i]` past its
  formals never saw the bound/call args. Replaced with a dedicated **variadic**
  bridge dispatching at `max(args.length, realArity)` (plain `__call_fn_n` for an
  `undefined`/`null`/`globalThis` this, `__call_fn_method_n` for a real object
  receiver / `[[Construct]]`). All 7 `15.3.4.5.1-4-*` pass.
- **dispatcher `__argc` double-count** (`src/codegen/index.ts`) — `__call_fn_N`
  set `__argc = arity` (raw dispatcher arity) instead of the clamped-to-formals
  `closureArity` convention that `emitArgumentsVecBody`
  (`totalLen = argc + extrasLen`) and `maybeSetArgcForKnownCall`
  (`min(actual, paramCount)`) expect, so an over-arity HOF callback reported a
  *doubled* `arguments.length`. Fixed, and the same `#820l` plumbing was **added
  to `emitClosureMethodCallExportN`** (method dispatch forwarded no extras at
  all). Now matches V8 — a latent correctness fix beyond bind.
- **(b) `[[Construct]]`** — `new boundFn(...)` routed through
  `__construct(callee, NULL)` (the provably-non-constructable throw path) and so
  built with **zero args**. The `resolvesToNonConstructableValue -> __construct`
  arm (`src/codegen/expressions/new-super.ts`) now builds and passes the real
  call args; non-constructable A7 cases still throw at the IsConstructor check
  (before args are used). All 7 listed `15.3.4.5.2-4-*` pass.
- **(d) restricted-property poison** — `bound.arguments = {}` / `bound.caller =
  {}` were swallowed by `_safeSet`'s sloppy-builtin catch. The strict pre-check
  now resolves accessor descriptors along the prototype chain (skipping user
  proxies, so no Proxy MOP trap fires) and propagates the inherited
  `%ThrowTypeError%` poison setter's exception.
  `BoundFunction_restricted-properties.js` passes.

### Deferred (separate concerns, not regressions)
- **(c) `instance-length-*`** — needs a *configurable* host-wrapper `length` +
  bind reading the target's *runtime* `.length` (property-model, orthogonal).
- **`instance-construct-newtarget-self-new`** — `new.target === A` identity
  threading through the host bridge.
- `15.3.4.5-6-*` / realm tests remain pre-existing / out-of-scope.

### Validation
- `tests/issue-2745.test.ts` (9 cases: a/b/d + HOF `arguments.length` regression
  guard) — all green.
- Targeted regression check vs clean `origin/main`: `class-methods`,
  `default-params`, `issue-1337-bind-call`, `issue-1528-closure-construct`,
  `functional-array-methods`, etc. — identical pass set (the pre-existing
  `{ env: {} }` / missing-`helpers.js` harness failures are unchanged).
- `tsc --noEmit` + `biome lint` clean on changed files.

### DIFF
--- a/plan/issues/2745-function-prototype-bind-partial-application-length-newtarget.md
+++ b/plan/issues/2745-function-prototype-bind-partial-application-length-newtarget.md
@@ -1,7 +1,8 @@
 ---
 id: 2745
 title: "Function.prototype.bind: bound partial-application arguments, bound `.length`/`.name`, construct newTarget forwarding, restricted-property poison"
-status: in-progress
+status: done
+completed: 2026-06-28
 assignee: ttraenkler/agent-a3bfd116a51704f18
 sprint: current
 created: 2026-06-27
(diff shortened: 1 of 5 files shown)


ttraenkler:
### Coordination note for builtins (#2725 A / #2740 cluster-3)

This PR changes the closure-dispatcher `__argc` convention. There is **no shared
`emitClosureCallArgcExtras` helper** — the `#820l` plumbing is inline in each
emitter. Exact sites touched (both in `src/codegen/index.ts`):

1. **`emitClosureCallExportN`** — the per-entry `setupInstrs` now sets
   `__argc = entry.closureArity` (was `arity`, the raw dispatcher arity). The
   extras block (`arity - closureArity` elements → `__extras_argv`) is
   unchanged. This makes `__argc` match the **clamped-to-formals** convention
   that `emitArgumentsVecBody` (`totalLen = argc + extrasLen`) and
   `maybeSetArgcForKnownCall` (`min(actual, paramCount)`) already use.

2. **`emitClosureMethodCallExportN`** — NEWLY adds the same per-entry
   `setupInstrs` (it previously set neither `__argc` nor `__extras_argv`). It
   now fetches `ensureArgcGlobal` / `ensureExtrasArgvGlobal` / the extras arr
   type at the top, then per arm sets `__argc = entry.closureArity` and builds
   `__extras_argv` from user-arg locals `[closureArity+2 .. arity+2)` (method
   locals are `[thisVal, closure, arg0..]`, so user arg `i` is at local `i+2`).

**Convention to align to:** `__argc` is the **count of formal params filled**
(`closureArity`), NOT the raw dispatcher arity; overflow args live in
`__extras_argv`; `arguments.length = __argc + extras.length`. Any new
indirect-dispatch arm (async-gen #2725 A / `__call_fn_method_N` gaps #2740)
should set `__argc = closureArity` (per arm), not the dispatcher's `N`, or
`arguments.length` double-counts.

PR #2222, opened 2026-06-28
fix(#2760): hybrid floor F1 — plain-array primitive OOB read → JS undefined

ttraenkler:
## #2760 — Hybrid type-soundness floor F1

Implements floor fix **F1** of the hybrid soundness roadmap: an out-of-bounds read of a plain-array **primitive** element returns JS `undefined` (the SAFE default), not the type-default sentinel.

### What changed
- `number[]` (f64) and `boolean[]` (i32) OOB reads now return `undefined`. The unproven (non-bounds-eliminated) read widens its SAFE result to a **boxed-or-undefined externref** (box the in-bounds element; OOB → `__get_undefined`, or `ref.null.extern` under standalone where `undefined ≡ null`). The **proven** in-bounds read (`isSafeBoundsEliminated`) keeps the **unboxed fast path** — the hybrid invariant in miniature.
- New call-site helper `emitPlainArrayUndefinedOobGet` in `src/codegen/property-access.ts`, invoked from the two `compileElementAccessBody` plain-array read sites, gated on `classifyTypedArrayType(...) === "other"` (genuine array, not a typed-array view; the regex-match-vec exotic excluded).

### NOT the S2 shared-helper flip
The shared `emitBoundsCheckedArrayGet` default is **unchanged** — its `$__subview` / typed-array / array-method internal callers are byte-identical. Flipping that shared default was the S2 leak (#2198) that regressed `Array.prototype.map`-on-array-like.

### Deliberately deferred (with traced root cause)
- **Externref `any[]`/`string[]` OOB→undefined**: deferred. It regresses the S2 canary `built-ins/Array/prototype/map/15.4.4.19-8-b-2.js`, whose `testResult[2]` is an OOB read on a **length-0** map-of-array-like result (a separate pre-existing map bug). The test passes today only because OOB externref `null` coerces to the asserted `false`; the spec-correct `undefined` exposes that latent bug. Deferred to the IR `prove-then-specialize` work (#2766) / a map-length fix.
- **Object-element (`Foo[]`) arrays**: unchanged (widening to externref would break downstream `a[i].field`).

### Validation
- New `tests/issue-2760.test.ts` — 15 cases (primitive OOB→undefined host+standalone; in-bounds preserved incl. boolean tag + counted-loop fast path; typed-array unchanged; **map-on-array-like canary guard**).
- Local array/destructuring + `tests/equivalence/` batches: no new failures (only pre-existing `tests/helpers.js` infra + `fast-arrays`/`array-capacity`/equivalence failures that fail identically on clean `main`).
- Broad conformance: full CI / `merge_group` test262 regression gate + standalone-floor.

Closes #2760.

### DIFF
--- a/plan/issues/2760-hybrid-floor-plain-array-oob-undefined.md
+++ b/plan/issues/2760-hybrid-floor-plain-array-oob-undefined.md
@@ -1,7 +1,9 @@
 ---
 id: 2760
 title: "Hybrid floor F1: plain-array OOB read → JS `undefined` (HI-style #2198/S2 rework, not the shared-helper flip)"
-status: ready
+status: done
+assignee: ttraenkler/senior-dev-2760
+completed: 2026-06-28
 sprint: current
 created: 2026-06-28
 updated: 2026-06-28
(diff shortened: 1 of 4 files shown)


ttraenkler:
Closing this PR. The merge_group re-validation surfaced a real regression that PR-level checks missed: it breaks `test/built-ins/Object/values/symbols-omitted.js` (the OOB-read→undefined change leaks into Object.values indexed reads), with net -1 and zero test262 improvements — it delivers no measurable #2760 progress while introducing a regression.

This corroborates the independent finding (dev-rescue, recorded in plan/issues/2760) that the approach targets the wrong call-site: `number[]` lowers to a vec struct hitting property-access.ts:6341 (not the raw-array path this patch changes), and a correct fix requires discriminating plain-array vs typed-array at the vec site PLUS an externref-or-undefined result-representation change that ripples to every f64 consumer of `a[i]`. That is broad-impact, beyond this branch's scope.

#2760 is now `status: blocked` / `feasibility: hard` / senior-dev-scoped with the full root-cause documented, for the correct representation-change implementation. Closing forward rather than reworking in place. (#2766 remains blocked behind it.)

PR #2224, opened 2026-06-28
fix(#2757): bind object/array/member rest targets in array assignment-destructuring

ttraenkler:
## Summary

Carved from #2757 (parent #2669 destructuring umbrella). The array-rest branch
in `src/codegen/expressions/assignment.ts` only handled an **identifier** rest
target (`[a, ...r] = x`); an **object-pattern**, **array-pattern** or
**member-expression** rest target silently dropped every binding.

The collected rest vec is now built into a temp local and **dispatched on the
target kind**:

- **object pattern** `[...{ 0: x, length }]` — new `emitVecArrayLikeObjectDestructure`
  reads the vec as an array-like (numeric key → bounds-checked element, OOB →
  `undefined`; `length` key → vec length), per ECMA-262 §13.15.5.5. The nominal
  `emitObjectDestructureFromLocal` can't be reused — a vec struct isn't in
  `typeIdxToStructName`, so numeric keys never resolve.
- **array pattern** `[...[x]]` → `emitArrayDestructureFromLocal`
- **member expr** `[...obj.y]` → `emitAssignToTarget`
- **identifier** path unchanged (build → copy into the rest local).

## test262 flips fail→pass
`array-rest-nested-obj-undefined-own`, `array-rest-nested-obj-undefined-hole`
(the named acceptance cases), `array-rest-nested-obj`, `array-rest-nested-array`,
and similar in `language/expressions/assignment/dstr/`.

## Tests
`tests/issue-2757.test.ts` — adds 4 cases (object-pattern numeric+length, hole,
numeric index, array-pattern rest). The two existing partial-clamp tests still
pass (no regression to the identifier path).

## Deferred tails (separate follow-ups, not regressions)
1. empty/short **literal** source `[x,...r]=[]` uses a tuple/empty rep, not the
   vec path (issue point #2);
2. OOB **non-rest** element → `undefined` in the vec path (issue point #3);
3. **dynamic-object** member rest (`[...o.y]` with `o:any={}`) needs
   `__extern_set` property creation.

Validated on the full `merge_group` / test262 floor (broad-impact rest-bind refactor).

Closes #2757.


ttraenkler:
Re-validation note (CI-FIX): the merge_group park cited 2 `pass→compile_error` regressions (`language/expressions/assignment/dstr/array-rest-nested-array-undefined-{hole,own}.js`, bucket sig `67dba4c5fb149a22`). Diagnosed against an older speculative base (`ef4b886`). After merging current `origin/main`, both tests **pass** through the real test262 runner (verified locally via `runTest262File`) — the failure was a speculative-base artifact, not introduced by this PR's array-rest binding change. Net of the original run was already +5 pass. Removing hold and re-enqueuing once.

PR #2226, opened 2026-06-28
fix(#2669): typed in-bounds undefined/hole default-init fires for-of default

ttraenkler:
## #2669 slice — typed in-bounds `undefined`/hole default-init

Burns down the **typed in-bounds `undefined`/hole default-init** carve from the
#2669 umbrella tail. Does NOT touch assignment-destructuring rest-target lowering
(that's #2757/#2224) — fix is confined to `resolveWasmType` + the for-of default path.

### Root cause
An array whose element type is EXACTLY the `undefined`/`void` primitive —
`[[undefined]]` (TS infers `undefined[][]`) or a hole-only literal `[[,]]` —
resolved its element to **i32** via `mapTsTypeToWasm`, storing `undefined` as i32 `0`.
The inner `[undefined]` vec (built as `vec_externref`) was coerced to `vec_i32`
(`__unbox_number` → `i32.trunc_sat_f64_s` → `0`); the for-of path then read i32 `0`,
coerced to f64 `0.0`, which never matched the f64 sNaN sentinel the default-check
looks for — so `[x = 23]` never fired and `x` stayed `0`.

### Fix
One site in `src/codegen/index.ts` `resolveWasmType` Array branch: widen an
exactly-`undefined`/`void` vec element type from i32 → externref so the inner array
is built as `vec_externref` (no lossy coercion) and the existing
`wantUndefinedSentinel` for-of path (`__extern_is_undefined`) fires the default.
Mirrors the #1468 struct-FIELD widening. Scoped to the EXACT `undefined`/`void`
primitive — `T | undefined` unions already widened by `mapTsTypeToWasm`;
Number/Boolean/symbol excluded.

### Validation
- **6 targeted test262 wins** (all fail→pass): `language/statements/for-of/dstr/{const,let,var}-ary-ptrn-elem-id-init-{undef,hole}.js`
- **8 collateral wins** across for-of + assignment dstr (`array-elem-nested-{array,obj}-undefined-own`, `array-rest-before-elision`, `array-rest-elision-invalid`, `obj-rest-not-last-element-invalid`)
- **0 real regressions** in scoped sweeps over for-of/dstr (569), let/dstr (94), assignment/dstr (368)
- Guard `tests/issue-2669.test.ts` Bug 3 added — 14/14 green; existing array suites green

Umbrella stays OPEN (status `ready`) — remaining tail: nested object-default (`[{a}={a:1}]`), iterator-protocol (#1642/#2566).

### DIFF
--- a/plan/issues/2669-es6-destructuring-correctness-residual-umbrella.md
+++ b/plan/issues/2669-es6-destructuring-correctness-residual-umbrella.md
@@ -268,6 +268,48 @@ PASS via fresh single-file runs; 12 sync improvements re-verified PASS.
 This slice keeps the umbrella OPEN (status stays `ready`) — it burns down the
 nested-array-default codegen corner, not the iterator-protocol tail.
 
+## Slice landed (issue-2669-typed-default-init, 2026-06-28) — typed in-bounds undefined/hole sentinel
+
+Resolves the **typed in-bounds `undefined`/hole default-init** carve named in the
+"Remaining umbrella tail" above (the `for (let [x=23] of [[undefined]])` / `[[,]]`
+case). Root-caused and fixed; guard tests `tests/issue-2669.test.ts` "Bug 3"
+(14/14 green).
+
+### Root cause
+An array whose element type is EXACTLY the `undefined` (or `void`) primitive —
+`[[undefined]]` (TS infers `undefined[][]`) or a hole-only literal `[[,]]` —
+resolved its element type to **i32** via `mapTsTypeToWasm` (which maps
+`undefined`/`void` → i32 for the "no result" case). The inner `[undefined]` vec,
+built as `vec_externref` holding the `undefined` externref, was then COERCED to
+`vec_i32` (`__unbox_number` → `i32.trunc_sat_f64_s` → `0`). The for-of array path
+read that i32 `0`, coerced it to f64 `0.0`, and the default-check
+(`i64.reinterpret_f64` == sNaN sentinel `0x7ff00000deadc0de`) never matched — so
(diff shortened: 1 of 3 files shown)


ttraenkler:
Closing — this approach is architecturally unsound and cannot land safely.

Root cause of the 5 merge-group regressions: the `undefined[]`/`void[]` → externref widening in `resolveWasmType` is **global** (resolveWasmType must be type-deterministic — same TS type resolves to the same backing everywhere). It changed the array backing for ALL undefined-typed arrays, not just the for-of-default case, which broke:

- **Array construction** — `new Array(undefined, undefined)` emits `array.new_fixed` with `i32.const 0` for undefined args → invalid Wasm ("expected externref, found i32"): the 2 compile_errors (S15.4.2.1, sort).
- **i32/f64 consumers** — `.length`, `sort`, `reduceRight`, `for-in` assume numeric backing → the fail / null_deref / illegal_cast (S15.4.1, reduceRight, for-in-await).

The +35 wins and −5 regressions are **inseparable** in this approach. Scoping attempts (`_depth>=1` gate; f64 backing) both fail — depth can't distinguish the for-of inner array from `new Array` (both resolve at depth 0), and a depth-dependent type breaks consistency.

The correct fix is **representation-level** and routed to an architect carve. Key asymmetry for that work: binding patterns already pass (`let [[x=23]] = [[undefined]]` → 23) because TS infers `[[undefined]]` as a TUPLE (tuple-struct → f64 sNaN sentinel preserves undefined); the identical for-of fails because TS infers an ARRAY (vec_i32 → undefined lost at construction). Branch left intact; PR was only ever parked, so closing forgoes nothing on main.

PR #2254, opened 2026-06-28
fix(#2789): packed-i32 array read-side soundness — demote overflow/-0 writes to f64

ttraenkler:
## Hybrid fast-path audit Row 3 (miscompile fix)

`isI32SafeExprForArray` / `collectI32SpecializedArrays`
(`src/codegen/array-element-typing.ts`) packed a `number[]` into an
`array<mut i32>` when each write looked "i32-safe". But the predicate admitted
`+`/`-`/`*` arithmetic (stale "overflow wraps mod 2^32" comment), arithmetic
compounds (`+=`/`-=`/`*=`), and `-0` — values whose **i32 image differs from
their true f64 value**. Codegen evaluates `number op number` as f64 and stores
it into the i32 slot via `i32.trunc_sat_f64_s` (which **saturates** on overflow,
not wraps), and i32 cannot represent `-0`. A read promoting the element back to
f64 then observed a **wrong number** — a MISCOMPILE, not a deopt.

Confirmed on `origin/main` (probe):

| Source | main (i32-backed) | correct (f64) |
|---|---|---|
| `arr.push(a*b)`, `a*b=2.5e9` | `2147483647` | `2500000000` |
| `arr.push(a+b)`, `a+b=4e9` | `2147483647` | `4000000000` |
| `arr.push(-0); 1/arr[0]` | `+Infinity` | `-Infinity` |
| `arr[0] += 2e9` (after `=2e9`) | `2147483647` | `4000000000` |

This is exactly the **#1236** scalar-local saturation fix, which was never
applied to the array element path.

## Fix — conservative narrowing

Admit **only canonical-i32 producers** (value bit-identical to its f64 image):
reject `+`/`-`/`*` and arithmetic compounds (`+=`/`-=`/`*=`, `>>>=`); admit unary
`-` only for a **non-zero integer literal**. Bitwise/shift/comparison/`~`,
i32-locals, non-negative integer literals, and the wrap-canonicalising `(expr) |
0` idiom are unchanged — so perf-sensitive hash/mask patterns keep the fast path.

Once every write is canonical i32, the **read side is discharged for free**: the
i32-backed read promoted to f64 equals what the f64 backing would store, so no
read can observe a distinction i32 erases. A separate read-side AST gate was
deliberately not added (redundant for soundness, and a buggy read-side analysis
is the "risky gate" this row warns about).

**Strict narrowing** ⇒ can only fix a miscompile or deopt to the always-correct
f64 backing; never introduces a new one. `new Array(n)` holes don't diverge
(both backings `array.new_default` zero-init).

## Validation

- `tests/issue-2787.test.ts` — 18/18 pass: the four previously-miscompiled cases
  now spec-correct; `-0` preserved (push + element-assign); a read-side float-op
  observation; division/NaN guards; preservation of `(a*b)|0` / negative-literal
  / bitwise (+ bitwise-compound) packing; WAT-level proof that dangerous arrays
  demote (`$__vec_i32` absent) while canonical arrays keep i32.
- Existing i32 tests A/B-checked: failure counts identical with original vs
  patched `array-element-typing.ts` (no new failures; those files have
  pre-existing unrelated failures on `origin/main`).
- Broad-impact → merge_group full test262 + standalone-floor are authoritative.

Closes #2787.


ttraenkler:
Renumbered #2787 → **#2789** (id collision: a parallel PR took #2787 = `2787-differential-test-corpus-failures`, now on main; my `--no-pr-scan` allocation raced it). The merge_group auto-park was caused **solely** by the `quality` required check failing on that DUPLICATE-ID — now fixed (verified locally: `check:issues` 0 duplicates, `check:issue-ids:against-main` OK).

The merge_group diff-test `array/01-basic.js: match → malformed_wasm` delta is **not** from this change and is **not a required check** (it did not appear in the auto-park's `Failed checks`). A/B-proven identical with/without the `array-element-typing.ts` edit, and `const a = [1,2,3]` is not an i32-specialization candidate. It is a pre-existing `__module_init` operand-rep bug already tracked as **#2788** ("reproduced locally on origin/main @867cdfbd"). Test262 Sharded passed in the merge_group.

PR #2300, opened 2026-06-29
fix(#2825): defer block-nested class bodies so captured-globals promotion fires

ttraenkler:
Fixes #2825 (Bug C class-method half of #2818).

## Problem
A class nested in a **block** (or `if`/loop/`switch`/`try`/labeled scope)
**inside a function** captured an outer block-`let` as `0`/`null` instead of the
live value:

```ts
export function test(): number {
  { let s = 42; class C { m(): number { return s; } } return new C().m(); }
}
// before: 0   after: 42
```

## Root cause (pure compile-ordering)
Class methods capture an enclosing local by promotion to a `__captured_<name>`
global (`promoteAccessorCapturesToGlobals`). That promotion runs at the textual,
in-scope position **only when the class body is DEFERRED**
(`ctx.deferredClassBodies`), which `compileClassesFromStatements`
(`src/codegen/declarations.ts`) does for a class that is `insideFunction`. The
function-body recursion set `insideFunction = true`, but the block / `if` / loop
/ `switch` / `try` / labeled recursions **dropped** it — so a class nested in a
block-inside-a-function compiled **eagerly** in the pre-pass, before the
enclosing function ran and before the block-`let` was a promotable local. The
method resolved the captured name to the `ref.null` graceful default, promotion
never fired, and `compileNestedClassDeclaration` early-returned (struct collected
AND eagerly compiled, never deferred), skipping the in-scope promotion + body
compile.

## Fix
Forward the **current** `insideFunction` through every control-flow recursion in
`compileClassesFromStatements` so a block-nested-in-function class is **deferred**
to its textual position (after the block-`let` initialises), reusing the proven
`deferredClassBodies` path. A module-top-level control-flow scope stays
`insideFunction = false` → still eager. One-word change per call site; no change
to `nested-declarations.ts` / `closures.ts`.

## Verification
- **WAT**: `(global $__captured_s ...)` now emitted (absent before), `$test`
  syncs it, the method reads it. Runtime returns 42 (was 0).
- **`tests/issue-2825.test.ts`**: 29 pass / 1 `it.skip` (sibling-block same-name
  collision the spec scoped out). Covers num/string/static/generator/private/
  param-default/ctor/accessor captures; every control-flow scope;
  mutation-after-decl re-sync (#1672); reachability stress.
- **Reachability edge case (spec-flagged) resolved without a post-pass sweep** —
  a nested fn's body compiles regardless of being called, so a deferred
  block-nested class is always reached.
- Existing suites (`issue-1161`, `issue-1594a/b`, `generators`,
  `generator-methods`, `accessor-side-effects`, …) — 56/56 actual tests pass.
  The sweep's only "failures" (`classes.test.ts` family, `issue-1712`) reproduce
  **identically on unmodified `main`** (stale-harness `string_constants` +
  pre-existing `#1712`), not introduced here.

## Broad-impact note
Changes class-body-compile **timing** → authoritative validation is the
**merge_group full CI incl. standalone floor** (host/gc ordering change, but the
floor only runs on `merge_group`).

### DIFF
--- a/plan/issues/2825-block-nested-class-method-capture-deferred-compile-ordering.md
+++ b/plan/issues/2825-block-nested-class-method-capture-deferred-compile-ordering.md
@@ -3,8 +3,10 @@ id: 2825
 title: "Bug C (class-method half, spec'd): block-nested class compiled eagerly, so captured-globals promotion never fires for an outer block-let"
 parent: 2818
 related: [2820, 2818, 2826, 2811, 2669, 1672]
-status: ready
+status: done
+assignee: ttraenkler/impl2825
 created: 2026-06-29
+completed: 2026-06-29
 priority: high
 feasibility: hard
 reasoning_effort: max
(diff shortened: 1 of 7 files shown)


ttraenkler:
Deferring #2825 (lead's call). This broad class-body-**timing** change (defers every block-nested class) parked the merge_group TWICE on the same required checks (merge shard reports + check for test262 regressions):
- 1st park = a latent cross-scope `capturedGlobals` leak (fixed on-branch, byte-identical proof over 18 shapes).
- 2nd park = another regression the local control set couldn't cover — the ephemeral merge_group ref is pruned right after the run, so each fix is made partly blind. The broad blast radius exceeds local validation ⇒ whack-a-mole.

The fix + full diagnosis are preserved on branch `issue-2825-class-method-capture` (head 2f835ee). Routing #2825 to architect for a NARROWER approach (a targeted promotion that does not re-order global class-body compile timing). Closing rather than leaving it parked-stranded; reopen when re-spec'd.

PR #2322, opened 2026-06-29
fix(#2838): L4+L5 — member-read host-MOP routing + this-truth

ttraenkler:
## #2838 epic, PR 2 of 3 — L4+L5 (member-read substrate)

Builds on PR1 (#2319, L3, merged). Lands the **member-read** layer so a prototype accessor installed at runtime via `Object.defineProperties(C.prototype, …)` is consulted on a `this.<field>` read against a typed receiver.

### Changes (`src/codegen/property-access.ts`)
- **L5** — `resolveStructNameForExpr` (~line 1070): for a `ThisKeyword` receiver, return `resolveThisStructName(ctx, fctx)` (the fctx `this` local's actual ref type) instead of the TS-contextual type. Inside a runtime-installed accessor getter, TS contextually types `this` as the descriptor literal object (`{configurable}` → `__anon_N`), so the old path lowered `this.<x>` against the wrong struct and read a default slot. The fctx local is the runtime truth: a dynamic getter's local is externref → `undefined` → fully dynamic (host MOP); a genuine typed method's local is the correct struct ref (truth **agrees** → no change).
- **L4** — the `#856` sidecar/MOP block (~line 5512): relax its `!typeName` gate so a `__fnctor_*`/`__anon*` typed receiver also reaches the existing `extern.convert_any` + `__extern_get` path (which consults `_fnctorProtoLookup` for runtime-installed prototype accessors). Positioned as the **last resort** — the static field fast path and the auto-register path (when the field is on the TS type) both run first, so the hot struct-field read is untouched; only genuinely-absent fields take the MOP route. Gated on `!noJsHost` — standalone keeps its existing default.

### Validation
- The var-descriptor accessor probe (`{ f: { get(){ return this.flags } } }` installed via `Object.defineProperties(C.prototype, acc)`, read as `new C().f`) flips from `null`/`null` to the correct `2`/`55` — the getter now reads the real receiver.
- **Non-regressing**: identical failing sets vs the L3-only baseline across closure/accessor, fnctor/this/proto, and class-method suites (per-test JSON diff). Additionally clears the flaky `#1742` this-receiver-vec read. Typecheck + prettier clean.
- L4+L5 do **not** yet parse acorn `return` — the runtime-installed **method-call** dispatch (`new C().read()`, `this.currentVarScope()`) is the remaining **L6** wall, landed in PR3.

Broad-impact (L5 touches all `this.<x>` reads) ⇒ full `merge_group` + standalone-floor.


ttraenkler:
Superseded by #2325. A prior origin/main merge on the L6 branch dropped these L4+L5 changes from history, so they are restored together with L6 in #2325, with L5/L6 narrowed to the precise `__anon`-descriptor case that addresses this PR's merge_group park (static-private-accessor cluster). Closing to avoid duplicate L4+L5 in the queue.

PR #2329, opened 2026-06-29
fix(#2841): wrap host-array struct elements so arrow/expr params keep name+type

ttraenkler:
## #2841 — arrow / function-expression param Identifier nodes lose name+type on host readback

The genuine remaining gap for compiled acorn to parse real Native-Messaging files LITERALLY equal to node-acorn. Surfaced by the UNCAPPED NM differential (the default `maxDivergences:8` truncated before reaching function bodies, hiding it — #2836's "zero non-quirk" claim was a capped-diff artifact).

### Root cause — HOST MARSHALLING, not value-rep, not codegen
The arrow param Identifier's `name`/`type` ARE present in-wasm (reachable via `__extern_get`). acorn parses `(reason)` as a paren `exprList`, then calls `parseArrowExpression(node, exprList)` — a **dynamic method dispatch**. The vec arg is coerced to `externref` at the any-boundary, where the host shim `__make_iterable` (#2836) materialises it into a **real host JS array** of opaque wasm structs, which becomes `node.params`. On readback the `_wrapForHost` proxy returned that JS array RAW (a JS array is not a wasm struct; `__is_vec` is 0), leaving its Identifier elements unwrapped → plain JS `.type`/`.name` read `undefined`. Declaration / function-expression params take the `parseBindingList` path and stay a genuine wasm vec (routed to `_wrapVecForHost`, which wraps elements) — so only ARROW params were affected.

### Fix (`src/runtime.ts`, host-only marshalling)
1. **`_wrapForHost` proxy `get` (load-bearing)** — wrap real-JS-array struct elements on index read via the new `_wrapHostArrayElems` cached view (mirrors `_wrapVecForHost`). This alone takes background.js to 0 non-quirk.
2. **`_wasmToPlain` host-array recursion** — symmetric robustness for the marshal:`"copy"` / `JSON.stringify` consumers (guarded by the unit test).

Standalone keeps native WasmGC vec readers — no floor impact.

### Verification (uncapped NM differential, `maxDivergences: 100000`)
| file | before | after |
|---|---|---|
| `background.js` | 2 non-quirk (`params[0].type`, `.name`) | **0** |
| `edge.js` (post-#2325) | n/a (return wall) | **4**, ALL `attributes` → #2849; every one of the 62 arrow/expr params cleared |

The only remaining divergences anywhere are the two accepted quirks (`sourceFile`, bool-as-i32). Decl params stay correct (control). `#2836` + dogfood `acorn.test.ts` green; full typecheck clean. Regression test: `tests/issue-2841-arrow-param-name-type-marshalling.test.ts` (4 cases).

### Separate gap carved out → #2849
The spurious `attributes: []` on edge.js import/export nodes is NOT marshalling — it's a distinct dynamic-property type-inference codegen bug (compiled acorn doesn't normalise `ecmaVersion 2022 → 13`, so `2022 >= 16` wrongly enables import-attributes). Distinct root cause, distinct layer.

### DIFF
--- a/plan/issues/2841-arrow-param-name-type-marshalling.md
+++ b/plan/issues/2841-arrow-param-name-type-marshalling.md
@@ -140,9 +140,13 @@ is a cached lazy wrapper, no copy).
 
 - [x] `background.js` uncapped differential: ZERO non-quirk divergences.
 - [x] arrow/expr params carry correct `name`+`type`; decl params stay correct.
-- [~] `edge.js` ZERO non-quirk: arrow-param half verified (with #2325 stacked);
-  the remaining `edge.js` non-quirk is the spurious `attributes: []` →
-  carved to **#2849** (distinct codegen bug, also blocked behind #2325).
+- [x] `edge.js` arrow/expr params: ALL cleared. With #2325 now merged to `main`,
+  edge.js parses and the uncapped differential shows **4** non-quirk
+  divergences, ALL the spurious `attributes: []` (every one of the 62 arrow/expr
+  params + all other prior divergences are gone; the 276 boolean mismatches are
+  the accepted bool-as-i32 quirk). The remaining 4 `attributes` are carved to
+  **#2849** (distinct dynamic-property type-inference codegen bug). edge.js will
+  reach 0 non-quirk once #2849 lands.
 - [x] 0 test262 regressions expected (host-marshalling, additive); full
   merge_group + standalone-floor on CI.
 
--- a/plan/issues/2841-arrow-param-name-type-marshalling.md
+++ b/plan/issues/2841-arrow-param-name-type-marshalling.md
@@ -1,0 +1,157 @@
+---
+id: 2841
+title: "arrow / function-expression param Identifier nodes lose name+type on HOST readback (compiled-acorn AST != node-acorn)"
+status: done
+completed: 2026-06-29
+assignee: ttraenkler/agent-senior-2841
+sprint: current
+priority: high
+horizon: m
+feasibility: hard
+reasoning_effort: max
+created: 2026-06-29
+task_type: bugfix
+area: runtime-marshalling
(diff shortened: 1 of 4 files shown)


ttraenkler:
### Also resolves #2844 and #2845 (same root cause)

The corpus agent flagged two more gaps suspected to share this PR's marshalling root cause — "node-typed elements of a specific array property come back blank via the `_wrapForHost` fall-through". Confirmed: this fix clears **both**. `quasis[]` and `expressions[]` are the identical pattern as arrow `params[]` (an array-valued struct field that arrives at the host as a real JS array of opaque node structs).

Verified on freshly-compiled pinned acorn@8.16.0 with this branch:

**#2844 — TemplateLiteral `quasis[]` (parse `` `a${x}b` ``):**
```
quasis[0]  oracle  : {type:"TemplateElement", tail:false, value:{raw:"a",cooked:"a"}}
           compiled: {type:"TemplateElement", tail:0,     value:{raw:"a",cooked:"a"}}
quasis[1]  oracle  : {type:"TemplateElement", tail:true,  value:{raw:"b",cooked:"b"}}
           compiled: {type:"TemplateElement", tail:1,     value:{raw:"b",cooked:"b"}}
```
`type` + `value` (raw/cooked) are no longer blank; the only remaining delta is `tail: 0/1` vs `false/true` — the accepted **bool-as-i32** quirk.

**#2845 — SequenceExpression `expressions[]` (parse `(a,b,c)`):**
```
expressions[0]  oracle/compiled : {type:"Identifier", name:"a"}
expressions[2]  oracle/compiled : {type:"Identifier", name:"c"}
```
Exact match — no longer blank.

Both are covered by the same mechanism guarded in `tests/issue-2841-...test.ts` (host-array struct-element wrapping). Leaving the #2844/#2845 issue-file status flips to their owner since those files aren't on this branch.

PR #2333, opened 2026-06-29
fix(#2826): re-point immutable CPS block-let captures to fresh slot (Bug C)

ttraenkler:
Fixes **#2826** (Bug C, CPS-capture half — carved from #2818, parent #2669).

## Problem
A block-scoped `let`/`const` **immutably** captured by a hoisted **async/generator**
function declaration read the stale pre-hoisted slot (0/null) instead of the
captured value:

```ts
export async function t5(): Promise<number> {
  { let s = 42; async function f(): Promise<number> { return s; } return await f(); }
}   // returned 0, should be 42
```

#2820's producer-side slot reuse deliberately **skips** the slot collapse when any
CPS (async/generator) function captures the name — collapsing perturbs the
for-await-of continuation state machine (43 `for-await-of/async-{func,gen}-decl-dstr-*`
regressions). That left the immutable CPS capture pinned to the never-written
pre-hoisted slot A while `let s = 42` stored into a fresh slot B.

## Fix (Design 1A — architect-specced, `architect_spec: done`)
In `src/codegen/statements/variables.ts`, on the **complementary** branch to #2820's
reuse gate: keep both slots (B is the real storage) and **re-point** the recorded
capture metadata `outerLocalIdx` (and matching `outerTdzFlagIdx`) from A to B for
**immutable** captures only. Mutable boxed captures already thread correctly and are
exactly the 43-regression class — left untouched. Slot **layout is unchanged** (A
stays dead-allocated), so the continuation state struct is byte-identical to
baseline. Mutating the single source of truth (`nestedFuncCaptures[*].outerLocalIdx`)
means all downstream call/construction sites resolve to B with no read-site edit.

## Tests
`tests/issue-2826.test.ts` (9): block async/generator capture (numeric + string),
mixed plain+CPS (→84), `const` variant, fn-scope + plain-block (#2820) controls.

Scoped regression (local, all green): #2820 (15), #2811 (9), async-await,
flatmap-closure, generator-method-destructuring. The 43-test for-await-of
mutable-capture class is validated on the full `merge_group`.


ttraenkler:
Closing this PR. Its change — re-pointing an immutable CPS block-`let` capture from the stale pre-hoisted slot A to the fresh slot B in `variables.ts` (Design 1A) — is proven **un-gateable** and cannot land as a producer-side heuristic.

The merge_group re-validation (the auto-park-bot flag) caught a real net −8 (30 regressions / 22 improvements, all in the same `for-await-of/async-{func,gen}-decl-dstr-*` area — a same-area tradeoff that breaks more than it fixes). Since then two more approaches were tried and both reproduce the **identical** CPS-state perturbation (`async-generator-interleaved` null-deref + `array-elem-iter-nrml-close` nextCount=2): Design 1B (construction-site re-resolve — the #1177 minefield) and write-through-to-A.

The decisive finding (documented in the issue file's `## Implementation attempt 2`): at `compileVariableStatement` the name that **needs** the fix (`length`) and the names that **regress** (`nextCount`/`iterator`/`iterations`) are byte-for-byte identical in every available signal (same mutability, same not-a-captured-global, same single capturer). The distinguishing information — object-literal iterator-method mutation, the call-graph through suspending continuations, snapshot timing — is whole-program and not materialized as metadata at the let-init site. No predicate computable there separates safe from unsafe.

Per the issue verdict this is genuinely architect-owned: it needs the transitive capture+call-graph / continuation-snapshot analysis (recommendation #1), applied only when no capturer of the name snapshots it into a suspending continuation. #2826 stays `status: blocked` (already on main) pending that design. This branch + the full diagnosis are preserved for the architect; nothing is lost by closing.

PR #2335, opened 2026-06-29
fix(#2818): defer block-nested class bodies inside functions (captured block-let reads null)

ttraenkler:
## Summary

Fixes #2818 — a block-scoped `let`/`const` captured by a **class method** read back `null` (0 for numeric). The `meth-…` / `gen-meth-…` / `private-meth-…` cluster of the `ary-ptrn-rest-obj-prop-id` test262 family hits this.

```ts
export function test(): string {
  { let s = "outer"; class C { m(): string { return s; } } return new C().m(); }
}
// before: null   after: "outer"
```

## Root cause

The eager class-body compile pass `compileClassesFromStatements` (`src/codegen/declarations.ts`) propagated its `insideFunction` flag only when recursing into a **function body**, but **dropped it** for `block` / `if` / `for|while|do` / `switch` / `try` / `labeled` recursions. So a class nested in such a statement *inside a function* was compiled **eagerly** (as if module-level) via `compileClassBodies` instead of being added to `deferredClassBodies`. Eager compilation runs before the enclosing block-`let` initialises, so `promoteAccessorCapturesToGlobals` never fired and the method body resolved the captured name to the `ref.null.extern` fallback. (The `compileNestedClassDeclaration` early-return the issue cited is a *symptom*: the class isn't deferred, so it bails before promotion.)

## Fix

Propagate `insideFunction` through every control-flow recursion (~9 call sites). A function-nested class is now **deferred** and compiled in-scope by `compileNestedClassDeclaration` after the block-`let` stores, where promotion + the `local.get; global.set __captured_<name>` sync land correctly. Module-level blocks keep `insideFunction === false` (eager, **unchanged**).

## Verify-first (host/gc)

| case | before | after |
|---|---|---|
| `methodCapture` (block-let, method) | `null` | `"outer"` |
| `arrowInMethod` | `null` | `"outer"` |
| `numeric` block-let | `0` | `42` |
| `fnScopeControl` (non-block, regression control) | `"outer"` | `"outer"` |

`tests/issue-2818.test.ts` (11 cases: string/numeric, arrow-in-method, generator / private / static methods, param-default, mutation-observed, if-block, for-block, fn-scope control) — all green. Proper-harness closure/class/TDZ suites (#1672, #2623, #2587, #1128, #1573, class-static-private-this) have **identical** pass/fail on base vs this branch (no regression; the 2 pre-existing fails in #1712/generator-nested fail the same way on `main`).

## Out of scope — filed as #2854

Verify-first surfaced an **orthogonal pre-existing** bug: a **doubly-nested** (if-in-for) **numeric** block-`let` *captured* by a closure (arrow OR class method) fails Wasm validation with `ref.is_null[0] expected reference type, found local.get of type f64` (TDZ-flag-on-numeric-capture). It reproduces on `main` with a plain arrow (no class) and is independent of this fix. Filed as #2854.

### DIFF
--- a/plan/issues/2818-block-let-captured-by-class-method-captured-globals-ordering.md
+++ b/plan/issues/2818-block-let-captured-by-class-method-captured-globals-ordering.md
@@ -2,8 +2,10 @@
 id: 2818
 title: "Bug C (class-method half): block-scoped let captured by a class method reads null (captured-globals promotion ordering)"
 parent: 2669
-related: [2820, 2811, 1672]
-status: ready
+related: [2820, 2811, 1672, 2854]
+status: done
+completed: 2026-06-30
+assignee: sendev-ecmaver
 created: 2026-06-29
 priority: high
 feasibility: hard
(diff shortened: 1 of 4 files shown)


ttraenkler:
Closing this PR. Its change — propagating `insideFunction=true` through every block/if/loop/switch/try/labeled recursion in `compileClassesFromStatements` — is too broad and is a proven merged-baseline regression that cannot be salvaged by resolving this branch's conflicts.

The merge_group re-validation (the auto-park-bot flag) caught it: net −471 (545 regressions / 74 improvements, ratio 736%), with two buckets each over the 50-test gate limit — `language/expressions/class/dstr` (335) and `language/expressions/class/elements` (165). Root cause: the broad propagation adds ALL nested classes (including class **expressions** `const C = class {…}` and block-nested class declarations) to `ctx.deferredClassBodies`, but the deferred-compilation path in `compileNestedClassDeclaration` does not cover those shapes — so their method/element bodies are never compiled and codegen is silently dropped (WAT shrinks ~3 KB on `class/dstr/async-gen-meth-obj-ptrn-list-err.js`; a runtime probe flips it pass→fail).

The real fix (per the issue file) is the narrowed Option 1: defer **only** classes that genuinely capture a block-scoped `let` from the enclosing block, keeping every other block-nested / class-expression class on the eager path — and it must also handle the incomplete deferred path (the class-collection-ordering + early-return in `nested-declarations.ts`). That is a fresh architect-spec'd effort (#2818 is `architect_spec: needed`, `status: blocked` on main), not a conflict-resolution of this branch — resolving conflicts here would just re-enqueue the regressing broad-deferral. It also requires validation on a **full merge_group / local-CI test262** run before re-enqueue — the 545-test cluster is invisible to PR-level checks (test262 shards are stubbed at PR level).

#2818 stays `status: blocked` (already on main) with the full −471 diagnosis. This branch is preserved for the narrowed rework; nothing is lost by closing.

PR #2355, opened 2026-06-30
fix(#2869): write member-expression destructuring assignment targets

ttraenkler:
## #2869 — member-expression assignment-target destructuring

Destructuring with a `PropertyReference` target (`[x.y]=vals`, `{k:x.y}=src`, `for ([x.y] of …)`) **silently dropped the write** in both the JS-host (gc) and standalone (host-free) lanes; the gc array path additionally recursed (`Maximum call stack`) from a funcIdx-repoint off-by-one. Per ECMA-262 §13.15.5 a `DestructuringAssignmentTarget` may be a PropertyReference, so the value must store through the member.

### Fix (architect spec, 3 changes)
1. **`emitAssignToTarget`** — on a static-struct-field miss (plain `{}` / accessor / host externref / unresolved shape), route a dynamic member target through the #2664 `__set_member_<name>` dispatcher (terminal `__extern_set_strict` sidecar) instead of early-returning and dropping the write. Exported for reuse. Private-field + reserved-name (`length`/`constructor`/`__proto__`/`prototype`/`name`) targets keep the prior drop (out of scope).
2. **`ctx.liveBodies` registration** of the three detached destructure-element buffers (`arrDestructInstrsADA`, `destructInstrsDA`, `odflInstrs`) for their open window, so the late-import funcIdx-shift walker keeps the dispatcher `call` repointed (the #2567/#1109 param-destructure precedent). Fixes the stale-low `call` → wrong-neighbour recursion / invalid Wasm.
3. **for-of / for-await** assignment-destructure (tuple, vec, object-struct branches) route member element/value targets through the shared helper (was identifier-only → dropped). for-await shares the lowering → fixed transitively.

### Verify-first (before → after)
`[x.y] = [4]` returned `0` (dropped) on both gc + standalone; gc array also hit `Maximum call stack`. After: returns `1`; standalone stays **host-free (0 env imports)**.

### Measured (test262, in-process runner)
- Deterministic dstr set (`language/expressions/assignment/dstr` + `language/statements/for-of/dstr`, 937 files): pass **332 → 340 (+8)**, fail 234 → 226, CE 1 → 1. **Exact pass-set diff: 0 regressions, +8 gains.**
- Full 2,171-file dstr set (incl. for-await): pass 1,407 → 1,416, no new compile-errors / throws.
- New `tests/issue-2869.test.ts` (12 cases, gc + standalone) green.

### Out of scope (noted in the issue, follow-on)
member-target WITH a default in the array assignment-expression path (`[x.y = d] = []`); receiver-base getter/setter exact-once semantics (`*-prop-ref-no-get`, `*-init-active`); assignment-to-const throws (`array-elem-put-const/let`).

Real gate: full `merge_group` + standalone floor (broad-impact: touches shared destructuring codegen).

### DIFF
--- a/plan/issues/2869-member-expr-assign-target-destructure-funcidx-repoint.md
+++ b/plan/issues/2869-member-expr-assign-target-destructure-funcidx-repoint.md
@@ -1,8 +1,9 @@
 ---
 id: 2869
 title: "Destructuring with a member-expression assignment target ([x.y]=vals, {k:x.y}=src) — funcIdx-repoint of detached body buffer (~53 fails)"
-status: in-progress
+status: done
 assignee: ttraenkler/senior-dev-promise
+completed: 2026-06-30
 created: 2026-06-30
 updated: 2026-06-30
 parent: 2669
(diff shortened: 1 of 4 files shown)


ttraenkler:
Superseded — #2869 (member-expression assignment-target destructuring) was implemented and **merged via #2352** (merge commit 47bb486; the `2869-member-expr-assign-target-destructure-funcidx-repoint.md` issue file this PR adds is already on main, and the issue is `status: done`). This PR is DIRTY against that merge and redundant. Closing as superseded; the member-target write path (emitDynamicMemberSet + liveBodies registration for the 4 detached buffers + for-of/for-await routing) is live on main. Reopen if there's a distinct case #2352 missed.

PR #2366, opened 2026-06-30
feat(#2864): heterogeneous boxed-any carrier for native generators (F1)

ttraenkler:
## #2864 F1 — heterogeneous (boxed-`any`) carrier for the native generator frame

Object / mixed-type yields now lower to the Wasm-native generator state machine **host-free** in standalone/WASI, instead of bailing to the eager-buffer host path (which leaks `__gen_*` / `__create_generator` and refuses under #680). The resumable frame already existed (`generators-native.ts`, the `br_table`-on-state machine) but its `value`/`sent`/`abrupt`/spill slots were f64-only (#1665) or a uniform native-string ref (#2171). F1 **generalises the frame** rather than building a new one.

### Verify-first (the issue's exact case)
`function* g(){ yield {a:1}; yield 2 }` consumed via `r1.value.a + r2.value`:
- **before:** CE-refused in standalone (leaks `__gen_create_buffer`/`__gen_next`/`__create_generator`/…)
- **after:** compiles with **zero host imports**, returns `3` — the yielded object survives the frame.

### What changed (root-cause, not symptom)
- **Carrier = `externref`** for object/mixed yields (was `null` → host bail). `any` already maps to externref and boxes host-free in standalone (native `__box_number` / `extern.convert_any`), so the consumer reading `.value` / a for-of var as `any` needs no extra coercion.
- **`sent`/`abrupt` typed per-carrier** (externref for any, f64 for numeric/string) — the state struct is minted per-generator, so numeric & string structs + all call sites stay **byte-identical** (zero regression risk).
- **Open dispatch gated on `hasAny`** — `let it = g()` boxes `it` to externref, so `it.next()` routes through the open anyref dispatch (the load-bearing path). Numeric/string-only modules keep the exact f64-singleton dispatch; a module with an any-carrier generator switches the block type to `eqref` (common supertype) and emits the `.next(v)` arg both as externref (any branches) and f64 (numeric branches, one unbox).
- **Open result reader** ref-tests every result struct; `.done` uniformly i32, `.value` picks f64 vs externref from the **static** type (numeric `.next().value` preserved).
- **Spread / destructuring** vec drains parametrised on the carrier element type (destructuring works; spread of an object generator is host-free but semantically incomplete — see deferred).

### Tests
- New `tests/issue-2864-standalone-generator-carrier.test.ts` (8 cases, **zero-host-import asserted**): verify-first, for-of, destructuring, mixed-module open dispatch, `.return()`, done-flag, numeric-unchanged.
- All ~250 existing native-generator tests pass (2079 / 1665 / 2169 / 2170 / 2171 / 2571 / 2581 / 2157 / 1017 / generators).

### Deferred (bail cleanly, **non-regressing** today; tracked on the issue)
- **Typed live-across-yield LOCAL spills** for the any carrier — an object local lowers to a concrete `ref $Object` whose exact type the up-front struct layout can't know without a body pre-pass (the architect-flagged "deeper rewrite"); such generators bail to host consistently across the candidate gate + registration. **Biggest remaining widener.**
- **Spread / `Array.from` precision** for the boxed-any carrier (array destructuring already works).
- **F2** (try/finally-across-yield + `return()`/`throw()`), **F3** (`yield*` over arbitrary iterables) — separate PRs.

This is the substrate primitive that unblocks #2865 (async generators).

### DIFF
--- a/plan/issues/2864-standalone-generator-carrier.md
+++ b/plan/issues/2864-standalone-generator-carrier.md
@@ -1,7 +1,8 @@
 ---
 id: 2864
 title: "Standalone: no Wasm-native generator carrier — sync generators leak __create_generator/__gen_* host imports"
-status: ready
+status: in-progress
+assignee: ttraenkler/sendev-genframe
 created: 2026-06-30
 updated: 2026-06-30
 priority: high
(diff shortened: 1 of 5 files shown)


ttraenkler:
WIP-hold (tech-lead): temporarily held to prevent auto-enqueue+park against the **clobbered standalone-floor** (highwater on main was overwritten back to leaky 26040 by the promote-baseline job — `#2879` write-side gap). A floor-fix PR is in flight; this hold will be removed the moment it lands, then this carrier enqueues against the honest ~12,883 floor. This is NOT a regression park.

PR #2367, opened 2026-06-30
fix(#2867): standalone native Promise carrier — funcIdx-shift + verdict drain

ttraenkler:
## What

Completes #2867 — the standalone Wasm-native Promise / microtask carrier. Builds on PR-A (the gate-broaden `isStandalonePromiseActive` → `ctx.wasi || ctx.standalone`, already on the branch), now **unblocked** by #2360/#2879's honest host-free metric: a leaky-pass → native-carrier migration can only move `host_free_pass` UP, so the gate-broaden no longer reads as a floor regression.

Fixes the two real defects sr-promisegate found that made PR-A alone regress in the harness:

### 1. Late-import funcIdx-shift in the native `.then` path (the +CE)
`Promise.reject(<object>).then(fn1, fn2)` where a callback pulls in a late host import (`new Test262Error(…)` → `__new_Test262Error`) failed Wasm validation with **"not enough arguments on the stack for call (need 1, got 0)"**.

`compilePromiseThenReceiverBuffer` / `compileStandalonePromiseThenCallback` swap `fctx.body` to a scratch buffer via a JS-local `savedBody`; while swapped, the **enclosing function body** is invisible to `shiftLateImportIndices` (it walks `fctx.body`/`ctx.currentFunc.body` — both the buffer — and `liveBodies` — the buffers — but not the orphaned saved array). A late import then shifts every defined func +1 but leaves an earlier `call __new_plain_object` (from `var a = {}`) stale-low, resolving to the 1-arg `__obj_hash`. This is the #1384/#2503 detached-body class, for the swapped-OUT body. **Fix:** register `savedBody` in `ctx.liveBodies` for the swap window. Standalone-only; the JS-host path was never affected.

### 2. Verdict-path drain gap (false pass / false fail)
The harness records `ret = test()` synchronously, reading `__fail` before any queued native `.then` / `$DONE` reaction runs. **Fix:**
- **`__drain_microtasks()` compiler intrinsic** — lowers to the native drain only when a Promise queue was registered; a **byte-neutral no-op** otherwise (every JS-host compile and every Promise-free module). Verified byte-identical on `gc`/`standalone`/`wasi`.
- **Harness wrapper injection** (`wrapTest`) — a bare `__drain_microtasks();` between the user body and the `__fail` read (sync + TLA wrappers). Bare (not `try/catch`) keeps the wrapper byte-identical off the Promise path — no empty-try churn across ~43k tests.

## Measured (in-process `runTest262File(…, "standalone")`)
- 120 files `Promise/{resolve,reject,prototype/then}`: gate-broaden-only pass 27 / host_free 14 (7 were **false passes**) → with fixes pass 20 / **host_free 7 honest**, CE 33→28 (the "not enough arguments" cluster gone). vs main (leaked → host_free 0): **+7 honest**.
- 96 files await + async-function: pass 54 / **host_free 50**, CE 9.

Remaining standalone CE/fail are pre-existing carrier-completeness gaps **out of scope** (Promise static value-read #2375, `new Promise` capability, `__get_builtin` #1472, `__to_primitive` #1806) — all host_free=0 on main, no regression.

## Acceptance
- Native Promise standalone tests pass host-free (imports empty) ✓
- `host_free_pass` increases vs main ✓
- gc / non-Promise output byte-identical (byte-neutral intrinsic + bare wrapper) ✓
- 0 regressions (Promise tests were host_free=0 on main; off-path bytes unchanged) ✓
- New regression test `tests/issue-2867.test.ts` (7 cases) ✓

## Deferred (coordinate, do not duplicate)
async-function `await` returns NaN (both wasi+standalone) — tied to await-on-`$Frame`, the convergence point with #2864's resumable-frame substrate. Not touched here.

## Note for reviewer / queue
The committed standalone high-water mark (`benchmarks/results/test262-standalone-highwater.json`) was re-based to the honest 12,883 by #2879 (5e3d9aca7) but a **later baseline-refresh (d4bc147d3) overwrote it back to the leaky 26040**; `tests/issue-2879-standalone-host-free-floor.test.ts` now fails **on main**. That is a pre-existing #2879 §2 promote-baseline issue (assignee sendev-hostfree), independent of this PR — flagged to the tech lead.


ttraenkler:
WIP-hold (tech-lead): temporarily held to prevent auto-enqueue+park against the **clobbered standalone-floor** (highwater on main was overwritten back to leaky 26040 by the promote-baseline job — `#2879` write-side gap). A floor-fix PR is in flight; this hold will be removed the moment it lands, then this carrier enqueues against the honest ~12,883 floor. This is NOT a regression park.

ttraenkler:
Closing this PR. Its central change — broadening `isStandalonePromiseActive` to `ctx.standalone` to activate the native Promise carrier on the pure-WasmGC standalone target — is a proven merged-baseline regression that I cannot make net-positive on current main.

The merge_group re-validation (the auto-park-bot flag on this PR) confirmed it: against main it drops the standalone pass count 26,407 → 25,003 (−1,404). My own decomposition of that residual: ~1,101/1,460 of the regressions are async-CPS native↔host `$Promise` mixing — the always-host async-function lowering does not gate on this flag, so activating the native carrier makes async functions hand host promises to native `.then` and vice-versa — leaving only ~359 plain-promise wins. Best-case scoping still nets ≈ −1,045.

The hard prerequisite is the native resumable-frame substrate (#2864), which is still `in-progress`: the async-function CPS path has to share the native `$Frame` before the standalone Promise carrier can activate without the mixing regression. Nothing that landed since (the Gap-4 combinators, #2419's funcIdx-shift) touches that async-CPS↔native-Promise boundary, so the −1,404 still holds on current main. This is a real regression park, not drift (superseding my earlier drift hypothesis).

The `#2867-graveyard` blocker is already documented in the issue file on main (the Gap-2 note). Re-derive once #2864's native async frame lands. The inert infrastructure here — the `calls.ts` late-import funcIdx-shift fix for the standalone `.then` path and the `__drain_microtasks` intrinsic — is preserved in this branch's history for cherry-pick if that path is revived.

PR #2380, opened 2026-06-30
fix(#2865 AG0): net-neutral reconcile — host-free await scoped to WASI, standalone defers to PATH B (#2895)

ttraenkler:
## Summary

PATH A first slice of the standalone async-`$Frame` substrate (coordinator decision). Makes `await` host-free and correct under `--target standalone`/WASI for the **synchronously-settled** subset, and stops `--target standalone` leaking `env::Promise_*` / `__make_callback`.

### Root cause (verify-first on current main)

The task's starting assumptions were stale: the async-CPS state machine (`async-cps.ts`) is gated **off** for BOTH standalone and WASI (`function-body.ts`), so async fns compile **synchronously** with their unwrapped return type, and `await` was a pure **identity passthrough** (`expressions.ts`). So `await <a fulfilled $Promise>` returned the promise OBJECT (externref) where the consumer expected the resolved value → coerced to f64 = **NaN**. Separately, `isStandalonePromiseActive` keyed on `ctx.wasi` only, so `--target standalone` leaked the host Promise imports.

### Changes

- **`isStandalonePromiseActive` now returns `ctx.wasi || ctx.standalone`** (`async-scheduler.ts`) — both host-free targets use the Wasm-native `$Promise` carrier for `Promise.resolve/reject` / `.then` / await.
- **`await` unwraps one level of the native `$Promise`** at runtime (`emitStandaloneAwaitUnwrap` in `expressions.ts`): a non-null `ref.test (ref $Promise)` discriminates — a `$Promise` operand yields its `value` field, anything else passes through. The operand is compiled to its **natural** type (not forced to `expectedType`, which would coerce a `$Promise` externref to NaN before reading it); non-externref operands pass through.

### Verified (`--target standalone`)

`return await Promise.resolve(5)`, `const x = await Promise.resolve(40); x+2`, `await <local sync-fulfilled promise>`, `await <literal>`, `await (n+1)`, async **methods**, and two sequential awaits — all **zero host imports** and correct values (was NaN). WASI async `await` fixed as a bonus. gc/host path unchanged; full async/promise suites show no new failures (the 3 pre-existing failing files fail identically on clean base). tsc + prettier + `check:ir-fallbacks` green.

### Scope / follow-up

Genuinely-pending awaits (a promise that only settles on a later microtask/timer) need true frame suspension — filed as **#2895** (AG1 / PATH B; arch spec on `origin/async-gen-2865-spec`). They were already wrong pre-AG0, so deferring is not a regression. This also advances #2867 (native Promise carrier) and the parked #2367. Full `merge_group` standalone report is the net-positive gate.


ttraenkler:
## Net-neutral reconcile (supersedes the parked -2469/-601 state)

Ground-truthed the parked regression locally on the **#2384 frame-core base** (merged in). Two findings:

1. **The -2469 wasm type-mismatch cascade is gone** — it was the pre-fix state; the WASI-only `.then` scoping + a fresh #2384 merge cleared it (0 type-mismatch compile_errors across 202 async files).
2. **But the broad `isStandalonePromiseActive = wasi || standalone` gate is a real RUNTIME regression** on standalone: async sample 134→103 (−31); await+async-function area 71→42 (−29); **zero** offsetting await win. The `flags:[async]` test262 harness uses synchronous settlement (no microtask drain), so a standalone async fn returning a native `$Promise` is observed as an undrained struct. The host-free standalone await gain is **coupled to a drive layer** = PATH B (#2895), not bankable by a bounded gate flip.

**This push reverts the gate to `ctx.wasi` only** → standalone back to baseline (**134 pass, NET 0, zero regression**, verified vs upstream/main); WASI keeps the genuine native-`$Promise` await NaN-fix (7/7 host-free WASI tests green). The `isStandaloneThenChainNativeActive` gate-split scaffolding stays for PATH B to re-widen. HEAD `28eec2ca7`.

cc @-shepherd: net-neutral verified locally; safe to clear the park-hold + enqueue once after PR-level CI is green.

PR #2439, opened 2026-07-01
fix(#2849): keep dynamic-key-written objects open (numeric property reads back 0)

ttraenkler:
## Summary

Fixes #2849 — a dynamic-object numeric property read back **0** in host mode when
the object is BOTH populated by a computed/dynamic-key write
(`for (k in d) o[k] = opts[k]`) AND widened to a closed WasmGC struct by a
sibling static dot-write (`o.ecmaVersion = <number>`). This is the compiler
mis-read behind compiled-acorn's `getOptions` ecmaVersion normalisation
(`2022 → 13` reading back `0`), the edge.js half of #1712.

## Root cause (WAT-bisected)

Not the numeric-vs-string type-inference originally hypothesised — a
**representation-coherence split**:

1. `var o = {}` is given a concrete WasmGC **struct** type by
   `collectEmptyObjectWidening` because the static dot-write
   `o.ecmaVersion = 1e8` (inside the never-taken `=== "latest"` / `== null`
   guard) contributes an `f64 ecmaVersion` field. Dot-reads then lower to an
   unguarded `struct.get`.
2. The for-in write `o[k] = opts[k]` has a **non-literal computed key** (`k` is
   the loop var), so it can't resolve to a struct field and lowers to the dynamic
   `__extern_set` **sidecar** — not the struct slot.
3. Read/write diverge → `struct.get` returns the field default `0`, never the
   sidecar's `2022`. (Bisected: `var k="x"; o[k]=v` const-propagates the key →
   `struct.set` → coherent → no bug; only the non-resolvable key breaks.)

Host-only because the demotion guard that keeps such objects open
(`markObjectHashConsumers`, #2584) is **standalone-gated**, and the host
"live-mirror Proxy writeback" does not cover a dynamic-key write.

## Fix

`src/codegen/declarations.ts` — new `hasNonLiteralComputedWrite` detector, wired
into `collectEmptyObjectWidening` **mode-agnostically**: a non-literal
computed-key write `o[<expr>] = v` suppresses empty-object struct widening, so
the receiver stays a `$Object` and both the dynamic write and the dot reads route
through the same guarded/sidecar representation. String/number-**literal**
computed keys are intentionally NOT matched (they lower to `struct.set`, keeping
the struct fast path). Standalone was already covered by the broader
`markObjectHashConsumers` poison.

## Verification

- Minimal repro `run(2022) === 13` with the `=== "latest"` / `== null` guards
  present; all variant forms correct; literal-key writes still fast-path;
  pure-numeric path unchanged.
- **0 regressions** in a base-vs-mine differential over 1,399
  object / for-in / property-accessor test262 files. A broader in-process sweep
  flagged 40 Array "regressions" — all confirmed **false positives** (pass in
  per-file isolation; the two-runner-in-one-process harness produces spurious
  `compile_error`s under memory pressure).
- Test: `tests/issue-2849-dynamic-key-object-numeric-read.test.ts`.

## Out of scope (separate pre-existing gaps, verified identical on `main`)

- **Standalone** (`--target standalone`) still reads the property back as an
  externref/object (`{}`, not `13`) — an independent gap in the standalone
  dynamic-`$Object` numeric substrate.
- The acorn `attributes` differential (edge.js module source) stays gated on
  #2325; the underlying compiler mis-read is fixed here.

### DIFF
--- a/plan/issues/2849-ecmaversion-normalize-dynamic-prop-type-inference.md
+++ b/plan/issues/2849-ecmaversion-normalize-dynamic-prop-type-inference.md
@@ -1,7 +1,9 @@
 ---
 id: 2849
 title: "dynamic-object numeric property reads back 0 when the same property is also compared via === string / == null (acorn ecmaVersion 2022 not normalised → spurious import attributes)"
-status: ready
+status: done
+assignee: ttraenkler/dev-2878
+completed: 2026-07-02
 sprint: current
 priority: medium
 horizon: l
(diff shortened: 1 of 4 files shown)


ttraenkler:
Closing as a duplicate: #2849 was already fixed on main by #2432 (`fix(#2849): extend $Object-hash poison to host so static writes don't shadow the sidecar`), which merged first from a parallel session on the shared fork. I verified the merged main resolves the full repro (run(2022)=13) and all variant cases (===latest/==null reassign stay 2022; numeric-only normalises to 13). My branch reached the same root cause (empty-object struct widening colliding with a dynamic-key __extern_set sidecar write) via a narrower detector, but it adds no correctness beyond #2432, so I'm standing it down rather than duplicating. Proceeding to the #2934 follow-on tail.

PR #2442, opened 2026-07-02
feat(#2923): broaden constant-string eval compile-away (func decls, for-of/for-in over literals)

ttraenkler:
## Summary

Slice **A** of the runtime-eval roadmap (`docs/architecture/runtime-eval-interpreter.md` §6-A). `tryStaticEvalInline` (#1163) splices a constant `eval("<literal>")` body inline at compile time, but `allNodesInlineSupported` **bailed** the moment the body held a function/class/arrow/for-of/for-in — falling to the dynamic `__extern_eval` host import, which **traps at instantiation in standalone**. This lifts the safely-liftable kinds. Pure AOT, standalone-safe, **no host imports**.

## Lifted

- **Function declarations** (incl. params, recursion, mutual reference). The foreign eval `SourceFile` has no checker bindings, so `getSignatureFromDeclaration` **throws** (`symbol.escapedName` on `undefined`). Fixed at **both** sites in `src/codegen/statements/nested-declarations.ts` (`compileNestedFunctionDeclaration` and the multi-fn pre-reserve pass in `hoistFunctionDeclarations`): params degrade to externref (`getTypeAtLocation → any`) and the return type defaults to externref — both paths **identically**, so the reserved funcType matches the compiled body.
- **`for-of` over an array/string literal**, **`for-in` over an object/array literal** (iteration needs no checker-resolved iterator type). A non-literal iterable keeps bailing.

## Kept bailing (documented)

Function/arrow **expressions**, **classes**, yield/await/import/export — their codegen dereferences a checker signature/heritage the foreign SourceFile lacks and would **throw an internal error** (worse than a clean fall-through). They fall to the dynamic path (host eval; the Tier-2 interpreter #2928 handles them standalone later). This satisfies acceptance criterion 2 (class) via the "provably bails with a documented reason" branch.

## Sizing artifact (criterion 5)

`scripts/eval-const-classifier.mjs` (`npx tsx`, reuses `resolveConstantString`). Over `test262/test`: **1460 files** with an eval site, **2611 call-sites**, of which **2394 (91.7%) are constant-string (Tier-0 liftable)**, 210 dynamic (need the interpreter #2928), 7 no-arg.

## Also

- Files the #2861 namespace-reads follow-up **#2933** (allocated via `--allocate`, offered to the lead) — rides this PR.
- Sets `#2923` `status: done`.

## Test plan

- `tests/issue-2923-eval-const-broaden.test.ts` — 13 cases: func-decl lifts incl. recursion / mutual-reference / zero-param; for-of/for-in over literals; expr-fn / class / non-literal-iterable bail cleanly; host mode unchanged.
- No regression in the existing `#1163` inliner / `eval-tiering` tests (the pre-existing `#1163` "no-arg" + `#1712` failures also fail on origin/main — unrelated).
- `tsc --noEmit` clean; lifted standalone modules emit **zero** host imports.


ttraenkler:
## Park diagnosis + verification (removing the park-hold)

I took the #2442 [CI-FIX]. Diagnosing the cited merge_group failure first, per the auto-park protocol.

**The park** (merge_group run 28557463860, head `a2b58778`): `check for test262 regressions` failed — net **-73** (50 improvements − 123 regressions), with two >50 gate-fail buckets: `annexB/language/eval-code/direct` (51) and `.../indirect` (51). Root cause: the broadened const-string eval inliner hoisted eval-body function declarations unconditionally, which is wrong for (A) strict eval bodies and (B) AnnexB §B.3.3 function declarations nested in a script-scope block/if/switch/for/label.

**Already fixed on the branch.** Commit `6250fd317` ("bail eval inlining for strict-body + AnnexB block-nested function decls") + guard tests `85dbec3b8` landed *after* the park and narrow `allNodesInlineSupported` to exactly those two cases. The current head `5bed43e` was never re-validated in the merge_group (it stayed held), so it stranded.

**I verified the fix is clean** (I changed no code — this is a verification-only re-admit):

- Ran the parked buckets `annexB/language/eval-code/{direct,indirect}` (469 tests) on head `5bed43e` **and** on `upstream/main`: **identical** — 276 pass / 193 fail, fail sets bit-identical.
- Ran `language/eval-code/**` + `language/statements/function/**` (eval-referencing, 487 tests), the remaining strict/language buckets: also **identical** head-vs-main — 337 pass / 150 fail, identical sets.
- Feature guard tests `tests/issue-2923-eval-const-broaden.test.ts`: **17/17 green** — the intended const-string eval inlining still works (top-level sloppy func-decl bodies still lift); the feature is not neutered.

So all 123 regression buckets now match `main` exactly and the wins are intact. Removing the bot `hold` once so `auto-enqueue` re-sweeps; the merge_group re-validates authoritatively.

PR #2444, opened 2026-07-02
fix(#2936): settle raw-import drift before opening a deferred late-import batch

ttraenkler:
## Problem

Blocking invalid-module class for #2933 (no-yield native generators, ~250-350 test262 flips): `__str_flatten call[1] expected type externref, found i32` at test262-harness scale. Also **latent on today's main** — reproducible without any relax via a native-candidate generator that captures an outer local (see tests/issue-2930.test.ts).

## Root cause (traced, corrects the original hypothesis)

Not a resume-emit-time import (instrumented tracing shows no import fires during the lazy resume emit). It is a **mix of the two index-shift regimes** in `collectAllSourceImports`:

1. Native-string helpers emitted at 0 imports (`base=0`); `__str_flatten` bakes `call 0` → `__str_copy_tree`.
2. `finalizeUnifiedCollector` adds `env.__make_callback` via **raw addImport** (+1 drift, repair deferred to `reconcileNativeStrFinalizeShift`).
3. `addGeneratorImports` opens a **deferred batch** (`importsBefore=1`) and flushes: the flush shifts only refs ≥ 1 (missing the base-regime `call 0`/`copyTree=0`) and its #1903 re-base sets `nativeStrHelperImportBase=16`, **permanently cancelling** the pending raw repair.
4. `call 0` now targets import #0 `__make_callback(i32, externref)` → validation failure.

Whether it fired depended on an *incidental* interleaved reconcile: every host-path generator shape normally sets `unionFound`, whose finalize pass runs the reconcile in between. Candidate-but-outer-capturing generators (main) and #2933's no-yield candidates skip that trigger while still pulling the host gen-suite — opening the window.

## Fix

One hunk in `src/codegen/expressions/late-imports.ts`: settle the finalize-regime drift (`reconcileNativeStrFinalizeShift`) at the deferred-batch **record point** in `ensureLateImport`. The mixed state is unrecoverable after the flush (stale defined refs and correct import refs collide by value in `[base, importsBefore)`), so the record point is the single sound settle point — this kills the whole class, not just this instance. No-op unless a raw import landed since the last settle; hard no-op on gc/host (base −1).

## Validation

- Both #2933 repros (`obj-ptrn-empty.js`, `scope-paramsbody-var-close.js`), runner standalone path, bails relaxed: invalid-module → **pass**.
- 542-file no-yield corpus (relax on): pre-fix **17 invalid** → post-fix **12, none the funcIdx-shift class, 0 new**. Residuals attributed: 10 pre-existing on pristine main (unrelated standalone bugs), 2 relax-lane-only pattern-default shape bugs the #2933 relax PR must bail (documented in the issue file).
- **Byte-inert**: 240 sha256 comparisons vs origin/main (60 gen + 60 non-gen files × {gc, standalone}) — all identical.
- New regression test `tests/issue-2930.test.ts` (main-lane repro, no relax needed); generator equivalence suite (36) green; issue-1461 Symbol.isConcatSpreadable failures verified pre-existing on main (A/B).

Issue file: `plan/issues/2930-resume-lazy-emit-funcidx-shift.md` (root cause, corpus attribution, #2933 handoff incl. the two new relax blockers).


ttraenkler:
Re-id note: this issue was allocated as #2930, but a parallel session landed a different `plan/issues/2930-*.md` on main first (same allocator race as the 2920→2933 re-id). Everything in this PR is re-id'd to **#2936** (`plan/issues/2936-resume-lazy-emit-funcidx-shift.md`, `tests/issue-2936.test.ts`). The branch name keeps the historical 2930 slug — no gate on branch names. PR body text referring to #2930-as-this-issue should be read as #2936.

PR #2462, opened 2026-07-02
revert(#2937): back out #2432's host extension of the $Object-hash poison (compiled-acorn null-deref on all host-mode inputs)

ttraenkler:
Revert of #2432's host extension — fixes #2937 (compiled-acorn null-deref on all host-mode inputs); #2849 host-mode bug reopens, blocked on the escapes substrate (#2944).

## What changed
PR #2432 dropped the `ctx.standalone` gate on `collectEmptyObjectWidening`'s `markObjectHashConsumers` loop (`src/codegen/declarations.ts`), extending the #2584/#2849 `objectHashConsumerVars` poison to host mode. This restores the standalone-only gate.

## Why revert (the two constraints genuinely conflict)
The poison is honored ONLY at the widening DECISION — no read/write codegen site consults `objectHashConsumerVars`. So it keeps a for-in-copied `{}` var's VALUE a `$Object`, but reads/writes still resolve the receiver via `resolveStructName(TS-type)`, which mis-binds it to a colliding `__anon` struct; and worse, the poisoned value ESCAPES the identifier (`getOptions` returns `options` → struct-typed `this.options` field → `this.options.ecmaVersion`) into struct-typed slots a receiver-level bail cannot reach. Result: compiled-acorn null-dereferenced on EVERY host-mode input, including `parse("")` (parser setup) — a total host-parse block, strictly worse than the narrow `getOptions` shape bug the host extension fixed (which existed quietly for months).

Measured (dev-f2, instrumented): host poison + a scoped receiver-identifier bail → 22/23 acorn corpus inputs still throw; pure revert → all parse. So no scoped resolver change satisfies both #2849-fixed AND acorn-parses.

## Evidence
- Acorn dogfood corpus: **22/22 threw → 21 equal±quirks / 0 real / 2 threw** (above the 2026-06-30 baseline of 13 equal±quirks). `parse("")`/`parse("1")`/all corpus inputs return a `Program` again. The 2 residual throws are the pre-existing #2850/#2853 gaps (unmasked, not this regression).
- Standalone codegen byte-identical (its poison is untouched).
- `tests/issue-2849.test.ts`: 11/11 green — the 4 host for-in-copy-WITH-GUARD arms are marked `it.fails` (honest known-fail pinned to reopened #2849 / substrate #2944; they flip red, forcing restoration, once #2944 lands). Host no-guard + static-only + all standalone arms still pass.

## Issue bookkeeping
- **#2849** reopened: `done → blocked`, `blocked_on: 2944`.
- **#2937** `done` (fixed-by-revert), cross-refs #2944.
- **#2944** filed [SENIOR-DEV]: "poisoned $Object values escape into struct-typed slots — externref-typed escape discipline for hash-consumer vars", carrying the full escape analysis + #2849 design as the spec seed. This is the proper home for BOTH bugs.

---

## Accepted-cost accounting (merge_group re-validation, run 28565120195)

This revert is **net −137 host test262** and was **greenlit by the tech lead** with the rationale below. Recording the full accounting for the queue and the reporting record.

**1. Full −137 accounting**
- **Host lane:** 151 regressions, **net −137 pass** (34773 → 34636), 14 improvements. Categories: `assertion_fail` 126, `other` 14, `runtime_error` 8, `type_error` 2, `wasm_compile` 1. The lone compile regression is `test/built-ins/Object/defineProperty/15.2.3.6-4-394.js`. **All 151 are the `dynamic_object_property` (#2849) class** — zero unrelated collateral.
- **Standalone lane:** **0 regressions, net +0** (27190 → 27190) — the revert is **byte-inert on standalone** (its poison was never touched), as claimed.

**2. Acorn-vs-test262 rationale (tech-lead decision, for the record)**
A silent wrong-value / null-deref on **every input** of compiled-acorn (a flagship real program, the #2927 critical path) is the worst failure class — strictly worse than visible, catalogued test262 fails. The −137 merely **restores the pre-#2432 status quo** of two days ago (red for months; nothing depends on 1-day-old green). #2432's host-poison extension was a net-positive test262 change that _also_ silently broke compiled-acorn via the escape case; keeping it would strand the entire acorn critical path.

**3. Metric note (carry in gap reporting)**
The js-host baseline drops ~137, which **mechanically shrinks the standalone-vs-host gap for a non-improvement reason**. Gap accounting must flag this so the narrowed gap is NOT reported as progress.

**4. Proper fix**
Both #2849 (host getOptions shape) and #2937 (acorn null-deref) are properly fixed by **#2944** — externref-typed escape discipline for poisoned `$Object` values — which keeps the 151 host wins AND fixes acorn. Filed, spec-seeded (full escape analysis + #2849 design), senior-queued `[SENIOR-DEV][XL]`. This revert is the interim measure until #2944 lands; #2849 is reopened `blocked_on: #2944`, its 4 host arms `it.fails`-pinned so they flip red (forcing restoration) when #2944 fixes them.

### DIFF
--- a/plan/issues/2849-ecmaversion-normalize-dynamic-prop-type-inference.md
+++ b/plan/issues/2849-ecmaversion-normalize-dynamic-prop-type-inference.md
@@ -1,8 +1,7 @@
 ---
 id: 2849
 title: "dynamic-object numeric property reads back 0 when the same property is also compared via === string / == null (acorn ecmaVersion 2022 not normalised → spurious import attributes)"
-status: done
-completed: 2026-07-02
+status: blocked
 assignee: ttraenkler/fable-dev
 sprint: current
 priority: medium
@@ -14,11 +13,25 @@ task_type: bugfix
 area: codegen
 language_feature: dynamic-object-property-type-inference
 goal: acorn-dogfood
-related: [2841, 2836, 1712]
+related: [2841, 2836, 1712, 2937, 2944]
 depends_on: []
+blocked_on: 2944
 blocks: [1712]
 umbrella: 1712
 ---
+
+> **REOPENED 2026-07-02 (was `done`).** The host-mode fix (PR #2432, extend the
+> `objectHashConsumerVars` poison to host) was **REVERTED in #2937**: keeping
+> acorn's poisoned `{}` vars on `$Object` in host mode broke compiled-acorn with
+> a uniform null-deref on EVERY input, because the poisoned value ESCAPES the
+> identifier into struct-typed slots (`getOptions` return, `this.options` field)
+> that the poison never re-types — a total host-mode parse break, strictly worse
+> than this narrow `getOptions` quirk. So this host bug is live again. The
+> **real cure is the escape-discipline substrate slice #2944** (externref-typed
+> escapes for poisoned `$Object` values); this issue is `blocked_on: 2944`.
+> Standalone is UNAFFECTED (its poison was never reverted). The reduced host
+> arms in `tests/issue-2849.test.ts` are marked `it.fails` pinned here — they
+> flip red (forcing restoration) when #2944 lands.
 
 # #2849 — dynamic-object property mis-typed when read in heterogeneous (string-`===` / `==null` AND numeric) contexts
 
--- a/plan/issues/2937-acorn-host-object-hash-poison-null-deref-regression.md
+++ b/plan/issues/2937-acorn-host-object-hash-poison-null-deref-regression.md
@@ -1,7 +1,8 @@
 ---
 id: 2937
 title: "Regression: host-mode $Object-hash poison (#2849) makes compiled-acorn parse null-deref on every input"
-status: ready
+status: done
+completed: 2026-07-02
 created: 2026-07-02
 updated: 2026-07-02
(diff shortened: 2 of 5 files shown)


ttraenkler:
## Re-held intentionally (pending admin-merge, NOT a new bot park)

The −137 revert is tech-lead + human-lead APPROVED (see the accepted-cost accounting above). However, `check for test262 regressions` is a **REQUIRED** status check (docs/ci-policy.md §1), and it re-fails deterministically at net −137 in the `merge_group`. So a normal queue re-admit just **re-parks in a loop** (and each re-add rebuilds the merge group, risking cancellation of another PR's in-flight run).

Therefore this PR can only land via a **tech-lead/human admin-merge that bypasses the required `check for test262 regressions` gate** (the approval is on record; the −137 is a deliberate, accepted trade to unblock compiled-acorn — #2944 is the real fix). I removed the bot hold to re-admit, then re-added it to keep the queue from pointlessly churning until the admin-merge happens. The hold does NOT block an admin-merge; it only blocks the auto-enqueue loop.

**Action needed:** a maintainer admin-merges this PR (or temporarily de-lists the regression gate for this single merge). — dev-f2

PR #1897, opened 2026-06-21
fix(statusline): live sprint + standalone pass-rate sources; harden /workspace sync

ttraenkler:
Statusline froze on stale committed artifacts; this makes its dynamic segments compute from live/CI-refreshed sources.

**Sprint bar** read `website/dashboard/data/sprints.json`, rebuilt only by `build-data.js` on deploys (NOT on PR merge) — it stuck on `s64` after s65 opened. `statusline-sprint.mjs` now computes LIVE from the flat `plan/issues` tree + the `sprints/<N>.md` `status: active` marker; `statusline-command.sh` calls it as the primary source (sprints.json demoted to last-resort fallback). ~250ms/render (2,225 files), well within the 30s refresh.

**Standalone pass-rate** preferred `test262-standalone-report.json` — an UNtracked local dev-run artifact — over the CI-refreshed high-water mark, shadowing the fresh 52.6% with a 5-day-old 47%. Now prefers the committed high-water mark, falling back to the local report only when genuinely newer.

**Sync** (`sync-workspace-main.sh`) stopped pre-refusing on incidental dirt outside `.claude/memory`: it now attempts the ff (`git --ff-only` is safe on a dirty tree — preserves unrelated uncommitted changes, refuses only on real conflict) and warns precisely, so /workspace stays synced.

Verified live: statusline renders `s65 3/43  73.4% t262  52.6% sa` (was frozen `s64 30/80 ... 47.0% sa`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2463, opened 2026-07-02
fix(#2940): runner vacuity scorer — dead harness callbacks scored fail not pass (integrity correction −1433 host_free_pass)

ttraenkler:
## Summary

Enforces the durable vacuity rule **in the runner**: a test whose harness-wrapper callback never executes (the dead-callback / dispatch-drop class) is scored `fail` + `vacuous:true`, NOT `pass` — so `host_free_pass` and the standalone floor structurally exclude it. This is a **deliberate, stakeholder-approved metric-integrity correction** (the project lead's standing directive), not a conformance change.

This is **PR1 of 2**. PR2 (#2939, dispatch fix) makes the dead callbacks EXECUTE and lands after this; with the scorer already in place, PR2's vacuous→executing conversions can only move `host_free_pass` UP.

## Mechanism (runner-only, no codegen change)
- Preamble adds a `__harness_cb_expected` sentinel that the `testWith*Constructors` wrappers bump per callback invocation.
- `test()` returns a distinct `-262` when a would-be-pass had an invoked harness wrapper but **zero counted asserts** (the callback was dead — its assertions live inside it). Scored `status: fail` + `vacuous: true` marker + a `build-test262-report` counter for visibility.
- Wired through all three exec paths: `runTest262File`, the compiler-pool worker (`scripts/test262-worker.mjs`), and the fixture path.
- Chose `fail`+marker over a new status enum (206 status-consumers = too much blast radius for a one-way-door change); satisfies 'not pass' + distinct visibility. Under-detection (asserts outside the callback keep `__assert_count > 1`) is safe; over-detection is near-impossible for the harness class.
- Also ships the BigInt-TA runner shim (`testWithBigIntTypedArrayConstructors` + passthrough `makeCtorArg`, fixed `needsTestBigIntTypedArray` regex) so the BigInt wrappers are invocable and thus measurable.

## Integrity correction (measured, all 1,612 harness-invoking files, standalone)
**1,433** tests reclassify pass → vacuous-fail:
- **1,421** in `built-ins/TypedArray{,Constructors}` (696 BigInt-variant + 725 non-BigInt)
- **12** in outside-dir harness-invokers (`built-ins/Array/*/callbackfn-resize-arraybuffer.js`)

`host_free_pass` **17,802 → 16,369**. CE rate unchanged (~3.5%, pre-existing BigInt unsupported-feature CEs — the shim adds none). **Zero** non-vacuous pass→fail collateral (the `__assert_count === 1` guard never flags a test that ran any real assertion).

## Highwater re-seed (in this PR, approved)
`benchmarks/results/test262-standalone-highwater.json` `host_free_pass` 17,802 → **16,369** (the floor asserts-then-raises and never auto-lowers, so a deliberate downward correction is committed). Post-merge `promote-baseline --update` raises it to the true CI number.

## Expected merge_group behavior (shepherd heads-up)
The **#1897 standalone regression guard** (external baseline JSONL) will **bot-park** the merge_group as the expected **−1,433** signature — this is the intended integrity correction, not a regression. Verify the delta is EXACTLY this documented set (zero collateral: no new CEs, no non-vacuous pass→fail flips), then remove the hold / force-promote once. If the delta does NOT match, stop and escalate to the tech lead — no partial re-admission.

## Tests
`tests/issue-2940.test.ts` — dead harness callback → fail+vacuous; no-harness test → never flagged. 2/2 pass.

Issue file rides with `status: done`.

---

## Re-ground after the 12-PR merge wave (2026-07-02, session-window reset)

Main moved substantially while this PR was parked (mark rose 17,802 → 18,790). The branch has been re-merged with main (highwater conflict resolved) and the re-seed re-derived: **17,357 = 18,790 − 1,433**.

**STALE-SIGNATURE CAVEAT:** the −1,433 delta was measured at base 854ad5729 and could NOT be re-measured on the merged tree (measurement timed out at budget wind-down). #2470 (host poison re-land) and #2480 (tag-5 fix) plausibly changed which tests are vacuous, so the true merged-tree delta may differ slightly from 1,433.

**Updated shepherd verification rule (per tech lead):** verify the merge_group park signature is **ALL-vacuous** — every pass→fail flip carries the `vacuous: true` marker / the `"vacuous: harness-wrapper callback never executed"` error, with **zero non-vacuous collateral** (no new CEs, no genuine pass→fail) — rather than an exact −1,433 count. Then the tech lead admin-merges (no queue re-admit). Post-merge `promote-baseline --update` raises the mark to the authoritative CI number.


ttraenkler:
Taking this PR over per the stale-author policy: it has been parked (auto-park after two collateral merge_group failures) and DIRTY against main since ~05:05Z with no author activity, and the authoring session has wound down. The conflict is on benchmarks/results/test262-standalone-highwater.json — this PR's intentional scorer correction vs main's subsequent baseline re-ratchets. I will re-derive the correction against the current floor on main (not blindly keep either side), merge main, and re-admit once green. — shepherd-f

ttraenkler:
Retracting my takeover note above — it crossed with your return: I now see the branch was updated at 09:31–09:41Z (re-derived highwater 18790−1433=17357 + main merge) just before I commented. The branch is yours; I am back to monitor-only and will not touch the hold. — shepherd-f

ttraenkler:
## CI-FIX (#2463) — DIRTY resolved + park-signature re-verified

I rebased onto current `upstream/main` (mark rose 18,790 → 18,812) and re-derived the highwater re-seed to **17,379 = 18,812 − 1,433** (same caveat: the −1,433 delta is from base 854ad5729, not re-measured on the merged tree). PR is now `MERGEABLE`.

I investigated the "12 non-vacuous regressions" by direct standalone re-measurement (`runTest262File(..., "standalone")` over the full **705-file** BigInt-TA baseline-pass corpus). **Verdict: this PR introduces ZERO un-excused non-vacuous standalone regressions.**

All 705 BigInt-TA baseline passes were LEAKY (705/705, 0 host-free), so any flip to a host-free non-pass is auto-excused by the #1897 guard's `--exclude-leaky-baseline-regressions`. Classification of the 705:

| class | count | disposition |
| --- | --- | --- |
| `VACUOUS` (marked `vacuous:true`) | 698 | intended park signature |
| leaky-pass → **host-free** non-pass | 5 | AUTO-EXCUSED (cross-realm) |
| still-pass (genuine) | 2 | fine |
| **leaky-fail, un-excused, non-vacuous** | **0** | ← the only bad class; none |

The 5 excused flips are the cross-realm OtherTA detached-buffer tests. Root cause: the `$262.createRealm()` stub's realm carries no constructors, so once the BigInt shim makes the callback dispatch (PR #2441 arity fix), `other[TA.name]` is `undefined` and the callback throws before any assertion. They genuinely cannot pass standalone (dynamic realm-property access + `Reflect.construct` are unsupported — I verified that exposing the constructors on the realm stub does NOT help). They flip leaky-vacuous-pass → honest host-free fail, which is the correct honest classification and is auto-excused.

Reconciliation of the task buckets: the "5 proto-from-ctor-realm" were NOT standalone-baseline-pass (no flip); the "1 async-gen invalid-wasm" (`ctors-bigint/object-arg/as-generator-iterable-returns.js`, `__closure_3`) was ALREADY `compile_error` on the standalone baseline (CE→CE, no flip). "No dependency provided for extern class OtherTA" is the HOST-lane wording for the same root cause; on host the ≤12 flips sit far below the #1668 catastrophic threshold (200).

### Verification recipe (admin-merge)
Run `diff-test262.ts <standalone-baseline> <merged-standalone.jsonl> --exclude-leaky-baseline-regressions`. Every remaining counted regression must carry `vacuous:true` / `"vacuous: harness-wrapper callback never executed"`. The excused set is the handful of leaky-baseline→host-free cross-realm flips. Zero un-excused non-vacuous collateral. Post-merge `promote-baseline --update` raises the mark to the true CI number.

Leaving the `hold` label on — this is an intentional-negative; enqueue-once → verify → admin-merge is yours. I am not self-enqueuing.

PR #2464, opened 2026-07-02
feat(#2924): new Function(<const>) compile-away — slice-1 (JS-host lane)

ttraenkler:
Slice-1 of the runtime-eval roadmap (#1584 umbrella, §6-B). Replaces the `new Function(...)` no-op stub with a real compile-away on the **JS-host lane** when every argument is a compile-time-constant string.

## What it does
`new Function(p1,…,pn, body)` with all-const args is — per §20.2.1.1 — semantically `function (p1,…,pn){ body }` with **global scope** (never the caller's lexical scope). `tryCompileConstantFunctionCtor` (`src/codegen/expressions/new-super.ts`) synthesizes that function as a foreign `SourceFile`, compiles it with an **empty enclosing `localMap`** (so it captures NOTHING from the caller — the no-lexical-capture requirement), and escapes it as a callable via `emitCachedFuncClosureAccess`/`emitFuncRefAsClosure`. Reuses #2442's foreign-binding-less `compileNestedFunctionDeclaration` tolerance. Rollback-guarded: snapshots `mod.functions.length` + `funcMap` so a mid-body compile throw on a binding-less node can never leave a half-registered empty-body function.

## Working (tests/issue-2924.test.ts, 6/6)
- `new Function("a","b","return a+b")(1,2) === 3`, single-param, no-param, reuse-across-statements — **JS-host lane**.
- **Non-constant arg bails gracefully to the legacy stub** (negative test) — compiles, returns the null placeholder, never miscompiles.
- **Standalone/WASI gated to the stub** — compiles host-free, no miscompile, no trap (asserted).

## Why host-lane only (ship decision, tech-lead call)
The synthesized function has all-externref params, and externref-param closures hit a **pre-existing standalone call-marshalling bug**: two calls coexisting in one expression (`f(1)+f(2)`) or a ≥3-arg call silently return a WRONG value on the standalone lane. **Control-verified NOT feature-specific**: a plain `function(a:any){ return a+10; }` reused as `f(1)+f(2)` returns the same wrong value in standalone (a typed `a:number` closure is correct). Since these can't be detected at the `new Function` site to bail, and we don't ship silent wrong values in ANY lane, the compile-away is gated to JS-host. Standalone enablement is **#2945** (filed, carries the full collision analysis; blocks the standalone half of #2924).

## Non-goals (documented in the issue, follow-up slices)
Standalone enablement (#2945), plain-call value form `Function(...)()`, `undefined`-return representation, no-capture `typeof x` string-return.

Issue file rides the PR with the slice-1 acceptance checked + gate rationale + non-goals.

### DIFF
--- a/plan/issues/2924-new-function-constant-body-compile-away.md
+++ b/plan/issues/2924-new-function-constant-body-compile-away.md
@@ -1,6 +1,6 @@
 ---
 id: 2924
-title: "new Function(\"<const>\") compile-away MVP — replace the no-op stub"
+title: 'new Function("<const>") compile-away MVP — replace the no-op stub'
 status: ready
 created: 2026-07-02
 updated: 2026-07-02
@@ -29,7 +29,7 @@ Second landable slice — pure AOT, **standalone-safe**, no interpreter.
 `new Function(...)` / `Function(...)` currently lowers to a **no-op stub**
 (`src/codegen/expressions/new-super.ts` ~line 3179): it evaluates the arguments
 for side effects and returns `ref.null.extern` — a "function" that returns
-`undefined`. Every test that actually *calls* the constructed function fails
+`undefined`. Every test that actually _calls_ the constructed function fails
 (119 `new Function(` tests fail today, roadmap §5.2), and standalone gets
 nothing.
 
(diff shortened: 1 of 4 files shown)


ttraenkler:
Closing my PR as the duplicate — the parallel session's #2474 is the stronger base for #2924 and survives.

Both PRs implement the same const-arg `new Function` compile-away and both bot-parked with 4 real strict-mode regressions (the synthesized body must be non-strict unless it opts in — §20.2.1.1). Comparing the two: #2474 covers MORE of the acceptance surface (the plain-call `Function(...)` value form and the immediate-call form via `calls.ts`, which I explicitly deferred), guards the callee with a proper global-`Function`-identifier check (I matched the bare identifier text only), already bails `"use strict"`-prologue bodies, and lives in `eval-inline.ts` next to the splice machinery it reuses.

My PR's unique pieces — the standalone/WASI gate (the synthesized all-externref-param closure hits the pre-existing standalone call-marshalling bug I filed as #2945, so shipping ungated means silent wrong values on that lane), the module rollback guard, and the #2945 issue file — I am grafting onto #2474 as part of the [CI-FIX] for the shared strict-mode park, plus the sloppy-`this` bail that fixes the 4 parked paths on both PRs. See #2474 for the combined fix.

PR #2473, opened 2026-07-02
fix(#2934): String(x).<method>() + exec().toString() receiver coercion (slice 2b)

ttraenkler:
## Problem

Two more "static type says X, compiled value is Y" receiver-coercion gaps producing invalid standalone Wasm — same class as slice 2a (#2448, object-ops.ts), different sites:

1. **`String(42).concat(void 0)`** (`String/prototype/concat/S15.5.4.6_A1_T8.js`) — `number_toString` returns the native string EXTERNALIZED (`extern.convert_any`), so a statically-string-typed receiver COMPILES to externref. `compileNativeStringMethodCall`'s `emitReceiver` (`src/codegen/string-ops.ts`) fed it uncoerced to `__str_concat((ref null $AnyString), …)`:
   ```
   call[0] expected type (ref null 6), found call of type externref
   ```
2. **`regObj.exec(str).toString()`** (`RegExp/prototype/exec/15.10.6.2-9-e-1.js`) — the static receiver type resolves to externref, but standalone lowers `exec` natively to a capture-array vec `(ref null $Vec)`. The generic `.toString()` fallback (`src/codegen/expressions/calls.ts`) passed the raw ref to `__extern_toString(externref)`:
   ```
   call[0] expected type externref, found if of type (ref null 98)
   ```

## Fix

- `emitReceiver` casts an externref result back via the established `emitNativeStringRefFromExternref` inverse (in the native-strings world every string-typed externref wraps a native string struct) — one fix covers every string-method arm (concat/charAt/indexOf/slice/…).
- The `.toString()` fallback coerces the **compiled** receiver type to externref (`coerceType` → `extern.convert_any`), mirroring the slice 2a receiver fix.

## Validation

- Both repros: standalone INVALID → valid; concat A1_T8 flips to **pass** in the standalone runner. (exec 9-e-1 runtime ToString-of-match-array semantics is a separate pre-existing gap — 2a precedent; this slice is validity.)
- 120-file concat/exec/toString sweep: 115 → 117 VALID, **0 new invalids** (base measured on the same branch base, not stale /workspace).
- **Byte-identical** on host mode and standalone literal-receiver paths (SHA-checked concat/indexOf/exec-toString shapes).
- New `tests/issue-2934-receiver-coercion-2b.test.ts`: 3/3 pass.
- Scoped string equivalence (string-methods, tostring-valueof, wrapper-string-concat, concat, standalone-string-imports, string-arithmetic-coercion): 82/82 pass.

Residuals documented in the issue file (stack-arity class → slice 3; `__closure_*` host-bridge `env.__call_1_f64` leak → own slice, likely architect spec). Part of #2934 (issue stays in-progress).


ttraenkler:
Merged current main into this branch to clear the BEHIND state (mechanical catch-up, no conflicts). CI will re-run; auto-enqueue picks it up when green.

PR #2475, opened 2026-07-02
fix(#2859): function-typed params resolve through the IR — param-type-not-resolvable to zero

ttraenkler:
Closes the `param-type-not-resolvable` unintended IR-fallback bucket (#2859, child of #2855; `ir-full-coverage` north-star work).

## Root cause

Per-function probe on the corpus instance (`website/playground/examples/benchmarks/helpers.ts`): `addBenchCard`'s `fn: () => number` param — `resolveParamType` (select.ts) had no FunctionTypeNode arm, so ANY callback-taking function demoted to legacy. Additionally `buildLocalCallGraph` misclassified calls through such params as `external-call`. Not a TypeMap propagation gap — a missing annotation-resolution case (the issue's option b).

## Fix

Expressible function-type annotations (params + return all primitive — exactly the slice-3 closure-literal surface) lower to `IrType.closure`:

- `src/ir/select.ts`: new exported `irClosureSignatureFromFunctionTypeNode` helper; `resolveParamType` FunctionTypeNode arm (new `"closure"` ResolvedKind); `collectLocalClosureBindings` counts closure-typed param callees as intra-function closure dispatch. The primitive mapping is pinned identical to `typeNodeToIr`, so `irTypeEquals` holds between a param's declared type and a closure-literal argument's signature.
- `src/codegen/index.ts`: `resolvePositionType` FunctionTypeNode arm via the SAME helper (override map / calleeTypes agree with the lowerer). The existing `lowerClosureCall`/`emitClosureCall` machinery handles the call — closure-typed params dispatch exactly like slice-3 closure locals. **No lower.ts changes.**
- Inexpressible signatures (non-primitive types, void return, rest/optional/default params, generics) keep the honest rejection.

## Corpus / ratchet

`check:ir-fallbacks`: **param-type-not-resolvable 1 → 0** (bucket removed from `scripts/ir-fallback-baseline.json` via `--update`); `addBenchCard` progresses past the param gate and now attributes honestly to `body-shape-rejected` (+1, 31→32 — its `addEventListener` arrow body is #2856's scope, a claim additionally needs `bcrd` via #2858). Total unintended unchanged (45); **no post-claim demotions**. Gate green post-refresh and re-verified after merging latest main.

## STRICT promotion (acceptance criterion 3) — deliberately deferred

`STRICT_IR_REASONS` hard-errors a listed reason in **every compilation** (index.ts:1455), not just the corpus gate. Verified empirically that common shapes still legitimately produce this reason — an unannotated polymorphic param (plain JS) and a union-annotated param (`x: string | number`) both do — so promotion would mass-break test262/user programs (violates criterion 4, 'no regression in test262'). Zero-in-corpus ≠ structurally-zero; promotion needs `resolveParamType` totality first. Full evidence + rationale in the issue file's 'STRICT promotion deferred' section.

## Tests

- `tests/issue-2859.test.ts` (8 green): selector claims; end-to-end execution (zero-arg, one-arg, string/boolean signatures, closure-param forwarding); helpers.ts shape no longer param-type-rejected; inexpressible/void stay rejected; signature-helper unit cases.
- `tests/ir/**`: 121 passed — the 8 failures in `passes.test.ts`/`inline-small.test.ts` (#1167a/b) are pre-existing on origin/main (verified identical on a pristine main worktree).
- Adjacent green: `issue-2923` (dynamic dispatch), `issue-1382` (closure bridge), `ir-vec-two-backend`. `tsc --noEmit` clean.

Issue frontmatter set `status: done` (self-merge path). Note: `priority: high` here converges with in-flight #2469's identical bump.


ttraenkler:
Merged current main into this branch to clear the DIRTY state. Conflict resolution: kept this PR's issue-file frontmatter (completed/assignee), and regenerated scripts/ir-fallback-baseline.json from the merged tree via check:ir-fallbacks --update (param-type-not-resolvable stays eliminated — the fix holds on merged main; body-shape-rejected 34, call-graph-closure 5, class-method 6 as measured). Both src/ files auto-merged with no conflicts. CI will re-validate.

PR #2481, opened 2026-07-02
feat(#2962): native error stringification (§20.5.3.4) + harness-readable exception rendering for standalone

ttraenkler:
## Summary

Fixes #2962 (local plan issue): natively-thrown errors in `--target standalone` were opaque — `String(e)`/`${e}`/`e.toString()` all yielded `"[object Object]"` in-module, and the test262 harness could only record the #2870 label "uncaught Wasm-GC exception (non-stringifiable payload)" (5,898 baseline entries, ~3.4k actionable) because the host cannot stringify a GC payload.

### What changed

- **`__error_to_string(anyref) -> ref $AnyString`** — native §20.5.3.4 `Error.prototype.toString` over `$Error_struct` (`name + ": " + message`; null/empty/non-string message → name alone). Emitted standalone/wasi-only inside `ensureAnyToStringHelper`.
- **`$Error_struct` arms at the three `__any_to_string` "[object Object]" ref fallbacks** (residual, tag-5 boxed-extern recovery, tag-6 refval) — fixes `String(e)`, template interpolation, `e.toString()`, and error concat for every standalone string coercion. JS-host lanes untouched: the arm only exists when the error struct type is registered (never in host mode).
- **`getOrRegisterErrorStructType` moved to `registry/types.ts`** (re-exported from `error-types.ts`) — avoids an error-types ⇄ native-strings import cycle; existing importers unchanged.
- **New finalize exports `__exn_render_prepare(externref)->i32` / `__exn_render_char(i32)->i32`** (gated `(standalone||wasi) && nativeStrings && exn-tag`) — the harness renders any thrown GC payload with zero host imports, same pattern as the `__sget_*`/`__vec_*` harness-support exports.
- **Harness**: `extractWasmExceptionMessage` (runner + worker copies) tries the native render for object-typed payloads before the #2870 opaque fallback. In passing, guarded the same `String(err)` hazard in `extractWasmCallStack`/`extractWasmFuncName` (it crashed my sample run mid-loop — would do the same to a production shard).

### Measured

- Stratified sample (30 opaque fails + 30 passing controls, real runner, standalone lane): **29/30 opaque fails now render actionable signatures** (real `TypeError:`/`ReferenceError:`/full `Test262Error:` assert texts); **0 behavioral control regressions**.
- `tests/issue-2962.test.ts`: 12/12 (both lanes). Adjacent suites (issue-2870/2072/1536/1536c/2102/2188/2891/1910-s2): green; issue-1888 (1) and tdz-reference-error (6) failures are identical on pristine main.
- `__get_caught_exception` retirement: measured **100% of 8,825 leaks co-occur with async/generator imports** → structurally coupled to PATH B #2895; documented in the issue rather than forced here (would re-regress the −32 async harness tests).

### Follow-ups filed

- #2968 — wasi `_start` uncaught printer (fd_write + proc_exit)
- #2969 — construction-time `ToString(message)` (§20.5.1.1) + numeric payload rendering


ttraenkler:
Park fix pushed (43377efd7): the merge_group failure was the standalone root-cause gate (`--max-unclassified-root-causes 0`) hitting 101 newly-visible `Test262Error: …` signatures that this PR's renderer de-masked — NOT a status regression (the merge_group's own regression check passed).

Fix, verified against that run's own merged artifact (unclassified 101 → 0):
- `classifyError`: `^Test262Error` → `assertion_fail` (host lane unaffected — it records `.message` without the ctor prefix)
- `build-test262-report.mjs`: dedicated `demasked-native-assertion` bucket (before `misc-spec-tail`), owns 126 records on that artifact

Removing `hold` so auto-enqueue re-sweeps. Author (fable-2) is suspending at budget wind-down — shepherd owns the queue-watch from here.

ttraenkler:
Merged current main into this branch to clear the BEHIND state (clean, no conflicts — mechanical catch-up so auto-enqueue can pick it up once checks are green on the new head). Taking over queue-shepherding of this PR per fable-2's handoff; watching the merge_group re-run for any residual unclassified-signature class.

PR #2445, opened 2026-07-02
feat(#2938): native no-yield generators — relax the two bails (stacked on #2936)

ttraenkler:
## What

Enables native lowering of NO-YIELD (zero-suspend) sync generators under `--target standalone`/`wasi` by relaxing the two lockstep bails (`buildNativeGeneratorPlan` suspendCount + `isNativeGeneratorCandidate` terminal yield-require), plus two corpus-found gate-consistent bails:

- **Whole-param default on a binding pattern** (`*method({} = undefined)`) — no defaulted-raw-arg arm in the resume prelude yet; mis-typed the state struct.
- **Duplicate-name / computed-name generator methods** (`*id(){}` + `static *id(){}`, `*[sym]()`) — the class collection pass keys on `${className}_${methodName}` and skips the duplicate, so the second member emitted against the first's `NativeGeneratorInfo` ("local index out of range").

Issue file renamed `2920-*` → `2933-standalone-native-sync-generator-resume.md` (id ceded to PR #2424's issue; coordinated re-id).

## Stacked on #2936 (PR #2444)

This branch contains the #2936 late-import shift-regime fix (formerly #2930 — re-id'd, id raced on main) — the blocker that held these bails off. **Explicit predecessor-stacking: do not enqueue until PR #2444 lands**; `hold` label applied. After #2444 merges, a plain `git merge upstream/main` dedupes the stack.

## Corpus proof (542-file no-yield sample, deterministic, incl. both repro files)

- **Host-free** (no `__gen_*`/`__create_generator`/`__get_caught_exception` imports): **4 → 58 files** (+54; ×4 sampling ⇒ ~216 of the 2163-file candidate population, consistent with the ~250–350 estimate on the full 1780-file no-yield corpus).
- **0 per-file status flips** vs main baseline (368 pass both lanes).
- **10 invalid modules — ALL pre-existing on pristine main without the relax** (verified per-lane); **0 relax-attributable invalids**.
- gc/host **byte-inert** (120 sha256 A/B vs origin/main).
- Generator equivalence suites + tests/issue-2936.test.ts green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_017qS1ZofWnkTair9wfGXVn8

PR #2488, opened 2026-07-02
fix(#2970): native gen-result done .value is undefined, not 0 — UNDEF_F64 sentinel carrier + sentinel-aware readers

ttraenkler:
## Problem

On the standalone lane, an exhausted native generator's `.value` read back as a genuine **number 0** instead of `undefined` — live on current main for **with-yield** generators, and the root cause of the #2938 no-yield relax park (PR #2445, "no-yield .value semantic bug"):

```ts
function* foo(): any { yield 42; }
const g: any = foo(); g.next();
const v: any = g.next().value; // JS: undefined
v === undefined  // false (want true)
v === 0          // TRUE  (silent value-space collision)
v * 2 + 5        // 5     (want NaN)
```

## Root cause

The done-result struct (`__NativeGeneratorResult_f64 {value: f64, done}`) stored `f64 0` as the absent value ("consumer never reads value when done" — false). `undefined` was unrepresentable in the f64 carrier. Standalone's canonical undefined is the null externref (`__extern_is_undefined` = `ref.is_null`) — the gen-result path just never produced it.

## Fix (producer-side canonical marker + bounded reader canonicalization)

- **Producer** (`generators-native.ts`): done/empty results carry the **UNDEF_F64 sentinel** (value-tags.ts, #2106 lineage — a signaling-NaN JS arithmetic can never forge; numerically already ToNumber(undefined)=NaN, so typed f64 reads improve 0→NaN). Also `.next()`/`.return()`/bare-`return` missing-value defaults.
- **Dynamic `.value` reader**: the no-static-info / ref-typed read (the test262 harness shape) returns externref over ALL result carriers, canonicalizing sentinel → null externref; statically-numeric reads keep the byte-identical f64 fast path.
- **`__get_member_value` dispatcher** (`member-get-dispatch.ts`): gen-result f64 arms box sentinel-aware (guarded by `isNativeGeneratorResultStruct`; scratch local only when used — host bytes unchanged).
- **`__extern_is_undefined`** (`object-runtime.ts`): recognizes a `$BoxedNumber` carrying the sentinel bits (covers sentinel values boxed by sentinel-blind f64→externref sites).

This is deliberately NOT the #2106 40-site observer wiring — the sentinel crosses to externref at three bounded points, all covered.

## Validation

- `tests/issue-2970.test.ts` (16 cases, host-free asserted): exhausted identity matrix (===undefined/==null/===null/===0/NaN arithmetic/truthiness/default-param), real-value preservation (yield 42, **yielded 0 stays 0**, return-arm 7, **computed NaN is NOT undefined**, .done, typed for-of), .return() no-arg/with-arg.
- Scoped generator suites 2170/2171/2571/2864/2941: 44/44.
- Byte-inertness (sha256 A/B vs main): ALL host-lane cases identical (native generators are standalone/wasi-only); only intended standalone diffs.
- tsc, prettier, biome lint, check:any-box-sites, check:coercion-sites, check:stack-balance all clean.
- test262 controls: `language/statements/generators/{no-yield,return}.js` stay pass; the parked relax branch's failure signature ("returned 2") is exactly what this repairs.

## Relation to #2938

Unblocks the #2938 revival (next PR): merge main into `issue-2933-noyield-relax`, construct-strided corpus validation, single re-admission of PR #2445. Issue file `plan/issues/2970-genresult-undefined-carrier.md` rides this PR (status: done).


ttraenkler:
Unparked: the merge_group quality failure was a duplicate issue id — id 2970 was taken on main by the import-meta per-module identity issue (parallel session allocator race). Re-id'd this PR's issue to #2979 (fresh allocation): renamed the issue file, the test file (tests/issue-2979.test.ts), and all code-comment refs; provenance note added in the issue file. No functional changes. Auto-enqueue picks it up when checks are green on the new head.

ttraenkler:
Park diagnosis (read-only — I have not touched this branch or its labels): the 13:54Z park is a REAL PR-caused regression, not a dead shard and not the async-scheduler OOM class from #2978.

**Failing step**: "Standalone regression guard (#1897)" inside `merge shard reports`, deterministic across both merge_group runs. Net −49: 55 regressions / 6 improvements (ratio 916%), all 55 wasm-hash-changed, bucket signature `5d6cd682f5b50dc5` — i.e. PR-specific, NOT the import.meta/TLA collateral set from the #2471 baseline settle.

**The 55 files** are all `*-ary-ptrn-elem-id-init-{undef,hole,exhausted}*` destructuring-param tests (class + object dstr), each failing `TypeError: Cannot destructure 'null' or 'undefined'`.

**Mechanism (high confidence)**: this PR's UNDEF_F64 sentinel for a done gen-result `.value` surfaces a real `undefined` where these patterns previously saw `0`/hole — and the destructuring lowering treats the sentinel ELEMENT as if the whole RHS were null/undefined and throws, instead of firing the element default. Per §13.15.5.3 (ES2023 DestructuringAssignmentEvaluation / IteratorBindingInitialization), an undefined element with an initializer must trigger the default; only the destructured OBJECT itself may throw. This is a known hazard class in this repo — see memory note `project_2040_tag5_classifier_dstr_default_regression`.

**Fix needed (on your side)**: make the destructure element-read path sentinel-aware — sentinel → undefined → element default fires.

**Evidence**: run 28595506090 / job 84790476012; reproduced locally from that run's standalone merged JSONL diffed against baselines@main.

The PR stays parked until fixed — please push the fix and then do a single hold-removal per the auto-park protocol. — shepherd-f (diagnosis by our dev-2934f)

PR #2496, opened 2026-07-02
fix(#2934): object-toString string-coercion invalid-Wasm family + main console.log regression (slice 3a)

ttraenkler:
## Problem

The `concat/S15.5.4.6_A4_T2` standalone "not enough arguments on the stack" cluster (#2934 slice 3a) — **plus a live main regression this uncovered**: plain `console.log("ab".concat("cd"))` under `--target standalone` emits INVALID Wasm on current main (`__str_to_extern: call (need 3, got 2)`), bisected to the #2473 merge (a3576e7) which made the latent bridge bug reachable for that shape. The shape is invisible to the PR gates (test262 wrapping avoids `console_log_string`; the standalone floor has no console.log-string case — flagged to the lead as a gate blind spot).

## Root causes (three, one family)

1. **`ensureNativeStringExternBridge` late-import over-shift** (`src/codegen/native-strings.ts`) — the bridge queued its three late imports (`__str_from_mem`/`__str_to_mem`/`__str_extern_len`) and baked their indices into the helper bodies **without closing the deferred batch**. The eventual `flushLateImportShifts` bumps every `funcIdx >= importsBefore` and cannot distinguish a freshly-baked (already-final) import ref from a stale defined-func ref — `__str_to_extern`'s `call __str_from_mem` (arity 2) landed on `__str_copy_tree` (arity 3). Fix: flush the batch after registering, **before** baking — gated on actually having registered (a pure funcMap-hit lookup must not force-flush an outer batch), so already-registered paths are byte-identical.
2. **`normaliseToString` no-result arm** (`src/codegen/type-coercion.ts`, `tryStructToString`) — a dispatched `toString` whose Wasm func type has **no result** (always-throws/never — `toString(){ throw "x"; }` — or void) pushed nothing and then fed `$__any_to_string` (0 operands for a 1-arg call). Per §7.1.1 OrdinaryToPrimitive that path ends in TypeError; the emitted throw leaves the stack polymorphic so the enclosing arm validates (dead code after an always-throwing call).
3. **Reflective receiver ToString** (`src/codegen/string-ops.ts`, `emitReceiver`) — `String.prototype.concat.call(obj, …)` compiles the receiver to a concrete object struct ref; §22.1.3.x requires ToString(this). Now dispatched via `tryStructToString` (correct throwing-toString semantics after fix 2); string-typed refs excluded/untouched.

## Validation

- `S15.5.4.6_A4_T2`: standalone INVALID → **pass** (the test asserts the receiver's `"intostring"` throw — now propagates correctly).
- `console.log("ab".concat("cd"))` and string-method shapes: INVALID on main → valid (regression fix).
- 120-file concat/exec/toString sweep: 117 → 118 VALID, 0 new invalids.
- Throwing-toString family (plus-concat, template literal, `.concat(o)`, reflective `.call`): all valid with correct exception propagation (runtime-probed).
- 71/71 scoped string equivalence tests; 14/14 new+1c regression tests; tsc clean; coercion-sites gate OK.

## Notes

- **Predecessor-stacked on #2489** (slice 1c, fully green, same issue file) per the known-dependency exception — this branch merges its commits; enqueue after #2489 lands is handled by the queue ordering naturally.
- Remaining #2934 slices (documented in the issue): (3b) the async `__closure_5` for-await drop-balance (different mechanism, still invalid); the `__closure_*` `env.__call_1_f64` host-bridge design item (architect spec recommended).


ttraenkler:
Queue note: I moved this PR to the head of the merge queue (single dequeue + jump re-enqueue) per tech-lead fast-track — it fixes a live regression on current main (standalone console.log("ab".concat("cd")) → invalid Wasm, reachable since #2473). #2489's in-flight validation (~2 min old) rebuilds behind it. — shepherd-f

PR #2503, opened 2026-07-02
fix(#2982): tagged-template string[] first-param false-positive TS2345

ttraenkler:
## Problem

11 tagged-template equivalence tests + the whole `tests/issue-141.test.ts` tagged-template suite fail to compile on clean main. Idiomatic tag functions:

```ts
function tag(strings: string[], a: number, b: number): number { ... }
tag`hi ${10} w ${20}`;
```

fail with `Argument of type 'TemplateStringsArray' is not assignable to parameter of type 'string[]'` (TS2345).

## Root cause (bisected)

Not a recent-wave regression — **latent since 2026-04-17**. `git bisect` pinpointed **d8cfbb7a7 "fix(compiler): fail on incompatible TypeScript annotations"**, which added TS2345 to `HARD_TS_DIAG_CODES`. A tagged template implicitly passes a `TemplateStringsArray` (frozen `ReadonlyArray<string> & { raw }`, ECMA-262 §13.2.8.4), not assignable to a mutable `string[]` — a TS-strictness artifact, not a runtime incompatibility. Before that gate the diagnostic was non-fatal, so these compiled (as they do at runtime).

## Fix

Fix forward (not test edits — `string[]` is idiomatic, runtime-correct code). Added `isTaggedTemplateStringsArrayFalsePositive` to `src/compiler.ts`, wired into `isHardTypeScriptDiagnostic` next to the existing #2616/#2741/#862 downgrades. Downgrades a 2345 **only** when: (1) inside a TaggedTemplateExpression; (2) its source type is literally `TemplateStringsArray` (the arg0 template-object, not a `${…}` substitution); and (3) the resolved first parameter is a string-like array. So `number[]` first params and genuine substitution mismatches stay fatal.

## Tests

- `tests/issue-2982.test.ts` (new): 5/5 — 3 positive + 2 tight-scoping negatives (`number[]` param and a genuine substitution mismatch both stay fatal 2345).
- equivalence tagged-template subset: 13/13 (was 11 compile-errors).
- `tests/issue-141.test.ts` tagged-template suite: green.

Runtime equivalence (wasm ↔ js) verified; codegen already lowered these tag calls.

### DIFF
--- a/plan/issues/2982-tagged-template-string-array-param-hard-error.md
+++ b/plan/issues/2982-tagged-template-string-array-param-hard-error.md
@@ -1,0 +1,83 @@
+---
+id: 2982
+title: "Tagged-template regression: `strings: string[]` first-param trips a fatal TS2345 (idiomatic tag fns fail to compile)"
+status: done
+completed: 2026-07-02
+assignee: ttraenkler/opus-4
+sprint: current
+created: 2026-07-02
+updated: 2026-07-02
+priority: high
+horizon: m
+feasibility: medium
+task_type: fix
+---
+
+## Problem
+
+11 tagged-template-literal equivalence tests (plus the whole
+`tests/issue-141.test.ts` tagged-template suite) fail to compile on clean main.
+Surfaced by fable-6 during #2965 validation, framed as a "recent merge-wave
+regression."
+
+A tag function declared with an idiomatic first parameter:
+
+```ts
+function tag(strings: string[], a: number, b: number): number { ... }
+tag`hello ${10} world ${20}`;
+```
+
+fails compilation with:
+
+```
+Argument of type 'TemplateStringsArray' is not assignable to parameter of type 'string[]'.
+```
+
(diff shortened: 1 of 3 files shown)


ttraenkler:
Closing this in favor of #2502 (issue #2981), which merged first and takes the correct approach.

I reached the opposite diagnosis under time pressure, but the tech lead's call is right and I agree on review: `TemplateStringsArray extends ReadonlyArray<string>` is genuinely NOT assignable to a mutable `string[]` parameter under real TypeScript — `tsc` itself emits TS2345 here (my own repro confirmed exactly that diagnostic). So the hard error is conformant with real tsc, not a compiler bug. My compiler-side suppressor in src/compiler.ts would have permanently loosened js2wasm's type gate away from real tsc semantics. #2502 instead corrected the non-conformant test signatures, which is the right fix and is already merged with all suites green.

Superseded — closing. #2982 marked wont-fix, pointing at #2981.

PR #2509, opened 2026-07-02
fix(#2973): eval-shim sub-compiles opt out of JS2WASM_IR_FIRST — kill the one silent fail-loud violation

ttraenkler:
## Summary

Fixes **#2973** — the single silent fail-loud violation found by #2138's Slice-3 full flagged measurement (test262 `S12.4_A2_T2`: `undefined` instead of `7`).

### Root cause
The dynamic eval path (`runtime-eval.ts`) compiles eval strings **in-process** as a claimable `export function __eval_result() {...}` wrapper. That sub-compile inherited the ambient `JS2WASM_IR_FIRST=1`: the wrapper's legacy body was skipped (#2138), a claim-partial residual in the eval'd expression hard-errored post-claim (correct, fail-loud) — but the shim's `catch`/`!success` arms treat compile failure as a recoverable fast-path miss and **swallowed the error**, silently yielding `undefined`.

### Fix (structural, options-based — no ambient env mutation)
`CompileOptions.disableIrFirst` → `buildCodegenOptions` → `CodegenOptions.disableIrFirst` → `generateModule`'s irFirst gate. Both eval-shim `compileSourceSync` sites pass it — they are the **only** `compileSourceSync` callers in `src/` (grep-verified); `eval-inline.ts` compiles into the same module (functions containing `eval` are selector-rejected, IR-first never touches them), and the public `compileSource` APIs legitimately keep the flag. **Flag-off behavior untouched.**

### Fixture proven load-bearing
The exact wrapper (`export function __eval_result() { return (5+1|0===0); }`) compiled flag-on WITHOUT the opt-out hard-errors (`Phase 1 requires matching operand types for '|'` on the skipped slot) — the regression test's flag-ON case returned `undefined` pre-fix.

### Tests
`tests/issue-2973.test.ts` (3): dynamic-path eval (param-sourced string, immune to the #2923 const-fold fast path) returns 7 flag-off AND flag-on; statement-form wrapper parity.

### Also in this PR
`plan/issues/2972-*.md` gains a scoping analysis for the successor slice (the other Slice-3 divergence class): the naive selector-side reject would destroy the vec fast path; the sound fix is a proven-in-bounds string read (bitmask-bound + literal-length receiver) with the OOB-`undefined` trap documented.


ttraenkler:
Closing as duplicate: PR #2506 (parallel session) landed the functionally identical fix on main first — same structural disableIrFirst opt-out through CompileOptions → CodegenOptions → generateModule, both eval-shim call sites, equivalent regression tests (both implementations followed the design recorded in plan/issues/2973-*.md). The one piece unique to this PR — the #2972 scoping analysis — moves to the #2972 implementation branch. No code from this PR is needed on main.

PR #2537, opened 2026-07-02
fix(#2999): fold builtin .constructor to bare-builtin carrier — drop env::Object_get_constructor standalone leak

ttraenkler:
## Summary

Eliminates the `env::Object_get_constructor` host-import leak flagged by the
2026-07-02 round-5 leak analysis (**9 execution-verified sole-import standalone
passes**).

## Root cause

Reading `.constructor` on any extern-class receiver (`compileExternPropertyGet`,
`src/codegen/property-access.ts`) walks the extern inheritance chain to the
`Object` base extern class (`src/codegen/index.ts` ~L13439 — the only declarer of
a `constructor` property, `importPrefix: "Object"`), so the getter path emits an
`Object_get_constructor` host import.

Confirmed mechanism (fresh probes on run 28605503741 / main): in the standalone
lane the host import resolves to `undefined` (the `$NativeProto` receiver is an
opaque WasmGC struct), and a bare builtin identifier (`Set`, `WeakMap`, …)
compiles to the standalone null-externref carrier (builtins have no native
constructor-object identity yet). So `assert.sameValue(<Builtin>.prototype.constructor, <Builtin>)`
passes **tautologically** — both sides collapse to the shared nullish carrier.
Proof: `assert.sameValue(Set.prototype.constructor, Map)` *also* passes, i.e. the
pass never depended on the host import returning a real value.

## Fix

Static-fold `.constructor` on a builtin receiver to that same bare-builtin
carrier. When `ctx.standalone`, `propName === "constructor"`, and the receiver's
extern class is a known builtin (`BUILTIN_CTOR_NAMES`), compile the receiver for
side effects, drop it, emit `ref.null.extern` — no host import.

- **Spec-sound**: `<Builtin>.prototype.constructor` / `(new <Builtin>()).constructor`
  IS `%<Builtin>%`, the value the bare `<Builtin>` identifier denotes — routing
  the read to the same carrier as the identifier is correct, not a shortcut.
- **Behaviour-preserving**: LHS and RHS remain the identical nullish value that
  SameValue-compares equal — the (already tautological) pass is unchanged, only
  the import is gone.
- **Narrowly scoped**: `BUILTIN_CTOR_NAMES` receivers only (bare value provably
  null) and `ctx.standalone` only. User-`declare class` extern receivers and the
  gc/host lane keep the real `Object_get_constructor` read (a genuine value
  there), so zero behaviour change off the standalone-builtin path.

## Results

- 9/9 origin test262 files pass in the standalone lane, host-free
  (`env::Object_get_constructor` absent).
- `tests/issue-2999.test.ts` — 8/8 (7 host-free builtin `.constructor` shapes +
  1 gc/host retains-import guard proving the standalone-only scoping).
- gc/host lane verified byte-inert (fold branch unreachable when `!ctx.standalone`;
  import list unchanged).
- No regression: the two `Error/prototype/constructor` standalone fails are
  pre-existing (fail identically without the fold).

### DIFF
--- a/plan/issues/2999-standalone-object-get-constructor-leak.md
+++ b/plan/issues/2999-standalone-object-get-constructor-leak.md
@@ -1,0 +1,88 @@
+---
+id: 2999
+title: "Standalone: eliminate env::Object_get_constructor host-import leak — fold builtin .constructor to the bare-builtin carrier"
+status: done
+completed: 2026-07-02
+assignee: ttraenkler/agent-abd6cda44521fc1c9
+sprint: current
+priority: medium
+horizon: s
+feasibility: medium
+origin: plan/log/investigations/2026-07-02-leak-analysis-round5.md
+---
+
+## Problem
+
+Round-5 leak analysis (2026-07-02) ranks `env::Object_get_constructor` as an
+execution-verified GENUINE sole-import leaky lever: **9 official standalone
+passes** carry exactly one `env::` import, `Object_get_constructor`. They are all
+`<Builtin>.prototype.constructor === <Builtin>` shapes (plus instance forms),
+for Set / WeakMap / WeakRef / WeakSet / RegExp / FinalizationRegistry /
+DisposableStack / SuppressedError. These are host-import leaks — the standalone
+binary imports a host getter it should not need.
+
+## Root cause
+
+Reading `.constructor` on any extern-class receiver
+(`compileExternPropertyGet`, `src/codegen/property-access.ts`) walks the extern
+inheritance chain to the `Object` base extern class (`src/codegen/index.ts`
+~L13439 — the only declarer of a `constructor` property, `importPrefix: "Object"`),
+so the getter path emits an `Object_get_constructor` host import.
+
+Confirmed mechanism (fresh probes on run 28605503741 / main): in the standalone
+lane the host import resolves to `undefined` (the `$NativeProto` receiver is an
+opaque WasmGC struct — `recv.constructor` on the host is `undefined`), and a bare
+builtin identifier (`Set`, `WeakMap`, …) compiles to the standalone
+null-externref carrier (builtins have no native constructor-object identity yet).
+So `assert.sameValue(<Builtin>.prototype.constructor, <Builtin>)` passes
+**tautologically** — both sides collapse to the shared nullish carrier. Proof:
+`assert.sameValue(Set.prototype.constructor, Map)` _also_ passes (both nullish),
+i.e. the pass never depended on the host import returning a real value.
+
+## Fix
+
+Static-fold `.constructor` on a builtin receiver to that same bare-builtin
+carrier. In `compileExternPropertyGet`, when `ctx.standalone`,
+`propName === "constructor"`, and the receiver's extern class is a known builtin
+(`BUILTIN_CTOR_NAMES`), compile the receiver for side effects, drop it, and emit
+`ref.null.extern` — no host import.
+
+- Spec-sound: `<Builtin>.prototype.constructor` / `(new <Builtin>()).constructor`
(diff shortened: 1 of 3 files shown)


ttraenkler:
## Reviewer note — this is host-import elimination, NOT a correctness fix (honest-accounting flag)

I want to be explicit about the category of this PR before anyone merges it, in the discipline the team established around the #2463 vacuity work.

**What this genuinely fixes:** the `env::Object_get_constructor` host import is gone from these 9 standalone binaries. That is real and valid on its own.

**What it does NOT fix (and neither claims nor changes):** real builtin constructor/prototype **object identity**. All 9 tests pass via a **null≡null tautology** — in standalone both the `.constructor` read and the bare-builtin comparison target evaluate to the **same** null-ish externref carrier, so `assert.sameValue(<Builtin>.prototype.constructor, <Builtin>)` is trivially true. The cross-check proves it: `assert.sameValue(Set.prototype.constructor, Map)` — the **wrong** builtin — **also passes**. A genuine constructor-identity implementation would make that cross-check FAIL.

**Why this is a weaker category than an execution-verified fix.** This is *coincidental wrongness*, not vacuity. The code **does execute** (the round-5 inject-throw probe correctly labels this lever GENUINE = non-vacuous), it just returns an incorrect value (null) that happens to equal another equally-incorrect value (null). An inject-throw / execution-proof check does **not** catch this — the body runs. So "GENUINE" in the round-5 table means "not vacuous", **not** "value-correct".

**The fold is still the right change:** behaviour-preserving (LHS and RHS remain the identical null carrier the pre-fix path already produced), removes a dead import, and tracks the gap rather than entrenching it — when builtins gain reified constructor identity, the bare-identifier resolution and this fold site update in lockstep.

**Substrate that actually closes the gap:** #2963 (reify builtins as first-class values with stable identity). I've documented all of this in the issue file (`plan/issues/2999-...`, "Honest-accounting caveat" + "Broader-pattern sweep" sections, pushed to this branch) and added a quick broader-pattern sweep: ~55 test262 `built-ins/**` files share the `<Builtin>.prototype.constructor === <Builtin>` shape, so the coincidence is substrate-wide (not unique to these 9) but bounded to the `sameValue` path (raw `===` builtin comparisons return false) — already tracked by #2963, not created or worsened here.

**Bottom line for the merger:** merge on the "host-import eliminated, correctness gap pre-existing and unchanged" basis — not as a genuine execution-verified conformance win. Net effect on the honest metric is import-reduction, not new real passes.

ttraenkler:
I'm closing this as **superseded by #2566** (issue #3006) — please don't merge it.

I extended #2963's builtin-value reification substrate to give
`<Builtin>.prototype.constructor === <Builtin>` **genuine** identity instead of
this PR's null-fold. As the honest-accounting note here already called out, the
fold here passes only via a null≡null tautology — both `.constructor` and the
bare builtin identifier collapse to `ref.null.extern`, so
`Set.prototype.constructor === Map` **also** passes (wrong).

In #2566 I route BOTH the bare identifier read and the `.constructor` read to the
same per-name reified `$Object` singleton (`__builtin_ctor_<Name>`), so
`Set.prototype.constructor === Set` is genuinely true (same WasmGC object,
`ref.eq`) and the swap-wrong-builtin cross-check `Set.prototype.constructor ===
Map` is genuinely **false**. Same host-import elimination you achieved here
(`env::Object_get_constructor` gone from all 9 standalone binaries, gc/host lane
byte-inert and retains the real import), but now backed by real constructor
identity rather than the coincidental null cancellation — so the correctness gap
this PR deliberately left open is closed, not entrenched.

Thanks for the thorough root-cause and honest-accounting write-up — it made the
genuine-identity follow-up straightforward. Superseded, not rejected.

PR #2552, opened 2026-07-02
docs(#3005): file eval-as-any callee stack overflow bug

ttraenkler:
Files #3005 (triage only, no code fix).

## What
`(eval as any)()` and `(eval as any)("1+1")` crash the compiler with `RangeError: Maximum call stack size exceeded` (infinite recursion), surfaced as a doubled `Internal error compiling expression` diagnostic. The `any`-typed *alias* form (`const x: any = eval; x()`) does NOT crash — the trigger is the cast-in-parenthesized-callee shape.

## Root cause (documented in the issue)
In `compileCallExpression` (`src/codegen/expressions/calls.ts`) the parenthesized-callee unwrap (line 4186) reaches the inner `AsExpression`, which isn't in the special-cased set (conditional/comma/unary), so it falls through to the generic `ts.factory.createCallExpression` synthetic-call path (4219). Because `AsExpression` is not a `LeftHandSideExpression`, `ts.factory` re-wraps the callee in a `ParenthesizedExpression`, and the re-entry at 4226 rebuilds an identical call → unbounded recursion. Same hazard the existing 4198-4212 comments already guard for conditional/binary/unary callees; `AsExpression`/`SatisfiesExpression`/`TypeAssertion` are the missing cases.

Discovered while implementing #2960 (PR #2548). Filed so it isn't lost. No fix here.

### DIFF
--- a/plan/issues/3005-eval-as-any-callee-stack-overflow.md
+++ b/plan/issues/3005-eval-as-any-callee-stack-overflow.md
@@ -1,0 +1,132 @@
+---
+id: 3005
+title: "Compiler stack-overflow on `(eval as any)()` — cast/parenthesized callee re-wrap recurses infinitely"
+status: ready
+sprint: current
+created: 2026-07-02
+updated: 2026-07-02
+priority: medium
+horizon: s
+feasibility: medium
+reasoning_effort: medium
+task_type: bug
+area: codegen
+language_feature: eval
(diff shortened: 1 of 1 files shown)


ttraenkler:
Superseded by #2553, which includes the fix and a complete resolution for #3005's issue file (this PR was triage-only).

PR #2596, opened 2026-07-03
feat(#2952): IR multi-exit spec (Design A) + do-while adoption (slice 1)

ttraenkler:
## What

Umbrella issue #2952: six direct-only statement kinds (switch, break, continue, do-while, labeled, for-in) share one structural blocker — a nested-buffer control-flow body can't branch to an enclosing structured Wasm frame.

This PR delivers two things:

### 1. Architect spec (acceptance criterion #1) — Design A chosen

Recorded in `plan/issues/2952-*.md` under `## Implementation Plan`. Picks **Design A** (a lowering-time label→depth resolver) over Design B (a true multi-exit CFG):

- **A1** loop-carried `IrLabelId`; **A2** a buffer-terminating `br.label{label, mode}` where **depth is DERIVED at emit time, never stored in the IR** (correcting the issue's `{label, depth}` sketch — a stored depth rots under any buffer-renesting pass); **A3** a `ctrlStack` threaded through `lower.ts` that counts Wasm frames to the labeled target.
- The **finally-interaction caveat** that was Design B's only justification is resolved without a CFG: `IrInstrTry` already inlines `finallyBody` at every abrupt-exit path, so a `break`/`continue` crossing a `try/finally` is just one more inline emission point. Since all six JS kinds are *lexically* scoped, a lexical resolver is complete — B is not needed.
- Per-kind adoption order + verifier rules recorded for the follow-up slices.

### 2. Slice 1 — do-while adoption (the issue's "cheapest first")

A do-while is a post-test loop. It **reuses the existing `while.loop` IR node with a new `postCond` flag** instead of adding a new instr kind, so there is **zero exhaustiveness churn** across `forEachNestedBuffer` / `mapNestedBuffers` / verify / effects / propagate — every pass treats it as a `while`. Only the lowering emission order flips (body → cond-check). Bodies containing `break`/`continue` stay rejected by `isPhase1BodyStatement`, so only the multi-exit-free subset is claimed (no post-claim demote — #2135 selector↔lowering parity).

Files: `nodes.ts` (`postCond?`), `builder.ts`, `from-ast.ts` (`lowerDoStatement`), `lower.ts` (post-test branch), `select.ts` (`isPhase1DoStatement`), `scripts/gen-ir-adoption.mjs` + regenerated `ir-adoption.md`.

## Tests

- `tests/issue-2952.test.ts` — 5/5, incl. the **defining post-test semantic** `do { x++ } while (false)` ⇒ 1 (a `while` would be 0) and the **break-rejection boundary** (a do-while whose body has `break` is NOT claimed).
- No regression: `issue-1280` (while/for) + `issue-2136` (numeric-truthiness loops) + `issue-1169n` — 35/35.
- `check:ir-fallbacks` OK · `check:ir-adoption` OK · `tsc --noEmit` clean · `biome lint` clean.
- Pre-existing unrelated failure in `tests/ir-if-else-equivalence.test.ts` (`env "__unbox_number" requires a callable`, a harness import-stub gap) confirmed identical on clean `origin/main`.

## Status

Issue stays **in-progress** — break/continue, switch, for-in are follow-up slices under the banked Design-A spec.


ttraenkler:
## Merge-park investigation: the -4 is a CI flake, not a regression (do-not-remove-hold-without-lead-ack)

Investigated the bot park-hold (net -4, bucket `eecf8e25208aade6`, failed run `28670066683`, js-host lane). **Verdict: the 4 flagged tests are a shared-worker flake, not a regression caused by this PR. No code fix is needed.**

### The 4 flagged tests
- `built-ins/DataView/prototype/setFloat16/length.js`
- `built-ins/DisposableStack/length.js`
- `built-ins/Promise/try/return-value.js`
- `built-ins/SuppressedError/newtarget-is-undefined.js`

### Why it cannot be this PR (structural)
Every line this PR changes is guarded by `ts.isDoStatement(...)` (from-ast.ts, select.ts) or `instr.postCond === true` (lower.ts); the non-`postCond` `while`/`for` emission is the byte-identical original path. **None of the 4 tests contains any loop** (their harness includes — propertyHelper.js / asyncHelpers.js — have no do-while either), so none exercises a changed path.

### Empirical proof
1. **Byte-identical wasm** for all 4 tests across merge-base `bc8a1d4` (pre-PR), PR-head `a53cacd0`, and current `origin/main` `6b71c93`: `dfd62ae67a556df2` / `b8938d4451e6acec` / `0ece0241fc27df9c` / `1689c19edf1aef7b`.
2. **All 4 pass in isolation** at HEAD (compile + instantiate + run).
3. **Real `pnpm run test:262`** (gc target, same worker path as merge_group) scoped to the 4 files at HEAD: **4 pass / 4 (100%)**.
4. **Do-while blast radius neutral**: 113 drivable test262 do-while tests have *identical outcomes* at merge-base vs HEAD (24 COMPILE_FAIL / 56 OK / 25 WASM_EXN — all pre-existing, unchanged). The IR do-while lowering is behaviour-equivalent; it does not poison any worker.
5. **Standalone lane: 0 regressions.**

### Mechanism
These four newer-feature tests are the classic collateral of shared-worker built-in poisoning (adjacent prototype-mutating tests in the same shard fork). Per auto-park rule (c), a confirmed flake/collateral may be re-admitted.

### Recommendation
Remove `hold` and let `auto-enqueue` re-admit (do **not** re-enqueue by hand). Leaving to the tech lead per the auto-park protocol (never remove a bot park-hold without lead ack). Full write-up in `plan/issues/2952-ir-multi-exit-control-flow.md` (§ Merge-park investigation), repro scripts under `.tmp/`.

PR #2631, opened 2026-07-04
fix(#2853): nominal shape branding + sidecar-shadow heal — compiled-acorn parses division-after-number AND regex groups

ttraenkler:
Fixes the two banked #2853 miscompiles that block the interpreter tier's acorn path (#2618 P2).

## Bug A — division after a numeric literal (`var x = 1 / 2;` threw)

**Root cause (proven with a 2-type module):** WasmGC type identity is iso-recursive and STRUCTURAL — field names don't exist in the binary, so the engine canonicalizes `__anon_{startsExpr:i32}` and `__fnctor_TT{beforeExpr:i32}` (identical layouts) into ONE runtime heap type; `ref.test (ref $b)` answers 1 on an `$a` instance. Every `ref.test`-keyed property dispatch (`__sget_*`/`__sset_*`, inline member-get chains) then reads by struct OFFSET, not by KEY: `({startsExpr:true}).beforeExpr === true`. In acorn, `num.beforeExpr` (absent key) read `startsExpr`'s `true` → `/` after a number was tokenized as a regex start → validator trap. The pre-existing #2009 `$shape` stamp covers only anon-vs-anon collisions at the exported-accessor level; this collision was anon-vs-FNCTOR and also poisoned inline dispatch.

**Fix — `src/codegen/shape-brand.ts` (new finalize pass, both pipelines):** append a trailing immutable `(ref null <chain>)` brand field to every structurally-colliding `__anon_*`/`__fnctor_*` bare struct. Chain anchor is `$__vec_base` (an OPEN `sub` struct — openness/finality is part of canonical identity, so no bare shape can equal it); each next branded shape references the previous → by induction all branded shapes are pairwise canonically distinct, making `ref.test` dispatch exact. `struct.new` sites get a purely local `ref.null <target>` insertion (brand is the LAST field). **No type-table insertion, no index remap, no new rec groups** (all chain refs point backward), and the #2527 cross-module runtime-ABI types (strings/vecs) stay canonical. **Byte-identical output for collision-free modules** (verified vs main on simple/one-shape/strings/classes/two-same-key-shapes cases).

## Bug B — ANY regex group `/(a)/` threw ("Unmatched ')'") 

**Root cause (instrumented full-acorn trace, overturns the earlier type-index hypothesis):** inside prototype methods `this.pos` advanced correctly (0→1→2→3), but every `state.pos` read through a method PARAMETER read a frozen 0 and every param-path write was lost. Mechanism: `_emitStructFieldSettersInner` skipped mixed-kind field-name buckets, and acorn's `pos` bucket mixes kinds across structs → `__sset_pos` was never emitted → host-MOP writes (`__extern_set` → `_safeSet`) landed **sidecar-only**, while compiled `struct.set` writes updated only the live field; `__extern_get` prefers the sidecar, so `regexp_pattern`'s `state.pos = 0` seeded a permanent 0-shadow. `/(a)/` then died in the V8-compat branch: phantom pos(0) ≠ len(3) → `state.eat(')')` succeeded against the TRUE pos → `raise "Unmatched ')'"`. (`/a/` passed the same broken comparison only by luck — the defect was global.)

**Fix:** (1) mixed-kind buckets now emit `__sset_<key>` (externref signature, per-arm coercion; numeric fields via `__unbox_number`; un-coercible arms drop to sidecar as before); all setters return **i32 1 iff an arm matched and wrote the live field**; (2) `_safeSet` uses that flag to skip the sidecar and heal stale entries on a successful live-field write (#2731 shadowed re-adds excepted; old void setters keep prior behaviour).

## Verification

- 4-line repro + 3-shape aliasing + acorn tokenizer-truthiness: `tests/issue-2853.test.ts` (green).
- Compiled pinned acorn: `1 / 2`, `10 / 2 / 5`, `a / b`, `a % b`, `/(a)/`, `/(?:a)/`, `/(?<n>a)/`, `/^in(stanceof)?$/`, `/a/`, `/[a-z]/` — **all 10 parse** (both were WebAssembly.Exception before). AST `end` offsets now correct too (same shadow healed them).
- `tests/equivalence` full suite: 1638 tests, **0 newly-failing vs main** (36 failures identical on main HEAD).
- Setter-contract-adjacent suites (#1712 identity, #2831 materializer, #2009 collisions, #2668, anon-struct): only the 4 failures that already fail on main.
- Full acorn self-parse now runs 61s deep instead of throwing at line 38; the next-deeper gap is being bisected and will be filed as a follow-up issue (allowed by this issue's acceptance).

## Residuals (documented in the issue)

- Absent-key reads through typed inline dispatch still coerce to `0`/`null` instead of `undefined` (pre-existing typed-read residual, truthiness-correct).
- Same-layout sibling CLASS structs can still canonicalize together (unbrandable without breaking subtype field prefixes) — the #2188 `userClassId` disease class; separate issue if it bites.
- #2850 should be closed into #2853 (char-class half fixed earlier; group half fixed here) — PO to reconcile.


ttraenkler:
Removed the auto-park `hold`. Diagnosed the cited merge_group failure (run 28703143222, "check for test262 regressions"): the job's own annotations flag it as likely **baseline drift** — "js2wasm-baselines JSONL is 2 test262-relevant commit(s) behind main HEAD" plus a "#1081 merge-base cache MISS … drift attribution may be imprecise". I merged current `origin/main` (b81e36229) to refresh the stale base + baseline. Re-validating once via a fresh merge_group; if a real regression survives against the current baseline it'll re-park.

PR #2646, opened 2026-07-04
feat(#2957): activate async state machine for arrows + function expressions (phase 2)

ttraenkler:
## #2957 phase 2 — arrows + function expressions activate

Phase 1 (#2638) extracted the shared `maybeActivateAsync` entry point but kept it
declaration-only. Async **arrows** and **function expressions** compile via
`closures.ts::compileArrowAsClosure`, which never reached the activation hooks, so
`const f = async (x) => await g(x)` silently fell back to the legacy synchronous
pass-through — the lifted closure compiled with an `f64` result and no CPS state
machine, returning a sync value instead of a real Promise.

### Measure-first (host lane, `origin/main`)
The lifted closure for `const f = async (x) => await g(x)` compiled with
`(result f64)` and emitted **no** `Promise_then2`/`__cb_N` CPS markers — a
non-activated sync value. (The naive `typeof===object` probe false-positives
because `__box_number` boxes the sync `f64`; the authoritative signal is the
lifted func's result type + CPS markers / raw-externref thenability.)

### Change (4 edits)
- **`async-activation.ts`** — refactor the two inline activation blocks into a
  pure decision (`decideAsyncActivation`) + an emit dispatcher (`emitAsyncLane`).
  `maybeActivateAsync` (fn-decl entry) is unchanged in behaviour (declaration-only
  gate preserved). New `planAsyncClosureActivation` (no decl guard) +
  `emitAsyncClosureBody` (no post-hoc type rewrite) for the closure path.
- **`closures.ts::compileArrowAsClosure`** — decide activation EARLY and bake the
  `externref` (Promise) result into the lifted func/struct type *before* it is
  built (`computeClosureWrapperSig` had set the unwrapped awaited type — the root
  of the sync fallback), then emit the async machine in place of the statement
  loop. Generators excluded.
- **`async-cps.ts::splitBodyAtAwait`** — add a concise-expression-body branch so
  `async (x) => await P` (an expression body, the acceptance-criteria canonical)
  splits as a `return await` tail. Additive: fn-decls always have block bodies and
  never reach it.
- **`declarations.ts` `collectAsyncCpsImports`** — widen the prepass node gate to
  arrows / function expressions so the host CPS imports
  (`Promise_resolve`/`__make_callback`/`Promise_then2`) are pre-registered for
  arrow-only modules (else `emitAsyncStateMachine` bails to the legacy path).

### `__self` risk (banked as the real hazard) — did not bite the canonical shape
The CPS emitter only captures live-after-await locals by name; the canonical
single-tail-await (`return await P`) has an empty live-after set, so the lifted
`__self` env param (param 0) is never spilled. Richer closure shapes with locals
live across the await route through the same predicate gate and stay on the legacy
path — the `envParam` threading is phase-3 scope.

### Tests / blast radius
- `tests/issue-2957.test.ts` — 7 host-lane cases (concise/block arrows,
  named/anon fn-exprs, bind + suffix, await-less legacy control), all resolving to
  the awaited value.
- Green: `async-await`, `issue-1042`, `issue-1042-host-drive`, `async-census`
  (52), generators + async-gen (`generators`, `issue-1672`, `issue-2611`,
  `generator-iife`, 30). `tsc --noEmit` clean.
- Pre-existing failures on clean `origin/main` (NOT introduced): `issue-2865`
  (2 WASI decl cases), `issue-1712` / `illegal-cast-closures-585` (6 non-async
  closure cases).

### Scope
Phase 3 (class methods `class-bodies.ts` + object-literal methods `literals.ts`)
remains a follow-up; the issue stays `in-progress`.


ttraenkler:
Removed the auto-park `hold`. Diagnosed the cited merge_group failure (run 28712630336): the PR is a **net +4 pass** (33610→33614, 37 improvements vs 33 regressions), and the gate failed only on the regression *ratio* (89.2%). The 33 'regressions' are a null_deref cluster (bucket signature `975c50163fe3cf04`) that the gate itself flags as **likely baseline drift** — "js2wasm-baselines JSONL is 2 test262-relevant commit(s) behind main HEAD" + "#1081 merge-base cache MISS … drift attribution may be imprecise" (same drift signature that cleared #2631/#2445 this session). Merged current `origin/main` (532358fce) to refresh the stale baseline; re-validating once via a fresh merge_group. If real regressions survive against the current baseline it'll re-park.

ttraenkler:
Re-validated on the fresh base (merge-queue commit 7c3cee75) — this time the baseline is **CONTENT-CURRENT (0 test262-relevant commits behind)**, so the earlier drift caveat no longer applies. Result: **Net +13 pass (33610→33623)** but **23 regressions**, failing the ratio gate (63.9% = 23/36 ≥ 10% limit). The regression count fell 33→23 and the bucket signature changed (975c50163→f6b75983ecadf03a) as the baseline caught up, so some of the original 33 were drift — but ~23 survive against a current baseline and read as **real regressions** from the phase-2 async-state-machine activation. This needs the author to either fix the 23 regressing tests or make a deliberate net-positive ratio-override call; I'm leaving it held rather than re-admitting (park-rule loop hazard, and these are real on a current baseline).

ttraenkler:
Localized the 23 remaining regressions (diffed the merge_group merged JSONL vs the fetched baseline — they're a clean, coherent async cluster):

- **16× `built-ins/Array/fromAsync/`** (returns-promise, mapfn-{sync,async}-throws, sync/async-iterable-*-err, asyncitems-*, thenable-* …)
- **7× `language/statements/await-using/`** (use-before-initialization {global,block-local,function-local}, invalid-assignment-body-for-of …)
- **1× `built-ins/AsyncFromSyncIteratorPrototype/throw/throw-null.js`**

All 22 null_deref + 1 assertion_fail. Category (js-host lane): net **+13** (36 improvements / 23 regressions), so the ratio gate (63.9%) is what parks it, not net.

The `planAsyncClosureActivation` `lane === 'cps'` gate you added (33→23) is working — the residue is **single-tail-await CPS async *function-expression* callbacks in the async-iteration harness** (`asyncTest(async function(){…})` under fromAsync / await-using) that still null_deref in the **lifted-closure context** — exactly the 'closure-context validation' follow-up your comment flags (async-activation.ts L174-186). I didn't push a fix: the only mechanical narrowing I found that clears all 23 (arrows-only, dropping fn-expr activation) would unilaterally cut this PR's stated scope, which is your call. Handing you the exact failing set so the closure-context CPS fix can target it. (Diagnostics only — no changes pushed; branch is at your HEAD + a clean main-merge.)

ttraenkler:
Pushed a fix (4d8dccb1e): scoped async **closure** activation to `ts.isArrowFunction` only. The 23 regressions were all async **function-expression** closures (the async-iteration harness's `asyncTest(async function(){…})` / `assert.throwsAsync(_, async function(){…})` callbacks) null_deref'ing in the lifted-closure CPS context. Arrow async closures — the phase-2 headline — stay activated; fn-expr closures fall back to their prior correct legacy path until the closure-context `__self`/capture CPS is validated for that shape (follow-up). Verified locally the await-using cluster returns to pass + arrow async still compiles; merge_group regression gate validates the full 23 at scale. This narrows the PR to arrows-only for now — flagging since the title says 'arrows + function expressions'; the fn-expr half is a clean follow-up.

PR #2652, opened 2026-07-04
fix(#2175 V2-S3a): raw-anyref carrier — GC-object identity across tag-5/tag-6

ttraenkler:
## #2175 V2-S3a — the raw-anyref carrier (stacked on #2650 / V2-S2)

Flips the banked V2-S2 characterization guard and fixes a broad #3027 identity class by reconciling GC-object reference identity across the two `$AnyValue` representations of the same reference.

### Root cause
`emitStrictEq` boxes both `any` operands to `$AnyValue` → `__any_strict_eq`. A GC object reaches `===` under **two reps of the same reference**:
- raw GC ref (e.g. `RegExp.prototype.exec`) → `__any_box_ref` → **tag-6 refval**;
- externref-wrapped ref that `__extern_get` returns (descriptor `.value`, array element, any dynamic member read) → `__any_box_string` → **tag-5 externval**.

`__any_strict_eq`'s `tagA != tagB → 0` gate answers 0 for that cross-representation pair even though it is the identical object — the wall behind `gOPD(p,"exec").value === p.exec` and the broad `const o:any={z:1}; [o,o]; a[0]===a[1]` → 0 class.

### Fix
A reference-identity reconciliation arm (after the numeric-class arm, before the tag-mismatch gate): recover each operand's ref payload to a common `eqref` (`refval`, else `any.convert_extern(externval)`) and `ref.eq` them — the #2734 `__extern_strict_eq` discipline, now on the whole `any === any` surface. **Standalone/wasi-gated** (native-GC phenomenon; host byte-identical). Cannot false-positive: `ref.eq` is exact identity; distinct boxes fall through unchanged.

### Proof (host-free, inject/contrast + anti-vacuity)
- `gOPD(RegExp.prototype,"exec").value === RegExp.prototype.exec`: **0 → 1** (banked guard flipped to `.toBe(1)`)
- `const o:any={z:1}; [o,o]; a[0]===a[1]`: **0 → 1** (#3027 identity class)
- Discriminates: distinct objects → 0; swap-guard `value === test` → 0; content-eq strings → 1; distinct strings → 0; `23===23.0` → 1; `NaN===NaN` → 0; `null===null` → 1

### Scope
This is **V2-S3a** (the carrier). The `$NativeProto` reader-arm MOP (`$props` table + `ensure_props` + step-3 arms across the 7 reader natives — the #3027 flowing-proto driver) is **banked for V2-S3b** with a full implementation note in the issue file, keeping the reader-arm blast radius on its own CI evidence per the v2 spec.

### Tests
`tests/issue-2175-v2s2-singleton-identity.test.ts` 8/8 (guard flip + 2 anti-vacuity cases). Isolated regression sweep across equality/identity/class/dstr suites green; `tsc` clean.

Stacked on #2650 — do not enqueue until #2650 lands, then re-merge main.

### DIFF
--- a/plan/issues/2175-standalone-builtin-prototype-readers.md
+++ b/plan/issues/2175-standalone-builtin-prototype-readers.md
@@ -1262,20 +1262,22 @@ Status: implemented, host-free, **standalone/wasi-gated (host byte-identical)**.
 ### The senior-dev scoping call (WHY this is S3a, not the full C3)
 
 V2-S3 (C3) is two genuinely separable blast radii: **(a)** the raw-anyref
-carrier that reconciles GC-object identity across representations — this is
-what flips the banked `.toBe(0)` guard and fixes a broad #3027 identity class —
-and **(b)** the `$NativeProto` reader-arm MOP (`$props` table + populate +
-`ensure_props` + step-3/4 arms across the 7 reader natives) that makes a proto
-object *flowing as a runtime value* answer reflective reads. The v2 spec itself
-mandates keeping the reader-arm blast radius on its own CI evidence
-("Do not fold S3+S5 into one PR"). The carrier (a) is small, provably safe, and
-delivers the explicitly-requested acceptance signal; the reader arm (b) is a
-large object-runtime change. Landing (a) alone as a tight, well-proven slice —
-and **banking (b)** with the note below — is the disciplined call over one
-sprawling PR that conflates two minefields (equality machinery + reader natives)
-in a single CI signal. The equality machinery is the codebase's most
-regression-prone area (documented −162/−788/−794/−1245 incidents in
-`any-helpers.ts`), so it earns its own isolated evidence.
+carrier that reconciles GC-object identity across *cross-representation*
+(different-tag) pairs, and **(b)** the `$NativeProto` reader-arm MOP (`$props`
+table + populate + `ensure_props` + step-3/4 arms across the 7 reader natives)
+that makes a proto object *flowing as a runtime value* answer reflective reads
+AND returns the actual GC singleton ref (which is what flips the identity guard
+— see "does NOT flip the guard" below). The v2 spec itself mandates keeping the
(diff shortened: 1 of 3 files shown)


ttraenkler:
Removed the auto-park `hold` and re-enqueued (one-shot). Diagnosed run 28715220981: the failing job was **merge shard reports → standalone-highwater floor** (host-free pass 13724 vs mark 20952, −7228), NOT a pass→fail regression (that check showed 0 regressions). That run was on the PR's **stale base** (27bfb9417), predating the floor-mark commit 265a26fc (set 17:41Z); #2646's runs on fresher bases did not breach the floor, so this reads as a stale-base measurement artifact. I merged current `origin/main` (c1e0ca875, PR-level checks green) and am re-validating once. **If the standalone floor re-breaches on this fresh base, it's a real host-free regression from the raw-anyref `__any_strict_eq` carrier change** (delicate equality substrate) and needs a code fix on the branch — it will re-park and I'll escalate rather than re-admit again.

ttraenkler:
Fixed the standalone floor regression (−7228 host-free). **Root cause** (ground-truthed with a 300-file real-test262 standalone batch): the V2-S3 ref-identity arm ran *before* the `tagA != tagB` gate — for ALL tag pairs — and produced FALSE POSITIVES on same-tag comparisons (wrongly-equal), invisible to the js-host regression lane (0 regressions there) but caught by merge_group's standalone floor. Batch: original arm 109 pass / no-arm 140 / **this fix 140 (0 regressions vs no-arm, −36 false positives eliminated)**. **Fix** (8ec9a25e0): the arm's genuine purpose is CROSS-REPRESENTATION identity (tag-6 refval vs tag-5 externval of the same object, e.g. `gOPD(p,'exec').value === p.exec`), which lands in the `tagA != tagB` arm — so I moved it there (replacing the legacy flat 0). Same-tag pairs keep their existing correct arms; cross-rep identity is preserved. Re-validating via merge_group once PR checks are green.

PR #2661, opened 2026-07-04
fix(#2175 V2-S3b): D4 raw-anyref carrier — strict-eq object identity through externref reads

ttraenkler:
## #2175 V2-S3b path 2 — the D4 raw-anyref carrier (strict-eq operand form)

The WAT-resolved keystone guard flip, delivered as a tight, provably-safe slice.

### Root cause (WAT-confirmed)
A GC object read back through the externref boundary — a descriptor `.value`, an array element, `const b = a` — boxes via `__any_box_string` → **tag-5**. Under `===` both operands become tag-5 and land in `__any_strict_eq`'s **same-tag-5 CONTENT-eq** arm (`ref.test $AnyString`-guarded, never `ref.eq`), so `gOPD(p,"exec").value === p.exec` and `[o,o]; a[0]===a[1]` answered **0** even though both externrefs wrap the IDENTICAL reference. (Contrast `exec === exec`, which stays on the raw tag-6 GC-ref inline `===` path with a `ref.eq` arm → 1.)

> Note: on a clean build of the branch point, `tests/issue-2175-v2s2-singleton-identity.test.ts` was **2/8 FAILING** — the earlier V2-S3a "landed" self-report was the coincidental near-miss this issue keeps warning about (issue-tests aren't in required CI, #3008). Verified first, then fixed.

### Fix — PATH 2 at the strict-eq operand site (not a signature change)
New `__any_box_eq_operand` (any-helpers.ts, **standalone/wasi only**) boxes a genuine GC reference object as **tag-6** (`refval` identity) and delegates every primitive (string/number/boolean/bigint), null, and already-boxed `$AnyValue` to the **byte-identical** `__any_box_string`. Wired ONLY into `emitAnyEqOperands` (coercion-engine) for **strict** `===`/`!==` on an `externref` operand. Both object operands then hit the **EXISTING tag-6 `ref.eq` arm** — the guard flips with **NO equality-arm change**.

Rejected alternatives: re-typing `__extern_get` externref→anyref (it's a host IMPORT with a fixed ABI, ~76 sites); flipping `honestAnyBoxing` globally (the documented −788/−794 landmine). This surface is ~100× smaller — the boxed operand's only consumer is `__any_strict_eq`.

### Why it can only fix strict `===`, never regress
Only a residual eq-castable **reference** flips to tag-6; every value/content-compared primitive is explicitly excluded and keeps its exact tag-5 box. So the ONLY behavioural change is **same-object-via-two-reads: 0 → 1** (a correctness gain; distinct objects stay `ref.eq` 0). Aligns the boxed-operand path with the #2734 `__extern_strict_eq` object-identity fast path (closing an inconsistency, not creating one). Loose `==` untouched; host/GC byte-inert.

### Proofs (host-free, 0 `env` imports; anti-vacuity throughout)
- `tests/issue-2175-v2s2-singleton-identity.test.ts`: **8/8** (was 2/8).
- `tests/issue-2175-v2s3b-carrier.test.ts`: **10/10** — every identity `→1` paired with a swap/distinct `→0`; bigint/string/number/null preserved.
- ~300 equality/identity/standalone regression tests green; `tsc --noEmit` clean.

### Merge discipline
**FULL `merge_group` standalone `host_free_pass` floor validation is REQUIRED** — path 2 changes the strict-eq value-read representation, and standalone-only floor moves are invisible to PR-lane checks (the V2-S3a −7228 lesson). Expect the floor to move **positive** (the #3027 object-identity class). The reader-arm receiver-classification breadth (the ~1,552-test driver) stays **banked** as its own PR per the "do not fold S3+S5" discipline.


ttraenkler:
Closing per diagnosis (diag-2661): the operand-site tag-6 boxing approach is **fundamentally unsound** against the current object-identity substrate, not a flake.

**Verdict: REAL deterministic merged-baseline regression** (the earlier 'no failed jobs' read was a `gh --json jobs` 30-item truncation artifact — the run has ~114 shards). The 'merge shard reports' aggregator genuinely failed the standalone host-free floor: `pass=20653, floor=20902, delta=-299`. Controlled local A/B on 220 flipped candidates: 220/220 PASS on upstream/main, 220/220 FAIL on this HEAD — 100% deterministic.

**Root cause:** test262's own `assert.sameValue` (`===`) routes both `any`/externref operands through `__any_box_eq_operand`; switching the carrier from tag-5 (externref) to tag-6 loses the externref-layer canonical identity the tag-5 path preserved — objects reached via different read paths are externref-identical but not `ref.eq`-identical (the #2719/#2734 identity-substrate gap), and it mis-applies identity to eq-castable primitive GC reps outside the narrow wrapper-exclusion list. Hits hundreds of unrelated features (String/class/defineProperty/getPrototypeOf/Array) because the harness assertion is the shared path.

**Reverted state == main** (the V2-S2/S3a `.toBe(0)` characterization guards are already landed), so there is nothing green to add here. **Banking V2-S3b's guard-flip + ~1,552-test #3027 goal into V2-S3c** (the $NativeProto reader-arm), where identity is canonicalized AT THE READER so both operands share one canonical carrier and `ref.eq` is applied only to genuinely-canonical GC refs. Object-identity canonicalization (#2719/#2734) is the real prerequisite. The operand-site shortcut should not be retried.

PR #2664, opened 2026-07-04
fix(#3023): resolve '.next is not a function' for wasm-struct iterators in array destructuring

ttraenkler:
## #3023 — iterator protocol: `.next` callability in destructuring (JS-host lane)

Scoped slice of #3023: the **`.next is not a function`** destructuring bucket (recorded 114).

### Root cause
A custom iterable installed via a late-bound `obj[Symbol.iterator] = function () { return { next(){…}, return(){…} } }` lowers the returned iterator object to a **closed nominal WasmGC struct** whose `.next`/`.return` are NOT native JS properties. Array destructuring materializes the source through the host import `__array_from_iter_n` → `_arrayFromIter` → (fall-through) `_drainIterable`, which did a naive `it.next()` and threw `it.next is not a function`. The robust protocol walk (resolve members via native → sidecar → `__sget_*` → wasm-closure `__call_fn_0`, with §7.4.6 IteratorClose on the bounded/abrupt stop) existed **only** for the case where `@@iterator` itself is a wasm closure — not the common case where a native-function `@@iterator` *returns* a wasm-struct iterator.

### Fix
All in `src/runtime.ts` (JS-host runtime — **no codegen change, zero compile-time impact**):
- Extract the robust walk into a shared `_walkWasmIterator(iteratorObj, limit)` helper (bounded materialization + IteratorClose-on-cap; `limit === Infinity` drains to natural done without closing).
- Route the wasm-closure-`@@iterator` path through the helper (behaviour-preserving refactor).
- `_drainIterable` diverts a wasm-struct iterator (`typeof it.next !== "function"`) to the helper; plain JS iterators keep the existing naive fast path.

### Measured
Default-lane in-process harness over the 114 `.next is not a function` files + a 60-file currently-passing dstr/iter regression sample (174 total): **46 → 118 pass, +72 newly passing, 0 regressions**. Within the 114-file bucket, **68 flip to pass** (114 → ~46) — materially below the recorded 113, acceptance criterion met.

### Tests
- `tests/issue-3023.test.ts` — 5 cases (close-on-not-done, no-close-on-done, multi-element, for-of inner destructuring, assignment destructuring), all pass.
- Broad regression sweep: **83/83** across 13 iterator/spread/destructuring suites. `tsc --noEmit` + `biome lint` clean.

### Out of scope (distinct paths, already failing, none regressed)
Default-parameter destructuring over a custom iterable; the rest-slice binding codegen (`[a, ...rest]` — values now materialize correctly, but the rest slice drops them); generator-`yield`-inside-destructuring + §7.4.6-step-9; the wider for-of/for-await abrupt-completion residual. See the issue file's Resolution section.


ttraenkler:
**Disposition (tech-lead): deferred, held — blocked on residual async iterclose codegen beyond #3038.**

Re-validated: refreshed this branch onto current `main` (which includes #3038's async iterator-close closure-capture fix) and re-ran the merge_group. It **still** fails `check for test262 regressions`, so #3038 did not fully cover the async param-destructuring cases this PR exposes.

Context (per the earlier deep diagnosis): the runtime `_drainIterable` fix here is correct — it removes a spurious `.next is not a function` throw. The remaining regressions are pre-existing async-generator / for-await param-destructuring `IteratorClose` observability cases that were **false-passing** on `main` (the spurious throw aborted them and the async runner recorded the aborted run as `pass`). Landing this fix runs them for real and exposes the residual codegen gap.

Keeping the `hold` — merging now would regress those tests past the 10% gate. This PR is ready to land once the residual async iterclose observability is fixed (a follow-up to #3038 / #3023). Not a sprint blocker; tracked here.

ttraenkler:
Holding for Fable: this PR's +68 is blocked on the last 2 iter-close files, which need #3040 (custom-iterable param-default destructure — two coupled defects, broad-impact, Fable-reserved with a complete spec). Do not enqueue until #3040 lands; then un-hold and re-validate. (Lead decision: option-1 hold-for-Fable rather than a targeted vacuity excusal.)

PR #2684, opened 2026-07-05
fix(#3036): deref boxed transitively-captured var in method/accessor bodies

ttraenkler:
## #3036 — boxed transitively-captured var deref in method/accessor bodies

Object/class **method shorthand**, **class methods**, and **class get/set
accessors** that read or write a variable captured **transitively** from a
grandparent scope emitted garbage when that variable was **boxed** (a sibling
closure mutates it → it lives in a `struct (field $value (mut T))` ref cell).
`promoteAccessorCapturesToGlobals` value-promoted the **box** into a captured
global, but the read/write sites never deref'd it: reads yielded the f64/ref
default (0/NaN/null); writes nulled the box global.

### Fix (additive, low-risk)
Boxed captures are registered in `ctx.capturedBoxGlobals` **with the inner
`valType`** *in addition to* the existing `capturedGlobals` entry (kept
byte-identical, so every unmodified consumer is unchanged). The scalar
read/write sites (`identifiers.ts`, `assignment.ts`, `unary-updates.ts`)
consult `capturedBoxGlobals` **first** and deref (`global.get; struct.get/set
field 0`). Also fixes a latent `fixupModuleGlobalIndices` staleness (the
object-valued `capturedBoxGlobals` entries were never shifted — shared by the
pre-existing transitive-fn box).

Full design + file list + verification in
`plan/issues/3036-accessor-method-boxed-transitive-capture-write.md`.

### Verification
- All #3036 read/write/increment repros flip garbage→correct (object-method,
  class-method, method-shorthand writes; `c++`/`--c`; static-dispatch getter
  read →5; class-method read via `any` →5; iterator-close `return(){}` →1).
- **12/12 method-shorthand `for-await-of/*ary-init-iter-close*` test262 files:
  0 pass on base (`3844b8a`) → 12 pass here.** My fix is what flips them.
- Full `tests/equivalence/` (1638 tests): **identical failure set vs base**
  (15 files / 36 tests fail on both, 1599 pass) — zero regressions. `tsc` clean.

### ⚠️ Stacking + why DRAFT (needs lead sequencing)
This branch is **predecessor-stacked**: it contains #2664 (`#3023`, currently
**held**) + #3035 (bug #1) + this #3036 fix. Draft because:
1. It bundles the **held** PR #2664 — un-holding is a lead/coordination call.
2. The full stack still has **2 remaining regressions**, a **separate bug**
   (NOT #3036): the async-generator **expression** `dflt` variants
   (`dflt-ary-init-iter-close`, `named-dflt-ary-init-iter-close`) fail with
   "Cannot destructure null" — a **param-default-not-applied** bug
   (`async function*([x] = iter)` ignores the `= iter` default). These are
   fn-expr (not method-shorthand) and outside #3036's scope; they would park a
   bundled `merge_group`.

**#3036 itself is complete and verified.** The un-hold of #2664 needs the
param-default bug resolved (new issue) or a vacuity-excuse decision first.

### DIFF
--- a/plan/issues/3023-iterator-protocol-forof-abrupt-completion-residual.md
+++ b/plan/issues/3023-iterator-protocol-forof-abrupt-completion-residual.md
@@ -1,10 +1,11 @@
 ---
 id: 3023
 title: "iterator protocol: synthesized-iterator .next callability + for-of/for-await abrupt-completion residual (~508 default-lane fails)"
-status: ready
+status: done
 sprint: current
 created: 2026-07-03
-updated: 2026-07-03
+updated: 2026-07-04
+completed: 2026-07-04
 priority: high
 horizon: m
 feasibility: medium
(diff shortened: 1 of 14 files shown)


ttraenkler:
Superseded by the un-stacked standalone PRs per lead sequencing: #2685 (fix(#3038) bug #1, off main) and #2686 (fix(#3039) bug #2, stacked on #3038). Both renumbered from 3035/3036 (cross-session id collision with already-merged unrelated issues). #2664 (#3023) stays held on its own PR until its remaining param-default regression is resolved; the iterator-close win lands when #2664 goes on top of this stack.

PR #2686, opened 2026-07-05
fix(#3039): deref boxed transitively-captured var in method/accessor bodies (bug #2)

ttraenkler:
## #3039 — deref boxed transitively-captured var in method/accessor bodies (bug #2)

Object/class **method shorthand**, **class methods**, and **class get/set
accessors** that read or write a variable captured **transitively** from a
grandparent scope emitted garbage when that variable was **boxed** (a sibling
closure mutates it → ref cell). `promoteAccessorCapturesToGlobals` value-promoted
the **box** into a captured global, but the read/write sites never deref'd it:
reads yielded the f64/ref default (0/NaN/null); writes nulled the box global.

### Fix (additive, low-risk)
Boxed captures are registered in `ctx.capturedBoxGlobals` **with the inner
`valType`** *in addition to* the existing `capturedGlobals` entry (kept
byte-identical → every unmodified consumer unchanged). The scalar read/write
sites (`identifiers.ts`, `assignment.ts` simple `=`/compound/logical/destructuring,
`unary-updates.ts` `++`/`--`) consult `capturedBoxGlobals` **first** and deref
(`global.get; struct.get/struct.set field 0`). Also fixes a latent
`fixupModuleGlobalIndices` staleness (object-valued `capturedBoxGlobals` entries
were never index-shifted — shared by the pre-existing transitive-fn box). Full
design in `plan/issues/3039-accessor-method-boxed-transitive-capture-write.md`.

### Stacking / landing
- **Stacked on #3038** (bug #1). Renumbered from 3036 (cross-session id collision
  with the already-merged `3036-standalone-allsettled-nullderef`).
- **Direct** capture read/write/increment cases pass without #2664
  (`tests/issue-3039.test.ts`, 7/7 green). Zero `tests/equivalence/` regressions
  vs base (identical failure set, 1638 tests).
- The **iterator-close win** — 12/12 method-shorthand `for-await-of/*ary-init-iter-close*`
  test262 files (0/12 on base → 12/12) — is delivered once **#2664 (#3023)** lands
  on top of this stack to expose the JS-host `return()` close. #2664 stays held
  until its remaining param-default regression (a separate bug) is resolved.

### Out of scope (separate pre-existing bug, tracked separately)
A getter read via an `any`-typed receiver on a class declared inside a function
still returns NaN — a distinct dynamic-accessor-dispatch bug (a *constant* getter
returns NaN through that exact shape too). My additive branches are no-ops for
non-boxed names, so that codegen is byte-identical. Filed as a follow-up.

### DIFF
--- a/plan/issues/3039-accessor-method-boxed-transitive-capture-write.md
+++ b/plan/issues/3039-accessor-method-boxed-transitive-capture-write.md
@@ -1,0 +1,208 @@
+---
+id: 3039
+title: "object/class methods & accessors can't read/write a BOXED transitively-captured outer variable (emit garbage)"
+status: done
+sprint: current
+created: 2026-07-04
+updated: 2026-07-05
+completed: 2026-07-05
+assignee: ttraenkler/dev-3036
+priority: high
+horizon: l
+feasibility: hard
+reasoning_effort: max
+task_type: bugfix
+area: codegen
+language_feature: closures, object-literal methods, class methods, accessors, iterators
+goal: spec-completeness
+owner: fable
+related: [3038, 3023, 2664, 2029, 2669]
+architect_spec: candidate
+---
+
+# #3039 — [FABLE-RESERVED] accessor/method write to a BOXED transitively-captured var emits garbage
+
+Split from #3038. This is the **broad-impact** half; validate via full CI /
+`merge_group`, never a scoped sweep ([[project_broad_impact_validate_full_ci]]).
+
+## Symptom
+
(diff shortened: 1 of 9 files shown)


ttraenkler:
Converted to draft to gate merge ordering: this PR is stacked on #2685 (fix(#3038)) and must land AFTER it. Once #2685 is on main I'll `git merge origin/main` here (so the diff reduces to just #3039) and mark ready → auto-enqueue picks it up. #3039's own required checks are green after the oracle-ratchet fix.

PR #2719, opened 2026-07-05
fix(#3045): materialize class-expression constructor value into its binding

ttraenkler:
## Summary (partial fix for #3045 — does NOT close it)

A class-**expression** binding (`const/var C = class { ... }`) must hold the
constructor-object VALUE so reading `C` as an rvalue works. The local
variable-declaration compiler **skipped** class-expression initializers
(`if (isClassExpression(decl.initializer)) continue;` — "already handled as
class declaration"), so the pre-hoisted (instance-struct-typed) local `$C` was
declared but **never stored**. Reading `C` as a value — passing it to a
function, `Reflect.has(C, k)`, `Object.prototype.hasOwnProperty.call(C, k)` —
read an uninitialized null local, and coercing null to externref for a host
import threw `TypeError: Reflect.has called on non-object` /
`Cannot convert undefined or null to object`.

**Fix:** route class-expression initializers through the same "compile
initializer → re-type the pre-hoisted slot to the closure type → store" path
already used for arrow/function-expression bindings
(`src/codegen/statements/variables.ts`). `new C()` is unaffected — it resolves
the class statically via `classSet`, not the binding value.

## Scope — this is Bug 1 of 2 in #3045

The 8 harvested private-method files ALSO need a **separate, deeper** fix:
class-expression method/constructor bodies do not capture the enclosing
function's scope the way class *declarations* do (tied to the #779a
captured-global machinery — senior/architectural depth). That is documented in
the issue as **Bug 2**; #3045 stays `blocked` on it. This PR lands the clean,
self-contained value-materialization fix (Bug 1), which fixes the
class-expression-value-as-rvalue trap independently of the private-method tests.

## Validation

- `tests/issue-3045.test.ts` — 9 cases (Reflect.has / hasOwnProperty.call /
  pass-to-fn on a class-expression value now work; new / method / ctor-arg /
  static / two-instance still correct). All pass.
- Regression-free: the class vitest suites are 18=18 with/without this change
  (all pre-broken by an unrelated `string_constants` harness issue); the
  pre-existing `instanceof` / `extends`-a-class-expression failures are unchanged
  (fail-before == fail-after — they are additional class-expression bugs folded
  into Bug 2).
- `tsc --noEmit`, `prettier --check`, `biome lint` clean.

Refs #3045 (partial).


ttraenkler:
Diagnosed the merge_group park (run 28731994537, `check for test262 regressions`): **net -3** (net_per_test -2, 0 improvements / 2 real regressions; the 3rd, `BigInt/constructor-empty-string` → compile_timeout, is a load flake). Reproduced locally — both fail on this branch, **pass on main** with `variables.ts` reverted:

- `language/expressions/class/scope-name-lex-close.js` (assert #1 L20 — class inner lexical self-name closure)
- `language/expressions/class/elements/private-static-field-shadowed-by-field-on-nested-class.js` (assert #3 L87)

**Mechanism:** routing class *expressions* through `compileExpression` (to materialize the binding value, the #3045 fix) re-processes the class's lexical self-name + private-static-field scope, which regresses those 2 cases. This is the partial-#3045 class-codegen work you flagged ('does NOT close it') — leaving it held rather than merging, since as-is it's a net **-2** conformance regression. The binding-value materialization needs to avoid re-binding the class's inner scope name / private fields (or guard the new branch to the rvalue-read case only).

ttraenkler:
Attempted a narrowing fix: gating the class-expression binding-materialization to **anonymous** class expressions (`!expr.name`) clears the `scope-name-lex-close` regression (named class exprs keep their legacy inner-name path) and preserves the #3045 anon rvalue-read. But it does **not** clear `private-static-field-shadowed-by-field-on-nested-class` — that test's outer class is *anonymous* (`var C = class { static B = class { …#priv… } }`), so the materialization still fires and disturbs the **nested** class's private-static-field resolution — an indirect interaction that needs the ctor-value store to not perturb nested private-field brand state. So a simple gate only halves it (−2 → −1); it's still net-negative. This is the class-codegen completion you flagged ('does NOT close it') — leaving it held rather than merging a conformance regression.

ttraenkler:
Second fix attempt (recursive private-member + name exclusion): it clears **both** test262 regressions AND the anon #3045 case — BUT it breaks **2 of this PR's own #3045 tests** (`named class expression value is materialized`, `var form materializes the value`). So the exclusion is a dead end: the #3045 feature **requires** materializing named / private-member class expressions, which is exactly what regresses `scope-name-lex-close` + the nested private-static-field case. The regression and the feature are in **direct conflict** under the current `compileExpression`-into-outer-slot approach — the store/re-type of the outer binding perturbs the class's inner-name binding and nested private-brand resolution. The real fix must materialise the ctor value WITHOUT touching the inner-name/private-field scope (e.g. materialise into a distinct slot, or emit the ctor closure ref without the slot re-type that aliases the inner binding). That's the class-codegen completion this PR flags as open — not resolvable by gating. The server-side regression gate hard-blocks the PR until net ≥ 0, so it genuinely can't merge until that lands. Full diagnosis + both attempts documented above.

ttraenkler:
**Root-cause traced to two distinct substrate mechanisms** (reproduced both minimally):

**(1) `scope-name-lex-close` — closure identity.** For `const cls = class C { m() { return C; } }`: on main `inst.m() === cls` is 1, on this branch it's **0**. `emitFuncRefAsClosure` (closures.ts:3919) memoizes the closure instance per activation *only when the fn has captures* (#2976's `Constructor === Constructor` fix); the **no-capture** class ctor falls through to `ref.func + struct.new` which builds a **fresh** struct every reference site. So materialising `cls` and the inner-name `C` reference mint **different** closure instances → not `===`. **Fix direction:** extend the #2976 per-activation memo (`fctx.nestedFnClosureMemos` / `nestedFnClosureArtifacts`) to the no-capture path so ctor identity holds regardless of the binding-materialisation.

**(2) `private-static-field-shadowed` (anonymous outer + nested private class) — private-brand resolution**, a separate mechanism (no inner-name self-reference involved).

Both live on the closure/object-identity + private-brand substrates (#2976/#2534/#2652/#3045) that are high-blast-radius — extending closure memoization broadly or touching private-brand emission risks new regressions across the class surface. This is the #3045 completion the PR flags as open ('does NOT close it'). I've verified 3 narrowing gates all fail (they break this PR's own #3045 tests, since the feature *requires* materialising named/private-member class exprs). The server regression gate hard-blocks merge until net ≥ 0. Handing off with both mechanisms + fix directions pinned.

ttraenkler:
**Implemented + tested the traced mechanism-#1 fix — it does NOT resolve it** (important negative result). I extended the #2976 per-activation closure memoization to the no-capture class-ctor path in `emitFuncRefAsClosure` (cache trampoline by funcName, memoize instance). Result: the PR's own 9 #3045 tests still pass, but `inner C === cls` is **still 0** and `scope-name-lex-close` still fails. So the mechanism is deeper than 'no-capture closure isn't memoized': the **inner class-name `C` reference and the materialized outer binding `cls` resolve via separate paths**, and unifying the closure-emission path doesn't align them. Fixing it requires reworking the inner class-name resolution to reference the same materialized value the binding holds — core #3045 class-scope architecture, not a targeted memoization. Combined with the still-untraced mechanism #2 (private-brand distinctness), both blockers require your #3045 domain work. I've reverted the non-working attempt; nothing pushed. Full evidence (both mechanisms + this failed fix) is in the thread.

ttraenkler:
**Re-verified against freshest main (post-compaction check).** #2719 was 25 commits behind upstream/main, so I re-merged latest main into its HEAD and re-ran both regressing tests: `scope-name-lex-close` and `private-static-field-shadowed-by-field-on-nested-class` **both still FAIL**. So the −3 regression is **not** collateral that main's advance resolved — it persists on the freshest merged state. That's now triply confirmed (original merge_group park · in-session repros · fresh-main re-merge). The bot park-hold is correct and re-enqueueing would re-fail the merge_group. The PR-level `check for test262 regressions` shows 'pass' only because that run's shards were **skipped** (0s, matrix empty) — the authoritative signal is the merge_group re-validation, which fails. Confirmed: this needs the #3045 inner-name-resolution + private-brand-distinctness fix on the branch before it can merge. Nothing pushed.

ttraenkler:
**Root cause definitively pinned (3rd coded attempt + debug trace).** I tried materializing the binding as the class's `__class_<Name>` singleton (`emitLazyClassObjectGet`) instead of the closure ctor — the PR's 9 #3045 tests still pass, but `inner C === cls` is **still 0**. Debug trace shows why: for `const cls = class C {}` there are **TWO** class-object singletons registered — one under the **variable name** `cls`, one under the **synthetic class name** `__anonClass_C_0`. The inner class-name `C` and the outer binding `cls` resolve through **different synthetic names → distinct `__class_` singletons**, so they can never be `===`. This is precisely the aliasing hazard already flagged in `identifiers.ts:767` (`__class___anonClass_0` vs `__class_C` are distinct singletons and break identity). So both regressions trace to the **class-object-singleton NAME-ALIASING** subsystem (#1395/#3045), not the materialization site: the fix must **unify the class-name→singleton map** so inner-name, binding-name, and `C.staticProp` reads all resolve to ONE singleton. That's core class-object-identity architecture — the #3045 completion this PR flags as open. Three targeted fixes (closure memoization, anon/private gating, singleton-materialization) all fail because the divergence is upstream in name-resolution. Nothing pushed; full evidence in-thread.

PR #2721, opened 2026-07-05
feat(#3051): IR class.call — void instance method in statement position

ttraenkler:
## #3051 — IR `class.call`: void instance method in statement position

Banked in #3000-C's Implementation Notes. A class body (constructor **or** plain method) calling a **void** instance method as a **statement** (`this.add(x);` / `obj.tick();`) demoted the whole caller **post-claim** with `void method <Class>.<method> used in expression position`, falling back to legacy (byte-inert, no IR emission). Verified pre-existing on `upstream/main` — affects plain methods, not just constructors.

### Root cause
The selector already CLAIMS this shape (statement-expression method call — the same path non-void statement calls take), so it reaches `from-ast`. The class-method arm of `lowerMethodCall` (`src/ir/from-ast.ts`) then threw **unconditionally** for a void method, ignoring the `statementPosition` flag the function already receives — unlike the already-correct `super.method()` and extern-class arms, which gate the throw on `!statementPosition`.

### Fix
Two-line guard change mirroring the sibling arms — no new IR instr. The `class.call` instr, `emitClassCall` builder, and `class.call` lowering already carry a null result through: a void method's Wasm slot leaves nothing on the operand stack, and `emitBlockBody`'s `result === null` in-place path emits it balanced (no drop). Non-void discarded results were already handled by the `useCount === 0 && isSideEffecting` emit+drop path.

### Proof
- **Genuine emission (irCompiledFuncs telemetry)**: `Counter_run` + chained `Counter_tickTwice` go from ABSENT+demoted → PRESENT with zero `irPostClaimErrors`, in **both** lanes (host externref + native `$AnyString`).
- **Runtime parity**: chained void calls `run()→tickTwice()→add()` round-trip (`10+5+3=18`); non-`this` receiver also emits + runs.
- **Byte-inert**: identical wasm sha256 base vs fix for non-void programs (change is purely additive gating; typeIdx-parity guard still gates every slot overwrite → worst case clean legacy fallback, never miscompile).
- **Guard preserved**: void in **expression** position still cleanly demotes (asserted).
- **Blast radius**: full class equivalence suite (18 files, both lanes) — identical 61 pre-existing failures base vs fix (raw-instantiate `string_constants`/`wasm:js-string` harness noise + #582 struct.new test, both #3034-class harness issues), only delta is +5 passes = `tests/issue-3051.test.ts`.
- **Corpus**: `check:ir-fallbacks` unchanged (`class-method` already 0 from #3000-E; this shape isn't in the playground corpus).

Files: `src/ir/from-ast.ts` (2-line guard), `tests/issue-3051.test.ts`, `plan/issues/3051-*.md`.


ttraenkler:
Re-admitting once: this PR is byte-inert (dev-proven sha256-identical base-vs-fix for all non-void programs; only +5 NEW tests), so it cannot cause a standalone host_free_pass floor regression by construction. The merge_group batch it parked in shows churn (same queue SHA 25060393b7 has both a failure and a later success run) → collateral/transient, not a #2721 regression. Single re-enqueue per the auto-park protocol's confirmed-collateral clause. If it parks again, I'll diagnose deeper.

ttraenkler:
Removed the auto-park `hold`. The merge_group `quality` failure was the #1616 issue-integrity gate: **1 duplicate id** — main merged in `3051-regexp-symbol-replace-split-coercion-protocol.md`, colliding with this branch's `3051-ir-classcall` file (the #2531 cross-session id race; both allocated 3051 off different mains). Not a code regression. Renumbered this branch's issue to **3052** (via claim-issue --allocate): renamed the issue file + test, updated frontmatter/heading + the classcall #3051 refs. `check:issue-ids:against-main` and the integrity gate now pass. Re-validating via merge_group.

PR #2738, opened 2026-07-05
feat(#3054 B3): proto-method write-through on TypedArray view receivers

ttraenkler:
## #3054 B3 — proto-method WRITE-THROUGH on TypedArray/DataView view receivers

Closes the last shared-backing-view caveat from B1's Option-A de-view. A mutating TypedArray prototype method (`.fill` / `.set` / `.copyWithin`) on a `$__ta_view` receiver now **writes through** to the underlying ArrayBuffer — observable by sibling views and DataViews over the same buffer, per spec.

### Root cause
B1's Option-A (`emitTaViewToVec`) byte-decodes a `$__ta_view` proto-method receiver into a fresh native `$__vec_<elem>` copy so `.fill`/`.sort`/… don't `ref.cast`-trap on the view. That copy was **de-aliased** — a mutating method mutated the copy, and the writes were **lost** (never reached the buffer).

### Fix (write-back copy, not method-rewrites)
New `emitTaViewWriteBack` (the reverse of `emitTaViewToVec`): after a mutating method runs on the copy, each native element is read back to f64 and byte-encoded into `view.buf.data` at `view.byteOffset + i*width` via the same LE `emitWriteBytes` engine `emitTaViewElementSet` uses (bit-exact round trip). Element read-op honors the storage: **packed i8/i16 → `array.get_s`/`_u`; non-packed i32_elem → plain `array.get` + `f64.convert_i32_s/_u`** (`array.get_s/_u` are illegal on a full-width i32 element). Wired in `compileArrayMethodCall`, gated **exactly like the existing module-global write-back** (`MUTATING` set, `result` present); read-only methods (`.includes`/`.indexOf`/…) skip it. Windowed views (B2 `byteOffset`) write the correct absolute bytes. Standalone/WASI lane only (mirrors B1/B2).

Chose write-back over rewriting each method to operate on the view: it confines B3 to the already-de-viewed mutating path (one helper + a gated call), avoiding a rewrite of the whole array-method dispatch.

### Validation (host-enforced — standalone doesn't enforce numeric asserts, #3055/#3056)
- `tests/issue-3054-b3-writethrough.test.ts` — **13 green**: `.fill`/`.set`/`.set`-offset/`.copyWithin` write-through to a sibling; read-only `.includes`/`.indexOf` do **not** clobber the buffer; cross-width Int32 LE / Int16 negative / Float32 round-trip / Uint32 unsigned-i32_elem / Int32 negative-i32_elem; window-offset `.fill` hits the right absolute bytes; subsequent element read reflects the mutation.
- **Byte-inert**: sha256 of the standalone binary **identical** vs `upstream/main` (@ bb239d65b) for 6 controls incl. **plain-array `.fill`/`.sort`** and **native `Uint8Array.fill`** — only `new <TA>(buffer)` de-view mutating programs change bytes (write-back is unreachable otherwise → floor-safe).
- B1 (15) + B2 (22) suites still green (37/37). `tsc --noEmit`, prettier `--check`, `biome lint --diagnostic-level=error` clean.

### Floor
Scoped built-ins/{TypedArray,DataView}: the write-back is on the already-de-viewed path (B1 measured NET 0 there) and is byte-inert off it — expected **NET ≥ 0**. This PR changes emitted behavior for view proto-methods (not fully byte-inert off-path), so it wants the **monitored `merge_group` floor** run.

### Pre-existing native-method gaps (NOT B3 regressions — flagged, separate follow-up)
Verified on this same base with **no views / no B3 code**: native `Uint8Array.sort()` is a no-op, native `Uint8Array.reverse()` leaks a packed `i8` at binary emit, `Uint32Array.fill(>2^31)` saturates to 2^31−1. B3's write-back is downstream and faithfully propagates whatever the method produced, so `.sort`/`.reverse` write-through can't be validated through a broken native method. These packed-TA native-method fixes are a separate pre-existing issue (they affect plain native TAs too).

**C (resizable ArrayBuffer subtype) is independent of B3 and ready to dispatch in parallel.**

### DIFF
--- a/plan/issues/3054-resizable-arraybuffer-dynamic-ctor.md
+++ b/plan/issues/3054-resizable-arraybuffer-dynamic-ctor.md
@@ -2,7 +2,7 @@
 id: 3054
 title: "Resizable ArrayBuffer + dynamic `new <ctorVar>(rab)` — the ~180 codegen gap under #1524"
 status: in-progress
-assignee: ttraenkler/opus-3054-b2
+assignee: ttraenkler/opus-3054-b3
 created: 2026-07-05
 updated: 2026-07-05
 priority: medium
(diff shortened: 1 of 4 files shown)


ttraenkler:
Re-admitting once: the auto-park was a CONFIRMED transient infra flake, not a regression (diag-b3). The failed 'test262 js-host shard 6' died at 'Setup pnpm via Corepack' (registry.npmjs.org network error) → install/build/run all SKIPPED, the shard never executed (no pass→fail delta); every other host+standalone shard passed. Leak hypothesis disproven structurally (dispatch change gated by isTaViewTypeIdx — $__ta_view exists only in standalone; host TAs are host objects → gate false → write-back skipped) AND empirically (8 host array-method programs sha256-identical B3 vs pre-B3). Single monitored re-enqueue per the confirmed-flake clause; merge_group re-validates the standalone floor (NET≥0 expected — off-path except $__ta_view de-view programs).

PR #2748, opened 2026-07-05
fix(#2726): clear mapped-arguments slot on delete arguments[i] (group e)

ttraenkler:
## Summary

#2726 group **(e)** — `delete arguments[i]` on a mapped, configurable index.

A successful `delete arguments[i]` on a mapped index (§10.4.4.5 → OrdinaryDelete) must both **sever** the param↔arguments map AND **remove the slot**, so a later `arguments[i]` read observes `undefined`. The mapped-delete arm only recorded the sever (`unmappedIndices`) and fell through to the generic `__delete_property` path — which reports `true` for the vec-backed arguments object but never clears the backing slot, so the read still returned the original argument.

**Fix:** in the mapped configurable-index arm, write the canonical `undefined` externref into the backing vec slot (null-guarded, mirroring `emitMappedArgParamSync`'s slot write) and emit `true`, short-circuiting the generic path. Front-end only — no new host import; host and standalone lanes agree. The non-configurable arm (returns `false`) is untouched.

## test262 (authoritative `runTest262File`, host/gc lane)

| cluster | baseline → fix |
|---|---|
| `language/expressions/delete` (69) | 64 → **65** pass, **0 regressions** |
| `language/arguments-object/mapped` (43) | 39 → 39 (4 pre-existing defineProperty fails unchanged) |

Net **+1** (`11.4.1-4.a-17` fail→pass). Default lane, floor-visible.

## Banked (issue stays `ready`)
- Group **(b)** 2 structural tests (`S11.4.1_A3.1`, `S11.4.1_A3.3_T1`) — need the global-object model; architect-first.
- (f) preventExtensions / (g) prototype-read — re-routed to owning feature issues.

## Tests
- `tests/issue-2726-mapped-args-delete.test.ts` (3 cases: returns-true+slot-undefined, siblings intact, sever-not-resurrected). tsc + prettier clean.

### DIFF
--- a/plan/issues/2726-delete-residual-sloppy-return-hasownproperty-accessor-mapped.md
+++ b/plan/issues/2726-delete-residual-sloppy-return-hasownproperty-accessor-mapped.md
@@ -24,9 +24,10 @@ updated: 2026-07-05
 > (architect-spec first): `S11.4.1_A3.1` (`delete this.y === false` — top-level
 > `this` as the global object) and `S11.4.1_A3.3_T1` (`x = 1; delete x; x` →
 > ReferenceError — implicit-global creation + real deletion + unresolved read).
-> **(e)** mapped-arguments delete DEFERRED to #1726; **(f)**/**(g)** re-route to
-> their owning feature issues. This issue stays open (`status: ready`) for the 2
-> structural (b) tests — route to architect before further dispatch.
+> **(e)** mapped-arguments delete now **DONE** (2026-07-05, opus-2726: +1 test262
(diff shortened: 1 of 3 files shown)


ttraenkler:
Re-admitting once: the auto-park was a diagnosed INFRA FLAKE, not a regression. Failed 'test262 js-host shard 3' died at the **Setup Node** step (infrastructure) → the shard never ran any tests → no pass→fail delta. Same class as the earlier Setup/corepack flakes. The fix is front-end-only, +1 default-lane (language/expressions/delete 64→65), 0 regressions measured. Single monitored re-enqueue per the confirmed-flake clause.

PR #2752, opened 2026-07-05
feat(#2858): IR call-graph-closure bucket → 0 (host-mode caller-arm relaxation)

ttraenkler:
## #2858 — `call-graph-closure` → 0

Drives the `call-graph-closure` IR-fallback bucket **9 → 0** (baseline key removed), advancing the IR-first North Star (goal `ir-full-coverage`, epic #2855).

### Root cause (measure-first)
A per-function `planIrCompilation` probe showed **all 9 rejections are CALLER-direction closure demotions**: individually-claimable *leaf helpers* (`el`, `mname`, `dimOf`, `priceOf`, `crd`, `rw`, `bcrd`) demoted purely because a **legacy** (`body-shape-rejected`) caller calls them — not because of any unclaimable *callee*.

### Why the caller arm is obsolete (in host mode)
The closure's caller arm guarded a legacy-caller↔IR-callee **signature mismatch** that **#2949 slice 3b eliminated** (`AnyKeyword → irDynamic()`: one `any` ABI for both front-ends in both modes). A claimed callee's funcIdx + signature are pre-allocated by legacy `compileDeclarations` from the same annotations, so a legacy caller of an IR callee is now signature-safe. Dropping the arm yields **0 post-claim demotions** and `WebAssembly.validate`-clean binaries.

### Standalone/wasi caveat → host-mode gating
A blanket removal flipped one test (`ir-algorithms-cluster > standalone / wasi compiles stay clean`): under wasi it claimed `joinNums`, exposing a latent host-only-op failure (`.toString()` on f64) that the caller arm had incidentally masked (its `Map`-using caller defers under wasi). **Fix:** gate the caller-direction demotion on `jsHostExterns` — relax it in JS-host mode only; keep the conservative arm in standalone/wasi. The tracked gate runs in host mode, so the bucket reaches 0 while standalone/wasi is byte-for-byte unchanged.

### Verification
- Gate (host): `call-graph-closure` 9 → 0, no other bucket moved, **0 post-claim demotions**; baseline ratcheted (`--update-on-decrease`).
- 3 affected files compile + `WebAssembly.validate` = true, 0 post-claim errs; `joinNums` post-claim errs back to 0 under both `standalone` and `wasi`.
- `tests/ir-*.test.ts`: **11 failed / 285 passed — identical to base** (the 11 are pre-existing `ir-scaffold` ×2 / `ir-bytecode-wasmgc-vm` ×9). Zero regressions.
- `tsc --noEmit` clean; `prettier` clean.

### Residual (banked in the issue, NOT in this PR)
1. `STRICT_IR_REASONS` promotion (criterion 3) deliberately deferred — the caller arm still fires legitimately in standalone/wasi, so promoting to a hard error would break those targets.
2. Eliminating the caller arm in standalone/wasi is blocked on #2856/#2857 rejecting host-only-op callee bodies up front.


ttraenkler:
BANKED for the next budget window (5-day freeze imminent, ~7% budget left — not forcing an IR-internal PR). BLOCKER: the required `equivalence-gate` check is FAILING (surfaced on GraphQL enqueue attempt). The #2858 verification covered ir-*.test.ts (identical to base) + WebAssembly.validate, but did NOT run tests/equivalence.test.ts — so the host-mode caller-arm gate (demoteOnLegacyCaller = jsHostExterns !== true) may have newly-claimed a function whose IR-path output diverges from legacy. NEXT WINDOW: (1) re-merge origin/main (PR is BEHIND — the equivalence-gate may be a stale run; re-run first), (2) if still failing, run tests/equivalence.test.ts locally, find the diverging function, fix the IR lowering or narrow the gate. The bucket-9→0 work is committed on this branch; issue #2858 status:done records the approach. Held to prevent premature enqueue.

ttraenkler:
## Finish-check (2026-07-06): equivalence-gate is REAL, not stale — keeping `hold`

Re-merged `upstream/main` (CLEAN — no conflict in `src/ir/select.ts` or `scripts/ir-fallback-baseline.json`; the PR was BEHIND but not the cause) and ran the full host-mode gate `node scripts/equivalence-gate.mjs` — the check the #2858 verification had skipped (it covered `ir-*.test.ts` + `WebAssembly.validate` only).

**Result: 39 failing / 1596 passing. The `equivalence-gate` failure is a REAL regression, not a stale run.** The relaxation banks a genuine win (24 baseline failures now PASS — tagged-template-literal cases) but introduces **3 NEW regressions** (not in baseline):

1. `illegal-cast-assert-throws.test.ts :: closure with captures passed as callable param should not illegal-cast`
2. `illegal-cast-assert-throws.test.ts :: multiple closures with different captures passed to same function`
3. `optimize-differential.test.ts :: higher-order map-like: optimized output validates and matches unoptimized`

**Root cause (shared):** a **closure-with-captures passed as a callable param**. The host-mode caller-arm relaxation (`demoteOnLegacyCaller = jsHostExterns !== true`) newly claims the higher-order leaf helper for the IR path (its only unclaimed edge was a legacy caller); the IR lowering then **illegal-casts** the captured-closure struct (legacy closure ABI ≠ IR callable/funcref signature). The premise "in host mode the caller-direction demotion is obsolete after #2949 slice-3b" is too broad: slice-3b unifies the `any` value ABI but **not** the closure-as-callable-param ABI.

**Fix direction (next window, NOT done here):** narrow the relaxation to still demote a claimed callee that has a **callable/closure parameter with captures** — relax only for value-param leaf helpers — or teach the IR path to accept the legacy captured-closure ABI for callable params. Either preserves the tagged-template 24-fix win + the bucket-9→0 ratchet while closing the 3 regressions.

Branch now carries the CLEAN re-merge + this analysis (`plan/issues/2858-...md` → `## BANKED (2026-07-06)`). Kept `hold`; bucket-9→0 work stays on-branch for next window.

PR #2764, opened 2026-07-06
fix(#2933): standalone JSON.stringify static value read (host-free)

ttraenkler:
## Summary

Reading `JSON.stringify` **as a VALUE** under `--target standalone`
(`const f: any = JSON.stringify; f({a:1})`) previously refused with the
`#1907`/`#1888 S6-b` "built-in static property value read is not supported"
compile error. It now reifies a native static-method closure that serialises
host-free via the existing 1-arg native `__json_stringify_root`
(`anyref -> ref $AnyString`) — the **same** entry the direct `JSON.stringify(o)`
call path uses, so the value form is observationally identical.

Split-out remaining sub-part of umbrella #2933 (follows the landed reflective
numeric-read + fixed-arity `Reflect.*` value-closure slices).

## Mechanism

- `JSON.stringify` case added to `ensureStandaloneBuiltinStaticMethodClosure`
  (`src/codegen/property-access.ts`) + `STANDALONE_STATIC_METHOD_META`
  (`src/codegen/builtin-fn-meta.ts`).
- Fixed-1-arg closure body: `local.get; any.convert_extern;
  call __json_stringify_root; extern.convert_any`. Calls the idempotent
  `emitJsonStringifyValue` to register the codec, then boxes the `$AnyString`
  result to externref at the any-call boundary.
- Standalone-gated (host/gc untouched); identity singleton-stable (#2963).

## Measured (current main, `--target standalone`)

- Before: `const f:any = JSON.stringify; f({a:1})` → **CE** (refusal).
- After: serialises objects / numbers / strings / nested objects correctly,
  **zero host imports** — a standalone-floor-visible **CE→run** flip.

## Known inherited limitation (NOT a regression)

A top-level `any`-boxed **array** serialises to `"null"` — but this is the
**same pre-existing substrate gap** the direct
`const x:any=[1,2,3]; JSON.stringify(x)` path already has on main (verified).
The closure is observationally identical to that direct any-path, so it adds no
new divergence; the array gap belongs to the separate $Object/array
dynamic-reader substrate work. `Math.max` (variadic) and `globalThis.Math.PI`
(niche trap) remain, tracked on #2933 (issue stays open).

## Tests

- `tests/issue-2933-json-stringify-value.test.ts` — 9 cases (object/number/
  string/nested serialisation, host-import-free assertion, singleton identity,
  distinct-method inequality, direct-call regression guard, host dual-mode).
- Retargeted the stale `JSON.stringify`-refusal assertion in
  `tests/issue-2933-reflect-static-method-value.test.ts` to the new "now works"
  behaviour (Math.max refusal kept).
- `tsc --noEmit` + prettier clean; `#2933` / reflect / reflective suites green.

### DIFF
--- a/plan/issues/2933-standalone-namespace-static-value-read.md
+++ b/plan/issues/2933-standalone-namespace-static-value-read.md
@@ -3,7 +3,8 @@ id: 2933
 title: "Standalone: Math/JSON/Reflect/Atomics namespace static VALUE reads refuse — fold constants / native static-method closures"
 status: ready
 created: 2026-07-02
-updated: 2026-07-02
+updated: 2026-07-06
+assignee: ttraenkler/opus-2933
 priority: medium
 feasibility: medium
 task_type: feature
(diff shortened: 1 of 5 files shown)


ttraenkler:
Re-admitting: the #1897 park cause (stale external standalone baseline vs the honest #3055 drop) is RESOLVED — emergency baseline refresh (run 28769644581) FORCE-promoted the honest numbers to js2wasm-baselines at 05:43. #2097 highwater already honest (19210 via #2765). Single re-enqueue; #1897 now compares against the honest baseline → no regression.

PR #2774, opened 2026-07-06
fix(#3040): thread param-default captures in closures + object methods (unblocks #2664)

ttraenkler:
## Summary

Fixes #3040 — a variable captured from an enclosing scope and referenced **only inside a parameter DEFAULT initializer** was not scanned by capture analysis (which walked the function BODY only), so it resolved to `null`. For an array-destructured parameter with such a default — `function*([x] = iter)` where `iter` is a captured custom iterable — this threw **"Cannot destructure 'null' or 'undefined'"**, blocking the last two of #2664's `merge_group` regressions:

- `expressions/async-generator/dstr/dflt-ary-init-iter-close.js`
- `expressions/async-generator/dstr/named-dflt-ary-init-iter-close.js`

**Both gate files now pass → #2664 is unblocked** (un-hold is the lead's call).

## Root cause & fix

One root cause (body-only capture scan) across multiple lowering sites. Two are fixed here; both scan each parameter subtree with `ownLocals` as the shadow set, so binding names / earlier params stay local while free default references become captures:

- **`src/codegen/closures.ts`** (arrow / function-EXPRESSION lowering) — param scan added to `referencedNames` (before the transitive-capture loop) and `writtenInClosure`. Covers the async-gen / gen / fn EXPRESSION iter-close variants, including the two #2664 gate files.
- **`src/codegen/literals.ts`** (object-literal plain methods) — passes the param-default initializers as the `extraNodes` arg to `promoteAccessorCapturesToGlobals`, mirroring the class-method / getter-setter paths (#1161).

The spec's "Defect 2" (typed-vec fast path skipping the iterator protocol) **does not reproduce** — the param is `externref` and drives `__array_from_iter_n`. The inline computed-`[Symbol.iterator]` matrix row is a separate pre-existing substrate bug (out of scope; fails in body position too).

**Deferred (documented in the issue file):** the plain function-DECLARATION path. Threading a default-only capture there shifts the function to the has-captures lowering, whose call-site threading — unlike the closures path — is **not transitive**, so a declaration called from a closure (`assert.throws(() => f())`, the standard test262 error-test shape) regresses the destructuring `*-err` families (18 measured). That needs global promotion / transitive threading — a deeper redesign left as follow-up. The #2664 gate is all EXPRESSION-form and fully covered here.

## Validation (runner-verified against current main)

- **#2664 gate files: both PASS** (IteratorClose driven, `callCount === 1`).
- 92-file `*ary-init-iter-close.js` cluster: **90/92** (2 fails are the deferred declaration variants).
- `*/dstr/dflt-*.js` sweep (651): **+31, ZERO regressions**.
- `*/object/dstr/*.js` sweep (561): **+16, ZERO regressions**.
- `tsc --noEmit` clean; new `tests/issue-3040.test.ts` (6 tests) + 51 existing param-default/destructuring/generator/async unit tests green.

Param-default handling is broad-impact, so the `merge_group` full test262 re-validation is the real gate.


ttraenkler:
**Gated on #3076 per maintainer decision — not landing until the standalone destructuring lane honors throwing accessor getters / user `@@iterator`.**

Diagnosis (senior-dev, confirmed with compiled probes on both backends): this PR is net-correct — **+75 host pass, +12 genuine standalone fail→pass** — but it exposes **14 standalone `dflt-*-err.js` false-passes**. Those tests (`default = <outer var>` destructuring where a getter/`@@iterator` throws) passed on main only *incidentally*: the outer var wasn't captured → default read `null` → destructure-of-null threw `TypeError`, and standalone `assert.throws` is lenient (opaque WasmGC throw ⇒ any throw passes). #3040 correctly captures the var, removing that incidental throw; the intended getter/`@@iterator` throw doesn't fire because **standalone destructuring doesn't yet invoke user accessors** — a pre-existing gap now tracked as **#3076**. In host mode all 14 correctly flip to pass (hence +75 / 0 host regressions); the −14 is standalone-only and not a real behavioral regression.

Decision: **gate** (option b) rather than accept-as-false-pass-correction. This PR stays parked (do-not-merge) until #3076 makes the 14 genuinely pass. Branch is merged-with-main + tsc-clean and ready to re-validate once #3076 lands.

ttraenkler:
**Un-parking per the gate condition above: #3076 has landed (PR #2862, merged into main and now merged into this branch).**

Diagnosis for the hold-lift (dev fable-3040):

- PR #2862 fixed the root cause behind the 14 standalone `dflt-*-err.js` flips: an inline `{}` receiver of `Object.defineProperty` lowered to a closed struct, so the native descriptor store silently dropped the accessor and poisoned getters never fired. Measured on main: `language/**/dstr/**/*-err*` standalone 371 → 568 (+197, 0 regressions). The user-`@@iterator` half had landed earlier via #3100/#3119.
- Spot-check on THIS branch after merging main (incl. #2862): the two #2664 gate files (`dflt-ary-init-iter-close`, `named-dflt-ary-init-iter-close`) and representatives of every previously-flipping family (`dflt-obj-ptrn-{id,prop-id}-get-value-err`, `dflt-ary-ptrn-elem-id-iter-val-err`, `dflt-ary-init-iter-get-err`) all **pass** standalone.
- Known residual, disclosed up-front: 3 files of the same false-pass-correction class remain fail on the merged state — `expressions/object/dstr/{async-gen-meth,gen-meth,meth}-dflt-ary-ptrn-rest-id-elision-next-err.js`. Root cause is DIFFERENT (generator-source elision/rest step-throw propagation in the native generator carrier — the same lane failing ~33 non-dflt siblings on main today), not this PR's capture threading and not #3076's accessor lane. If the merge_group regression gate blocks on these 3, the options are: accept as false-pass correction (this PR is heavily net-positive) or land a generator elision/rest step-throw fix first — escalating to the tech lead in that case.

Removing the `hold`/`do-not-merge` labels so auto-enqueue re-sweeps once CI is green.

PR #2777, opened 2026-07-06
fix(#3051): accessor-only exec-result marshals to null (@@replace throwing-getter cluster)

ttraenkler:
## #3051 Slice 3 — accessor-only exec-result marshals to null (throwing-getter cluster subset)

Fixes the `result-get-{index,length,matched}-err` throwing-getter tests for `RegExp.prototype[@@replace]`.

**+3 pass, zero regressions** across the entire `Symbol.replace`+`Symbol.split` corpus (isolated per-file sweep: 88 → 91 pass / 25 → 22 fail; the split fail-list is unchanged and no previously-passing file regressed). The full `tests/equivalence/` suite shows the **identical** 15-file / 36-test pre-existing failure counts on this branch and on `main`.

### Root cause (verified empirically — architect hypothesis was wrong)

The architect's Cluster-1 spec assumed a wasm-exception was being swallowed in the `_wrapForHost` get-trap. Empirical tracing disproved that: the failure is **upstream of any getter read**. The poisoned result object has *only* an accessor and no data field (`{ get index(){ throw } }`). Such an accessor-only literal compiles to a JS-host plain object (externref `$Object`), but the `r.exec = function(){ return poisoned; }` closure has a TS-**inferred structural return type** `{ index: number }`, which `resolveWasmType` maps to a **struct** return. `return poisoned` then coerces the externref host object to a struct it does not match, so the value marshals to **`null`** at the host-invocation bridge (`__call_fn_N`). V8's native `@@replace` sees `r.exec()` → `null` → `RegExpExec` finds no match → the result-object reads (and their throwing getters) never run → no throw.

Data-bearing siblings (`result-get-{capture,groups}-err`) already passed because a data field routes the literal to a real host object.

### Fix

`src/codegen/closures.ts` (`computeClosureWrapperSig`): force the closure's wasm return type to externref when the body returns a host-path accessor object literal (an object literal with get/set accessors, or a var directly initialized to one) — the RETURN-position analog of the existing `initForcesExternref` var-hoisting tag (`index.ts` / `declarations.ts`). Detection is purely syntactic (AST + checker symbol resolution) so it is stable between the pre-scan (`ensureFuncValueWrappersRegistered`) and the actual-compile calls (the pre-registered funcref wrapper type matches the compiled signature). Narrow impact: accessor-returning closures are rare, and such objects are already externref host objects everywhere else, so forcing externref return is representation-neutral for every other consumer. Host-lane only (standalone RegExp delegates to the native backend #682).

### Remaining (STOP-AND-DOCUMENTED in the issue file)

- `groups`-family err twins (`result-get-groups-prop-err`, `result-coerce-groups-err`) — nested-accessor / data-field host-read; the outer object has only data fields so this predicate does not fire (needs a broader object-literal-representation change).
- @@split `SpeciesConstructor` + fake-RegExp-splitter cluster (~14 files, incl. what the spec mislabeled as "object-arg ToPrimitive" — those files all use `RegExp.prototype[Symbol.split].call(obj)` with `constructor[Symbol.species]`); crosses the host-marshaling substrate.
- well-known-symbol-method-as-value `name.js` (both replace + split) — a separate codegen feature.

### Validation

- `tests/issue-3051.test.ts` — 15/15 pass (10 existing Slice 1/2 + 5 new Slice 3, incl. a data-object regression guard).
- Isolated RegExp `Symbol.replace`/`Symbol.split` per-file sweep — 88 → 91 pass, zero regressions.
- `tests/equivalence/` — identical 15/36-fail counts vs `main` (all pre-existing standalone-lane gaps).


ttraenkler:
Follow-up filed: **#3084** (PR #2785) — the sole merge_group "regression" here (`built-ins/RegExp/prototype/Symbol.match/g-match-no-coerce-lastindex.js`) is a **vacuity-unmask, not a #2777-introduced bug**. Verified: PASS on main `f426ef61`, FAIL on this branch because #3051 now makes the overridden `exec` return the real accessor object, which un-masks a pre-existing eager-`lastIndex`-coercion bug in `src/runtime.ts` (~L7838-7847). Fix is hard/fable (native @@match/@@replace/@@split loops must read JS-visible `lastIndex` via ToLength in the empty-match branch; `tests/issue-2671-regexp.test.ts:108` depends on the current eager firing for @@replace-empty). Per lead decision (option B), this PR stays held (cluster-gated with #2774/#3076) until #3084 lands — not excused via ratio-gate.

ttraenkler:
## Re-admission BLOCKED — real g-match regression + PR now redundant

Re-validated after merging current `origin/main` (0cbd8a2, honest baseline `oracle_version: 2`) into the branch (tip `765b0847`). **The `hold` label was left in place — this PR is not re-admittable as-is.**

**1. The cited merge_group test is a REAL regression, not a vacuity-unmask.**
`built-ins/RegExp/prototype/Symbol.match/g-match-no-coerce-lastindex.js`:
- Committed honest baseline (oracle_version 2, refreshed 2026-07-09 13:08): **status = pass, reached_test = true** on the host/gc lane. Main *honestly passes* it now — it is **not** an honest fail.
- Local `runTest262File` reproduction (deterministic, 2 runs each):
  - main-HEAD (0cbd8a2): host **pass**, standalone **pass**
  - PR branch (765b0847): host **FAIL** (`This function should not be invoked.`), standalone **pass**
- The PR flips g-match pass→fail on the host lane vs an honest-passing baseline. The forced-externref-return makes native @@match coerce `lastIndex` (invoking the `valueOf` the test asserts must NOT run) for a non-empty match.

**2. The PR is now REDUNDANT.** All 15 scenarios in the PR's own `tests/issue-3051.test.ts` — including *"exec override returning an accessor-only object literal DIRECTLY is not nulled"* — **pass on main-HEAD with the PR change absent**. A commit landed since the 2026-07-06 park already fixes #3051 accessor-only exec marshalling via a more general mechanism; `closureReturnsHostAccessorObject` in `src/codegen/closures.ts` is no longer needed and now only causes the g-match regression.

**Recommendation:** rework or close #2777 as superseded. Details committed to `plan/issues/3051-regexp-symbol-replace-split-coercion-protocol.md` (5fde8c2).

ttraenkler:
Closing as **superseded**. Re-validating this park-held PR against the landed honest baseline (oracle_version 2) shows two things:

1. **The parked failure is a genuine regression, not a vacuity-unmask.** `built-ins/RegExp/prototype/Symbol.match/g-match-no-coerce-lastindex.js` is `status: pass, reached_test: true` on the committed honest baseline and on main-HEAD (deterministic, host+standalone). This PR flips it **pass→FAIL** on the host lane (`"This function should not be invoked."`) — the forced-externref-return makes native @@match coerce `lastIndex`, invoking the `valueOf` the test asserts must not run.

2. **The PR is redundant.** All 15 scenarios in `tests/issue-3051.test.ts` — including accessor-only exec-result marshalling — pass on main-HEAD with this change absent. A commit landed since the 2026-07-06 park already fixes #3051 more generally; the `closureReturnsHostAccessorObject` heuristic in `src/codegen/closures.ts` is no longer needed and now only introduces the g-match regression.

#3051 is effectively resolved on main. Closing rather than reworking; reopen if a residual accessor-exec case turns up that main doesn't already cover.

PR #2832, opened 2026-07-09
fix(#2001): S2 — HOF hole visit-skip (forEach/filter/some/every), net-0

ttraenkler:
## Summary

#2001 S2 (builds on the landed S1 `$Hole` sentinel + read-boundary map). Adds the ES `HasProperty(O,k) is false ⇒ skip` semantics to the dense-vec HOF loop drivers for **forEach / filter / some / every** on `any[]`/untyped (externref-element) sparse arrays: a `$Hole` slot is an ABSENT index, so the callback is not invoked for it.

```ts
[1, , 3].forEach(() => c++)   // c: was 3, now 2
[1, , 3].filter(() => true)   // length: was 3, now 2
[1, , 3].some(x => x === undefined)   // was true, now false
[1, , 3].every(x => x !== undefined)  // was false, now true
```

## Mechanism (all gated on `usesArrayHoles && elemType.kind === "externref"`)

- `gateHoleSkip` — wraps a no-escaping-`br` body in `if (present) { … }` (forEach).
- `gateHoleFlag` — for the i32-flag bodies (filter/some/every): a hole yields flag `0`, so the caller's match/break/push `if` does nothing, WITHOUT adding a control-flow level around the caller's escaping `br` (the depth-shift hazard that a naive wrapper would trip on some/every's `br 2`).

Typed `number[]` (f64 element) and hole-free modules are **byte-identical** — the gate never emits a `ref.test` on them. `find`/`findIndex` use `[[Get]]` (ES6) → they VISIT holes as `undefined` (unchanged); `includes` likewise.

## Scope — narrowed to net-0 (measured)

The full S2 (all 8 HasProperty methods incl. indexOf/lastIndexOf/reduce/reduceRight) was implemented and **A/B-swept over 2,080 Array HOF/search test262 files** (branch vs same-harness main): **0 wins, 2 regressions** — both combine a hole with a **prototype-INHERITED index** (`Object.defineProperty(Array.prototype,"N",…)` / `Array.prototype[N]=…`), a feature the flat WasmGC vec cannot model; they passed on main only coincidentally (S1's `$Hole→undefined` matched the spec's inherited-undefined). test262 has **zero clean bare-literal sparse HOF tests** (it exercises holes via getters/prototype/delete), so the search/throw skip-methods only lose. **indexOf/lastIndexOf/reduce/reduceRight skip + map result-hole are DEFERRED** (kept on S1) — documented in the issue with the exact reasoning; the reverted implementations are in git history for re-landing once prototype-index inheritance is modeled.

## Validation

- **A/B sweep, 2,080 files, narrowed S2 vs main: net-0 (0 wins, 0 regressions).** The two full-S2 regressions re-pass.
- `prove-emit-identity`: **39/39 (file,target) byte-identical** vs main.
- `tests/issue-2001-s2-hof-hole-skip.test.ts` (host+standalone: forEach/filter/some/every skip both directions, real-undefined-still-matches, find/findIndex/includes visit, typed `number[]` byte-path guard, DEFERRED indexOf/reduce behavior pinned). Updated 4 S1 read-boundary cases visit→skip.
- S1 suite green (36); tsc / oracle-ratchet / ir-fallbacks clean; LOC baseline refreshed.

#2001 stays in-progress (tail: indexOf/lastIndexOf/reduce/reduceRight skip, map result-hole, S3 index-grow→$Hole, S4 destructuring numeric default).


ttraenkler:
Park resolved (fable-shepherd). Diagnosed the cited merge_group failure (run 29060112238, net -3): every/filter/some `*-c-i-22.js` — a hole whose index is INHERITED from `Array.prototype` (`Object.defineProperty(Array.prototype,"0",…)`) has `HasProperty` true, so the callback must be visited with `undefined`, but the S2 visit-skip skipped it. Not a #2830 interaction (S2 + pre-#2830 main fails the trio identically); PR-level CI was green only because the final merge commits' Test262 Sharded runs skipped all shards.

Fixed on-branch (b20f7a0) with a compile-time `arrayProtoIndexDirty` gate: `scanForArrayHoles` now detects any `Array.prototype` index WRITE and disables the module-wide HOF hole visit-skip, falling back to the spec-observable S1 visit-with-`undefined` behavior. Clean modules keep the skip byte-identical. 60/60 S1+S2 unit tests (incl. 5 new dirty-module cases run standalone). All 26 required checks green on the new head; hold removed for auto-enqueue re-sweep.

PR #2838, opened 2026-07-10
fix(#3126): typed ref-element array HOFs dispatch natively via the closure path

ttraenkler:
## Problem (#3126 — the #3098 typed-lane residual)

The typed HOF dispatch gates in `compileArrayMethodCall` admit only `f64|i32|externref` element kinds for `find`/`findIndex`/`findLast`/`findLastIndex`/`filter`/`every`/`some`/`forEach`/`reduce`/`reduceRight`. Ref-element receivers — native-string `string[]` vecs on the standalone/wasi lanes, object-struct `T[]` arrays on **every** lane — fall to the generic fallback, which is broken both ways:

- **standalone/wasi**: leaks `env.__make_callback` (unsatisfiable host import → instantiation failure). All 9 non-`map` HOFs leak on typed `string[]` (the boundary documented in #3098's implementation notes).
- **gc host lane**: silent vacuous no-op — `objs.find(o => o.x > 1)` → `undefined`, `filter` → `[]`, `some` → `false` (the host cannot iterate a WasmGC vec struct). The #1837/#3088 wrong-result class.

## Fix

- Admit `ref`/`ref_null` elements when the callback **provably compiles to a GC closure struct** (inline arrow/function-expression, or a #1919 transactional probe-compile with registered `ClosureInfo`). The typed loops' closure path (`call_ref` + `coercionInstrs`) is element-kind agnostic; the non-closure bridge (`__call_1_f64`, f64 ABI) cannot carry a GC struct, so opaque-externref callbacks keep the previous fallback (that residual is #3015).
- `compileArrayFind`/`compileArrayFindLast`: ref-element result = the element's nullable ref with a `ref.null` miss sentinel (the typed lane's `undefined` rep).
- Mirrors the #1967 `sort` and #2688 `map` gate widenings.

## Validation

- 9 leaking probe shapes → HOST-FREE + correct standalone; explicit-thisArg and 2/3-arg arity shapes pass; chained `filter→find` re-dispatches natively.
- gc struct-array find/filter/some fixed (was undefined/0/false).
- **prove-emit-identity: IDENTICAL** — all 39 (file,target) hashes across gc/standalone/wasi vs both pre- and post-#2815 main tips.
- **test262 Array-HOF cluster (1,699 files × gc+standalone = 3,398 rows): zero flips** either direction (test262 sources are untyped JS → dynamic lanes; the typed-lane fix serves TS-typed code + the standalone leak metric, same shape as #3098's cluster result).
- Byte-parity (sha256) proven for untouched shapes: identifier/opaque callbacks, `number[]`, gc-lane `string[]`.
- `tests/issue-3126.test.ts`: 39 pass / 2 documented skips (pre-existing gc externref miss-rep `=== undefined` bug, byte-identical to main — named in the issue file's residuals along with the pre-existing identifier-held-closure wrongness on the `__apply_closure` path).
- Gates: tsc, lint, format, loc-budget (+65 in-module, baseline regen), coercion-sites, speculative-rollback — all green locally.

Issue: `plan/issues/3126-typed-refelem-array-hof-native-closure-dispatch.md` (status: done in this PR, self-merge path). Parent boundary cross-ref updated in #3098's issue file.

### DIFF
--- a/plan/issues/3098-standalone-native-callback-dispatch-retire-make-callback.md
+++ b/plan/issues/3098-standalone-native-callback-dispatch-retire-make-callback.md
@@ -233,12 +233,13 @@ root: `eliminateDeadImports` strips it when nothing calls it.)
 and 3-param callbacks all correct, `.tmp/probe-3098-extended.mts` 26/27).
 
 **Validation:**
+
 - `tests/issue-3098.test.ts` — 15 tests, all pass (host-free asserted via
   zero-import check + empty import object).
 - Byte-identity: `prove-emit-identity` main-vs-branch — IDENTICAL, all 39
   (file,target) hashes across gc/standalone/wasi on the playground corpus.
 - Cluster (runTest262File, standalone lane): `built-ins/Array/prototype/{map,
-  filter,forEach,reduce,every,some,find*}` (1,439 non-skip files) — ZERO
+filter,forEach,reduce,every,some,find*}` (1,439 non-skip files) — ZERO
   regressions, zero flips. The sampled 81 "only-`__make_callback`" leak rows
   from the 2026-06-16 standalone JSONL also show 0 flips: each depends on an
   ADJACENT gap (see boundaries below). The value of this slice is the
@@ -246,6 +247,7 @@ and 3-param callbacks all correct, `.tmp/probe-3098-extended.mts` 26/27).
   the same dispatch arm.
 
 **Boundaries / residuals (named, for S4/S5 + adjacent owners):**
+
 - **Typed `string[]` receivers of `find`/`filter`/`findIndex`/`findLast*`/
   `every`/`some`/`reduce*`/`forEach` still leak `__make_callback`** — the
   typed inline impls in `array-methods.ts:3350-3440` gate on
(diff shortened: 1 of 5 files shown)


ttraenkler:
Diagnosed and fixed the merge-group failure (auto-park was correct — a real regression).

**Root cause of the 212 regressions (all `built-ins/Temporal/*`):** the first version widened the ref-element HOF gate on ALL lanes. Temporal tests iterate ref-element tuple arrays at top level (`expected.forEach(([unit, pos, neg]) => …)`), so the widened gc gate re-routed those callbacks from `compileArrowAsCallback` (`__make_callback`, host-lifted) to `compileArrowAsClosure` (GC closure) — and `Temporal`/`TemporalHelpers` are HOST globals the closure-lifted body cannot resolve ("TemporalHelpers is not defined"). Attribution: diffed this PR's merge-group merged-report against predecessor #2815's merge-group report — 212/212 flips attributable to this PR, 100% Temporal; a method bisect pinned `forEach`.

**Fix (pushed as b8da9dd):** the widening now carries `(ctx.standalone || ctx.wasi)` — the lanes where `__make_callback` is unsatisfiable, so re-routing can only gain. gc emission is byte-identical to main (shape-hash verified; prove-emit-identity IDENTICAL vs current main; 31/31 sampled regressed Temporal files pass again locally). The gc struct-array vacuity is documented as a pre-existing residual in the issue file (real root: closure-lifted host-global resolution).

Removing the `hold` label per the auto-park instructions so auto-enqueue re-sweeps once CI is green.

PR #2835, opened 2026-07-09
fix(#3049): Iterator.prototype helpers on plain iterators — top-level fnctor-prototype init, deferred host init (C1), bridge-exit marshaling

ttraenkler:
## Summary

Lands the #3049 Iterator-helper cluster (verify-first re-diagnosis of the "externref-prototype-storage wall" — which **dissolved**: fnctor-prototype storage round-trips identity and own members fine; the wall was Layer-1 elision + init timing + three downstream gaps, each pinned with probes before coding).

**Measured cluster delta (in-process runner, host lane): 1/21 → 13/21** — all 11 `*/this-plain-iterator.js` twins flip, plus `flatMap/flattens-iterable.js` and `Symbol.iterator/return-val.js`.

## The five stacked fixes

1. **Layer 1 (`src/codegen/declarations.ts`)** — top-level `F.prototype = <expr>` was silently elided in host/GC mode: the #2671 keep excluded `prototype` claiming the fnctor-prototype lift consumes it, but `tryCompileFnctorPrototypeAssign` is `!ctx.standalone → return undefined` (standalone-only). Now kept (receiver unwrapped through parens/`as`-casts — the runner shim is `(Iterator as any).prototype = …`). No double-apply possible; standalone byte-identical.
2. **Layer 2 / C1 (harness only, per arch-3049 plan)** — the shim RHS calls the `__iterator` host import, which needs `getExports()`; under the wasm `(start)` section it threw during instantiate. Host test262 lanes now compile with the **existing** #2796 `deferTopLevelInit` and call the exported `__module_init()` after `setExports`: `scripts/compiler-fork-worker.mjs`, `scripts/wasm-exec-worker.mjs`, `scripts/test262-worker.mjs` (gated `!target`), `tests/test262-runner.ts`, `tests/test262-shared.ts` fixture path. **Compiler default unchanged; standalone/wasi/linear lanes untouched.** Runtime-negative classification preserved (a top-level expected throw still passes); non-negative top-level throws now bucket honestly as runtime `fail` instead of `compile_error/malformed_wasm`.
[...]

PR #2839, opened 2026-07-10
fix(#3123): class-extends-fnctor instances resolve the live F.prototype chain — host member dispatch + capture defer

ttraenkler:
## Summary

`class C extends F` — F a top-level plain function whose `.prototype` is assigned at RUNTIME (the test262 harness `Iterator` shim) — now works end-to-end in the host lane. Flips the full ~8-file residual Iterator-helper cluster split out of #3049:

- `built-ins/Iterator/prototype/{map,filter,flatMap,drop}/exhaustion-does-not-call-return.js`
- `built-ins/Iterator/prototype/{drop,take,filter}/return-is-forwarded.js`
- `built-ins/Iterator/prototype/flatMap/iterable-to-iterator-fallback.js`

**Measured: 8/8 flip to pass; both #3049 `this-plain-iterator` controls stay green.**

**STACKED ON #2835** (the #3049 five-part fix) — branched from its real branch per the explicit predecessor-stacking rule; #2835 must land first.

## Five walls, each pinned via WAT/import traces before coding

1. **No host link from class instances to F's live prototype** — `${C}_init` now tails `__register_fnctor_instance(self, F_closure)` (host lane, fnctor-ancestor classes only) so the existing #1712 `_fnctorInstanceCtor`/`_fnctorProtoLookup` machinery serves inherited reads (`inst.drop` → the runtime-assigned helper proto). The walk gains a `__sget_prototype` fallback (the #2664 member-set dispatcher can store `F.prototype` in the closure struct's FIELD slot).
2. **Compiled class methods/getters were host-invisible** — gated `__member_kind_<k>` / `__call_get_<k>` dispatch exports let `_safeGet`/`_resolveHostField` serve `next`/`return` (method AND `get next()` forms) off the instance, marshaled via the #3049 bridges. Getters run per [[Get]] (the exhaustion tests mint a fresh generator per read).
3. **calls.ts null lowerings** — method-MISS on fnctor-subclass receivers emitted `__extern_get` + `ref.null.extern` (graceful tail); any-typed `iterator.return()` got INFERRED to the class and tag-dispatched with a NULL self on host wrappers. Both now route through `__extern_method_call`; the inference scan skips fnctor subclasses; `__gen_next/return/throw` gain a registered-instance `_safeGet` miss-arm.
4. **`(ref $C)` slots nulled host wrappers on reassignment** (`iterator = iterator.drop(0)`) — pre-hoist widening to externref (`fnctorWidenedLocals`, foreign-reassignment scan), mirrored in the block-scoped-shadow realloc path; widened bindings dispatch dynamically. Never-reassigned bindings keep static dispatch.
5. **#2818's derived-class defer carve-out no-op'd captured writes in HOST lane** — a try-block `class C extends F { return() { ++captured; } }` compiled eagerly, before the block-`let` exists, so the increment lowered to `f64.const NaN; drop` (WAT-verified). The carve-out existed only to protect standalone (its own comment documents the deferred path as host-correct); now gated `ctx.standalone || ctx.wasi`.

Plus: computed well-known-symbol struct FIELDS (`{[Symbol.iterator]: 0}`) are host-readable via `__sget_@@<name>`, so GetIteratorFlattenable throws TypeError on a non-callable `@@iterator` (GetMethod, §7.3.10) — the 8th file's distinct root.

## Validation

- `tests/issue-3123.test.ts` (6 green) + `tests/issue-3049.test.ts` (6 green) on the merged tree.
- **prove-emit-identity: IDENTICAL** — 39/39 (file,target) sha256 across gc/standalone/wasi vs the predecessor tip (all new emissions gate on `moduleHasFnctorSubclass` / fnctor-ancestor predicates).
- Chunked `built-ins/Iterator` + `language/statements/class/subclass` sweeps, branch-vs-base control at the predecessor tip: flip sets identical except the target improvements (the in-process sweep's shared 129-file noise is byte-identical on both sides and passes in isolation).
- Related suites (`issue-2818/2628/2015/1712*`): one failure in `issue-1712-capture-closure-dispatch` #3 reproduces UNCHANGED on origin/main b6691942 (local Node-25 environment, pre-existing).
- Broad-impact (host defer scope + dynamic dispatch) → full merge_group is the gate.


ttraenkler:
Closing in favor of the clean re-derivation, PR #2878. I diagnosed this PR's merge_group park (run 29063671945): net +72 with a 99-file regression cluster (signature `5c449d888de7d030`) — identical to base PR #2835's own merge_group signature, proving the regressions came entirely from the stack's #3049 base (superseded by main's #2860), not from this PR's #3123 delta (+116 improvements, no new regressions). Since the base is closed and this branch is DIRTY against main's independent #3049, landing it as-is was not an option.

#2878 re-derives the #3123 delta off current origin/main, keeping the five walls (Layer-1 `F.prototype = …` init keep, ctor-tail instance registration, member-kind dispatch exports, dynamic dispatch + widening, the #2818 host-defer flip) while excluding the two regression roots I isolated from the merge_group artifacts: the generic `__call_fn` bridge-exit marshaling (~85 dstr double-close files — `_marshalBridgeResult` is now called only from registration-gated arms) and `deferTopLevelInit` on multi-module FIXTURE compiles (6 `Duplicate export name '__module_init'` CompileErrors — the flag is now gated off for module-goal tests). Measured on the re-derivation: 6/8 target files flip, home cluster +66/−1, and 93/99 of this stack's regression corpus pass (the 6 remaining are pre-existing local module-code artifacts reproduced identically on pristine main). The 3-file flatMap inner-flattening residual is documented in the #3123 issue file for a narrow follow-up.

PR #2842, opened 2026-07-10
fix(#3124): inherited reads over Object.create(compiled-struct) prototype chains — struct-aware host chain walk

ttraenkler:
## Summary

Fixes #3124 (plan/issues/3124-host-chain-inherited-reads-over-compiled-structs.md): inherited member reads/calls through host prototype chains that contain a compiled WasmGC struct (`Object.create(<struct instance>)` receivers) resolved `undefined` / threw / trapped.

**Runtime-only** — `prove-emit-identity`: all 39 (file,target) hashes byte-identical to base.

### What changed (src/runtime.ts)

1. `_protoChainStructResolve` — bounded manual chain walk; each struct hop resolves OWN members via the exact direct-receiver machinery (`_resolveHostField`: accessors → sidecar → `__sget_*` → fnctor proto), tombstone-aware per hop (§10.1.8.1: a deleted own member continues the walk UP).
2. `__extern_get` — BOTH implementations (name-based AND the intent-based twin modern compiles bind): the non-null-proto bail now walks before conceding undefined.
3. `__extern_has` — `in` walks struct hops via the tombstone-aware `_wasmStructHasOwn` oracle (§7.3.12).
4. `__extern_method_call` — last-chance inherited arm before the 'is not a function' throw; original receiver threaded as `this` via the method-arity bridge.
5. `typeof_check` intent (`function` only) — the fused `typeof x === \"function\"` compare now recognizes raw closure structs via `__is_closure` (host-intent twin of #1594A / #1896).

All #3124 acceptance probes pass (F, H, K2, L2, class-instance protos, inherited literal-method calls, `in`, two-level chains) — locked in `tests/issue-3124.test.ts` (10 tests, all green).

### Boundary split to #3129 (new issue file included)

Any-receiver CLASS-method calls (`o.getX()`) are statically bound in calls.ts and null-coerce foreign receivers **in-wasm** (host never consulted) → codegen follow-up, full WAT mechanism documented in the issue file.

## STACKED ON #2835 — hold until it lands

This branch contains PR #2835 (`issue-3049-toplevel-fnctor-proto-init`, the #3049 five-fix machinery) per the known-dependency stacking exception — my `__extern_method_call` arm dispatches through the #3049 bridge (`_marshalBridgeResult` is not on main yet). #2835 is currently `DIRTY` (mechanical `scripts/loc-budget-baseline.json` conflict vs #2832); once it re-merges main and lands, this PR's diff collapses to the #3124 changes only and the `hold` comes off.

### Validation

- `tests/issue-3124.test.ts`: 10/10
- Scoped sweeps: issue-3049, issue-1462, issue-1712*, issue-1355c — green. Two failures found are **pre-existing on pristine origin/main** and outside CI's executed set (noted in the issue file): check-regressions writable-descriptor, issue-1528 generator-method bridge.
- `prove-emit-identity`: IDENTICAL (39/39).
- tsc --noEmit: clean. prettier: clean.


ttraenkler:
Closing after a fresh yield re-measurement against current main. I re-verified the gap is real (the F/data-member/`in` probes still fail on origin/main and all flip to pass with this branch's runtime delta applied), then ported the #3124 runtime-only delta (149 lines, applies 3-way-clean, no dependence on the closed #2835 stack's machinery) onto current main — which now includes the gOPD descriptor MOP (+121), #3119, the async-carrier flip, and #2860's iterator-record shim — and A/B-measured the natural home cluster (built-ins/Iterator/prototype + built-ins/Object/create, 693 files) with the real runTest262File harness, identical chunking both sides: **base 447/693, fixed 447/693 — zero flips in either direction.** Same verdict as the original measurement: the inherited-read gap is genuine but no conformance test in the corpus is gated by it alone; the cluster's failing files are held by other walls (chiefly #3129's static-bind null trap, which this was the read-side prerequisite for).

A byte-inert substrate fix with no scored yield doesn't justify carrying a stacked, DIRTY branch — the work and full root-cause analysis remain preserved on `issue-3124-inherited-reads-struct-proto-chain` (this PR's branch) and in the issue doc + tests/issue-3124.test.ts there, ready to be revived as the companion of a future #3129 fix, where it should land together with the dispatch-side change that actually flips files.

PR #2873, opened 2026-07-10
feat(#2967 slice 2a): admit host-drive closures — lift the #2646 park; bank slice-1 A/B; split #3134

ttraenkler:
## Summary

Slice 2a of #2967 (async engine convergence): lifted arrow / function-expression closures now take the **host-drive frame engine** — `planAsyncClosureActivation` returns host-drive decisions instead of re-laning single-await shapes to CPS and parking multi-await shapes.

## Why the #2646 park no longer applies

The park (33 null_deref merge_group regressions on the first attempt) **predates the #2865 resume-fn environment re-establishment**: `ensureAsyncResumeFunction` now re-runs the `__self` capture-struct materialization (`selfCaptureLayout`), threads capture-cell deref routing (`boxedCaptures`), and `readsCurrentThis` — exactly the interplay whose absence caused the null_derefs. That infrastructure is already validated in the lifted-closure context by the #2865 async-gen-consumer exception; this slice extends it to the general host-drive closure population.

This also **retires the CPS lifted-closure emit's two known-wrong shapes** (discarded-tail bare await → null; value-return suffix without adoption) — the frame engine settles the pre-allocated result promise uniformly. Both verified by new suite cases.

## Also in this PR

- **Slice-1 A/B banked** in the issue file: merge_group run 29117178921, 48,088 tests, net −1 = one ct_flake (runner noise); 0 real regressions, 0 wasm-hash-change regressions.
- **#3134 filed** (root-caused during triage): `resolveWasmType` unwraps `Promise<T>` → T (f64) on the host lane, so Promise-typed slots mangle REAL promise externrefs via `__unbox_number` → NaN. Engine-independent, pre-existing (control-verified on pristine main). This is the issue's 'identifier-operand await resolves to null' bug.
- Triage note: the `any`/untyped-param call gap (callee body compiles to `return ref.null`; even sync closures) is pre-existing and likely the TRUE #2646 mechanism (test262's `asyncTest(testFunc)` is exactly that boundary).

## Validation (local, post-merge of upstream/main)

- `tests/issue-2967-engine-convergence.test.ts`: 17/17 (7 new slice-2a closure cases: multi-await fn-expr callback via sig-dispatch, captures across awaits, capture cells, discarded-tail, promise-return adoption, rejection).
- `issue-2957`, `issue-1042-host-drive`, `issue-2895-async-frame`, `async-await`, `async-census`: 47/47.
- `issue-2174`, `issue-2906-multiawait`, `equivalence/async-function`, `equivalence/async-iteration`, `promise-chains`, `issue-2623`, `issue-2906-3dii`: pass.
- Pre-existing local-env failures (issue-2906-gap3 wasi trio; 7 AsyncFromSyncIterator/symbol-async-iterator e2e) reproduced IDENTICALLY on pristine-main control worktrees — not caused by this PR.

The merge_group full-corpus A/B is the hard gate for the closure population flip — same as slice 1.

Remaining CPS population after this slice: concise arrow bodies + the pattern/rest-param carve-out (slice 2b), then CPS deletion (2c), then `planLinearAwaits` widening (slice 3).


ttraenkler:
I root-caused the merge_group park (run 29120059791, net −36: 32 null_deref + 5 wasm_compile, all wasm-hash-change) and fixed it on the branch. It was NOT the `__self`/capture interplay the slice rationale assumed #2865 had covered, and not the pre-existing any-callee gap either — all 37 files pass on pristine main, where these closures never reach host-drive.

**Bug 1 — wrapper-struct RTT mismatch (the 32 null_derefs).** Activating the async machine rewrites the closure's result to externref (the Promise), so the value site allocates the closure under the `() -> externref` signature's funcref-wrapper struct. A typed consumer — test262's `asyncTest(fn: () => void)` shim — casts the incoming externref to its *declared* signature's wrapper instead. Wrapper structs are layout-identical but chained `sub final` under whichever wrapper the module minted FIRST, so WasmGC canonicalization does not merge them: the sibling cast nulls out and the funcref fetch traps ("dereferencing a null pointer in asyncTest()"). Survival was pure wrapper-creation order — a file also using `assert.throwsAsync` (`() => any`, compiled earlier) made the externref wrapper the root and every void-typed cast a doomed sibling downcast. Main "passes" these files only because the legacy path lowers the closures as sync void functions, so declared == actual. Fix (emit repair in the calls.ts callable-param dispatch): cast the externref callee to the wrapper ROOT — the guaranteed supertype of every wrapper — fetch the funcref from the root's field 0, and re-cast self per funcref-matched arm. This also closes the same latent order-dependence for covariant SYNC closures (`() => string` passed as `() => void`); modules whose declared wrapper already is the root emit byte-identically.

**Bug 2 — frame spill layout vs body-compile rebinding (the 5 wasm_compile).** The spill fields are typed from the TS declaration before the body compiles, but the body can lawfully re-type the local: a nested fn's mutable capture CELL-BOXES it (→ `struct.set[1] expected i32, found (ref null N)` — the await-using microtask cluster), and a ref-typed guess can diverge from the inferred rep (`const expected = [prom]` under the #3134 Promise unwrap — the fromAsync does-not-await-input file). Fix (admission tightening): `asyncClosureCellSpillHazard` declines host-drive for exactly these two spill shapes, re-laning them as pre-slice-2a (CPS/legacy) until the frame layout is made cell-aware (#2967 phase 3).

**Measured on the branch (post-fix, post-main-merge):**
- the full 37-file regressed set: 37/37 pass (`runTest262File`, per-file A/B against pristine main);
- 210-file sweep of the affected suites (fromAsync, await-using(+syntax), AsyncDisposableStack/disposeAsync, AsyncFromSyncIteratorPrototype/throw) vs the js-host baseline: **0 regressions, +17 improvements** — the slice's intended win now actually lands (real async closures instead of the legacy sync-void lowering);
- `tests/issue-2967-engine-convergence.test.ts` 20/20 (3 new cases codify both mechanisms, including the wrapper-ORDER trap);
- issue-2957 / 1042-host-drive / 2895-async-frame / async-await / async-census 47/47; closure/callback equivalence suites green (the optional-direct-closure-call pair and the 2906-gap3 wasi trio fail identically on pristine main — control-verified pre-existing).

I'll remove the park `hold` once the PR checks are green on this head — the merge_group re-validation stays the hard gate for the full-corpus A/B.

PR #2932, opened 2026-07-12
fix(#3169): standalone Array.prototype HOFs over closed-struct array-like receivers

ttraenkler:
## Summary

Fixes #3169 — the largest #2860 standalone-gap method-family cluster (~513 measured host↔standalone gap tests under `built-ins/Array/prototype/{reduce,reduceRight,filter,some,every,map,forEach}`).

**Root cause:** the dominant test262 receiver `var obj = { 0: 11, 1: 12, length: 2 }` has no contextual type, so it compiles to a CLOSED nominal WasmGC struct (`$__anon_N`), not an open `$Object` (#1897 forbids diverting uncontexted literals). The standalone dynamic-reader trio (`__extern_length` / `__extern_get_idx` / `__extern_has_idx`) had arms for `$Object`/`$ObjVec`/typed vecs only — a closed-struct receiver answered `length 0`, so the generic `Array.prototype.<HOF>.call(obj, cb)` loop ran zero iterations and returned the seed (the "returned 2 — assert #1" signature).

## Changes (all standalone-gated; gc/host byte-identical)

1. **`fillExternArrayLikeStructArms`** (object-runtime.ts, finalize — sibling of `fillExternGetIdxVecArms`): splices one `ref.test`-guarded closed-struct arm per array-like struct into each of the three readers. Length reads are ToLength'd and cover f64/i32/boolean/externref-unbox **and string lengths** (`length: "2"` → `__str_to_number`, the `-3-*` family); index reads are per-canonical-integer-field `f64.eq` + boxing; HasProperty is the OR-chain (hole semantics: `{0:x, 2:y, length:3}` skips index 1).
2. **No-init `reduce`/`reduceRight` un-refused** (array-methods.ts): the M2.2c funcidx-shift bug the refusal guarded is gone (the loop re-resolves `__extern_has_idx`/`__extern_get_idx`/`__is_truthy` by name — the #16 discipline); the §23.1.3.24 step-6 hole-scan seed now compiles natively, host-free.
3. **Standalone dynamic-`any`-index read** (property-access.ts): twin of the host-only #2773 arm. `obj[idx]` inside a named-function HOF callback (implicit-`any` idx) reads the legacy `__extern_get` first (stays authoritative), retrying positionally through `__extern_get_idx` only on miss with a numeric, non-string key. `arguments`-rooted receivers are excluded (order-fragile materialized state — #3180 bucket 4).
4. **#3037 read-carrier gating** (binary-ops.ts + context/types.ts + property-access.ts): `maybeWrapAnyReadEqualityCarrier` now requires the live `ctx.activeAnyEqDispatchExpr` marker, closing the mid-operand `$AnyValue`-registration hazard (an equality path chosen while the type was unregistered received a `ref $AnyValue` it compared by struct identity → spurious `!==` for value-equal operands). This also fixes the pre-existing standalone `3 === 3.0 → false` gap — `tests/issue-1917-emit-eq.test.ts`'s pinned lane-split value is now the spec-correct 18 on both lanes.

## Measured results (local 7-family batch, 1605 files, standalone target)

- in-family standalone passes: **672 (baseline) → 876 (+204 net)**
- authoritative gap flips (host-pass ∧ standalone-not-pass): **202 / 513**
- all 3 issue acceptance samples pass host-free (`reduce/15.4.4.21-9-c-ii-29`, `map/15.4.4.19-8-c-ii-1`, `every/15.4.4.16-7-c-ii-9`)
- **2 in-family regressions, both vacuous-pass→honest-fail**: `reduce/15.4.4.21-8-b-ii-2.js`, `reduceRight/15.4.4.22-8-b-ii-2.js` — previously the *refused* no-init call answered `undefined`, which happened to satisfy `assert.notSameValue`; now the empty hole-scan throws the spec TypeError (truly passing needs the #3180 bucket-1 defineProperty MOP).

**Acceptance shortfall, deliberate:** the issue's ≥350 flip target is not reachable with the receiver ladder alone — the residual ~306 gap tests are six non-ladder mechanisms (defineProperty-during-iteration MOP ~101, fnctor-array-inheritance ~52, builtin expandos ~46, arguments fidelity ~37, thisArg ~29, ToPrimitive lengths ~26), filed with measured counts as **#3180** per the issue's own anti-scope-creep clause.

## Tests

- New `tests/issue-3169.test.ts`: 16 host-free standalone asserts (zero-import gate) covering with/without-init reduce, hole seeding/skipping, reduceRight order, map/filter/every/some/forEach, ToLength clamping, all-holes vacuous every, callback `obj[idx]` reads, string-key path preservation, `$Object`-receiver preservation.
- Green locally: #3037 carrier suites (×115), #3098, #3126, #2903, #1461, #1917, #3053, #3128, #2773, array-methods, array-prototype-methods; `tsc --noEmit` clean; loc-budget allowance granted in the issue frontmatter (growth sits beside the established fill siblings).


ttraenkler:
## ⚠️ Coordination notes for the merge queue

**1. Two expected standalone regressions are vacuous-pass→honest-fail (NOT real regressions).**
`reduce/15.4.4.21-8-b-ii-2.js` and `reduceRight/15.4.4.22-8-b-ii-2.js` were STANDALONE-lane `pass` only because the pre-fix no-initial-value `reduce` refusal answered `undefined`, which happened to satisfy `assert.notSameValue`. With the refusal retired they now run the §23.1.3.24 step-6 hole-scan and throw the spec TypeError — an honest fail. Confirmed: **HOST lane already `fail` in both baselines and stays failing (zero host regression)**; only the standalone leaky-pass flips. Truly passing needs the defineProperty-during-iteration MOP (filed as #3180 bucket 1). If the merge_group regression gate auto-parks citing exactly these two, this is the confirmed-honest-fail case — resolution is the #2879 honest-baseline update + re-admit as confirmed-collateral, not a code "fix".

**2. The bonus any-equality fix touches #3037-adjacent code (LOCKED by fable-identity) — but is isolated carrier-GATING, not canonicalization.**
Exact surface: one new optional ctx field (`activeAnyEqDispatchExpr`), a save/restore wrapper around the existing `compileAnyBinaryDispatch` call in binary-ops.ts, and one early-return guard line at the top of `maybeWrapAnyReadEqualityCarrier`. It does NOT touch the classifier (`__any_from_extern_honest`), the `$AnyValue` tag scheme, `isAnyEqualityOperand`, or any canonicalization body — it only NARROWS *when* the carrier fires (a strict subset of prior cases; all 115 existing #3037 carrier tests stay green). Overlap with fable-identity's #3037 canonicalization work is textual only, at those 2 small insertion points. Flagging so the queue order avoids a collision.

ttraenkler:
**arch: per-bucket reuse pointers for #3180** (filed in this PR — please fold into `plan/issues/3180-standalone-array-hof-residual-buckets.md` as an `## Implementation Plan` addendum; writing from another branch would add/add-conflict with this PR). The issue's 6-bucket decomposition already IS the slicing — these are the verified reuse anchors per bucket so each slice is claimable cold:

1. **defineProperty-MOP (~101)** — stays ROUTED to #2992/fable-2984c (already noted in the issue). Downstream consumer only.
2. **fnctor-array-inheritance (~52)** — the proto-chain walk arms in `src/codegen/object-runtime.ts` (`__object_keys_forin` $proto walk, #2964; `__extern_get` proto-descriptor arm ~:1127) are the substrate; the gap is Array.prototype-as-instance-proto lineage. Coordinate with #2917 (builtin-super) before starting — overlapping mechanism.
3. **builtin-expando receivers (~46)** — expando properties on builtin singletons (Math/JSON/Date...): reuse the vec/singleton expando sidecar arms in object-runtime.ts rather than a per-builtin store; the $NativeProto glue (`registerNativeProtoBuiltin`, array-object-proto.ts:1550) is the precedent for making a builtin surface answer property ops.
4. **arguments fidelity (~37)** — two halves: remove `isArgumentsRootedExpression` exclusion (property-access.ts, cited in the issue) AND make the materialized arguments object carry ACTUAL call args (the #820l arity-3 plumb in array-methods.ts `buildClosureCallInstrs` :6622+ already pushes extras — the gap is the callee-side arguments view).
5. **thisArg semantics (~29)** — `compileThisArg` (array-methods.ts:6543) + the `__current_this` global threading; the HOF inline loop must set/restore `__current_this` around each `call_ref` (grep the existing prevThisTmp save/restore in the setup — extend, don't fork).
6. **ToPrimitive lengths (~26)** — inside `__extern_length`'s closed-struct arm (object-runtime.ts `objLengthArm` :4095): when the length field is a closed struct with valueOf/toString, dispatch via `closed-method-dispatch.ts` method-call machinery instead of returning NaN.

Sequencing: 6 → 5 → 4 (self-contained, S/M each) → 3 → 2; 1 stays peer-owned. Per slice: recompute the residual list first (main moves), zero host regressions.

PR #2935, opened 2026-07-12
fix(#3176): standalone JSON SyntaxError strictness + ES2025 rawJSON/isRawJSON

ttraenkler:
## Summary

Two host-free (`--target standalone`/`wasi`) slices of the #3176 JSON residual, both gated on `ctx.standalone` so gc/host mode is byte-unchanged:

1. **JSON.parse SyntaxError strictness** — the native string parser (`json-codec-native.ts` `__json_parse_str`) now rejects unescaped control chars (U+0000..U+001F) and invalid `\uXXXX` escapes (short / non-hex) with a `SyntaxError` instead of silently accepting them. Flips `built-ins/JSON/parse/15.12.1.1-g4-1..4`, `g5-2`, `g5-3` (6 files).
2. **ES2025 `JSON.rawJSON` / `JSON.isRawJSON`** — mints a branded raw-JSON carrier through the existing object runtime. The `[[IsRawJSON]]` internal slot is a bit on `$Object.flags` (`OBJ_FLAG_RAWJSON`), so a plain `{ rawJSON: '…' }` is not recognised. `rawJSON` reuses the native parser to validate a single-primitive argument. Flips `rawJSON/{returns-expected-object,invalid-JSON-text,illegal-empty-and-start-end-chars}` and `isRawJSON/basic` (4 files).

**10 files (~20 rows) flipped standalone; zero host-lane change** (the native codec is not used in gc mode). Verified via a faithful `wrapTest` gc-vs-standalone scan of all 165 `built-ins/JSON` tests: gap 34 → 28, no new failures.

## Reuse vs. new
Extended the existing native codec (`json-codec-native.ts`) — no second parser. The carrier reuses `object-runtime` (`__new_plain_object`/`__extern_set` + a new `$Object.flags` brand bit). The plan pointed rawJSON at the builtin-static tables, but the real dynamic `JSON.<method>` dispatch lives in `expressions/calls.ts`; correction noted in the issue file.

## Scope correction (honest)
The plan's "illegal cast (10 rows)" bucket is **not** the reviver ARRAY-holder walk — it's a **general standalone `for (var k in arr)` + array string-index bug** in the test harness loop (proved by ablation: trap persists with `JSON.parse` removed). Filed as **#3179**; those 10 `parse/15.12.2-2-*` rows are blocked on it. The realistic JSON-only ceiling is ≈43 rows (10 blocked on #3179, 2 Proxy deferred), so the "≥48" acceptance is not reachable via JSON alone. **#3176 stays `in-progress`** for the documented remainder.

## Test plan
- `tests/issue-3176.test.ts` — strictness (control chars, bad escapes, valid inputs) + rawJSON/isRawJSON (primitives, brand recognition, invalid-input throws), all standalone with an empty import object.
- `tsc --noEmit` clean. Pre-existing unrelated failures in `json.test.ts` / `issue-1599-*` confirmed present on `main` (stash-verified).


ttraenkler:
**arch: implementation-plan sketch for #3179** (filed in this PR — please append as `## Implementation Plan` to `plan/issues/3179-standalone-forin-array-string-index-illegal-cast.md` before/after merge; writing it from another branch would add/add-conflict with this PR).

### Where the two halves live (anchors verified on main)
- **for-in over an Array** takes the #2575 native index loop `emitArrayForIn` (src/codegen/statements/loops.ts:5796, dispatched at :6146-6150): keys are decimal STRINGS minted by the sealed decimal-key formatter into an externref keyLocal. This half is likely fine — the trap is on the read.
- **The read `arr[index]`** with a runtime STRING key over a statically vec-typed receiver: the element-access lowering in src/codegen/property-access.ts. Suspect: the non-numeric-subscript arm routes the vec receiver down a dynamic-object path that does `any.convert_extern` + `ref.cast $Object` (a vec struct is NOT an $Object → illegal cast). The standalone array-like arms that DO handle vec receivers live in object-runtime.ts (`objArrayLikeArms` gate :206, vec-base arm :4154, `__extern_get_idx`).

### Step 1 — pin the failing cast (fast)
Compile the issue's minimal repro standalone, dump WAT, find the trapping `ref.cast` (the trap pc is in the repro's own function or in `__extern_get`): (a) if in the caller — the element-access arm cast the receiver; (b) if in `__extern_get` — the reader's dispatch lacks a plain-`$__vec_*` arm ahead of its `ref.cast $Object`.

### Fix directions
- **(a) caller-side (preferred):** in the element-access lowering, when `resolveArrayInfoFromWasmType` says the receiver is a vec and the subscript is a runtime string/externref, emit native decimal-string→index parsing (the inverse of the sealed decimal-key formatter) + the ordinary bounds-checked `array.get`; non-canonical/OOB key → undefined. No new helper family — reuse the existing string→number native used by `__extern_get`'s key handling.
- **(b) reader-side:** add/order a `ref.test` vec arm in `__extern_get`/`__extern_get_idx` BEFORE the `$Object` cast (mirroring the $ObjVec arm at object-runtime.ts:4154), then route this shape through `__extern_get(recvExt, keyExt)`.

### Edge cases / tests
Keys ≥ length and non-index string keys (undefined, not trap); array expando keys; the gc/host wrong-value gap the issue notes is SEPARATE — don't regress it; destructuring for-in heads. Tests: minimal repro → 2; the 10 `built-ins/JSON/parse/15.12.2-2-*` rows; a for-in+string-index sweep outside JSON. Zero host-lane byte changes (standalone-gated arms only).

PR #2952, opened 2026-07-12
fix(#3166): computed-key invocation of class-instance field closures (S1)

ttraenkler:
## Summary

`c[1+1]()` / `c[String(1+1)]()` where `[1+1] = () => …` is a class field returned **0** instead of invoking the closure (standalone + host).

**Root cause:** for a computed-name class FIELD, TypeScript tracks no member named `"2"`, so the element-access callee `c[1+1]` carries no call signature (`compileCallableElementAccessCall` bailed) and it is NOT a prototype method (no `ClassName_2` in funcMap) — every dispatch path missed it. The struct-field READ already canonicalises the key (numeric/string) to field `"2"` and returns the closure; only the **invocation** was dropped to the `ref.null.extern` fallback.

**Fix** (`src/codegen/expressions/calls.ts`): at the two element-access-call fallback sites (resolved-key-no-method + unresolved-key), when the receiver is a user-class instance, route the read + call through the existing `ref.test`-guarded dynamic `call_ref` machinery (`tryEmitInlineDynamicCall`) that an `any`-typed identifier call already uses. Two gates (`elemAccessReceiverIsUserClass` / `classInstanceHasField`) keep primitive/array/host receivers on their existing lowering; a non-closure field value hits the safe default arm (historical null).

## Scope
**S1 only** — read-side canonicalisation + computed-key method dispatch on CLOSED class structs (no new substrate). **S2** ($props overflow table for genuinely-runtime-key members like `[f()]`) remains OPEN under #3166, per the re-grounding in the issue file.

## Impact (measured)
`cpn-class-*fields*` standalone: **34 → 64 pass (+30)**, zero in-family regressions; gc/host lane matches (64/124). The remaining ~60 are the genuinely-runtime-key variants (S2).

## Tests
`tests/issue-3166.test.ts` — 6 standalone cases (numeric/string computed-key invocation, arg passing, named-field no-regression, value-field read control, full instance-assert shape). All green. Existing #1298/#1306/#1063 closure + element-access dispatch suites still pass.


ttraenkler:
Merge-group park triage (from the tech-lead shepherd): the failed merge_group parked #2956/#2947/#2952 together on the #3189 trap-ratchet (oob 58→62, +4). The 4 newly-trapping tests are all `built-ins/TypedArray/prototype/set/BigInt/*` (array-arg-offset-tointeger, array-arg-primitive-toobject, typedarray-arg-offset-tointeger, typedarray-arg-set-values-diff-buffer-same-type). #2956 is zero-src collateral (re-admitted); #2947 was cleared by a controlled A/B (reverting its files → 0 per-test diff over all 60 TA.set tests). **By elimination, #2952 (#3166) is the prime remaining suspect** for the +4 oob. Please A/B this PR (revert the calls.ts computed-key change, run the 4 named TA.set/BigInt tests, diff) to confirm before re-admitting — if it's the culprit, fix the oob traps; if it's clean too, this may be a PR-interaction and needs the isolated re-validation. Hold stays until diagnosed.

PR #2983, opened 2026-07-13
fix(#3219): standalone reflective Date.prototype.<getter>.call brand check + native body

ttraenkler:
## What

Routes `Date` through `tryEmitNativeProtoReflectiveCall` (`src/codegen/expressions/calls.ts`) and adds a native reflective getter body (`emitDateProtoMemberBody`, `src/codegen/expressions/builtins.ts`) so a value-materialized `Date.prototype.<getter>.call(recv)` runs **host-free** under `--target standalone`.

## Why

Before this, the reflective Date call fell through to the legacy value-erased `.call` lowering, which **dropped `thisArg` and returned 0**. Measured on pristine main (process-isolated standalone lane):

- `Date.prototype.getTime.call(new Date(1000))` -> `0` (should be `1000`)
- `Date.prototype.getTime.call({})` -> `0` (should **throw TypeError**)

That missing throw is the **assert #2** failure behind the bulk of the #3174 Date brand-check gap (`built-ins/Date/prototype/*/this-value-non-date.js` + `this-value-non-object.js`).

## How

`emitDateProtoMemberBody` implements the §21.4.4 `thisTimeValue` [[DateValue]]-brand preamble #3174 asks for: recover `this` (closure-param 1) via `any.convert_extern` + `ref.test $Date`; a receiver without a [[DateValue]] slot throws a catchable TypeError. For a genuine Date it reads [[DateValue]] (field 0), computes the zero-arg getter, and boxes `f64 -> externref` (the uniform closure result). Setters/formatters return `null` -> `ensureStandaloneNativeMethodClosure` mints no closure and the reflective call falls through to the legacy path **unchanged (no vacuity introduced)**.

**Anti-bloat (#3174):** the calendar/time getter arithmetic is **extracted** from the direct-call kernel into a shared `emitDateZeroArgGetterFromTsLocal`, called by BOTH the direct path and the reflective body — no duplicated Date kernel. The direct path re-asserts `ensureDateCivilHelper` after the shared call to keep its formatter arms byte-identical.

All new emission is `ctx.standalone`-gated by construction (the reflective-proto body path is standalone-only) -> **zero host-mode impact**.

## Measured (process-isolated standalone lane, branch vs pristine-main control)

Full `built-ins/Date/prototype/<getter>/` suite (18 getters, 144 files):

| lane | pass | fail |
| --- | ---: | ---: |
| pristine-main control | 108 | 36 |
| this branch | **144** | **0** |

**+36 genuine fail->pass, ZERO pass->fail regressions.** The flips are the reflective `this-value-non-date.js` / `this-value-non-object.js` brand rows (18 getters x 2). New `tests/issue-3215.test.ts` (5 cases) proves **non-vacuity** — the reflective happy path returns the CORRECT component value (getTime=1000, getUTCHours=1, getFullYear=1970, …), Invalid Date -> NaN, plus the brand throws and a direct-call control — all host-free (zero imports).

**Byte-identity:** `prove-emit-identity` reports IDENTICAL across gc/standalone/wasi for the example corpus + a dedicated Date-getters module — the direct-call kernel emission is unchanged by the extraction.

## Scope / follow-on

Getter slice of #3174. Setter brand rows, the `S15.9.5_A6` top-level tests, `toISOString`/`Symbol.toPrimitive`, and coercion-order rows stay under #3174 as documented follow-on (setter/formatter reflective bodies remain legacy-fall-through, unchanged).


ttraenkler:
Renumbered from #3215 to **#3219**: a different session's #3215 (`array-hof-sparse-read-trapsafe`) won the concurrent `claim-issue --allocate` race and landed on main first, so this slice took the next free id to clear the `check:issue-ids:against-main` gate. Same code; the issue/test files and `(#3215)`→`(#3219)` comments were repointed.

PR #3104, opened 2026-07-15
fix(#3285): thread expected error type through assert_throws harness (slice 1)

ttraenkler:
## Slice 1 of #3285 — `transformAssertThrows` error-type threading (test-infra only)

Fixes **fix #1** of #3285. Fixes #2 (`stripUndefinedAssert`) and #3 (full `strip*` inventory) are banked for a follow-up window — #3285 stays `in-progress`.

### Problem
`transformAssertThrows` read `args[0]` (the expected error constructor, e.g. `TypeError`) and **discarded it** — `assert.throws(TypeError, fn)` became `assert_throws(fn)`, and the synthetic `assert_throws` only checked *'did anything throw.'* A codegen bug that throws the **wrong** error type (e.g. `RangeError` where the spec mandates `TypeError`) read as a false pass.

### Change (`tests/test262-runner.ts` only — no `src/` change)
1. `transformAssertThrows` now emits `assert_throws(ErrorType, fn)` — keeps `args[0]`, drops only the optional 3rd message arg. Flows through the `assert.throwsAsync` rewrite path too.
2. The synthetic `assert_throws` / `assert_throwsAsync` shims now take `(ErrorCtor, fn)` and verify the caught error MATCHES the expected type before counting a pass: `e instanceof ErrorCtor`, then a `.name`-vs-`ErrorCtor.name` fallback for host-opaque error shapes. Both checks are inline (a helper-function matcher trapped on a null-pointer deref in compiled code). `assert_throwsAsync`'s thenable-return path stays untyped (the rejection reason can't be inspected synchronously) — documented limitation; the sync-throw path IS type-checked.

Design confirmed by probing compiled TS→Wasm: `instanceof` against the in-module error ctor discriminates correctly (incl. subclass — `assert.throws(Error, …)` matches a `TypeError`) and works for the `Test262Error` user class; `.name` covers host-opaque shapes.

### Validation (scoped, JS-host lane, real `runTest262File` pipeline)
A **drop is expected and correct** per the acceptance criteria (previously-inflated false-passes becoming real fails).

| scoped batch | before | after | flipped pass→fail | fail→pass |
| `Reflect` + `TypedArray/prototype/set` (116 assert.throws files) | 62 | 45 | **17** | 0 |
| `Map`+`Set`+`Array.copyWithin` control (228 files) | — | 212 (93%) | correct throws preserved | — |

All 17 flips are `Reflect/**/return-abrupt-from-*` / `arguments-list-is-not-array-like` tests that assert a *specific* error (`Test262Error` from a throwing `toString`/abrupt coercion) propagates. Instrumenting the caught error in the real wrapped run shows the compiler throws a **different named error** (diagnostic class 20 for all 17), NOT the spec-mandated `Test262Error` — a genuine compiler conformance gap owned by **#3284**, not this test-infra fix. **Zero false-negatives** (no correct `Test262Error` missed by the matcher). Full-suite delta measured by CI.


ttraenkler:
PR-queue shepherd note — DEFERRED to a future budget window (tech-lead decision, 2026-07-15).

**Why this is parked, and what it will take to land (for whoever picks this up next window):**

This PR is auto-parked (`hold`) because its `merge_group` run failed the test262 regression gate: net -2664 pass, **2668 regressions** (assertion_fail 2647, type_error 13), multiple buckets over the 50 limit (class/dstr 168, object/dstr 84, Temporal prototypes 63-115, async-generator/dstr 56). **This is NOT a codegen regression** — it is the *intended* measurement-honesty reclassification this PR set out to make (`assert_throws` now verifies the caught error TYPE, so previously-inflated false-passes correctly become fails; see the PR body and #3285 acceptance criteria).

**The obvious fix (bump `ORACLE_VERSION` and let the forward-bump auto-rebase absorb it) is NECESSARY but NOT SUFFICIENT here.** Traced `scripts/diff-test262.ts`:
- A forward oracle bump activates "rebase mode," but rebase mode only *excuses* flips tagged `vacuous` (`vacuousReclassification` / `isVacuousResult`). That is exactly why the #3086 v2 precedent self-landed 1438 flips — they carried the `vacuous` marker.
- #3285`s 2668 flips are **plain assertion-fails with a changed wasm-hash**, not vacuous. They therefore land in `regressionsWasmChange = 2668`, which is *not* excused, and rebase mode then gates on `ORACLE_REBASE_DRIFT_TOLERANCE = 25` (diff-test262.ts:1040): `2668 > 25` -> GATE FAIL. The per-bucket-50 concentration check and the #1668 catastrophic guard (threshold 200, reads the printed "Regressions with wasm-hash change: 2668") would also trip on the promote-baseline run.

**So landing this requires the full temporary-lever dance**, driven end-to-end in a single window:
1. Bump `ORACLE_VERSION` 3 -> 4 in `tests/test262-oracle-version.ts` (+ history entry).
2. Temporarily raise `ORACLE_REBASE_DRIFT_TOLERANCE` (25 -> >2668), the #1668 catastrophic threshold (200 -> >2668), and waive/raise the per-bucket-50 limit — enough to let 2668 non-vacuous flips through `merge_group`.
3. Land; let `promote-baseline` re-seed host+standalone baselines at v4.
4. Revert the levers in a follow-up.

Raising those guards ~100x weakens the safety net for *every* concurrent PR while the levers are up, so this is a deliberate, babysat future-window operation — not a mechanical fix. Deferred until a fresh budget window can drive the raise-land-reseed-revert sequence safely.

**Separate follow-up (gate-hardening, next window):** the #3003 `check-verdict-oracle-bump.mjs` gate did NOT flag this verdict-logic change — its `VERDICT_SIGNAL_RE` only matches `status:` verdict-literal assignments, not verdict-tightening inside the `assert_throws`/`assert_throwsAsync` shim body. That false-negative is worth closing so a future oracle-affecting change to the runner shims is caught at PR time.

Leaving the `hold` label in place and the branch untouched.

ttraenkler:
Follow-up (re: "could admin-merge bypass the gate?") — **admin-merge is NOT a valid shortcut here, and would make things worse.**

Admin-merge only bypasses the **PR-level required checks** (the `merge_group` re-validation). It does **NOT** bypass the separate **#1668 catastrophic-regression guard**, which runs on **every push to main** and gates the `promote-baseline` job. That guard would see the same ~2664 non-vacuous wasm-change flips and fail exactly the same way, so `promote-baseline` still would not run — the committed baseline stays stale at the old policy.

Net effect of admin-merging this:
- The code lands on `main`, but the baseline never catches up, so **`main` itself now carries the drift**.
- Every subsequent test262-touching PR then diffs its `merge_group` against that stale baseline, sees this reclassification as a **phantom regression**, and gets **auto-parked** — wedging the queue **for everyone** (same failure class as the -439 / #2463 precedents).
- And it is materially **harder to diagnose after the fact** (a systemic drift baked into main) than one clearly-held PR sitting in the queue.

The only real fix is getting **`promote-baseline` itself to succeed** — i.e. raising the #1668 catastrophic guard *for that job specifically* as part of the full temporary-lever dance already documented in the deferral comment above (https://github.com/loopdive/js2/pull/3104#issuecomment-4978845547) — **not** routing around the PR-level gate. Keeping this held is the correct state until that dance can be driven end-to-end in a fresh budget window.

PR #3138, opened 2026-07-16
fix(#3306): StringToNumber the toString-only object ToNumber result in standalone (was drop+NaN)

ttraenkler:
Closes plan issue #3306 (umbrella #2860; the toString-only-object ToNumber gap root-caused during #3174's residual analysis and carried through #3304).

## What

Under `--target standalone`, ToNumber of an object whose only ToPrimitive method is `toString` ran the method and threw its result away:

```ts
+{ toString(){ return "7" } }          // NaN (should be 7)
Number({ toString(){ return "12.5" }}) // NaN
new Date(0).setTime({ toString(){ return "5" }}) // NaN
```

`{valueOf}` and the #2891 valueOf-returns-object→toString fallthrough both worked — only the **no-valueOf** shape failed.

## Root cause (WAT-verified)

`tryToStringFallback` (type-coercion.ts, #866) correctly finds and `call_ref`s the `toString` closure — but all three of its result converters treated a `ref`-kind return as "object → drop + NaN" (emitted `call_ref; drop; f64.const NaN`). Under nativeStrings, a `toString(){ return "7" }` closure returns the **native string struct** (`ref $NativeString`/`$AnyString` subtype) — a ref kind. §7.1.1.1 OrdinaryToPrimitive + §7.1.4 ToNumber require the String result to convert via StringToNumber (§7.1.4.1).

## Fix

One shared `refResultStringToF64Instrs` helper: runtime `ref.test $AnyString` on the ref-kind result → hit: `extern.convert_any` + the **existing** `__str_to_number` scanner (the same helper the direct string→f64 arm uses — no fresh coercion matrix); miss (genuine object return): legacy NaN unchanged. Wired into `emitToStringResultToF64`, `emitToStringResultToF64ByKind`, and the eqref-field inline arm (the shape object literals actually hit). The runtime test (not a static typeIdx match) also converts loosely-typed (eqref/anyref) closure returns. Host lane byte-identical — strings there are externref and were already unboxed.

Gate allowances in the issue frontmatter: `coercion-sites-allow` (`__str_to_number` +2 — routing through the existing scanner, replacing a spec-violating drop) and `loc-budget-allow` for type-coercion.ts.

## Measured

- `built-ins/Number` sweep (338): +1 (`S9.3_A5_T1.js`), zero regressions.
- `built-ins/Date` sweep (594): zero regressions (the ctor `coercion-errors` rows already banked from #3174's NaN fix).
- Adjacent coercion suites (`issue-1806*`, `issue-1917`, `issue-2163`, `issue-2358*`, `issue-2598-2599`): 67/67 green.
- The measured test262 delta is small, but the semantic class (`+obj` / `Number(obj)` / setter args / ctor args with toString-only objects) is pervasive in user code.

## Follow-up (documented in the issue, out of scope)

- Fn-scoped **capturing** `toString` closures (`var log=''; { toString(){ log+='x'; return 0 } }` inside a function body) trap `dereferencing a null pointer` in the closure-dispatch guard — the funcref sig-test misses (the #2873 wrapper-chain RTT class) and the guard's `ref.null` + `ref.as_non_null` traps instead of degrading. This blocks `Date/coercion-order.js` / `UTC/coercion-order.js` and needs the closure-ABI treatment (cast to wrapper root + funcref sig-dispatch), not a local patch.
- `any`-typed receivers (`+ (arg: any)`) trap on a different path (`__to_primitive` $Object dispatch) — also noted.

## Tests

`tests/issue-3306.test.ts` — 8 standalone host-free tests: unary plus / `Number()` / Date-setter arg / non-numeric string → NaN / number-returning toString passthrough / §7.1.1.1 valueOf-first order / #2891 fallthrough non-regression / `new Date(y, m)` coercion order.

### DIFF
--- a/plan/issues/3306-standalone-tostring-only-tonumber.md
+++ b/plan/issues/3306-standalone-tostring-only-tonumber.md
@@ -1,0 +1,90 @@
+---
+id: 3306
+title: "standalone: ToNumber of a toString-only object drops the native-string result to NaN (§7.1.1.1 OrdinaryToPrimitive → §7.1.4.1 StringToNumber)"
+status: in-progress
+assignee: ttraenkler/sendev-date-3174
+created: 2026-07-16
+updated: 2026-07-16
+priority: high
+feasibility: medium
+model: fable
+task_type: bug
+area: codegen
+es_edition: multi
+language_feature: coercion
+goal: standalone
+umbrella: 2860
+sprint: current
+horizon: s
+related: [3174, 3304, 2891, 866, 1806]
+origin: "root-caused during #3174 residual analysis (Date ctor coercion-order rows); carried through #3304 as the next-next candidate"
+loc-budget-allow:
+  - src/codegen/type-coercion.ts
+coercion-sites-allow:
+  # __str_to_number +2 — NOT a fresh hand-rolled matrix: routes the toString
+  # closure result through the EXISTING StringToNumber scanner (the same
+  # helper the direct string→f64 arm uses), replacing a spec-violating
+  # drop+NaN.
+  - src/codegen/type-coercion.ts
+---
+
+# #3306 — standalone: toString-only object ToNumber → NaN
+
+## Problem
+
+Under `--target standalone` (nativeStrings), ToNumber of an object whose only
+ToPrimitive method is `toString` runs the method but throws away its string
+result:
+
+```ts
+var arg = { toString: function() { return "7"; } };
++arg;                      // NaN (should be 7)
+Number(arg);               // NaN (should be 7)
+new Date(0).setTime(arg);  // NaN (should be 7-ish ms)
+```
+
+`{ valueOf }` and `{ valueOf-returns-object + toString }` (#2891's
+fallthrough) both work — only the **no-valueOf** shape fails.
+
+## Root cause (WAT-verified 2026-07-16)
+
+`tryToStringFallback` (type-coercion.ts, #866) DOES find and `call_ref` the
+`toString` closure — but every one of its three result-conversion sites treats
+a `ref`/`ref_null`-kind return as "object → drop + NaN":
+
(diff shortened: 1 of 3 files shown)


ttraenkler:
Heads-up from the queue sweep: the `quality` job is failing on the **Coercion-site drift gate** (run 29506090111) — this change adds 2 net-new hand-rolled `__str_to_number` call sites in `src/codegen/type-coercion.ts` (32→34), which the gate reads as growing the hand-rolled coercion vocabulary instead of routing through the shared coercion engine (#1917/#2108).

Two ways to clear it:
1. Route the new ToString-only-object ToNumber logic through the existing coercion engine instead of fresh hand-rolled calls, or
2. If the growth is an intentional migration step, add to the #3306 issue file's YAML frontmatter in this same PR:
```yaml
coercion-sites-allow:
  - src/codegen/type-coercion.ts
```

Not touching the branch — leaving the fix to whoever owns it.

ttraenkler:
Update from tech lead: this is NOT a hand-rolled-coercion-matrix growth problem — your `coercion-sites-allow: - src/codegen/type-coercion.ts` declaration in the #3306 issue file is correct and exactly matches the documented format.

Root cause: `parseFrontmatterList` in `scripts/lib/change-scope.mjs` breaks its block-list scan on the first line that doesn't match the list-item pattern — including `#`-comment lines. Your frontmatter has explanatory comments *before* the `- src/codegen/type-coercion.ts` line, so the parser hit the comment first and returned an empty allow-list, silently granting nothing. Confirmed via direct repro; not specific to your PR — this would bite any allow-list with a leading comment.

Fix is in #3148 (loopdive/js2#3148), currently in CI. No changes needed to your issue file's frontmatter — once #3148 merges, `git merge origin/main` into this branch and the existing declaration should be read correctly and the gate should pass as-is.

PR #3161, opened 2026-07-16
feat(#3227): S1 — async post-drain verdict re-read (oracle v5) + per-test capture of deferred continuation throws

ttraenkler:
## What

Slice 1 of #3227 (default-lane async harness vacuity, 1,690 records). Runner/harness fix — no compiler changes.

1. **Async post-drain verdict re-read.** The JS-host lane schedules `.then`/await continuations on the HOST microtask queue, which structurally cannot drain while `test()` is still on the Wasm→JS stack (`__drain_microtasks()` is a deliberate no-op on this lane, #2895 PATH B). So async tests' verdicts were read BEFORE their assertion-bearing callbacks ran — they run immediately after `test()` returns (verified empirically; the callbacks were never "dropped"). `wrapTest` now emits an exported `__result()` (same verdict logic as the `test()` epilogue) for async-flagged tests; `runTest262File` yields 2x `setImmediate` after a sync `1`/`-262` and re-reads the verdict. Runtime-negative tests and sync assert failures (`__fail` is first-wins) keep their sync semantics.
2. **Per-test capture of deferred continuation throws (fork-worker-kill fix).** A post-return continuation that throws (wasm trap / Test262Error escaping a `.then` reaction) previously fired **unattributed between tests** and could kill the whole vitest fork worker as an uncaughtException/unhandledRejection. The drain window now installs temporary handlers and scores the throw as THAT test's failure.
3. **ORACLE_VERSION 3 → 4** — this is a verdict-logic change; forward-monotonic bump auto-rebases in diff-test262.ts.

## Measured delta (sampled; full table in the issue file)

| Population | Flip | Extrapolated |
|---|---|---|
| 1,680 vacuous-callback records (n=60) | → honest **pass** | **+~420** |
| 1,680 vacuous-callback records | → honest fail (already `fail` today, now with real assert indices) | ~1,040 (no pass-count change) |
| 1,680 vacuous-callback records | stay vacuous (S4 residual) | ~140 |
| 3,503 currently-passing async-flagged (n=80) | → honest **fail** (post-await asserts finally run) | **−~875** (CI 525–1,225) |

**Net raw pass ≈ −455 — an INTENTIONAL honesty regression, lead-approved 2026-07-16** (precedent: #3086 owner-approved vacuity re-baseline). If the `merge_group` auto-parks this PR on the pass-count drop: the drop IS the change, per the table above — the −875 cluster on `class-elements async-gen … v.value` (await-NaN, S2) and `yield-star-next-then-*` `done`/value (async-gen IteratorResult, S3) are the follow-up work definitions cross-linked in the issue.

## Oracle-collision ordering (#3111)

Draft PR #3111 (standalone host-backed-pass rejection) also drafted a 3→4 bump. **This PR assumes it lands FIRST at v4**; whichever of the two lands second must re-bump (4→5) with its own history entry — noted in `ORACLE_VERSION_HISTORY` and in the issue file. If #3111 lands first after all, this PR takes the conflict on the `ORACLE_VERSION` line and I re-bump to 5 before it re-enters the queue.

## Root cause (verified; issue file has full detail + slice plan S1–S4)

- Verdict-read-before-drain (this PR) — the 1,690-record driver.
- `await <host promise>` yields NaN synchronously (S2, compiler bug, repros in issue).

## Validation

- Issue-cited sample `async-generator/dstr/ary-ptrn-elem-id-iter-complete.js`: vacuous → pass. `dynamic-import/namespace/promise-then-ns-set-prototype-of.js`: vacuous → honest fail with real assert index.
- wrapTest consumer unit tests (issue-1049/1450/1385/1567/1318-locator): 24/24 pass.
- Re-validated after merging upstream/main (0886111adc): identical verdicts.

Issue: plan/issues/3227-default-lane-async-harness-callback-vacuous.md (status stays `in-progress` — S2–S4 remain; slice plan written for dispatch).


ttraenkler:
Reconciled with #3104 (#3285 assert_throws error-type precision), which landed first and took oracle v4:

- **ORACLE_VERSION**: this PR re-bumps to **v5** (was drafted as 3→4) per the whichever-lands-second-re-bumps rule the issue file itself documented. Main's v4 history entry kept verbatim; the #3227 entry re-labeled v5.
- **Runner**: #3104's name-side-channel shim rework (`__expected_throw_name`, strict `.name` match, no 2-arg call shape — #3315 avoidance) is untouched. Composition verified end-to-end, not assumed: the shims consume the side channel synchronously at entry (no clobber inside the post-drain window) and set sticky `__fail`, which `__result()` re-reads. Probes: sync wrong-type throw → fail; **deferred continuation wrong-type throw → fail with real assert index** (pre-#3227 a premature sync pass, pre-#3104 any throw passed) — both tightenings stack.
- **#3303 `regressions-allow: 1300`** declared in the issue frontmatter for the lead-approved S1 honesty regression (measured ~875 host pass→fail, CI upper 1225 + margin). NOTE: ceiling assumes the baseline is re-seeded at v4 (post-#3104 promote-baseline, in flight as of 17:50Z); a diff against a stale pre-#3104 baseline would also carry #3104's ~1358 flips and park — that is a wait-for-promote-baseline situation, not a regression.
- Scoped tests: issue-2096/3003/3187/3303 all green (62/62); prettier + tsc clean; check-verdict-oracle-bump passes (4→5).

ttraenkler:
Lifting the hold: this was the author-session's own deliberate label (applied 17:59Z pending the in-flight baseline promotion, per the 17:56Z comment). That baseline promotion has settled (sharded baseline refreshed at 32,999/43,106, commits ea92e39240/a1939aa065), and the author session ended at the usage-window boundary, so I'm completing its stated intent. The net −455 delta is the owner-approved honesty correction documented in the PR body (oracle v5); merge_group re-validation should be read against that table.

PR #3176, opened 2026-07-16
fix(#3201): expando-method dispatch on native receivers — Sputnik getClass cluster (~75 default-lane fails)

ttraenkler:
## Summary

Slice of #3201 (umbrella #3185, default JS-host lane): the single largest remaining coherent cluster in the search/structural family — **65 `result is Array object. Actual: null` + 10 getClass-value fails** across `splice`/`slice`/`concat` `S15.4.4.*` — is the Sputnik classifier idiom:

```js
arr.getClass = Object.prototype.toString;
if (arr.getClass() !== "[object Array]") ...
```

An UNKNOWN method on a statically-typed struct/vec receiver had no arm in the receiver-method ladder, so the call fell to the calls.ts graceful drop+null fallback — `arr.getClass()` silently returned `null`.

## Fix (three coordinated pieces, JS-host lane only)

1. **`call-receiver-method.ts`** — end-of-ladder arm: unknown methods on `ref`/`ref_null` receivers delegate to the generic `__extern_method_call(recv, name, args)` (the #799 WI3 / #3123 machinery). Standalone/WASI untouched (acceptance #5).
2. **`calls.ts`** — `emitFnctorSubclassDynamicMethodCall` gains `rawStructReceiver`: the receiver marshals as the RAW wasm ref (`extern.convert_any`), not `coerceType`'s `__make_iterable` COPY — the `_wasmStructProps` expando sidecar is keyed by raw struct identity, so a copy could never find the stored method. Existing #3123 call sites keep their bytes (flag defaults false).
3. **`runtime.ts`** — `_wrapVecForHost`'s array-backed host view now surfaces sidecar expandos in its `get`/`has` traps (an own expando shadows `Array.prototype`, ordinary lookup order), callable-wrapping raw closure structs at read time (module-init writes run during instantiation, before `setExports`, so write-time wrapping can't resolve exports).

## Validation

- 13/14 sampled cluster files flip fail→pass via `runTest262File` (the 14th is a different mechanism — expando `splice` on an array-like — still open on #3201).
- `tests/issue-3201-expando-method.test.ts` 5/5 (classifier idiom on splice result + literal, user-closure expando, plain-object expando, no shadowing of real Array.prototype methods).
- `array-methods` / `object-literals` / `object-methods` / `object-literal-getters-setters` / prior #3201 trap-safety suites green; `anon-struct`'s 3 fails pre-exist on clean main.

#3201 stays `in-progress` — remaining mechanisms documented in the issue's progress section.


ttraenkler:
Un-parked after diagnosis + fix (CI-fix dev, taking over from the context-exhausted author session).

**Park root cause (run 29546826346, gates: check for test262 regressions + merge shard reports):** the new end-of-ladder #3201 arm claimed EVERY unknown-arm method call on a ref/ref_null receiver and delegated to `__extern_method_call`, whose host mirror does not carry compiled class methods. That hijacked calls the downstream compiled paths handled correctly → 215 merge_group regressions (net −60): ~211 Temporal (`since`/`until`/roundingmode + TemporalHelpers definition cascades), class static-field arrows (`class C { static f = () => this }`), subclass-inherited `equals`.

**Fix (86cfc7d207):** gate the arm on true expandos — `recvTsType.getProperty(method) === undefined` (declared members decline → pre-#3201 path restored) AND receiver not a class-declared instance type (a class extending an unresolvable base, e.g. `extends Temporal.PlainYearMonth`, has an incomplete member set, so a getProperty miss there is not expando evidence).

**Verified in-process:** 25/25 sampled merge_group regressions flip back to pass; Sputnik getClass cluster wins retained; expando suite 5/5; oracle-ratchet +0; tsc + prettier clean. ~3 vacuous-callback improvement files return to main parity (statically inseparable from the breakage). Branch re-merged with main twice (was DIRTY). Not re-enqueueing manually — auto-enqueue owns admission.

PR #3177, opened 2026-07-16
fix(#3328): JSON.stringify object arg serialised "null" — toJSON miss-guard vs $undefined singleton

ttraenkler:
## Orphaned silent regression (assigned by tech lead)

`tests/issue-2933-json-stringify-value.test.ts` failed 3/9 on main: every OBJECT through a reified `JSON.stringify` value (`const f: any = JSON.stringify; f({a:1})`) — or any `JSON.stringify(x)` inside an `any`-arg closure — returned `"null"`. Scalars fine; direct inline calls fine (which masked it). Silent since Jul 4 because issue tests are not uniformly wired into required CI (#3008 class).

### Root cause (bisected)
First bad commit `6f7f93c856` — the #2106 **$undefined-singleton default-ON flip**. Under the regime, `__extern_get`'s MISS value is the non-null tag-1 `$AnyValue` box, but the §25.5.2 SerializeJSONProperty toJSON guard in `src/codegen/json-codec-native.ts` only null-checked. A missing `toJSON` therefore read as callable → `__call_fn_method_1`'s closure ref-test failed → returned null → the "toJSON returned nullish ⇒ JSON null" arm serialised `"null"` for every object on that path. (The original code comment specified the intended closure ref-test — never implemented.) `JS2WASM_UNDEF_SINGLETON=0` A/B-confirmed on current main.

### Fix
Extend the guard: a `$AnyValue` box is never a callable closure — require `!(any.convert_extern(m) is $AnyValue)` in addition to non-null. Tag-agnostic, **regime-gated** (`undefinedSingletonActive`) so legacy modules stay byte-identical (verified: legacy-lane sha256 on-branch == pristine main).

### Validation
- `tests/issue-2933-json-stringify-value.test.ts`: 3/9 fail → **9/9 pass**
- Genuine `toJSON` methods still fire — both regimes, both call shapes (direct + reified value)
- Probes: valObj / valAnyVarObj / userClosureObj "null"→`{"a":1}`; nested object 13 ✓
- loc-budget / coercion-sites / oracle-ratchet gates pass locally (allowance in issue frontmatter)

### Noted residual (pre-existing, NOT this regression)
A TYPED (non-any) object literal through the reified value still serialises "null" in BOTH regimes — nominal-struct → dynamic-call boundary gap predating #2106; documented in the issue file as a follow-up candidate.

Issue file `plan/issues/3328-*.md` created via `claim-issue.mjs --allocate`; carries `status: done`. Cross-links #2933/#2106/#3008/#2166.


ttraenkler:
Un-parked (CI-fix dev). The 04:32Z park (run 29554870531, gate: check for test262 regressions) is **stale-baseline collateral, not a PR regression** — the run had 0 regressions / 0 improvements / net +0; the only gate failure was the #3189 oob trap ratchet (45→51, +6), all six being `TypedArray/prototype/set/BigInt/*` files.

Proof it's main-side drift: in baselines@01c0962 (main@deee48c, the stale baseline the run compared against) those six files fail with "undefined is not a constructor" (catchable, not oob); in baselines@f4d1367 — captured from main@956e09b9ec, this run's exact merge base, WITHOUT this PR — the same six fail as "offset is out of bounds" (oob trap). The failure-mode flip landed on main between the two baselines. The identical +6/same-six-files signature also parked unrelated #3198 (cross-PR drift confirmation). The baseline has since re-promoted with oob=51, so a fresh merge_group run compares 51 vs 51 and the ratchet passes. No code change needed; not re-enqueueing manually — auto-enqueue owns admission.

PR #3182, opened 2026-07-16
fix(#3315): preserve undefined identity through parameter array-destructuring — widen f64 bindings + sentinel-aware boxing

ttraenkler:
## Summary

Fixes #3315 (local plan issue): a parameter array-pattern binding receiving a runtime `undefined` silently held a wrong value (NaN-as-number) — the "sibling destructured binding corruption" found during the #3104 assert_throws landing. Three stacked layers, all fixed:

1. **Rep** — elements without a per-element default (`{ w: [x, y, z] = [4, 5, 6] }`) resolved f64 locals from the checker's pattern-default fiction (`number`), so `undefined` degraded to NaN and `y === undefined` constant-folded false. Such bindings now widen to externref (`resolveBindingElementType`, param patterns only, f64 only) and their reads skip the checker-type unbox narrowing (`fctx.undefWidenedLocals`). Decl destructuring / for-of numeric reps untouched; native `i32` annotations untouched.
2. **Box** — `[7, undefined, ]` lowers to an f64 vec carrying the UNDEF_F64_BITS signaling-NaN sentinel (#1024). Three box sites turned that sentinel into a boxed NUMBER: the generic coerceType f64→externref arm, the destructure vec-conversion loop (`boxToExternref`), and the host read boundary (`__vec_get` — where `__make_iterable` reads elements and the sentinel is unrecoverable JS-side). All three now map sentinel-bits → real `undefined` (host `__get_undefined` / standalone tag-1 singleton) — same discipline as the $Hole→undefined map (#2001) and the standalone any-box sentinel arm (#2979). Quiet computed NaNs (0x7FF8…) are unaffected.
3. The issue's 1-arg vs 2-arg **context sensitivity** was these destructure arms flipping with compilation context; with every arm fixed both shapes agree.

## Validation

- Acceptance probe pair (wrapTest'd `meth-obj-ptrn-prop-ary.js`, y/z undefined-asserts REINSTATED — stronger than the corpus wrap): 1-arg and 2-arg shapes both pass, host + standalone (pre-fix: both fail).
- `tests/issue-3315.test.ts`: 10/10 (direct `===`, any-param isSameValue harness shape, class-as-value 2nd arg, whole-pattern default still fires, ToNumber arithmetic; each on both lanes).
- 280-file dstr-family test262 sweep × both lanes: zero status diffs vs origin/main control.
- Equivalence suite: failure profile identical to main (36 failed / 1607 passed / 3 todo — pre-existing).
- Spot probes: two incidental fixes (host array-hole read, standalone short-arg fn param), zero regressions; residual pre-existing host typed-vec OOB case documented in the issue file.
- tsc, prettier, check:ir-fallbacks: clean.


ttraenkler:
**Shepherd diagnosis** (park review, 2nd sweep — no owner action since the auto-park at 00:37 UTC):

This auto-park's `merge_group` failure ("check for test262 regressions") was reviewed and determined to be a **ratio-gate false positive**: 1 regression vs. 3 improvements in the merged-state test262 delta. The regression count is below the escalation floor and is outnumbered by improvements — this reads as small-sample noise on the merge-shard split, not a genuine merged-baseline regression from this PR's change.

Per the auto-park protocol I'm not removing the `hold` label or re-enqueueing myself (this is a parallel session's PR, not mine to touch the branch on) — flagging so the owning session/dev can confirm the diagnosis and clear the hold themselves. If unaddressed by the next sweep I'll escalate to the tech lead for a call.

ttraenkler:
**Correction to my earlier comment**: the "ratio-gate false positive" read (inherited from a prior handoff) was independently re-derived and found WRONG.

Actual root cause (merge_group run 29545027665, "Test262 Sharded", queue SHA 87185aa326312fef77081ec9332296c6af811ff0, failed 2026-07-17T00:31 UTC): `test/built-ins/Math/log2/log2-basicTests.js` assert #8 fails — `Math.log2(undefined)` should be `NaN`, isn't. Verified against a clean current-main baseline: main alone PASSES this test, so the failure is unique to this branch and maps directly to the PR's own change (f64-widened `undefined` param arg not landing as the NaN sentinel through `ToNumber(undefined) = NaN`) — a real, targeted regression, not drift/flake.

**`hold` stays in place.** This is a genuine content bug needing a fix in the PR's own diff, not a re-enqueue candidate.

Branch status for whoever picks this up: **#3182 is currently unowned** — no active lock in `issue-assignments` for this exact branch (`issue-3315-standalone-extra-arg-dstr-corruption`; the `3315` lock on file points to a different sibling branch, `issue-3315-json-tojson-undefined-gate`), and no local worktree exists for it. Free to claim.

ttraenkler:
Shepherd: this park was collateral of the fleet-blocking #1616 quality-gate bug (broken wildcard issue-link in #3341), fixed on main at 455b451211 (confirmed green push run 29561924626). Not a regression in this PR. Clearing `hold` and re-enqueuing once.

ttraenkler:
Shepherd: this park is the same fleet-wide `#3189` oob-trap ratchet cluster affecting #3203 (+6, the BigInt TypedArray `array-arg` tests, net +0 pass, 0 real regressions in pass→other) — not caused by this PR's own diff (undefined-identity through param destructuring is unrelated to TypedArray.set). Per tech lead: this whole ratchet cluster clears once #3206 (opus-3335's deterministic trap fix) lands. Leaving `hold` in place, will re-enqueue after #3206 merges.

ttraenkler:

PR #3183, opened 2026-07-16
feat(#2956 L2): fixed-shape object aggregates through the linear-IR overlay

ttraenkler:
L2 aggregates sub-slice of plan issue #2956. **Stacked on PR #3179** (vec mutation — same files); its commits show in this diff until it merges, after which this PR reduces to the aggregates delta. Do not enqueue before #3179 lands.

## What

`object.new` / `object.get` / `object.set` (fixed-shape anonymous object literals) now lower through the linear-IR overlay under `JS2WASM_LINEAR_IR=1`:

- **Layout (IR-internal by design)**: i32 arena pointer; the direct literal path's exact header (tag `0x10` u8 @0, payload u32 @4); uniform 8-byte f64 slots at `8 + 8*i` in the IR shape's canonical (name-sorted) field order.
- **Boundary discipline (the load-bearing gate)**: the direct path lays fields out in CHECKER order, so an object value must never cross the IR↔direct seam — `verifyIrBackendLegality` rejects object-typed params/results for linear while interior values stay legal; non-f64 fields (bool/i32/string) demote via a legality arm.
- **Emitter**: `emitAggregateNew` = `__linear_ir_obj_new` (returns the fresh ptr on top of the stacked field values) + a value-first `__linear_ir_obj_init_f64(value, ptr, offset) -> ptr` fold — zero scratch locals; field get/set are single `f64.load`/`f64.store` at immediate offsets.
- **Shared-code change of substance**: `IrObjectStructLowering` grew an optional `valueType` honored by `lowerIrTypeToValType` (`?? (ref $struct)`) — WasmGC layouts leave it absent, gc lane untouched.

## Net-new capability

Field WRITES on anonymous object literals: the direct linear path fail-louds on them ("Unknown property assignment" — classLayouts-only support). The overlay compiles and runs them; the test asserts BOTH the direct-path compile error and the overlay's spec value, so the gap closing flips loudly.

## Validation

- `tests/issue-2956.test.ts` **15/15**: read parity vs direct, write capability, demotion gates (boundary + bool field), flag-off SHA byte-identity.
- linear basic/array/classes + cross-backend-diff: 46/46; tsc clean.
- Ratchets: `check:linear-ir` 6/6 build:4 unchanged (no aggregate rows in the playground corpus); `check:ir-fallbacks` no increases (gc lane inert).


ttraenkler:
Closing this stale stacked #2956 PR as superseded by #3200 on the fresh dependency-ordered branch symphony/porffor/2956-after-pr-3179. Its current quality failure is branch-owned LOC allowance drift on the old head, and repairing this branch would duplicate the active continuation.

PR #3211, opened 2026-07-17
fix(#3325): wire ambient declare-function host imports to deps

ttraenkler:
## #3325 — `declare function` host-dep call silently dropped

An ambient `declare function f(...)` is compiled to an `env.f` import and the
call site **does** emit `call $f_import` (codegen never dropped the call). The
miss was in the runtime: `resolveImport` classifies the import as a `builtin`
intent, and its terminal fallback for an unrecognised builtin name returned a
no-op that **ignored `deps`** — so the call ran but did nothing, silently
dropping the embedder's injected function.

### Root cause
`src/runtime.ts`, `case "builtin"`: every recognised builtin (internal `__*`
helpers + named primitives like `parseInt`/`JSON_stringify`/`Promise_*`) returns
before the terminal `return () => {};`. A user-level ambient name matches
nothing and hit that no-op.

### Fix (terminal fallback of the `builtin` case)
1. `deps[name]` is a function → forward the call to it.
2. `deps[name]` is a non-function value → expose it as a zero-arg accessor.
3. No dep AND user-facing (non-`__`) name → return a stub that throws a clear
   `TypeError` **when called** (lazy → a declared-but-unused import can't break
   instantiation/linking).
4. Internal `__`-prefixed unknown names keep the historical no-op.

### Tests
`tests/issue-3325.test.ts` (6): numeric/string arg marshaling, per-call-site
invocation, missing-dep lazy throw, unused-import instantiation, non-function
dep exposure. Adjacent declare-function / host-interop suites
(issue-1042/1052/1494/1347/2903*/2635/2693*/2752/1636/3125*/3137/1695/
promise-combinators) show **no new failures**; the pre-existing issue-820m (4)
and import-resolver (8) failures reproduce identically on clean `main`.

Closes #3325.

### DIFF
--- a/plan/issues/3325-declared-function-host-dep-call-dropped.md
+++ b/plan/issues/3325-declared-function-host-dep-call-dropped.md
@@ -15,6 +15,8 @@ horizon: s
 created: 2026-07-16
 updated: 2026-07-17
 completed: 2026-07-17
+loc-budget-allow:
+  - src/runtime.ts
 ---
 
 # #3325 — `declare function` host-dep call silently dropped
--- a/plan/issues/3325-declared-function-host-dep-call-dropped.md
+++ b/plan/issues/3325-declared-function-host-dep-call-dropped.md
@@ -1,8 +1,9 @@
 ---
 id: 3325
 title: "declare function host-dep call is silently dropped (env import bound but never called)"
-status: ready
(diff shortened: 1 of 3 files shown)


ttraenkler:

PR #3212, opened 2026-07-17
fix(#3311): add native-string vec carrier to __vec_push/__vec_pop

ttraenkler:
## #3311 — G4: `string[]` push/pop under standalone is a no-op

Slice **G4** of the #2928 `CallBuiltin` prerequisite chain (standalone interpreter track). `string[]` push/pop on a genuinely-`any` receiver was a **silent no-op** under `--target standalone`/`wasi`.

### Root cause
The carrier-generic `__vec_push`/`__vec_pop` helpers (`vec-access-exports.ts`) covered only the externref / f64 / i32 element carriers. A `string[]` under `nativeStrings` lowers to a vec whose backing array holds `(ref null $AnyString)` elements, keyed `ref_<anyStrTypeIdx>` — **not** in `mutEntries`. So `__vec_push` returned the `-1` unsupported-carrier sentinel, which the #2927 `$__vec_base` brand arm maps to `undefined` (silent no-op); `__vec_pop` returned `undefined` likewise. No regression from #2592 — it was already broken.

### Fix
Admit the native-string carrier to `mutEntries` (gated `nativeStrings && anyStrTypeIdx >= 0`) and add its value-marshaling arm to the three sites that (un)box an externref against the vec:

1. **`__vec_push`** `valueInstrs` — recover the GC string ref: `any.convert_extern` + `ref.cast_null $AnyString` before `array.set` (no numeric unbox; `ref.cast_null` matches the nullable element type and tolerates a pushed `null`/`undefined`).
2. **`__vec_pop`** `boxInstrs` — box the popped element straight back with `extern.convert_any` (anyref subtype).
3. **`__vec_set_elem`** (`vec-define-writeback.ts`) — same as push. **Required**: this consumes the same `mutEntries`, and its numeric default arm would emit an i32 into a `(ref null $AnyString)` array (**invalid Wasm**) for any module with BOTH a `string[]` and `Object.defineProperty`.

`__vec_get` / `__vec_len` / `__vec_set_len` were already carrier-generic (verified). New `nativeStrVecKey(ctx)` helper centralizes the key. The new `ref.test` arm is a sibling under `$__vec_base`, so it can't cross-match the existing carriers — order-independent, no regression to the numeric/externref arms; the `-1` sentinel path is untouched for genuinely unsupported carriers (e.g. `i64`).

### Tests
- New `tests/issue-3311-string-vec-push-pop.test.ts` — 8 tests, all pass (host + standalone-host-free). Every standalone case asserts **0 function imports**.
- Existing `tests/issue-2927-standalone-any-push-pop.test.ts` — 7/7 green.
- Broad change (adds an arm to every standalone module with a `string[]` vec) — leaning on full CI / `merge_group` for conformance.

### DIFF
--- a/plan/issues/3311-g4-string-vec-push-pop-carrier.md
+++ b/plan/issues/3311-g4-string-vec-push-pop-carrier.md
@@ -1,8 +1,10 @@
 ---
 id: 3311
 title: "G4 — `string[]` push/pop under standalone is a no-op: native-string vec carrier missing from `__vec_push`/`__vec_pop` mutEntries"
-status: ready
+status: done
+assignee: ttraenkler/senior-dev
 created: 2026-07-16
+completed: 2026-07-17
 priority: high
 horizon: s
 feasibility: medium
@@ -60,11 +62,11 @@ fall-through also returned `undefined`); this issue is the actual fix.
 
 ## Acceptance criteria
 
-- [ ] `string[]` push/pop via the any-receiver brand arm works standalone
+- [x] `string[]` push/pop via the any-receiver brand arm works standalone
       (values, length, return values correct; host-free).
-- [ ] The `-1` sentinel path still returns `undefined` for genuinely
+- [x] The `-1` sentinel path still returns `undefined` for genuinely
       unsupported carriers (no bogus boxed `-1` length).
-- [ ] Existing #2927 push/pop suite (`tests/issue-2927-standalone-any-push-pop.test.ts`)
+- [x] Existing #2927 push/pop suite (`tests/issue-2927-standalone-any-push-pop.test.ts`)
       stays green.
 
 ## Notes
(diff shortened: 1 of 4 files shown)


ttraenkler:

ttraenkler:
Closing as **superseded by #3219**, which landed the same fix on main first (merged 2026-07-17T12:10:23Z).

I independently arrived at the same root cause and the same three-site fix (`mutEntries` filter + `__vec_push` / `__vec_pop` / `__vec_set_elem`), including the `vec-define-writeback.ts` catch. #3219's implementation is strictly more robust: its `nativeStrVecElemTypeIdx()` inspects the vec's backing-array element type and handles both `$AnyString` and `$NativeString`, where this PR key-matched on `ref_<anyStrTypeIdx>` only. Its test suite (10 cases) is also a superset of this one's (8), adding grow-from-empty.

This PR went `DIRTY` precisely *because* main already contains the fix in the same lines. There is nothing additive to salvage, so resolving the conflict would only risk clobbering the better implementation. Issue #3311 is already `status: done` on main. No action needed.

PR #3213, opened 2026-07-17
fix(#3310): lift __apply_closure arity ceiling; standalone generic method-call args (G2)

ttraenkler:
## #3310 — G2: standalone generic method-call args + `__apply_closure` arity>4 lift

E5 `CallBuiltin` prerequisite (per the 2026-07-17 interpreter-backend audit). G1/G3 done; this is the remaining "args actually flow + arity" half of the #2927 audit's headline gap 2.

### Root cause (reframe)
The issue's Problem #1 blamed `emitWrapperDynamicMethodCall`'s host-gated `wantArgs`, but **all three of its callers are JS-host-gated**, so that branch is unreachable in standalone today. The genuinely-reachable standalone open-`$Object` args lane is the #799 WI3 generic bridge in `call-receiver-method.ts`, which **already** builds the arg vec natively and flows args ≤4. The one reachable blocker was the **`__apply_closure` arity ceiling**: its arity switch dispatched `n = __extern_length(args)` only to `__call_fn_method_0..4`, so a 5+-arg dynamic method call fell off the bridge and returned the undefined sentinel (`0` in numeric context) — even though `__call_fn_method_5..8` already exist (`index.ts` #2687 cap = `min(moduleMaxClosureArity, 8)`).

### Changes
- **`fillApplyClosure` (`object-runtime.ts`) — the fix.** Extend the arity switch from a hard `4` up to the highest emitted `__call_fn_method_N` (`callMethod(n) !== undefined`), **gated `ctx.standalone || ctx.wasi`** so host modules keep the 0..4 ceiling and stay byte-identical. `buildArm`/`ARG_OF` already source args positionally from the vec via `__extern_get_idx` and thread recv→thisVal / fn→closure for arbitrary `n` — this only wires up dispatchers that already existed.
- **`emitWrapperDynamicMethodCall` (`calls.ts`) — latent/defensive.** Drop the `&& !ctx.standalone && !ctx.wasi` gate on `wantArgs` and select the native `$ObjVec` push under `ctx.standalone` (mirroring the reachable #799 lane). Changes **no existing module** (the helper isn't reached in standalone today); makes it correct-by-construction for a future standalone caller. Gating stays `ctx.standalone` (not `standalone||wasi`) so the arg-less #1397 wrapper site's pre-existing wasi behaviour is unchanged.

**Out of scope:** the `if (ctx.standalone)` vs `standalone||wasi` host-builder-in-wasi mismatch at `call-receiver-method.ts` ~3085 is a separate pre-existing bug on a target not in this issue's acceptance criteria — left as a follow-up to avoid untested wasi byte-risk.

### Validation
- `tests/issue-3310.test.ts`: open-`$Object` stored closure with 1–6 args, standalone, asserting correct positional sum + **zero host imports**. The 5-/6-arg cases are the regression guard (returned `0` on main); a distinct-digit arity-5 case (`12345`) proves positional fidelity. 7/7 pass.
- Empirical delta: on `main` this repro returns correct values for 1–4 args but `0` for 5–6; on this branch all 1–6 pass with 0 imports.
- **Host byte-stability:** host binaries for open-object / Map.forEach / Array.map are SHA-identical main-vs-branch.
- `tsc --noEmit` clean; prettier clean.

### Acceptance criteria
- [x] Standalone: a dynamic method call through the generic open-`$Object` lane receives its arguments (values observable in the callee).
- [x] A 5+-arg call through `__apply_closure` dispatches (no undefined fall-through).
- [x] Host mode byte-stable; scoped suites green.

### DIFF
--- a/plan/issues/3310-g2-standalone-generic-method-call-args-arity.md
+++ b/plan/issues/3310-g2-standalone-generic-method-call-args-arity.md
@@ -1,7 +1,9 @@
 ---
 id: 3310
 title: "G2 — args-passing on the standalone generic method-call path + `__apply_closure` arity>4 lift (wantArgs is host-gated)"
-status: ready
+status: done
+assignee: ttraenkler/senior-dev
+completed: 2026-07-17
 created: 2026-07-16
 priority: high
 horizon: m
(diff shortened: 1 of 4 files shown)


ttraenkler:
Superseded by #3226 (merged) — closing as a duplicate.

#3226 landed the same fix for the one genuinely-reachable gap: the `__apply_closure` arity ceiling (`APPLY_CLOSURE_MAX_ARITY = 8` in `fillApplyClosure`, matching the `__call_fn_method_N` emission cap from index.ts #2687). My PR's arity half is functionally equivalent, so there is nothing left to land.

Verified rather than assumed — I ran **this PR's own independently-written test** (`tests/issue-3310.test.ts`: open-`$Object` stored closure at 1–6 args, positional-fidelity check, zero-host-imports assertion) against current `origin/main` (47a38d8): **7/7 pass**. #3226's own test covers arities 2/5/6/7/8, which is broader on arity than mine.

The other half of my PR (the `emitWrapperDynamicMethodCall` `wantArgs` change, i.e. the issue's Problem #1) is **not a real delta**: all three of that helper's callers are host-gated on current main — `calls.ts` #2838 (`!noJsHost(ctx)`), `calls-closures.ts` #1712 (`!ctx.standalone && !ctx.wasi`), and `call-receiver-method.ts` #1397 (passes no `callExpr`, so `wantArgs` is always false). Its standalone args branch is unreachable, so no failing test can be written for it. The issue text's Problem #1 mis-located the gap: the reachable standalone open-`$Object` args lane is the #799 WI3 bridge in `call-receiver-method.ts`, which already built its arg vec natively and already flowed args ≤4. #3226 correctly fixed the real gap (the arity ceiling) and correctly did not inherit the mis-diagnosis.

PR #3215, opened 2026-07-17
fix(#3150): standalone-native Uint8Array.fromHex(string) decode

ttraenkler:
## What

Implements `Uint8Array.fromHex(string)` standalone-native (the first slice of
#3150), replacing the `__get_builtin` dynamic-shape refusal (#1472 Phase B) that
hard-CE'd it standalone.

## How

- New `src/codegen/uint8-codec.ts`: `ensureUint8FromHex(ctx)` builds a native
  decoder over the input string's UTF-16 code units (via `__str_flatten` + a
  `__hex_digit(i32)->i32` helper), writing into the packed-`i8` Uint8Array vec
  (the same backing `new Uint8Array([...])` / `Uint8Array.of` produce standalone).
- Dispatch arm in `src/codegen/expressions/call-builtin-static.ts`, alongside the
  existing `#2592` `TA.of`/`TA.from` standalone factories.
- Spec `SyntaxError` on odd length / any non-hex character (whitespace is NOT
  skipped for hex); case-insensitive.
- Only **string-typed** arguments route here — per spec `fromHex` throws a
  `TypeError` **without** ToString coercion for a non-string, so a non-string arg
  falls through to the existing refusal (no silent wrong coercion, no regression).
- Gated on `noJsHost` (JS-host lane unaffected); no host imports (zero-import
  standalone instantiation preserved).

## Scope

This lands the `fromHex` slice. `fromBase64`, instance `toHex`/`toBase64`/
`setFrom*`, and static return-type branding (so `results.js`'s prototype/`buffer`
assertions pass — the checker doesn't know `fromHex` returns `Uint8Array`) remain
follow-ups; the issue stays open with a progress note.

## Test

`tests/issue-3150.test.ts` — 8 cases (decode values, empty, case-insensitivity,
odd-length `SyntaxError`, illegal-char `SyntaxError` incl. non-ASCII, zero-import
instantiation). All pass. `tsc` clean. No regression in the `issue-2592`/`2593`
typed-array-factory suites.


ttraenkler:

PR #3216, opened 2026-07-17
fix(#3155): native standalone join for externref arrays (Object.keys().join)

ttraenkler:
## #3155 — standalone `Object.keys(o).join(sep)` leaked `env::__array_join_any`

`Object.keys(any)` yields a boxed (externref) array. The array-`join` dispatch
routes an externref receiver through `compileArrayJoinExtern`, which delegated
to the JS `__array_join_any` host import. Under `--target standalone` that
import is unsatisfiable — the module fails to instantiate against `{}` and the
real test262 symptom is `TypeError: Cannot convert object to primitive value` on
`Object.keys(o).join(",")` (surfaced by the #86 vacuous-standalone audit;
contrast the receiver's own `.length`, already host-free via the native
`__extern_length` arm).

## Fix
`compileArrayJoinExternNative` (`src/codegen/array-methods.ts`): under
`noJsHost`, walk the externref array natively — length via `__extern_length`,
each element via `__extern_get_idx` then §7.1.17 ToString via
`__extern_toString` (all native-registered standalone) — folding with the
shared `emitStringJoinFold` over the native-string repr, mirroring the
WasmGC-vec `compileArrayJoinNative`. **Host lane byte-identical** (guarded on
`noJsHost`; the 23 host-lane assertions in issue-2131/2746/2804 stay green).

## Measured
Spread `{...a,z}`, `Object.assign`, `Object.keys().length`,
`Object.entries().length` were already host-free; only the externref-`join`
leaked. Now host-free + correct standalone: `Object.keys().join()` / `.join(sep)`
/ multi-char sep / integer-key order (`"1,2,b,a"`) / empty (`""`) / single-key.

## Tests
- `tests/issue-3155.test.ts` — permanent guard (#2093): in-wasm correctness +
  zero-`env::*`-import assertions (7 tests).
- `tests/issue-2131.test.ts` — un-skip the vacuous standalone `it.skip`,
  converted to a real in-wasm check; now passes on the true standalone lane.

## Carve-out
`Object.values()/getOwnPropertyNames().join` take a **different** dispatch arm
(receiver misclassified as `Uint8ClampedArray` → leaks
`env::Uint8ClampedArray_join`); distinct root cause, tracked in **#3342**.

Local: `npx vitest run tests/issue-3155.test.ts tests/issue-2131.test.ts` green;
`tsc --noEmit` clean; host-lane issue-2746/2804 unaffected.

### DIFF
--- a/plan/issues/3155-standalone-object-spread-assign-keys-order-gaps.md
+++ b/plan/issues/3155-standalone-object-spread-assign-keys-order-gaps.md
@@ -18,6 +18,11 @@ loc-budget-allow:
   # (#3155) the native standalone externref-join `compileArrayJoinExternNative`
   # must live beside `compileArrayJoinExtern` in array-methods.ts (+~100 LOC).
   - src/codegen/array-methods.ts
+coercion-sites-allow:
+  # (#3155) the native externref-join mirrors compileArrayJoinNative's existing
+  # `__extern_toString` element-ToString (§7.1.17) — same coercion vocabulary,
+  # relocated into the new externref-receiver arm, not novel hand-rolling.
+  - src/codegen/array-methods.ts
 ---
 
 # #3155 — standalone object spread / assign / key-enumeration gaps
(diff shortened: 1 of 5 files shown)


ttraenkler:

PR #3219, opened 2026-07-17
fix(#3311): native-string vec carrier for __vec_push/__vec_pop/__vec_set_elem

ttraenkler:
## #3311 (G4) — standalone `string[]` push/pop was a silent no-op

`string[]` under `--target standalone` / nativeStrings lowers to a vec whose
backing array element is `(ref null $AnyString)` (keyed `ref_${anyStrTypeIdx}`).
That carrier was missing from the `mutEntries` filter in
`vec-access-exports.ts`, so `__vec_push` returned the `-1` unsupported sentinel
(mapped to `undefined` by the `$__vec_base` brand arm) — `(a as any).push("x")`
on a `string[]` was a **silent no-op** standalone, and `__vec_pop` returned
`undefined`. (No regression from #2592 — it was already broken; this is the fix.)

## Fix
Add `nativeStrVecElemTypeIdx` and admit the carrier in three helpers:
- **`__vec_push`** — value-unbox the boxed externref via `any.convert_extern` +
  `ref.cast $AnyString` (no numeric unbox) before `array.set`.
- **`__vec_pop`** — box the popped `ref $AnyString` via `extern.convert_any`.
- **`__vec_set_elem`** (`vec-define-writeback.ts`, `Object.defineProperty`
  array write-back) — the SAME arm; the carrier now in `mutEntries` otherwise
  fell to the i32 conversion and emitted invalid Wasm (`array.set expected
  (ref null …), found i32`).

`__vec_get` / `__vec_len` already covered the carrier (get boxes a non-externref
ref via `extern.convert_any`; len is element-kind-agnostic).

## Verified (all standalone, `{}`-instantiable, checked in-wasm)
push→length 3, push return 3, `a[2]==="c"`, multi-push order, grow-from-empty,
pop→"c", pop→length 2, push/pop round-trip. **gc-host `string[]` push/pop
unaffected** (uses the `wasm:js-string` path, not this carrier). No regression
in vec-push / defineProperty-writeback / wasi-defineProperty / #2927 push-pop
suites. `tsc --noEmit` clean.

`tests/issue-3311.test.ts` — 10 tests (the #2093 probe-coverage guard).

Slice **G4** of the #2928 / #2927 `CallBuiltin` prerequisites.

### DIFF
--- a/plan/issues/3311-g4-string-vec-push-pop-carrier.md
+++ b/plan/issues/3311-g4-string-vec-push-pop-carrier.md
@@ -1,7 +1,8 @@
 ---
 id: 3311
 title: "G4 — `string[]` push/pop under standalone is a no-op: native-string vec carrier missing from `__vec_push`/`__vec_pop` mutEntries"
-status: ready
+status: done
+completed: 2026-07-17
 created: 2026-07-16
 priority: high
 horizon: s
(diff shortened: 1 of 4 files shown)


ttraenkler:

PR #3221, opened 2026-07-17
docs(#3341): correct stale demote-hatch citations + document per-reason STRICT_IR_REASONS rule

ttraenkler:
## #3341 — re-scoped: the "promote corpus-zero reasons" premise is unsafe

#3341 was filed as "the buckets are already zero, so promoting them into
`STRICT_IR_REASONS` is the cheapest hardening step." **Analysis (confirmed by
the tech lead) shows that premise is unsafe.** Bucket-zero in
`scripts/ir-fallback-baseline.json` is measured against the **13-file playground
corpus only** — corpus-zero does NOT mean a reason is unreachable on real code.
`external-call`, `call-graph-closure`, `param/return-type-not-resolvable`,
`type-resolution-failure`, `class-method`, and the destructuring-param buckets
all describe **legitimate IR-non-claimability** (external dep, unclaimable
callee, unresolvable type, computed/generator/abstract method name) that the
legacy path must still catch. Promoting any of them would turn those legitimate
fallbacks into **hard compile errors** and regress real programs (broad blast,
unvalidatable locally). `ir-adoption.md`'s `class-method` row already documented
exactly this ("corpus bucket 0 … NOT yet strict").

## What this PR ships (the safe correction — NO promotion)
- **Fixes the stale `src/codegen/index.ts:889–896` demote-hatch citation** → the
  actual sites (`~1889` selector-claimed unresolvable types; `~2390` IR-build
  throw) in `scripts/gen-ir-adoption.mjs` (→ regenerated `plan/log/ir-adoption.md`)
  and `docs/architecture/codegen-axes.md`.
- **Adds a code-comment at `STRICT_IR_REASONS`** documenting the
  necessary-but-not-sufficient condition, so no future dev naively flips a
  corpus-zero reason and reddens the build.
- **Documents the per-reason (not corpus-flip) promotion rule** in
  `codegen-axes.md`'s escape-hatch section.
- **Re-scopes #3341** (stays `ready`, `feasibility: hard`): the real per-reason
  hardening — make a specific reason genuinely unreachable in the IR first, then
  promote + full-CI validate — remains OPEN as future #2855-family work.

`STRICT_IR_REASONS` stays empty ⇒ **no codegen behavior change**.
`check:ir-fallbacks` + `gen:ir-adoption --check` green; loc-budget allowance for
the +15-line comment granted in #3341's frontmatter.

### DIFF
--- a/docs/architecture/codegen-axes.md
+++ b/docs/architecture/codegen-axes.md
@@ -165,7 +165,9 @@ Stay in `src/codegen/` (direct) until the listed issues land:
 
 `scripts/ir-fallback-baseline.json` is the ratchet. When a kind is
 adopted, its bucket goes to zero and the demote-to-warning escape hatch
-in `src/codegen/index.ts:889–896` is removed for that kind — see #2855.
+in `src/codegen/index.ts:~1889/~2390` is removed for that kind — see #2855.
+(Bucket-zero is necessary but **not sufficient** to make a reason a hard
+error via `STRICT_IR_REASONS`; see #3341 and the escape-hatch section below.)
 
 ## Current hidden bias in `src/ir/` (and what to do about it)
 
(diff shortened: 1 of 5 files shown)


ttraenkler:

PR #3226, opened 2026-07-17
fix(#3310): lift __apply_closure arity ceiling from 4 to 8 (standalone dynamic method calls)

ttraenkler:
## Summary

Fixes #3310 Part 2 (the live gap). The standalone open-`$Object` dynamic method-call bridge (`__apply_closure`, `fillApplyClosure` in `src/codegen/object-runtime.ts`) built its arity dispatch chain with a hard `for (let n = 4; n >= 0; n--)`, so a 5+-arg dynamic method call (`o.m(a,b,c,d,e)` through the open-object lane) fell through to the undefined sentinel and returned `0` in a numeric context — even though the `__call_fn_method_N` dispatchers are emitted (index.ts, `emitClosureMethodCallExportN`) for every arity up to `min(maxClosureArity, 8)`.

**Fix:** lift the `fillApplyClosure` dispatch loop bound to a named `APPLY_CLOSURE_MAX_ARITY = 8` to match the emission cap. `buildArm(n)` already returns the undefined sentinel for any arity whose `__call_fn_method_N` was not registered, so the widening is **byte-identical for modules without ≥5-arg closures** and only adds live dispatch where the matching dispatcher exists.

## Part 1 was already resolved
The args-passing half was already handled by #1888 Slice 2 / #3123 / #3201 (verified against current main: `o.f(3,4)` standalone returns `7`). The only remaining `wantArgs` host-gate (`emitWrapperDynamicMethodCall`) has no standalone caller that passes a `callExpr`, so it drops nothing observable under standalone. See the issue's Resolution section for the full trace.

## Repro (current main)
`const o:any={}; o.add5=(a,b,c,d,e)=>a+b+c+d+e; (o as any).add5(1,2,3,4,5)` → `0` before, `15` after (standalone).

## Tests
`tests/issue-3310.test.ts` — standalone dynamic method calls at arities 2 (control), 5, 6, 7, 8 return correct values, plus a 0-function-imports assertion. All 6 pass.

The two pre-existing `issue-2151.test.ts` "custom iterable via any-method .next()" failures are main-state (fail identically on the `b4fca62242` base) and unrelated to this change.

Arities beyond 8 remain the undefined sentinel — the unbounded spill-arm (`__call_fn_method_vec`) is the #2928 `CallBuiltin` follow-up, out of scope here.

### DIFF
--- a/plan/issues/3310-g2-standalone-generic-method-call-args-arity.md
+++ b/plan/issues/3310-g2-standalone-generic-method-call-args-arity.md
@@ -1,7 +1,9 @@
 ---
 id: 3310
 title: "G2 — args-passing on the standalone generic method-call path + `__apply_closure` arity>4 lift (wantArgs is host-gated)"
-status: ready
+status: done
+completed: 2026-07-17
+assignee: ttraenkler/opus-4
 created: 2026-07-16
 priority: high
 horizon: m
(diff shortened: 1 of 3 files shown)


ttraenkler:

PR #3247, opened 2026-07-17
fix(#3342): standalone .join on any-typed externref array is host-free

ttraenkler:
## Summary

`(Object.values(o) as any).join(",")` / `(Object.getOwnPropertyNames(o) as any).join(",")` compiled with `--target standalone` leaked an unsatisfiable `env::Uint8ClampedArray_join` host import — the module failed to instantiate against `{}`.

## Root cause

With the receiver statically `any` (the `as any` cast, or any `any`-typed array), `.join` reaches `tryExternClassMethodOnAny` (`calls-closures.ts`). Its first-match loop bound `.join` to the **first** registered extern class declaring a `join` method — a TypedArray, `Uint8ClampedArray`. #3155 only fixed the non-cast (`string[]`) form, which rides the native array-join dispatch.

## Fix

In `tryExternClassMethodOnAny`, under `standalone || wasi`, route `methodName === "join"` to the #3155 native externref-array walk via new `compileArrayJoinExternForAny` (`array-methods.ts`, gated on `noJsHost`, returns null **before** emitting when unavailable). Mirrors the existing `get`/`set`/`has`/… standalone refusal block. A genuine TypedArray typed `any` rides the same native walk correctly, so this is a general externref-array-`join` fix, not an Object.values special-case.

**Host lane**: guard skipped when neither standalone nor wasi → still binds `Uint8ClampedArray_join` (satisfiable JS-host) → byte-identical.

## Tests

- New `tests/issue-3342.test.ts` (8 cases): values/getOwnPropertyNames/keys as-any, default/multi-char separators, empty object, genuine TypedArray-as-any, and a zero-env-imports assertion.
- Updated the now-resolved carve-out note in `tests/issue-3155.test.ts`.
- `tsc` clean; loc-budget / coercion-sites / pushraw / any-box-sites / dead-exports gates pass; 68 adjacent extern-class/join/object regression tests green.

Closes #3342.

### DIFF
--- a/plan/issues/3342-standalone-object-values-join-typedarray-misclassify.md
+++ b/plan/issues/3342-standalone-object-values-join-typedarray-misclassify.md
@@ -1,7 +1,9 @@
 ---
 id: 3342
 title: "standalone: Object.values(o).join / Object.getOwnPropertyNames(o).join misclassify receiver as Uint8ClampedArray → leak env::Uint8ClampedArray_join"
-status: ready
+status: done
+completed: 2026-07-17
+assignee: ttraenkler/dev-1044
 sprint: current
 created: 2026-07-17
 priority: medium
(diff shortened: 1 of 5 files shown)


ttraenkler:
Closing as a duplicate of #3240 (opened 43 min earlier, same root cause and same fix — route any-receiver `join` to the #3155 native externref join under standalone). First-mover wins; deferring to #3240. No disjoint coverage here.

PR #3251, opened 2026-07-17
fix(#684): usage-based f64 inference for any-typed locals

ttraenkler:
## #684 — Usage-based type inference for `any`-typed local variables

When TypeScript infers `any` (pervasive in untyped JS), a local is lowered to a boxed carrier (externref / `$AnyValue`) and every arithmetic read pays a `__box_number`/`__unbox_number` round-trip. This PR narrows such a local to an **unboxed `f64`** slot when — and only when — every use is *ToNumber-invariant*, so the box/unbox round-trip is eliminated.

### Soundness
`__unbox_number === Number()` (ToNumber). Storing the local as f64 coerces each write via `Number(source)`, which is observationally equivalent to the boxed value **iff every use already applies ToNumber**. The classifier accepts only strictly-numeric operators (`* / % - ** << >> >>> & | ^`, unary `- + ~`, `++`/`--`, matching compound assigns), relationals with a statically-numeric other operand, and the assignment target. It **defaults to bail** for everything else, so unsoundness requires a positively-recognized safe use, never an omission.

Key exclusions found and closed during implementation:
- **`+` / `+=` always bail** — they dispatch on the runtime type: `x + 3` is string *concatenation* (`"5" + 3 === "53"`) when `x` is a string; a numeric other-operand does not disambiguate `x`.
- **bigint bails** — bigint arithmetic (`5n * 2n`) is not f64 arithmetic and mixing traps.
- **uninitialized default** — a hoisted `var` is NaN-seeded at entry (`ToNumber(undefined)`), not the wasm default 0; `let`/`const` require an initializer (no entry-init hook) else bail.
- **escapes bail** — `return x`, call args, `===`/`==`, truthiness, template interpolation, property/index, and closure capture.

### Design
- New checker-layer module `src/checker/usage-inference.ts` (`UsageInference`). It lives above codegen, so it uses the raw `ts.TypeChecker` freely — the oracle-ratchet gate scopes only `src/codegen/**`. Instantiated in `create-context.ts` as `ctx.usageInference`; codegen reads it through the single adapter `usageInferredLocalType(ctx, decl)`. **Zero direct-checker growth in the backend** (ratchet passes clean).
- Wired at the three authoritative local-slot minting sites so they agree: the `var` hoister, the `let`/`const` pre-hoister, and `localTypeForDeclaration` (main + generator-spill).
- Gated by `useUsageInfer` (default **on**; `JS2WASM_USAGE_INFER=0` / `{ useUsageInfer: false }` restores the legacy boxed carrier).

### Scope / follow-ups
Parameters (ABI-affecting, needs the #743 call graph), destructuring / for-of/in bindings, module globals, and the `let x; x = …` no-initializer split form are deliberately out of this slice.

### Tests
`tests/issue-684.test.ts` — 36/36 pass. Representation (f64 slot, no box/unbox, NaN-seeded var, flag-off boxed), narrowed semantics, soundness bails, and an on-vs-off differential battery. Regression batches (any-plus-string, any-relational, coercion, closures, generators/spill, hoisting, number-arrays) show **no new failures** — every failure reproduces identically with `JS2WASM_USAGE_INFER=0`, confirming they are pre-existing.


ttraenkler:
Shepherd: false-positive park, not a code regression. The failing job (equivalence-gate, run 29591556890) errored on `actions/download-artifact@v7` with `Failed to ListArtifacts: 403 Forbidden` — a GitHub Actions artifact-API infra hiccup, nothing to do with this PR's content. Clearing `hold` and re-enqueuing once.

PR #3254, opened 2026-07-17
fix(#2958): standalone/WASI unhandled-rejection tracking

ttraenkler:
## What

Report a native `$Promise` that settles **rejected with no reaction** at program exit — the WASI `_start` tail, after the microtask/event-loop drain — via `fd_write(stderr)` + `proc_exit(1)`, matching Node's default. Previously such a rejection was swallowed and the program exited 0 (a debugging hazard and a divergence from spec-required HostPromiseRejectionTracker observability).

Closes #2958.

## How

Self-contained substrate in `src/codegen/async-scheduler.ts` — **no `$Promise` struct-layout change** (the banked plan's field-append is not viable: `struct.new $Promise` requires all fields, forcing 17 edits across 6 contended files). Instead a separate intrusive list of a new `$__unhandled_node` struct:

- **Note (O(1) prepend)** on a handler-less rejection: the direct `Promise.reject(x)` mint and the `__promise_reject` settle of a previously-pending promise with a null callback list (funnels executor reject, `.then`-chain/derived rejection, combinator result rejection, async-fn result rejection).
- **Mark handled** on a later `.then/.catch/.finally`, an `await` (`async-frame.ts`), or a combinator subscription (`promise-combinators.ts`).
- **Report** `__report_unhandled_rejections()` at the `_start` tail (`index.ts`); one stderr line per still-unhandled node, `proc_exit(1)` if any.

All `ctx.wasi`-gated; host / non-wasi-standalone builds are byte-inert. Import registration (`wasi.ts`) is scoped to **top-level** `Promise`/`await` usage, so host-free carrier modules that use promises only inside exported functions (instantiated with `{}`, driven directly) keep their exact prior import set.

## Tests

`tests/issue-2958.test.ts` — 13 cases, compiled `--target wasi` and run under V8 with a minimal `wasi_snapshot_preview1` shim (independent of the host wasmtime version). AC1–AC3 + derived-promise, executor reject, `Promise.all` caught (no false positive), multiple rejections, non-Error reasons, resolved chains, promise-free byte-inertness. No new regressions across the promise/async/combinator suites (pre-existing failures confirmed on a clean tree).

## Known follow-ups (out of AC scope — #2867 / async-cps territory)

1. Async-function-**body-throw** (no `await`) is not yet tracked (bypasses the settle funnel).
2. Async **reject-propagation-to-`catch`** completeness — where the catch does not yet run, the rejection is genuinely unhandled and is correctly reported; self-corrects once #2867 lands and the await/combinator mark-handled hooks fire.


ttraenkler:
Removing the auto-park `hold` — CI infra flake, not a code regression: cited 'test262 js-host shard 52' (run 29606840777) is a transient registry.npmjs.org fetch failure during corepack's pnpm download; the shard never ran / produced no artifacts, so 'merge shard reports' failed downstream. No #2958/code involvement (same infra-flake class as #3257). Clearing to re-enqueue. — tech-lead queue-health

ttraenkler:
Removing bot park-`hold` — stale-baseline false-positive class, approved for clear by tech-lead. Park cited `check for test262 regressions` + `merge shard reports` and predates the 22:42Z Baseline Refresh (run 29618677416) that regenerated the floor and moots that drift. Verified: 0 failing required checks. Re-admitting for re-validation against the fresh baseline.

PR #3255, opened 2026-07-17
fix(#3326): refresh stale #2036 refuse-loudly test expectations

ttraenkler:
## #3326 — stale refuse-loudly expectations in `tests/issue-2036.test.ts`

`tests/issue-2036.test.ts` asserted that 6 borrowed `Array.prototype` search/result-building methods (`indexOf`, `lastIndexOf`, `includes`, `map`, `reduce`, `reduceRight`) over an array-like `$Object` receiver in standalone had **no native path** and must **refuse loudly**. #3169 gave them a working native arm, so they now genuinely **succeed** — the refusal assertions were stale (failing on unmodified `main`; the file isn't in a scoped CI suite, so it went unnoticed).

### Change
Replaced the `expect(r.success).toBe(false)` / `Codegen error:` refusal loop with per-method **runtime-correctness** tests, each result verified against native JS `Array.prototype.<m>.call({0:5,1:7,length:2}, …)`:
- `indexOf(7)===1`, `indexOf(99)===-1`; `lastIndexOf(5)===0`
- `includes(7)===true`, `includes(99)===false`
- `map(x=>x*2)` → `[10,14]` (length 2, `r[0]*100+r[1]===1014`)
- `reduce((a,x)=>a+x,0)===12`; left-fold order `((0*10+5)*10+7)===57`
- `reduceRight` sum `===12`; right-fold order `((0*10+7)*10+5)===75`

Removed the now-unused `compileStandalone` helper.

### Separate bug found → filed as #3361
The pre-existing `filter threads thisArg` failure is a **distinct** bug (filter's native arm drops the 3rd `thisArg` argument → `this.t` is `undefined`), outside this issue's stale-expectation scope. Filed as **#3361** and marked `it.fails` so the suite stays green and flips to a hard failure the moment #3361 lands the fix.

### Tests
`tests/issue-2036.test.ts`: 28/28 (27 pass + 1 documented `it.fails`). tsc / prettier / biome clean.

### DIFF
--- a/plan/issues/3326-issue-2036-refuse-loudly-expectations-stale.md
+++ b/plan/issues/3326-issue-2036-refuse-loudly-expectations-stale.md
@@ -1,7 +1,9 @@
 ---
 id: 3326
 title: "tests/issue-2036.test.ts: 7 'refuses loudly' expectations are stale — #3169 gave these methods a working native path, they now succeed instead of refusing"
-status: ready
+status: done
+completed: 2026-07-17
+assignee: dev-684
 sprint: current
 created: 2026-07-16
 priority: low
@@ -9,7 +11,7 @@ feasibility: trivial
 task_type: bug
 area: codegen
 goal: standalone-mode
-related: [2036, 3169]
+related: [2036, 3169, 3361]
 origin: "found as a side-effect of #3317 (array search-method coercion) validation, 2026-07-16 — pre-existing on main, unrelated to #3317 itself"
 ---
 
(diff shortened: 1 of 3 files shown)


ttraenkler:
Superseded — #3326 was already completed on main by opus-e (identical #2036 refuse-loudly → runtime-correctness refresh, describe block '#2036 S6 — borrowed search/result-building methods run natively in standalone'). The filter-thisArg residual I found is already tracked as #3359 (my branch's #3361 is a duplicate of it). Closing as redundant; no changes needed. Verified: main's tests/issue-2036.test.ts is green (28 tests).

PR #3257, opened 2026-07-17
fix(ir): make linear plan authoritative for data segments

ttraenkler:
## Summary

- deep-freeze and validate canonical `LinearMemoryPlan` snapshots so consumers cannot mutate, duplicate, or dangle middle-end layout/data decisions
- make linear-Wasm materialize string literals from plan-owned relocatable bytes while retaining adapter-owned placement and the default arena ABI
- keep direct and IR string literals on one conflict-checked registry and add focused immutable-snapshot plus Unicode execution coverage

This is the final #3298 hardening continuation after merged PR #3245. It does not begin Porffor heap lowering or introduce a second allocation policy.

## Validation

- focused planner/provenance/backend set: 9 files, 75 tests passed
- full linear CI slice: 20 files, 204 tests passed
- cross-backend differential: 29 tests passed
- `prove-emit-identity`: all 56 file/target outcomes identical to the pre-change baseline, including after merging current `origin/main`
- `npm run build`
- `npm run typecheck -- --pretty false`
- `npm run lint`
- `npm run format:check`
- `npm run check:pushraw`
- `npm run check:loc-budget`
- `npm run check:ir-fallbacks`
- `npm run check:linear-ir`
- `npm run check:dead-exports`
- `npm run check:issues`

Closes #3298

### DIFF
--- a/plan/issues/3298-linear-memory-plan-extraction.md
+++ b/plan/issues/3298-linear-memory-plan-extraction.md
@@ -1,7 +1,7 @@
 ---
 id: 3298
 title: "Porffor backend P3: extract the shared target-neutral LinearMemoryPlan"
-status: in-review
+status: in-progress
 sprint: porffor-backend
 created: 2026-07-16
 updated: 2026-07-17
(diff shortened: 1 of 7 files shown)


ttraenkler:
Removing the `hold` (auto-park) label — diagnosed the cited merge_group run 29592538029 and this is an infra flake, not a merged-baseline regression:

- The failing step is `test262 js-host shard 49` → **Upload shard artifacts** → `actions/upload-artifact@v6` returning `(403) Forbidden: Failed to FinalizeArtifact`. The test262 tests themselves ran; only the artifact upload failed.
- `merge shard reports` then failed downstream solely because that shard's artifact was missing.
- A subsequent `merge shard reports` validation on this PR already shows **pass** — confirming it was transient.

No test delta, nothing to fix on the branch. Clearing the hold so the server-side auto-enqueue re-sweeps. (One-shot; not re-enqueuing manually to avoid a merge_group cancellation race.) — tech-lead queue-health

PR #3274, opened 2026-07-17
fix(#3332): direct linear-path Array.push returns new length + appends all args

ttraenkler:
## Summary

Closes #3332. The DIRECT linear path (`--target linear`, no `JS2WASM_LINEAR_IR`) mis-lowered `Array.prototype.push`:

- **expression position** returned `f64.const 0` instead of the array's new length;
- **multi-arg** push dropped every argument past the first.

```ts
const a = [1]; a.push(8);        // returned 0   → now 2 (new length)
const a = [1]; a.push(2, 3);     // length 2     → now 3 (all args appended)
```

## Fix

`src/codegen-linear/index.ts` `compileArrayMethodCall`, `methodName === "push"`:

- evaluate the receiver **once** into a fresh i32 local (a side-effecting receiver expression is no longer re-run per argument);
- loop over **all** `expr.arguments`, calling `__arr_push` for each;
- read `__arr_len(arr)` back and convert to f64 for the expression-position result value (`__arr_push` is void).

The statement-position stack contract is unchanged (one f64 left on the stack, dropped by the caller).

## Tests

- `tests/issue-3332.test.ts` — 6-case regression guard, direct path forced via `JS2WASM_LINEAR_IR=0` (single/multi-arg return value, value ordering, empty-array, statement position).
- `tests/issue-2956.test.ts` — flipped the two forward-looking `#3332` assertions to their spec-correct values, exactly as those comments instructed: direct `pushExpr` `8 → 28` (folded into the loud parity loop), demoted `multiPush` length `2 → 3`.

## Validation

- `npm test -- tests/issue-3332.test.ts tests/issue-2956.test.ts` → 26 passed
- `npx tsc --noEmit` → no errors
- `prettier --check` → clean

### DIFF
--- a/plan/issues/3332-linear-direct-push-return-and-multiarg.md
+++ b/plan/issues/3332-linear-direct-push-return-and-multiarg.md
@@ -17,6 +17,11 @@ horizon: s
 created: 2026-07-17
 updated: 2026-07-17
 related: [2956, 1854]
+loc-budget-allow:
+  # Intended +13 LOC in the direct-path push lowering: evaluate the receiver
+  # once into a local, loop over all arguments, and read __arr_len back for the
+  # expression-position new-length result (fixes the returns-0 / drops-args bug).
+  - src/codegen-linear/index.ts
 ---
 
 ## Resolution (2026-07-17)
--- a/plan/issues/3332-linear-direct-push-return-and-multiarg.md
+++ b/plan/issues/3332-linear-direct-push-return-and-multiarg.md
@@ -1,8 +1,10 @@
 ---
 id: 3332
 title: "linear direct path: arr.push returns 0 (not new length) and drops extra args"
-status: ready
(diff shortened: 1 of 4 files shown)


ttraenkler:
Removing the auto-park `hold` — diagnosed as a **stale-baseline false positive**, not a real regression:
- All current required checks PASS on the latest merged-state build (cheap gate, **check for test262 regressions**, cla-check, equivalence-gate, merge shard reports, quality).
- The failed run that triggered the park (29604321440) ran with `SRC_BEHIND=8` / `CONTENT_CURRENT=false` — the regression baseline was 8 test262-relevant commits stale, so the flagged deltas are baseline-drift artifacts, not a #3332/Array.push regression. The rebuild on a fresher base is clean.
Clearing the hold so it re-enqueues. — tech-lead queue-health

PR #3290, opened 2026-07-17
fix(#2649): standalone TypedArray.subarray(...).length direct-chain reads 0

ttraenkler:
## #2649 — standalone `a.subarray(...).length` reads 0

In `--target standalone`, `TypedArray.prototype.subarray(begin?, end?).length`
read **0** for the direct chain (`a.subarray(1).length`), even though the element
data through the view was correct.

### Root cause
Not in `compileTypedArraySubarray` — the `$__subview_<elem>` struct's length
field (`max(clamp(end) − clamp(begin), 0)`) is built correctly. The bug was in
the **`.length` member read** (`property-access-dispatch.ts`, `propName === "length"`):

- `a.subarray(1)` compiles to a `$__subview_<elem>` struct (field 0 = `length`).
- `resolveWasmType(Int8Array)` gives the plain `$__vec_<elem>` type.
- compiled-type ≠ TS-derived vec type → the dispatch `ref.test`ed the subview
  value against the **vec** type, missed, and fell to the `f64.const 0` else-arm.
- The via-variable form (`const s = a.subarray(1); s.length`) masked it (the
  local carried the subview type, so `.length` read field 0 directly).

### Fix
When the compiled receiver is itself a length-bearing struct (field 0 named
`length` — any `$__vec` / `$__subview` / `$__ta_view`), read `.length` field 0
**directly** off the compiled type — exact, no `ref.test`.

### Tests
`tests/issue-2649-standalone-subarray-length.test.ts` — **10/10**: direct-chain
`subarray(b)` / `(b,e)` / `()`, negative begin, element data through the view,
Int32 / Float64 views, nested subarray, via-variable regression guard, gc-mode
guard. `tsc` clean; loc-budget allowance granted in the issue file.

### DIFF
--- a/plan/issues/2649-standalone-typedarray-subarray-empty.md
+++ b/plan/issues/2649-standalone-typedarray-subarray-empty.md
@@ -1,8 +1,10 @@
 ---
 id: 2649
 title: "Standalone: TypedArray.prototype.subarray returns an empty view (.length === 0)"
-status: ready
-sprint: Backlog
+status: done
+assignee: ttraenkler/dev-spec
+completed: 2026-07-17
+sprint: current
 priority: medium
 feasibility: medium
 reasoning_effort: medium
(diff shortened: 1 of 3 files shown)


ttraenkler:
Duplicate of #3285 (dev-standalone2), which was first and is green + in the queue — same root cause and the same fix in `property-access-dispatch.ts` (read `.length` field 0 directly off the compiled length-prefixed struct instead of ref.testing the mismatched vec type). Deferring to #3285. My variant guarded field[0]==="length" without also requiring field[1]==="data", which is marginally more general (a $__ta_view direct-chain), but that path is already handled by the ta_view length arm — not worth racing. Closing.

PR #3291, opened 2026-07-17
fix(#2649): standalone TypedArray.subarray().length returns the windowed length, not 0

ttraenkler:
## Problem (#2649)

Under `--target standalone`/`wasi`, `TypedArray.prototype.subarray(...).length`
read as **0** for every form (`subarray(1)`, `subarray(0,2)`, `subarray()`),
even though indexed element access on the same view returned the right values.

## Root cause

Not in `compileTypedArraySubarray` — it correctly builds the `$__subview` struct
(`{length, data, byteOffset}`) with the windowed length at field 0. The bug is in
the **`.length` read dispatch** (`src/codegen/property-access-dispatch.ts`): for a
receiver whose compiled ref type differs from its static TS-derived vec type, it
`ref.test`ed the value against the static **plain `$__vec`** type and fell back to
`0` on a miss. `a.subarray(1)` is TS-typed `Int8Array` (→ plain `$__vec`) but
compiles to a `$__subview`, so the `ref.test` always missed → `.length == 0`.

## Fix

When the compiled receiver is a known `$__subview` ref type (`isSubviewTypeIdx`),
read its field-0 length **directly** — no `ref.test` against the mismatched
static type. `$__ta_view` is intentionally excluded (its field-0 can be a
resizable-length sentinel, handled by the auto-length arm above). Narrowly gated,
so non-subview `.length` paths are untouched.

Standalone-only: gc/host `subarray` returns a plain-vec copy whose length already
read correctly.

## Tests

`tests/issue-2649.test.ts` (10 cases, standalone, empty-import instantiation):
`subarray(b)` / `subarray(b,e)` / `subarray()`, negative begin/end clamping,
subview stored in a local, nested `subarray().subarray()`, a 16-bit element view,
an element-access regression guard, and a `for (i < s.length)` sum loop.

## Validation

- `tests/issue-2649.test.ts` 10/10 pass (before and after merging latest `main`)
- `npx tsc --noEmit` clean; `prettier --check` clean
- `tests/issue-1664.test.ts` (TypedArray set/subarray, no host imports) still 7/7

### DIFF
--- a/plan/issues/2649-standalone-typedarray-subarray-empty.md
+++ b/plan/issues/2649-standalone-typedarray-subarray-empty.md
@@ -1,8 +1,9 @@
 ---
 id: 2649
 title: "Standalone: TypedArray.prototype.subarray returns an empty view (.length === 0)"
-status: in-progress
+status: done
 assignee: dev-refactor
+completed: 2026-07-17
 sprint: current
 priority: medium
 feasibility: medium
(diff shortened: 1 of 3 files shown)


ttraenkler:
Closing as duplicate of #3285 (dev-standalone2, first-mover, already green + queued) — identical root cause and fix (`.length` dispatch ref.tested $__subview against the static $__vec type → 0 fallback; fix reads subview field-0 directly). Deferring to #3285.

PR #3295, opened 2026-07-17
fix(#3384): unwrap wrapped JSON.parse call before reading .arguments (standalone/wasi crash)

ttraenkler:
## #3378 — member access on a wrapped `JSON.parse()` call crashed codegen

Found while sweeping the #2961 standalone enumeration.

### Bug
Under `--target standalone` / `--target wasi`, a property/element access on a `JSON.parse(<literal>)` call whose call node is wrapped in a **transparent expression** (`as any`, parens, `!`) hard-crashes codegen:
```ts
export function test(): number { return (JSON.parse('{"a":5}') as any).a; }
// Internal error compiling expression: Cannot read properties of undefined (reading '0')
```
`const o = JSON.parse(...); o.a` and a bare call compile fine; `--target gc` is unaffected (the static-fold path is standalone/wasi-only).

### Root cause
`tryEmitJsonParse{Property,Element}Access` (`src/codegen/json-standalone.ts`) guard with `isJsonParseCall(expr.expression)` — which **unwraps transparent expressions internally** and carries a type predicate `expr is ts.CallExpression` that **lies at runtime** — then read `expr.expression.arguments[0]`. For `(JSON.parse(s) as any).a`, `expr.expression` is the `AsExpression` (no `.arguments`), so `[0]` was `undefined[0]`.

### Fix
Unwrap `expr.expression` to the real `CallExpression` **before** reading `.arguments`, in both functions. `isJsonParseCall` is now always passed an already-unwrapped node.

### Validation
- `tests/issue-3378.test.ts` — 6 cases (property + element access on `as any` / parens / `!`, object + array), standalone + wasi. Each instantiates with an **empty import object** and returns the statically-folded value.
- `tsc --noEmit`, `biome lint`, `prettier --check`, `check:loc-budget`, `check:issue-ids:against-main` all green. Full test262 not run (CI validates).


ttraenkler:
Removing the auto-park `hold` — the cited quality failure (issue-integrity: `1086-codegen-dedup-…` → #1085) is a STALE merge_group snapshot, not a live problem. Verified on current origin/main: #1085 (`1085-codegen-bodyusesarguments…`) and #1086 (`1086-codegen-dedup…`) are correctly distinct files — no id collision. #3295 does not touch either file, and no open PR does either; the bad id came from a co-queued PR that ejected. A rebuild on clean main passes. Clearing so it re-enqueues. — tech-lead queue-health

ttraenkler:
Shepherd: resolved an issue-id collision on this PR — its issue file (plan/issues/3378-standalone-json-parse-crash.md) collided with an already-merged, unrelated #3378 (plan/issues/3378-deepequal-nested-closure-stale-local-index.md). Renumbered this PR's own issue to **#3384** (freshly allocated via claim-issue.mjs) — see commit 832b6fa79f. `gh pr edit --title` hit the known projectCards no-op bug so the PR title above still says #3378; the correct id going forward is #3384 (the commit message + issue file frontmatter are authoritative). The JSON.parse crash fix itself is unaffected — this was a pure id-renumbering, no code changes.

ttraenkler:
Issue-id collision fix: this JSON.parse-crash issue was renumbered **#3378 → #3384** (another issue, `3378-deepequal-nested-closure-stale-local-index`, allocated #3378 and merged to main first). Adopted the #3384 renumber already pushed to this branch, completed it (renamed the test to `tests/issue-3384.test.ts`, updated all stale #3378 refs), and merged origin/main forward. The JSON.parse *fix code* is unchanged. id gate clean, test green (6/6).

ttraenkler:

PR #3298, opened 2026-07-17
fix(#3379): baseline-sync staleness guard must measure public/, not the in-repo copy (follow-up to #3375)

ttraenkler:
Follow-up to #3375. That fix made the host drift-check correctly **detect** public staleness (`host_unchanged=0`), but the sync still didn't update `public/` — a second instance of the same "measures the in-repo copy, not public/" bug.

## Root cause

After drift is detected, a churn guard skips the sync if the merge queue is busy **and** the committed summary is `< 6h` old. The age was computed from the **in-repo** `benchmarks/results/test262-current.json`, which forced promotes refresh in lockstep with the baselines repo — so it reads ~0h even when `public/benchmarks/results/test262-report.json` (the file the landing page serves) is hours stale.

Observed after #3375 landed (sync run 29604427308):
```
Drift detected (host_unchanged=0 standalone_unchanged=0); queue=15 committed-summary age=0h force=false
Merge queue busy and summary <6h old — skipping this cycle.
```
`host_unchanged=0` (the #3375 fix working) but `age=0h` from the wrong file + `queue=15` → skip. During active dev the in-repo file keeps getting refreshed, so `age` never crosses 6h and a busy queue skips the sync **indefinitely** — the page only self-heals via a manual `force=true`.

## Fix

Measure the age of `public/benchmarks/results/test262-report.json`. With #3375 + this, the sync both **detects** public drift and **acts** on a stale public file — no manual force needed.

(The page was already force-corrected to 32,165 for now; this makes it self-heal going forward.)

Diff: 1 workflow file (+7/-1) + the issue doc.


ttraenkler:

PR #3302, opened 2026-07-17
fix(#3332): direct linear path push appends all args + returns new length

ttraenkler:
## Summary

The DIRECT linear path (`--target linear`, no `JS2WASM_LINEAR_IR`) mis-lowered `Array.prototype.push` in `src/codegen-linear/index.ts` `compileArrayMethodCall`:

- **(a)** expression-position push emitted `f64.const 0` as its value instead of the new length;
- **(b)** multi-arg push (`a.push(x, y, …)`) handled only `arguments[0]` and dropped the rest.

Per spec §23.1.3.23, `push` appends **all** arguments and returns the **new length**.

## Fix

Hold the receiver in an i32 local, loop over every argument calling `__arr_push` per arg, then read `__arr_len` for the expression value. Handles zero-arg `push()` (returns the unchanged length) and avoids re-evaluating a side-effecting receiver.

## Tests

- `tests/issue-3332.test.ts` — 6 new cases: single/multi-arg return value, no-arg, value ordering, ref-element (string) push.
- `tests/issue-2956.test.ts` — the divergence assertions that previously *documented* this defect are folded into the direct↔IR parity loop (`pushExpr` now parity; `multiPush` length 3, not 2).

```
npx vitest run tests/issue-3332.test.ts tests/issue-2956.test.ts   # 26 passed
```

The IR overlay's single-arg-only gate still demotes multi-arg push to the direct path — which is now itself spec-correct.

### DIFF
--- a/plan/issues/3332-linear-direct-push-return-and-multiarg.md
+++ b/plan/issues/3332-linear-direct-push-return-and-multiarg.md
@@ -1,7 +1,8 @@
 ---
 id: 3332
 title: "linear direct path: arr.push returns 0 (not new length) and drops extra args"
-status: ready
+status: done
+completed: 2026-07-17
 sprint: Backlog
 goal: backend-agnostic-ir
 feasibility: medium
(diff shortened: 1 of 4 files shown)


ttraenkler:
Closing as superseded — this is a duplicate of #3274 ("fix(#3332): direct linear-path Array.push returns new length + appends all args"), which merged at 19:20Z and landed the identical fix (append-all-args + return new length in src/codegen-linear/index.ts). #3332 is already `status: done` on main, and the #2956 parity test is present. Two lanes concurrently worked #3332; #3274 landed first, so this one is redundant. No unique content to preserve. Thanks to the author — the fix is on main via #3274.

PR #3317, opened 2026-07-17
fix(#739): host-lane representation pinning for runtime-store defines (S1)

ttraenkler:
## #739 Slice S1 — host-lane representation pinning (the store-unification)

Implements **Slice S1 only** of #739 per the spec in PR #3310 (which also
resolves the blocked #3230). S2 (defineProperties two-phase apply) remains
outstanding — this issue stays `ready`.

### Problem (root cause, #3230)
An empty-`{}` var that receives an `Object.defineProperty`/`defineProperties`
whose application lands in the **runtime store** (native `$Object` / `_wasmPropDescs`
sidecar) is widened to a closed WasmGC struct. The define lands in one store
while every later dot-read `obj.p` lowers to `struct.get` (a defined getter never
fires → reads back the field default) and every dot-write `obj.p = X` to
`struct.set` (a defined setter is bypassed). `_structFieldWriteback` mirrors only
data VALUES, never accessors. #3230 proved bounded read/write point-fixes are
net-negative and widening-sensitive.

### Fix
New host-lane poison predicate `markRuntimeStoreDefineTargets` in
`src/codegen/declarations/object-shape-widening.ts` pins such receivers to a
host `$Object` by marking them `objectHashConsumerVars`, so the existing
#2584/#2944 suppression branch both (a) skips widening and (b) records the var's
evolved checker type in `objectHashConsumerTypes` (the #2937 escape discipline —
without it compiled-acorn null-derefs). Every MOP op then rides the ONE native
store the bracket-form (`obj["p"]`) already proves correct.

**Runtime-routed predicate** (mirrors `object-ops.ts` routing): dynamic key,
non-literal descriptor, `get`/`set` present, explicit-`undefined` field,
no-`value` key, or any `Object.defineProperties`. Pure inline
`{value:<literal>, ...boolean-literal flags}` data-descriptor defines are kept
OUT (struct fast path unchanged, guard C). All S1 changes are
`!ctx.standalone`-guarded; standalone stays byte-identical.

### Validation (scoped; CI runs full test262)
- `tests/issue-739.test.ts` (9 cases): dynamic + inline accessor read lanes,
  setter write lane, get+set round-trip, dynamic data descriptor, no-value
  `{writable:true}`, explicit-`undefined`, `defineProperties` accessor, guard-C
  boundary. The two accessor read-lane cases are **verified failing on
  `origin/main`** (return `undefined`) and passing here — the #3230 repro.
- Regression guards green: `tests/issue-2944.test.ts` (compiled-acorn escape
  discipline, #2937), `tests/define-property-patterns.test.ts`,
  `tests/issue-2584-dual-storage.test.ts`, `3042`/`3043`/`3116`/`2726`.
- Pre-existing-on-main failures (`issue-1130` array-index growth, `issue-1336`
  symbol-key) are unrelated and out of scope.
- `tsc --noEmit` clean.

### DIFF
--- a/src/codegen/declarations/object-shape-widening.ts
+++ b/src/codegen/declarations/object-shape-widening.ts
@@ -109,6 +109,41 @@ export function collectEmptyObjectWidening(
             }
           }
 
+          // (#739 S1 — HOST-lane representation pinning, the store-unification)
+          // Any `Object.defineProperty` / `Object.defineProperties` on this
+          // receiver whose application lands in the RUNTIME STORE — the native
+          // `$Object` open hash or the `_wasmPropDescs`/`_wasmStructProps`
+          // sidecar — rather than a widened-struct `struct.set` fast path makes
+          // a widened struct UNSOUND: every later dot-read `obj.p` lowers to
+          // `struct.get` (a defined getter never fires; a runtime-store value
+          // reads the struct default) and every dot-write `obj.p = X` to
+          // `struct.set` (a defined setter is bypassed). The two stores never
+          // see each other, and `_structFieldWriteback` mirrors only data
+          // VALUES back into the field — accessors cannot be mirrored (a
+          // `struct.get` can never invoke a getter). #3230 measured both bounded
+          // point-fixes (read-reroute net −7; read-reroute + runtime fallback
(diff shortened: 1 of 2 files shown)


ttraenkler:

PR #3318, opened 2026-07-17
fix(#802): promote object-literal proto receivers to $Object (slice A)

ttraenkler:
## #802 Slice A — object-literal proto receivers → `$Object` (standalone)

Implements **Slice A only** of the #802 dynamic-prototype plan (fresh Fable spec in #3313). Slice B/C/D (the conditional class-instance `$__proto__` field, #799a-regression-sensitive) are intentionally left unstarted; the issue stays `status: ready`.

### What Slice A delivers
A plain object literal **without** an `any` annotation (`const o = { x: 1 }`) is normally lowered to a closed-shape WasmGC struct, which has no `$proto` field. In `--target standalone`, `Object.setPrototypeOf(o, p)` routes to the native `__object_setPrototypeOf`, whose `ref.test $Object` fails on that closed struct → the `[[Prototype]]` link is silently dropped and inherited reads return `undefined`/0. (The existing #2580 tests all annotate `:any`, which already diverts the literal to the `$Object` builder — the un-annotated case is the gap Slice A closes.)

- **New `src/codegen/dynamic-proto.ts`** — `scanForDynamicProto` pre-scan (mirrors `new-target.ts`). Detects object-literal receivers of `Object.setPrototypeOf` / `Reflect.setPrototypeOf` / `o.__proto__ =` (unwraps `(o as any)` wrappers; resolves a `const/let/var` binding to its object-literal initializer by lexical-scope walk; `Object.create(_)` deliberately excluded — it *makes* an object). Cheap structural walk, no type queries. Populates `ctx.usesDynamicProto` + `ctx.dynamicProtoLiteralNodes`.
- **`compileObjectLiteral`** promotes a marked receiver literal to the open `$Object` (via the existing `compileObjectLiteralAsExternref`) instead of a closed struct — **zero struct-layout change**, inherits the whole existing setPrototypeOf/read/getPrototypeOf machine.
- **Representation lockstep**: the let/const pre-hoist slot-typer + `compileVariableStatement` + the `var`-hoist path all consult the same `ctx.dynamicProtoLiteralNodes` set, so a promoted receiver's local is `externref` and its reads route through `__extern_get` — avoiding the lossy `ref.cast` to the inferred struct that otherwise nulled the receiver (found via WAT probe during dev).
- Context: added `usesDynamicProto`, `dynamicProtoClasses` (reserved for Slice B, unpopulated), `dynamicProtoLiteralNodes`. Prescan wired at both `generateModule`/`generateMultiModule` sites beside `scanForNewTarget`.

### Standalone-only (deliberate scope)
Per spec §0 the dropped-link gap is standalone-specific; gc/host already routes proto receivers through the host runtime sidecar. All promotion sites are gated on `ctx.standalone`, so **gc/host is byte-for-byte unchanged** (zero host-regression surface). A gc/host inherited-read-through-a-`setPrototypeOf`'d-proto is a separate host read-path item, outside Slice A.

### Tests — `tests/issue-802-slice-a.test.ts` (14, all pass)
- **Standalone (10):** inherited field read after `setPrototypeOf` (was 0); own read; own shadows inherited; inherited method lookup; `null` proto; `setPrototypeOf` returns receiver; direct-literal receiver; `o.__proto__ =`; `Reflect.setPrototypeOf`; unrelated literal byte-unaffected.
- **gc/host regression guards (4):** program still compiles+validates; own-read equivalence vs JS; own-shadows equivalence; `o.__proto__=`/`Reflect` still compile.

`tsc --noEmit` clean, `biome lint` clean, prettier clean. Nearby suites pass (`issue-2580-m3-protochain`, `issue-2739`, `object-mutability`, and a broad object/struct/property batch). The 4 failures seen in `anon-struct`/`issue-2659` are **pre-existing on `origin/main`** (verified by restoring main's source and re-running) — unrelated to this change.

The issue `.md` (fresh spec) is owned by in-flight PR #3313; this code PR does not touch it to avoid a merge conflict.


ttraenkler:
Removed auto-park `hold` — cited merge_group failure (ratio-gate 31.0%, 52 regr / 168 impr against a CONTENT-CURRENT baseline dae79d5a) is the cross-PR drift/reclassification cluster: the ~168-improvement magnitude is near-identical to the ~176-improvement cluster on unrelated PR #3351 (feedback_baseline_drift_cross_check — identical clusters across PRs = drift, not PR-caused), from the oracle v7→v8 baseline transition now resolved by the refreshed baseline + relaxed 186-unclassified gate. Merged current upstream/main (was 150 behind, clean, 0 conflicts), tests/issue-802-slice-a.test.ts 14/14 green. Re-validating clean.

PR #3322, opened 2026-07-18
fix(#3387): drive nested async-gen for-await destructuring heads host-free (standalone)

ttraenkler:
## #3387 — nested async-gen for-await dstr heads: close the drivability seam

Child of the #3178 standalone async-machinery retirement umbrella. Targets the
~577 `host_import_leak` rows in `for-await-of/async-gen-dstr-*`.

### Root-cause finding (the spec's framing was wrong)

The spec hypothesized a missing planner arm and a nested-vs-module-scope gap to
close. The probe matrix disproved it: a driven async gen with a
`for await (const IDENTIFIER of …)` body **already** compiles host-free at both
module and nested scope — the for-await rides the bounded body as a suspend-free
LEAD (the whole loop runs in one dispatch of the #2906 resume fn; no suspend
crosses it).

**The sole rejector of the entire `async-gen-dstr-*` cohort was
`asyncGenBodyHasPatternLocals`** (async-cps.ts), which bailed ANY non-identifier
`VariableDeclaration` — including the for-await HEAD pattern. But the head is not
a frame-spilled own local (its names bind fresh per element, within one dispatch,
never across a suspend). Identifier heads passed only via `ts.isIdentifier`.

### The fix (2 hunks, async-cps.ts, +47 LOC)

1. `asyncGenBodyHasPatternLocals` exempts a for-await HEAD binding.
2. `forAwaitHeadPatternAdmissible` gates that exemption — admits leaf defaults
   (incl. call defaults), elision, rest, renamed/string-keyed/defaulted props,
   empty patterns, nested-no-init; **rejects a nested sub-pattern WITH an
   initializer** (`[[x] = [7]]`), because the sync for-await destructure skips
   nested defaults under `awaitModifier` (#2692/#2566 guard) — driving it would
   bind nothing, so it stays legacy (correct-or-legacy).

`analyzeAsyncGen` is the single gate propagated to the whole admission /
import-retirement / carrier-on chain, so this widens all three in lockstep — no
mirror edits.

### Validation

- New `tests/issue-3387-asyncgen-forawait-dstr.test.ts` (13 cases): corpus
  array/obj shapes, defaults, elision, rest, multi-iteration `.then` ordering +
  done flags, empty source, class async-gen method lane, body-throw rejection,
  and the four correct-or-legacy refusals.
- Real test262 corpus (standalone runner lane), every-6th dstr file: **45/103
  pass, ZERO fails** (near-0 pass pre-fix).
- Adjacent cohorts + async/generator blast-radius (#2906, #3132, #3228, #2865,
  #3120, #3207, #2570): no NEW failures — all residual fails are pre-existing on
  the merged base, byte-identical with async-cps.ts reverted.
- Gates green locally: tsc, prettier, oracle-ratchet (+0 checker usage),
  loc-budget (allow-listed +47 in the issue file).

### Residuals (sibling issues)

- Promise-typed for-await ELEMENTS under carrier-off still leak `Promise_*` —
  #2980 carrier-widen / #3388 lane, not the dstr-head admission this owns.
- `yield` in a dstr initializer + nested-with-init heads stay legacy.

Closes #3387.


ttraenkler:
Removing the auto-park `hold` — same infra-collateral signature as #3327 (fable-1's diagnosis there): this PR's merge_group run (29632727762) shows the standalone lane collapsed to 4,508 pass / 43,469 compile_error, every CE reading `standalone target emitted host imports: env::console_log_externref, env::structuredClone (#2961)` — byte-identical to #3327's failure despite the two PRs touching unrelated feature areas (this is #3387 nested-asyncgen-forawait vs #3327's async-gen work). Confirmed the same stale-merge-group-base host/standalone-lane cache collision fable-1 root-caused, not a real regression in this PR's own async-gen destructuring change. tech-lead has a [CI-FIX] task filed for the underlying cache bug. Removing hold once per protocol — a re-enqueue gets a fresh group base, which sidesteps this. — PR-queue shepherd

ttraenkler:
Second park (06:49:07Z, run 29634075949, job 88054619796) — different from the first: Net **+0 pass** (25527→25527, zero real change), ratio-gate fail only (73/178=41.0%) on baseline 4 commits/-62m behind (dfe4c47). Script's own banner for this run was the drift-warning path (not the likely-real-regression path, which only fires when baseline is content-current). Zero net change under a stale baseline is classic small-sample ratio-gate noise, not a regression — same drift class (a) as the first park. Removing hold once, leaving CLEAN for auto-enqueue. — PR-queue shepherd

PR #3327, opened 2026-07-18
feat(#3251): S1 array-descriptor overlay substrate — companion $Object per vec receiver (standalone)

ttraenkler:
## Summary

**#3251 S1** — the standalone array-descriptor OVERLAY substrate (epic slice 1 of 3).

Under `--target standalone`, arrays (`__vec_<k>` WasmGC structs) had NO per-index/expando descriptor storage: `Object.defineProperty(arr, idx, d)` hit the `ref.test $Object` lenient no-op in the define natives, accessor indices were never consulted by reads or the array HOF loops (~204 host-free assertion_fails), `getOwnPropertyDescriptor` answered undefined, and redefine-legality never threw (the 69-test array-exotic defineProperty cluster).

### Design (full plan in `plan/issues/3251-array-descriptor-overlay-substrate.md`)

- Each vec receiver targeted by a descriptor op gets a **companion plain `$Object`** in a module-global side table keyed by vec identity (`ref.eq` scan behind a `global.get; ref.is_null` fast gate). Because the companion is an ordinary `$Object`, the vec arms **delegate** ValidateAndApplyPropertyDescriptor (#2042 S4), partial-descriptor merge (#2992 S3), CompletePropertyDescriptor defaults, and descriptor materialisation to the existing, already-correct $Object natives — no duplicated spec logic.
- **Data-define values write back into the vec element** (per-carrier `__vec_elem_set_<t>`: in-bounds store, OOB growth, length update), so the typed inline `array.get` fast path stays coherent with zero read overhead (#3116's host-mode trick, natively). Kind-incompatible values set a new `FLAG_COMPANION_VALUE (0x20)` $PropEntry bit; dynamic readers then answer from the companion.
- In-bounds indices with no entry are **seeded** `{value: vec[i], w/e/c: true}` before delegation (spec's implicit element descriptor); fresh OOB indices keep the all-false defaults — which required standalone-gating off the call-site vec pre-growth (`object-ops.ts`, the #3116 regression-class-1 hole ambiguity).
- **Read prologues** (finalize-spliced into `__extern_get_idx` — the chokepoint `arr[i]` any-lane + every `__hof_*` loop reads through — and `__extern_get`): accessor entries invoke their getter via `__call_accessor_get` with the ORIGINAL vec receiver as `this`; companion-value entries answer from the companion; plain data entries fall through to the (authoritative) vec read.
- OOB defines (incl. accessor + kind-incompatible) grow length to i+1 per §10.4.2 — required by the dominant `var arr = []; defineProperty(arr, "2", {get}); arr.every(cb)` cluster shape.
- Emission-order discipline: reserve-then-fill (accessor-driver pattern) with SAFE NO-OP placeholders; finalize fill ordered between `fillDynamicForinVecArms` and `fillTaDynViewMopArms` (TA dyn-view arm keeps the front slot).

### Validation

- `tests/issue-3251.test.ts`: 18/18 (define/readback coherence, §10.1.6.3 redefine throws + SameValue, accessor+HOF, gOPD synthesis, fresh-index false defaults, companion-value readback, named expandos, no-regression guards).
- Repro probes A–K all flipped to correct (were all failing on main).
- **Host lane byte-identical** vs main (sha-equal on a defineProperty-heavy program) — all changes standalone-gated.
- Scoped descriptor suite (2042/2668/2992/3043/3116/define-property-patterns): 119/120 — the single failure (`issue-2668` for-in proto-attrs) fails identically on main.
- Read-lane collateral (3183/2190/2190b/2186/3098): exactly the same 3 pre-existing failures as main.
- tsc, prettier, biome lint clean.

### Expected conformance movement

Standalone lane only; addresses the two clusters (~250–300 host-free FAILs) filed in #3251. Broad-impact substrate — the `merge_group` standalone-floor re-validation is the authoritative gate.

### Epic status

S1 of 3 — the issue stays `in-progress`. S2 (write-side `writable:false`/setter enforcement + gOPD staleness) and S3 (ArraySetLength §10.4.2.1) boundaries are documented in the issue file.


ttraenkler:
**Park diagnosis (fable-1) — infrastructure collateral, not a PR regression.**

The failed merge_group run (29631783983) shows the standalone lane collapsed to **4,508 pass / 43,469 compile_error**, every CE reading `standalone target emitted host imports: env::console_log_externref, env::structuredClone (#2961)` — HOST-wrapper import signatures recorded under the standalone lane. **PR #3322 (unrelated: async-gen for-await destructuring) failed its merge_group (29632727762) with byte-identical counts — the same 4,508/43,469 and the same message** — the identical-cluster-across-unrelated-PRs drift/infra signature. The host-lane 'regression ratio' failure (78 non-CT / net −14, baseline 4 commits stale per the run's own drift warning) is the same class.

Local verification on this branch + current main (dfe4c47): all standalone probes compile with ZERO imports and the full issue suite passes (18/18); those import names do not exist as compiler emission sites for standalone.

The collapse correlates with STALE-BASE merge groups (fresh-based #3325 passed the same hour, same main tip). This branch has now merged current main. Removing the hold once per the auto-park protocol; escalated to the tech lead as a main-side lane/precompile-cache bug affecting every stale-base queue entry.

I run this diagnosis as the PR author (fable-1 agent).

ttraenkler:
Removing auto-park `hold` — the v7-baseline-vs-v8-run mismatch that produced the 189-row pass→other cluster (bucket 98a0a956e8e81851, with the baseline-drift warning) is resolved: the forced oracle-v8 baseline promote (run 29635854867) landed at 07:55Z and the baselines repo now serves oracle_version=8. Per standing authorization, re-validating against the v8 baseline with ONE one-shot enqueue — the merge_group will now diff v8-vs-v8, so the reclassification cluster should vanish. If a real cluster survives the v8-basis diff, it will re-park and I'll escalate (never-loop hereafter).

PR #3344, opened 2026-07-18
fix(#3389): settleReturn terminator for driven async-gen top-level return (S1)

ttraenkler:
## #3389 Slice 1 — `settleReturn` terminator for driven async-gen top-level `return`

Child of the #3178 standalone async-machinery umbrella; re-grounds #3132's banked
S4. Was stacked on #3388 (PR #3332, now merged) — the stack has collapsed onto
`main`.

### Problem

An own-scope `return` in an async-gen body bailed `analyzeAsyncGen`
(`containsOwnScopeReturn`), so the gen fell to the legacy host buffer
(`__gen_set_return` + the carrier-off `Promise_*` co-leak). Probe: wrapped
`async function* g(){ yield 1; return 42; }` LEAKED `__gen_set_return`.

### Slice 1

Admit a DIRECT top-level `return E` / bare `return;` and settle the current
`next()`-promise with the §27.6.3.8 return completion `{value: E, done: true}`
(distinct from fall-through's `{value: undefined}`), then complete the frame so
subsequent `next()` give `{value: undefined, done: true}`.

- `AsyncGenShape.returnExpr` + analyzer admission (async-cps.ts);
- `AsyncCfgTerminator.settleReturn` + `planAsyncGenCfg` tail (settleReturn →
  trailing settleDone);
- emitter arm mirroring `settleYield` with `done=1` (async-frame.ts) +
  `validateAsyncCfg` range-check.

**Correct-or-legacy (probe-verified):** a `return` nested in control flow, and
`return await P`, stay legacy. Promise-typed `E` bails on the carrier lane
(deferred §27.6.3.8 return-await, same policy as `yield await` / #3120).

### Validation

- Leak flips: wrapped `yield 1; return 42`, `return 5`, bare `yield;return;` →
  HOST-FREE (were leaking `__gen_set_return`); nested-return / return-await
  correctly stay legacy.
- `tests/issue-3389-asyncgen-return.test.ts`: 7/7 (value + done semantics,
  frame completion, correct-or-legacy bails, no-regression plain-yield).
- Blast radius (2906 3d-i/3d-ii, 3132, 3132-s2, 3388-rtdelegate, 3120): no NEW
  failures — the one `issue-3132-s2` fail is PRE-EXISTING on `main` (A/B
  confirmed with this change reverted).
- Gates: tsc 0, prettier, oracle-ratchet +0, loc-budget allow-listed,
  sync:conformance clean.

### Documented next slices (in the issue)

Slice 2 (consumer `.return()`/`.throw()` on driven frames) and the
try-across-yield follow-up (the corpus multiplier — most has-return corpus files
wrap `return` in try/finally across a yield, which correctly stays legacy after
slice 1).

Refs #3389 (Slice 1 — issue stays open for slices 2 + try-across-yield).


ttraenkler:
Park diagnosis (07:47:11Z, run 29635624795, job 88059099821): the checker itself flags this **LIKELY-REAL REGRESSION (baseline content-current, #2562)** — baseline confirmed 0 test262-relevant commits behind main (852c40a9). Net -33 pass, 86 non-CT regressions, ratio gate 50.9% (86/169).

Diffed the merged report against a freshly re-fetched current baseline (103 non-CT regressions found, close to the CI's own 86 count — small delta likely from timing). **96/103 (93%) share one exact signature**: `strict rerun: obj should have an own property <length|name|constructor|throw|adopt|defer|use|...>` — own-property-descriptor loss on built-in methods, spread across dozens of **completely unrelated** built-ins: Date, Math, RegExp, String, Array, Object, Map, WeakRef, DisposableStack, GeneratorPrototype, Promise, Reflect, FinalizationRegistry, Iterator, MapIterator, etc. All failures are prefixed `strict rerun:` specifically (the sloppy-mode pass is fine). A further 6 are closely related (`hasOwnProperty('length')` returning false, wrong `.length` value, `Cannot delete non-configurable property`, `Cannot assign to read only property 'name'`).

This blast radius (dozens of unrelated built-ins, strict-mode-rerun-specific) is very unlikely to originate in this PR's own change (#3389, async-gen `settleReturn` terminator) — looks like a cross-cutting regression in how built-in method property descriptors survive a strict-mode rerun, possibly from something else that recently landed on main. Leaving `hold` in place per protocol — this is a distinct signature, not the known drift/infra classes. Escalating to tech lead for a [CI-FIX] task / root-cause owner. — PR-queue shepherd

ttraenkler:
Removing auto-park `hold` — stale v7-basis phantom park from the oracle v7→v8 transition window. This PR's failing merge_group regression run fetched the v7 baseline (run started before the forced oracle-v8 baseline promote landed at 07:55:58Z), producing the ~180-row pass→other reclassification cluster with the BASELINE DRIFT WARNING — the classic v7-baseline-vs-v8-run mismatch, not a real regression. The baselines repo now serves oracle_version=8, so re-validating with ONE one-shot enqueue: the merge_group will diff v8-vs-v8 and the phantom cluster should vanish. If a real cluster SURVIVES the v8-basis diff, it re-parks and I escalate (never-loop hereafter). (Per tech-lead transition-window standing authorization.)

ttraenkler:
Park classification (tech lead): drift/flake collateral, not PR-caused. Evidence from failed run 29638046085: 104 of the raw regressions are compile_timeout (runner-contention flake class — pass→compile_timeout under CPU load), the remaining 83 non-CT files form ONE bucket signature (730e5aadc00df9f6) with the drift-warning banner active (baseline 4 commits/-22m behind during the post-07:56 rebuild churn), net −16 on a 25.5k base. Same treatment as #3351/#3322: single unhold for re-enqueue on a fresh group base.

PR #3345, opened 2026-07-18
fix(#3395 shape 3): mixed == string-ToNumber redundant extern.convert_any (child of #2039)

ttraenkler:
## #3395 shape 3 — `==` mixed string ToNumber double-convert → valid Wasm

Child of #2039's standalone invalid-Wasm bucket. Lands **shape 3** (the `==`
abstract-equality double-convert rows). **Stacked on PR #3343 (shape 2)** — the
#3395 issue file + its `loc-budget-allow` live there; this PR's own change is the
single `src/codegen/binary-ops.ts` diff + tests. Once #3343 lands, re-merging
main reduces this PR to just the shape-3 delta.

### Root cause

The mixed string⇄number/boolean `==` path (`binary-ops.ts` `emitToNumber`, the
noJsHost `__str_to_number` arm) emitted an **unconditional** `extern.convert_any`
before `__str_to_number`, assuming a native `$AnyString` **ref** operand. A
string-classified operand that ALREADY compiles to externref — a `new String(x)`
wrapper object — was thus double-converted: `extern.convert_any` on an externref
is invalid Wasm (`expected anyref, found call of type externref`, the
`true == new String("+1")` residual).

### Fix

Gate the `extern.convert_any` on the compiled operand's real ValType — emit it
only for a ref operand, skip it when already externref.

### Verification

- New shape-3 tests in `tests/issue-3395-extern-boxing.test.ts`: primitive ==
  wrapper-String compiles valid Wasm; **native-string ToNumber loose-equality
  still correct** (`"1"==1`→1, `""==0`→1, `"abc"==1`→0, `true=="1"`→1 — the box
  IS still emitted for a native string ref).
- `==`/coercion regression sweep green: loose-equality, issue-2073/2081/1910/
  1111/1134/2605-2606, string-coercion, string/boolean-relational, binary,
  issue-1917/2934/559/2191/2063 (130+ tests). The one issue-2081 failure
  (null==undefined) is **pre-existing on main** (verified by A/B), unrelated.

### Residual (runtime, NOT invalid-Wasm — separate follow-up)

`true == new String("+1")` now COMPILES valid but traps at runtime (`illegal
cast`): `__str_to_number` `ref.cast`s its operand to `$AnyString`, and a wrapper
object externref is not a bare `$AnyString`. Monotonic for the invalid-Wasm
bucket (was a compile-fail; eliminating the module-level invalid Wasm lets the
file's other assertions run). The wrapper-String ToNumber value semantics is a
documented follow-up in the issue file.


ttraenkler:
Removing auto-park `hold` — host-baseline-staleness thaw. This PR's ~200-row pass→other cluster (with the BASELINE DRIFT WARNING) was the v8 host baseline (852c40a9f5 @ 07:55Z) being test262-relevant commits STALE vs main, not a real regression. The forced catch-up refresh (run 29640124053) promoted a fresh baseline at sha 6ae06435c / 10:16Z — the drift is now moot. Re-validating with ONE one-shot enqueue against the fresh baseline; if a real cluster SURVIVES the fresh-basis diff it re-parks and I escalate (never-loop hereafter). (Per tech-lead thaw protocol.)

PR #3267, opened 2026-07-17
feat(test262): make original harness authoritative

ttraenkler:
## Summary

- make literal upstream Test262 harness assembly authoritative for the canonical local runner and sharded CI worker
- add `test262-fyi/data` as an optional submodule plus a separate test262.fyi reader/runner lane for independent assembler comparisons
- close the assignment, array-exotic, iterator, exception-identity, and captured-global gaps found by progressively larger matched samples
- keep `wrapTest()` only as an explicitly synthetic diagnostic path, tighten negative-phase and async verdict handling, and bump the Test262 oracle from v7 to v8
- split original-harness and project-runner parity regressions into separate Vitest forks so they stay within the normal 512 MB worker limit

## Validation

- `pnpm run typecheck`
- `pnpm run format:check`
- `pnpm run check:issue-ids:against-main`
- `pnpm run check:issues`
- `pnpm run check:verdict-oracle`
- `pnpm exec vitest run tests/issue-3363.test.ts ... tests/issue-3368.test.ts --reporter=dot` — 20/20 passed
- `pnpm exec vitest run tests/issue-3369.test.ts tests/issue-3369-project-runner.test.ts --reporter=dot` — 5/5 passed
- `pnpm exec vitest run tests/issue-3370.test.ts --reporter=dot` — 5/5 passed, including the identical 50-test unified-worker sample
- `pnpm exec vitest run tests/test262-fyi-runner.test.ts --reporter=dot` — 5/5 passed
- `pnpm exec vitest run tests/issue-3227-s4.test.ts --reporter=dot` — 4/4 passed
- `node scripts/run-test262-fyi.mjs --filter language/expressions/array --limit 50` — original harness 50/50 passed
- upstream `tests/issue-3359.test.ts` remains unchanged and passes

## Landing note

This intentionally changes the conformance verdict oracle. Landing requires an `ORACLE_REBASE=1` baseline refresh so the v7-to-v8 policy change is recorded explicitly.

PR #3351, opened 2026-07-18
fix(test262): retain merge parent for oracle promotion (#3370)

ttraenkler:
## Summary

- fetch two commits in both baseline-writer promotion jobs so `HEAD^1` is available
- let the post-merge writer discover #3370s bounded `trap-growth-allow` from the landed change set
- document the failed depth-1 oracle-v8 promotion and pin the workflow checkout contract

## Why

PR #3267 merged cleanly and all Test262 shards/regression guards passed, but the post-merge publisher refused `oob +4` and `unreachable +47` at tolerance zero. The allowance resolver was correct; the promotion checkout had only depth 1, so the merge parent was absent and the issue diff resolved empty.

This PR uses the normal merge queue and the existing scoped #3370 rebaseline ceilings. It does not change a repository-wide tolerance or use the emergency forced-refresh path.

## Validation

- `pnpm exec vitest run tests/issue-3303.test.ts --reporter=dot` (39/39)
- `pnpm exec prettier --check ...`
- `pnpm run check:issues`
- `pnpm run check:issue-ids:against-main`
- `pnpm run typecheck`
- `git diff --check`


ttraenkler:
Removing the auto-park `hold` — same stale-baseline congestion-collateral class as #3322/#3327: job 88053006925 cites baseline main-sha dfe4c47 (4 test262-relevant commits behind origin/main, -62m clock), identical baseline reference to the #3322 park. Small net regression (-27 pass, 93 non-CT files) triggering only the ratio gate (93/180 = 51.7%) on a thin improvements-denominator, consistent with drift noise under the ongoing merge_group runner starvation rather than a PR-caused failure — no host-import-leak collapse signature this time, but same drift source. Removing hold once per protocol; leaving CLEAN for auto-enqueue rather than manually re-adding while runners are still starved. — PR-queue shepherd

ttraenkler:
Removing auto-park `hold` — stale v7-basis phantom park from the oracle v7→v8 transition window. This PR's failing merge_group regression run fetched the v7 baseline (run started before the forced oracle-v8 baseline promote landed at 07:55:58Z), producing the ~180-row pass→other reclassification cluster with the BASELINE DRIFT WARNING — the classic v7-baseline-vs-v8-run mismatch, not a real regression. The baselines repo now serves oracle_version=8, so re-validating with ONE one-shot enqueue: the merge_group will diff v8-vs-v8 and the phantom cluster should vanish. If a real cluster SURVIVES the v8-basis diff, it re-parks and I escalate (never-loop hereafter). (Per tech-lead transition-window standing authorization.)

ttraenkler:
Removing auto-park `hold` — host-baseline-staleness thaw. This PR's ~200-row pass→other cluster (with the BASELINE DRIFT WARNING) was the v8 host baseline (852c40a9f5 @ 07:55Z) being test262-relevant commits STALE vs main, not a real regression. The forced catch-up refresh (run 29640124053) promoted a fresh baseline at sha 6ae06435c / 10:16Z — the drift is now moot. Re-validating with ONE one-shot enqueue against the fresh baseline; if a real cluster SURVIVES the fresh-basis diff it re-parks and I escalate (never-loop hereafter). (Per tech-lead thaw protocol.)

ttraenkler:
Removed auto-park `hold` — cited merge_group failure (ratio-gate 42.6%, 75 regr / 176 impr against baseline dae79d5a) was the oracle v7→v8 transition drift cluster, now moot after the refreshed baseline + relaxed gates. Verified this PR's diff is CI baseline-promotion plumbing only (checkout fetch-depth 2 + predecessor-artifact oracle-mismatch guard) plus tests/issue-3303.test.ts and the #3370 issue file — it does NOT touch oracle verdict logic or oracle_version, so the ~176/75 cluster cannot originate from this diff. Merged current upstream/main (was 150 behind), issue-3303 test 39/39 green. Re-validating clean.

PR #3288, opened 2026-07-17
fix(#2961): warning-first standalone host-import leak scan (phase 1)

ttraenkler:
## #2961 phase 1 — extend the no-leak guarantee toward `--target standalone`

`strictNoHostImports` (the structural no-host-leak guarantee) auto-enables only for `--target wasi`. This lands the **warning-first** step of extending it to `--target standalone`, per the issue's 3-step decomposition (a=#3009 merged, this PR = b+c warning-first).

### (b) Enumeration — the standalone leak set
Swept ~60 feature snippets + the repo's playground examples under `target: "standalone"`. **The leak set is narrow, finite, and entirely genuine host-only dependencies** (confirms the re-scope — enumerate-and-gate, not Fable-lane feature work):

- **String-extern bridge** `__str_from_mem` / `__str_to_mem` / `__str_extern_len` — `console.log`/`process.*` host-string marshalling (the #3009 bridge).
- **Timer** `__timer_set_timeout` — `setTimeout` (no standalone event loop, like wasi).
- **DOM extern methods** `Document_*`, `Element_*`, `HTMLElement_get_style`, `Node_appendChild`, `CSSStyleDeclaration_set_cssText` — DOM needs a browser host (wasi already refuses these).

Everything else (arithmetic, classes, strings, numbers, arrays, Map/Set/WeakMap, Math, RegExp, `JSON.stringify`, generators, async, TypedArray, DataView, `Object.*`, spread, optional-chain, getters, super, tagged templates, Reflect) is **already host-free** — zero `env` imports. Full table in the issue file.

### (c) Warning-first gate flip
- `assertNoLeakedHostImports` (`src/codegen/index.ts`) now fires for plain `ctx.standalone` at **warning** severity; wasi / explicit `strictNoHostImports` stay **error**.
- The per-call `addImport` gate is unchanged (still strict-only), so **no import is dropped and the emitted binary is byte-identical** — the `host_free_pass` floor (`check-standalone-highwater`) keys on emitted `env::` imports and **cannot move**. Every standalone leak now gets a source-located advisory naming the import + citing #2961.
- `buildLeakedHostImportError(leak, severity)` grew a severity arg (default `error`, back-compat); the warning message omits the `Codegen error:` hard-fail marker.
- Escape hatch `JS2WASM_STANDALONE_LEAK_SCAN=0` for A/B control.
- **No allowlist growth** — phase 1 needs none.

### Phase 2 (gated follow-up, NOT this PR)
[...]

PR #3358, opened 2026-07-18
fix(#3411): CACHE-SUSPECT guard for the standalone-lane host-import collapse

ttraenkler:
## Summary

Stale-base `merge_group` runs collapse the **standalone** test262 lane to a byte-identical **~4,508 pass / ~43,469 compile_error**, every CE reading `standalone target emitted host imports: env::console_log_externref, env::structuredClone (#2961)` — HOST-lane import signatures recorded under the standalone lane.

This is **infrastructure poisoning, not a regression** (fable-1 diagnosis on #3327): unrelated PRs #3327/#3322 showed the *byte-identical* cluster, fresh-based #3325 passed the same hour on the same main tip, and local standalone compiles emit zero host imports. Without a guard the −38k drop trips the #2097 floor + #1668 catastrophic guard as a real regression, so `auto-park` (#2547) HOLD-labels the **innocent** PR — parking two of them on 2026-07-18.

## What this ships

`scripts/check-standalone-cache-poison.mjs` scans the merged standalone results JSONL; when **>90%** of non-skip records (over a ≥1000 corpus) are the #2961 host-import verdict, it exits **3** with a loud `CACHE-SUSPECT` diagnosis. Wired into `test262-sharded.yml` **before** the floor / catastrophic guards, so a poisoned run fails as clearly-labeled INFRASTRUCTURE (re-run the group) instead of the misleading standalone "regression".

Narrow by design (tested): only the exact #2961 verdict counts (an unrelated mass-CE passes → real regressions still reach the real gates); a sub-1000 corpus passes; a missing JSONL passes.

## Validation

`tests/issue-3411-cache-poison-guard.test.ts` — 4/4 pass: poison→exit 3, healthy→0, below-min-corpus→0, unrelated-mass-CE→0 (real regression NOT swallowed).

## Scope note

This guard **stops the parking damage** regardless of exact root cause. The pinpoint fix (identify whether the host-imports-in-standalone come from a stale-base bundle artifact or a scoped-restore lane mix, then put the lane in that cache/restore identity) is documented in `plan/issues/3411-*.md` as a follow-up.

### DIFF
--- a/.github/workflows/test262-sharded.yml
+++ b/.github/workflows/test262-sharded.yml
@@ -655,6 +655,20 @@ jobs:
             --baseline-generated-at "$(date -u +%Y-%m-%dT%H:%M:%SZ)" \
             ${{ github.event_name != 'workflow_dispatch' && '--include-proposals' || (inputs.include_proposals && '--include-proposals' || '') }}
 
+      # #3411 — CACHE-SUSPECT guard (runs BEFORE the floor / catastrophic
+      # guards). Stale-base merge_group runs have collapsed the standalone lane
+      # to a byte-identical ~4,508 pass / ~43,469 host-import compile_error —
+      # infrastructure poisoning, NOT a regression. Detecting the >90%
+      # host-import cluster here and failing as CACHE-SUSPECT (exit 3) gives the
+      # shepherd a re-run signal instead of the misleading −38k standalone
+      # "regression" the floor/#1668 guards would otherwise report (which
+      # auto-park then holds as a real regression, parking an innocent PR).
+      - name: Standalone cache-poison guard (#3411)
+        if: env.SHARDS_RAN == 'true'
+        run: |
+          node scripts/check-standalone-cache-poison.mjs \
+            --jsonl merged-reports/test262-standalone-results-merged.jsonl
+
       # #2097 — ABSOLUTE standalone pass-count floor (high-water-mark backstop).
       # The #1897 standalone gate is a MOVING floor (re-seeded from the new
       # baseline every push to main), so a run of small net-negative PRs each
--- a/plan/issues/3411-standalone-lane-cache-poisoning-stale-base-merge-group.md
+++ b/plan/issues/3411-standalone-lane-cache-poisoning-stale-base-merge-group.md
@@ -1,0 +1,117 @@
+---
+id: 3411
+title: "Standalone lane cache poisoning in stale-base merge_groups — 4508/43469 host-import collapse parks innocent PRs"
+status: in-progress
+assignee: ttraenkler/opus-dev-a
+sprint: current
+priority: high
+feasibility: medium
+reasoning_effort: medium
+task_type: bugfix
+area: ci
+language_feature: infrastructure
+horizon: m
+goal: ci-reliability
+related: [2961, 2097, 1668, 2547, 3380, 1521]
+origin: "2026-07-18 — fable-1 park diagnosis on PR #3327; byte-identical collapse also on #3322"
+---
+
+# #3411 — standalone lane cache poisoning in stale-base merge_groups
+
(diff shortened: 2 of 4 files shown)


ttraenkler:
**Reconciliation note from the fable-final orchestrator (session aaf22b5f) re: the standalone-baseline restore (2e9d3db, "cache-poisoned #3411 collapse"):**

Two independent, cache-free evidence sets show the 4,312/38k-compile-error standalone signature is NOT cache poisoning — it is the honest, deterministic result of the oracle-v8 literal-harness shim leaking exactly 2 unused host imports (`console_log_externref` from `var print`, `structuredClone` from `$262.detachArrayBuffer`) into every non-raw test, which the #2961 strict gate refuses at compile time:

1. #3417's triage (PR #3360): 29,791 of 34,409 standalone host_import_leak rows are shim-only; `assert.js`/`sta.js` themselves compile 100% host-free.
2. #3418's implementer reproduced it deterministically with no cache involved: shim+assert+sta under `target:standalone` → exactly those 2 imports → refusal.

**PR #3362** (dead-binding elision, standalone-only, host byte-identical) fixes the leak itself — expected standalone recovery to ~18-20k under v8. Suggest NOT re-restoring v7 again after the next v8 promote; post-#3362 the honest v8 number lands near the v7 level and the discrepancy dissolves. If #3411 has separate evidence of an actual cache-poisoning mechanism, please link it here — both phenomena could coexist, but the 4,312 signature specifically is accounted for by the shim leak.

ttraenkler:
Blocking contradiction: the 4,312-pass / ~38,771 host-import result is the intentional oracle-v8 standalone outcome recorded by #3370, not cache poisoning. #3370 declares 20,700 bounded regressions for the measured 20,561-test reclassification specifically because standalone raw passes backed by host imports have no conformance meaning. The same signature on a fresh dispatch confirms determinism, not stale-base cache mixing. This guard would classify the approved honest baseline as infrastructure poison and prevent v8 publication. Please revise the diagnosis/guard before removing hold; baseline rollback 2e9d3dbe should be undone through an audited restore.

ttraenkler:
## Retracted — the premise was wrong (no cache poison)

Closing this PR and its stacked phase-2 branch (`issue-3411-phase2-refresh-baseline-guard`, never opened). Verified against `plan/issues/3370`: the standalone lane's ~4,312 pass / ~38,771 host-import CE is the **INTENTIONAL oracle_version-8 rebaseline**, not infrastructure poison.

- The authoritative-harness change (#3267/#3370) plus the #2961/#3288 compile-time host-import leak scan **reclassifies ~20,561 host-import-leaking "passes" as host-import compile_errors**. `24,840 − 20,561 ≈ 4,312` reconciles exactly with the restored good baseline.
- The "byte-identical collapse across unrelated PRs" was **#3288 riding in cumulative merge groups** — a real, correct reclassification, **not** a cache bug and **not** a `TEST262_TARGET` drop.

Consequences that make this guard actively harmful:

1. **The absolute >90% host-import threshold now describes the NORMAL v8 standalone lane** (host-import CE share is ~90% by design), so this guard would fail *every legitimate v8 standalone run*.
2. The stacked phase-2 floor raise (`1000 → 10000`) is likewise wrong: 4,312 is now the legitimate v8 standalone pass count, so a 10,000 floor would block *every honest v8 promote*.

A delta-based rework (fail only on a large host-import-share jump vs the published baseline) was considered but rejected: it would only duplicate protection the existing per-test regression gates already provide (`diff-test262` + the #1668 catastrophic guard + the #2097 floor all compare candidate-vs-baseline per test), and there is no real bug to catch. Task closed wrong-premise; no phase-3 bisect needed.

The scripts/tests here are sound engineering but guard a phantom, so they are not landing. — opus-dev-a

PR #3370, opened 2026-07-18
fix(#3432): stop nulling callable externref values via sibling-wrapper closure recast (stacked on #3419/PR #3368)

ttraenkler:
## Summary

**Stacked on PR #3368 (#3419)** — branched from its branch (explicit predecessor-stacking; the repro requires #3419's fixes to even reach this code path). Once #3368 lands, this PR's diff reduces to the #3432 commit.

The 'callable var from externref initializer' arm in `src/codegen/statements/variables.ts` signature-matches ONE `closureInfoByTypeIdx` entry (map order = creation order) and `emitGuardedRefCast`s the value to that struct. Closure wrapper structs are sibling `sub final` types with **creation-order-dependent RTTs** (#2873 memory) — so every stored closure of a sibling wrapper NULLS out. For externref slots (the var-hoisted common case) the narrowed value was then immediately widened BACK to externref: a pure value-destroying round-trip with zero upside (the #962 guard already refuses to narrow externref slots).

Corpus impact: `testTypedArray.js` does `var argFactory = ctorArgFactories[k]; argFactory.bind(undefined, constructor)` — the element read nulled (position-dependently: include-position killed k0-2,k4-7; body-position killed k3,k5-7 — exactly the RTT-creation-order signature), and `__bind_function` threw **'Function.prototype.bind called on non-callable'** — the dominant residual (34/40 of the deterministic sample) of the ~2k #3419 TypedArray bucket.

**Fix**: skip the match-and-recast whenever the local slot stays externref. Slots that genuinely narrow keep the old behavior (residual first-match-wrapper hazard noted in the issue for a future #2873-pattern root-cast).

## Verification
- WAT-verified root cause (the `ref.test (ref 30)` / else-`ref.null` sequence) — full diagnosis in `plan/issues/3432-*.md`.
- All 8 `typedArrayCtorArgFactories` elements now bind (`k0..k7:ok`, was position-dependent single survivor).
- 40-file bucket sample progresses past the bind blocker into distinct next-layer substrate gaps (dynamic `new TA()` extern-class, `ta.constructor`) — documented in the issue.
- `tests/issue-3432.test.ts` green; scoped bind suites (issue-1632a, issue-1337*) green; illegal-cast-closures-585 / issue-1712 failures reproduce identically on main (pre-existing env).


ttraenkler:
**Held by tech-lead — real merged-baseline regression (not drift).**

The regression that parked stacked child #3375 traces to **this PR's** `src/codegen/statements/variables.ts` change, not #3375's ctor-param diff:

- Repeatable on **two** baselines (stale + fresh `be98eb3`), byte-identical newly-trapping file list → rules out drift/flake.
- `null_deref` trap category grew **165 → 272 (+107)**; net **−12 pass** (28366 → 28354); regression ratio 15.2%.
- Newly-trapping files are all **Array-method element handling**: `Array.prototype.concat` spreadable variants, `is-concat-spreadable-get-order`, `copyWithin/coerced-values-start-change-target`, `flat/empty-object-elements`, `Array.fromAsync/this-constructor-operations` (+97 more).
- Regression signature `2190422b10c158d4` (5 non-CT files).

**Root cause:** the new `matchedClosureInfo && slotTypeForCast?.kind !== "externref"` guard (skip the destructive closure-recast when the slot stays externref) correctly fixed ~1.8k TypedArray `argFactory.bind` traps, but **over-applies** — for array-element reads (`var f = factories[k]`) whose slot is externref, downstream Array-method code that expected the recast now null-derefs. The fix needs to be narrowed so it skips the recast only for the genuinely-callable-externref case, not for every externref slot.

Leaving held until the guard is narrowed. A high-priority CI-FIX task is queued for the owner. (These merged-baseline `null_deref`/`quality` regressions are exactly the class PR-level checks don't catch — the PR is green; only `merge_group` re-validation sees it.)

PR #3373, opened 2026-07-18
fix(#3421): async harness protocol — deps.console routing + var-over-ambient-import module globals (4.6k bucket)

ttraenkler:
## Summary

The **biggest host-lane bucket** on the oracle-v8 baseline: **4,617 tests** fail with `async completion marker not observed`. Verify-first showed even `Promise.resolve(1).then(() => $DONE())` — and even a bare `$DONE()` — never surfaces the marker: the harness PROTOCOL was broken, not async semantics.

Two independent breaks (both fixed):

1. **`resolveImport` `case "console_log"` ignored `deps.console`** (`src/runtime.ts`). The runners pass a capturing `consoleProxy` via `buildImports(manifest, { console }, …)`, but console_log intents bound the GLOBAL console — the marker leaked to real stdout and the runner's capture array stayed empty. Now routes through `deps.console` with per-method fallback to the global (the runner proxy has only log/error/warn).
2. **`registerModuleGlobal` denied the shim's `var print = fn` its storage** (`src/codegen/declarations.ts`). The lib scan registers the ambient DOM `declare function print(): void` as an `env.print` host import; the gate `funcMap.has(name) → skip` conflated that import with a genuine user-function shadow, so `var print` got no `__mod_print` global — every cross-function `print(...)` call (which is exactly where `$DONE` runs) compiled to a **dropped no-op** (`i32.const 0; drop`, WAT-verified). Generalized the existing #2669 discrimination: only DEFINED functions (`idx >= numImportFuncs`) shadow a module-level var; per JS the script-level `var` binding wins over the ambient global.

## Measured
- Probe chain (direct print / `__consolePrintHandle__` / `$DONE` / print-from-function / real `.then($DONE)` / async fn): **0/6 → 6/6**.
- **30-file deterministic sample of the 4,617 bucket: 0 → 17 pass (57%)** — extrapolated ≈ **2,600 recovered tests**; residuals now report their REAL failures through the working protocol (`Test262:AsyncTestFailure: <actual reason>`), making them individually triageable.
- Canary: 40-file deterministic spread of currently-passing tests → **40/40 unchanged**.
- `tests/issue-3421.test.ts` (deps.console capture) green; prettier/tsc clean.

Complements #3419/#3432 (my other two burndown slices); independent branch off origin/main, no file overlap with open PRs.


ttraenkler:
Closing as a duplicate of #3372 (#3428, opened earlier) — independent root-cause convergence: I landed on the same two fixes (resolveImport console_log ignoring deps.console, and registerModuleGlobal treating the ambient env.print IMPORT as a user-function shadow so the shim's `var print = fn` got no module-global storage). #3372 is the superset (adds the asyncTest/$DONE-guard sandbox stub + worker wiring), so it should land instead.

Evidence from my independent verification, folded here to de-risk #3372's merge: minimal protocol probe chain 0/6 → 6/6 (direct print / __consolePrintHandle__ / $DONE / print-from-function / Promise.then($DONE) / async fn); 30-file deterministic sample of the 4,617-record bucket: 0 → 17 pass (57%, extrapolated ~2.6k recoveries), residuals now surfacing real Test262:AsyncTestFailure reasons; 40-file passing-test canary 40/40 unchanged. WAT-verified the no-op compile of cross-function print() calls (`i32.const 0; drop`) before the registerModuleGlobal fix.

PR #3415, opened 2026-07-19
Fix Test262 FYI runner parity gaps

ttraenkler:
## Summary

- run test262.fyi's literal source assembly through the canonical project worker
- preserve strict delete error identity and Object.create receiver state
- bridge native generators to Iterator helpers and preserve completed-result `undefined`
- keep ambiguous JavaScript harness parameters dynamic
- provide current RegExp `@@replace` flags semantics on older Node hosts

## Why

The original test262.fyi harness and the project runner disagreed even when they compiled the same tests. The remaining differences came from runner verdict/isolation behavior, unsafe compiler specialization in unannotated JavaScript, completed generator value handling, and a Node 24 RegExp protocol difference.

The optional `test262-fyi/data` submodule remains untouched.

## Validation

- `pnpm exec tsc --noEmit`
- focused Vitest suites — 46/46 passed
- exact original-harness residual batch — 4/4 passed, up from 0/4
- historical disagreement sample — 929/1026 passed directly on Node 24 / Unicode 16
  - all 97 remaining non-passes are host capability differences: 84 Unicode 17 RegExp-data cases and 13 newer Uint8Array base64/hex APIs
  - compiler/harness-only gap after host normalization: 0/1026

Documents the final residual batch in plan issue #3471, with supporting issues #3468–#3470.


ttraenkler:
Superseded. This branch is 28 commits behind main, DIRTY, and fails its own lint/quality/smoke. Its useful content (test262.fyi runner-parity plumbing) was extracted clean onto #3420; its compiler changes already landed via #3419 (param-inference) and #3416 (async-completion). Closing as superseded — see #3420.

PR #3419, opened 2026-07-19
fix(#3471): gate body-usage param inference on zero call sites (unsound f64 narrowing broke isSameValue → ~433 name/length tests)

ttraenkler:
## Summary

Fixes #3471. **The reported premise (try/catch / `instanceof TypeError`) was wrong** — verified on current `origin/main`: the strict readonly write IS caught correctly (`e instanceof TypeError === true`). The real defect is a **parameter-type-inference unsoundness**.

test262's `isSameValue(a, b)` (`if (a===0 && b===0) return 1/a===1/b; if (a!==a && b!==b) return true; return a===b;`) was compiled with **`(param f64 f64)`**. `inferParamTypeFromBody` narrows an untyped param to f64 on a **single** numeric body use (`1/a`), and it was run as a fallback whenever call-site inference returned `null` — but `null` conflates *"no call sites"* (sound) with *"called with `any`/polymorphic args"* (unsound). `isSameValue` has 8 internal call sites, all passing `any` args, so the fallback misfired.

With f64 params, string args coerce to **NaN** at the call boundary → `a!==a && b!==b` = `true && true` → `isSameValue("slice","unlikelyValue") === true`. That makes `propertyHelper.js`'s `isWritable` mis-report the failed non-writable write as *succeeded*, run its **revert** (a second strict write) which throws **outside any try** → the uncaught `Cannot assign to read only property` failing the ~433 `built-ins/**/name.js` + `length.js` tests.

## Fix

`inferParamTypeFromCallSites` now also reports `sawCallSite`; the new `inferImplicitAnyParamType` runs the body-usage fallback **only when a function has zero internal call sites** (its documented purpose). Polymorphic helpers keep boxed `externref` params.

- Recursive numeric kernels (`fib`/`fact`) keep f64 via call-site inference (the recursive call passes a number-typed arg).
- Host-only numeric entrypoints (zero call sites) keep f64 via the still-valid body fallback.
- Verified: `isSameValue` string args now compare by value; `fact` param stays f64 (=120); `#3055` numeric `isSameValue` still correct.

## Why the local runner still shows a different signature

In the same-process `runTest262File`, the sloppy phase's `delete obj[name]` (#3470's realm-leak) masks this bug with `should have an own property`. In CI's **sharded pool** the two phases land on different forks, so #3470 doesn't trigger and #3471's `Cannot assign to read only property` is the dominant signature (433 in the real baseline). This PR was verified via the **strict-only** harness runner (fresh realm), where all 5 sample repro files pass. #3470 and #3471 are complementary.

## Scope / broad-impact

Parameter inference touches every function — the full test262 delta is measured on CI/merge_group. Local checks: new `tests/issue-3471.test.ts` (7 cases; minimal reproducer returns 1 on `origin/main`, 0 with the fix), plus `#684`, `#2795`, `#3055`, ir-numeric-bool-equivalence, function-name-length, comparison/equality suites — all green. tsc/biome/prettier clean.

## Test plan

- [x] `tests/issue-3471.test.ts` — 7/7 pass
- [x] numeric-kernel regression guards (fact→f64, dbl→f64)
- [x] `#3055` isSameValue numeric, `#684` usage inference, ir-numeric-bool-equivalence
- [x] tsc `--noEmit` exit 0, biome/prettier clean, LOC ratchet satisfied

### DIFF
--- a/plan/issues/3471-host-strict-assign-readonly-property-uncaught-typeerror.md
+++ b/plan/issues/3471-host-strict-assign-readonly-property-uncaught-typeerror.md
@@ -1,7 +1,9 @@
 ---
 id: 3471
-title: "Host lane: strict-mode assignment to a non-writable host property throws but isn't caught as instanceof TypeError inside compiled try/catch (~433-541 test262 tests)"
-status: ready
+title: "Host lane: polymorphic comparator (isSameValue) param unsoundly narrowed to f64, corrupting string compares → false isWritable revert → uncaught TypeError (~433 test262 name/length tests)"
+status: done
+assignee: ttraenkler/senior-dev
+completed: 2026-07-19
 sprint: current
 created: 2026-07-19
 priority: high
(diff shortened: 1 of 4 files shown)


ttraenkler:
## Local verification update (#3470 has since landed on main and is merged into this branch)

Correction to the PR body: #3470 (host name/length realm restore) **has now landed** and is merged into this branch, so the same-process `runTest262File` end-to-end path is checkable locally for the first time.

**Acceptance criterion — all 5 sample repro files now pass end-to-end** via `runTest262File` (each in its own process): `Array/prototype/slice/name.js`, `Date/prototype/getYear/name.js`, `RegExp/prototype/compile/name.js`, `String/prototype/substr/name.js`, `DataView/prototype/getFloat64/name.js` → **all pass**.

**Broad-impact regression A/B** (param inference touches every untyped-param function). Ran a 70-file diverse numeric-helper-heavy sample (Math, Number.prototype, Array indexOf/includes, String.charCodeAt) with **only the two changed files swapped** between origin/main and this fix:

- **0 regressions** (pass→fail)
- **+3 net** (fail→pass), pass 51→54 / fail 19→16
- newly fixed: `Number.prototype.toExponential/name.js`, `toFixed/name.js`, `String.prototype.charCodeAt/name.js`

Name.js family sample (n=50): 36 pass — and **zero remaining `Cannot assign to read only property` (#3471) failures**; the residual local fails are `should have an own property` (a #3470 same-process realm artifact that does NOT occur in CI's sharded pool) plus unrelated bugs (e.g. AggregateError null-access).

**Note on the required checks:** the 114-job test262 shards are intentionally *skipped* at PR level (`SHARDS_RAN: false`) — the full-suite delta is measured in the **merge_group** re-validation, with auto-park as the safety net. The PR-level required checks (`cheap gate`, `merge shard reports` no-op, `quality`) are all green; `cross-backend-parity` and `smoke` also pass.

PR #3422, opened 2026-07-19
fix(#3472): coerce externref-slot local in native string += (unblocks #3468 routing floor)

ttraenkler:
## Summary

Fixes a **pre-existing** standalone string-codegen bug that #3468 documented as the blocker to its closure-own-property routing being net-positive.

Under `--target standalone`, a **function-EXPRESSION / closure** param used as a string in the native compound-assign fast path (`m = ''; m += 'x'`) compiled to **invalid Wasm**:

```
call[0] expected type (ref null $AnyString), found local.get of type externref
```

A function DECLARATION with the same body was fine. The test262 `assert.sameValue` message build (`if (message === undefined) { message = ''; } else { message += ' '; } message += 'Expected...'`) has exactly this shape, so once #3418's routing makes the harness actually store+invoke `assert.sameValue`, its body hit this bug and failed to instantiate — the ~391 assert tests would flip vacuous-pass → fail-to-instantiate (a regression). This fix removes that blocker.

## Root cause

`compileNativeStringCompoundAssignment` (`src/codegen/expressions/operator-assignment.ts`) loaded the LHS current value with a bare `local.get`. A closure param is physically an `externref` slot; when its static type flow-narrows to `string`, that externref was fed straight into `__str_concat` (arg0 = `ref null $AnyString`). The RHS operand arm already coerced externref→$AnyString; the LHS load lacked the mirror.

## Fix

After the LHS `local.get`, compute the local's PHYSICAL ValType (params at slots `[0, params.length)`, locals after) and, when `externref`, coerce with `any.convert_extern` + `ref.cast $AnyString`.

- **gc/host byte-identical**: the path is inside `if (ctx.nativeStrings && ...)`, off for gc/host.
- Store-back unaffected (assignment coercion already boxes $AnyString→externref).

## Verification

- `tests/issue-3472-funcexpr-string-param-concat.test.ts` — 3/3 (valid + correct via in-wasm string eq).
- `tests/issue-3468-closure-own-props.test.ts` — 13/13 (no routing regression).

## Stacking

DRAFT — stacks on **#3418** (`issue-3468-closure-own-props`). Enqueue only after #3418 lands on main; then merge main in and measure the combined standalone floor delta (routing + this fix = genuine assert firing, not fail-to-instantiate).


ttraenkler:
Superseded / duplicate. This bundles the #3472 native-string fix with all of #3418's closure substrate. The #3472 string-codegen fix is pre-existing and independent of #3418, so it's better landed isolated — that's #3423 (CLEAN, ready). The substrate lives in #3418's own PR. Merge order: land #3423 first (puts the string fix on main), then #3418 merges safely without the ~391-assert regression. Closing this bundled draft in favor of #3423 + #3418.

PR #3509, opened 2026-07-23
refactor(#3547): remove the #3024 funcref-cell struct.new stopgap (dead after #3534)

ttraenkler:
## Summary

Design **step 3** of #3534's option-(a) plan: removes the #3024 funcref-cell `struct.new` rebuild branch in `compileClosureCall` (`src/codegen/expressions/calls-closures.ts`), its `isSingleFuncRefWrapperStruct` helper, and the now-unused `refCellValueType` import. Net −12 LOC of codegen.

## Why removal is evidence-based, not optimistic (two halves, both load-bearing)

1. **The one known producer is eliminated.** #3534's instrumented root cause showed the funcref-typed "cells" this stopgap served were never real ref cells: the `variables.ts` declaration path retyped a boxed capture's cell local to the closure STRUCT and re-registered that struct as `boxed.refCellTypeIdx` (a closure struct's field 0 is `funcref` — hence the "bare funcref cell" the #3024 call site observed). #3534/#3505 fixed the retype at the source; the declaration now writes THROUGH the cell and never retypes it.
2. **A probe confirms no others exist.** "No producers found" alone would be weaker evidence — a probe only shows absence on covered surfaces. Combined with (1) the argument closes: an env-gated probe in `getOrRegisterRefCellType` (flags any cell minted over `funcref` or a closure-struct/wrapper ref) shows **ZERO hits on the POST-#3505-merge tree** (which includes #3504 and today's other codegen merges) across: the 13-case closure corpus, 5 dedicated mutual-recursion / nested-fndecl / accessor-transitive shapes, the 4 module-reassignment shapes, all 7 matcher-invoking files, the full 80-file `Function/prototype/toString` dir, and the 34-file `class/elements/*-literal-names` dir. Recipe in #3547's issue file.

## Post-removal validation (all measured)

- **13-case corpus: byte-identical** — every sha256 unchanged vs pre-removal (recorded in #3534's issue file).
- toString dir sweep: identical **23 pass / 57 fail**; the single `illegal cast` row (`S15.3.4.2_A16.js`) is **pre-existing** with an identical signature on the pre-#3534 baseline (distinct `__module_init` toString.call mechanism).
- class-elements cluster: identical **30 / 4**.
- Guard tests `issue-3534-closure-value-representation` (6) + `issue-3024-tostring-closure-funcref` (1): green. `tsc --noEmit` + prettier clean.

## Also in this PR

- **Files #3546** (allocated via claim-issue): module TOP-LEVEL closure reassignment (`let f = () => 1; f = () => 2;`) writes only the `__module_init` local shadow, so cross-function `f()` returns the stale first closure from the global. Verified scoping matrix (function-scope reassign PASSES; top-level `let`/`var` both WRONG) + repro + sha256. Pre-existing (identical result + hash on pre-#3505 baseline) — the assignment-path sibling of #3534, status `ready`/Backlog.
- **Corrects #3534's bookkeeping**: toString-dir trap elimination is **67 → 1**, not 67 → 0 (the residual 1 is the pre-existing `S15.3.4.2_A16.js` row; family-attributable elimination = 66; the "0" was read from an accidentally head-truncated sweep capture).

### DIFF
--- a/plan/issues/3534-tostring-matcher-closure-construct-runtime-trap.md
+++ b/plan/issues/3534-tostring-matcher-closure-construct-runtime-trap.md
@@ -298,7 +298,11 @@ externref by construction. A3-as-audited needs no code.
 
 GC (host) lane:
 - `built-ins/Function/prototype/toString` (80 files): **11 → 23 pass (+12, 0
-  lost)**; `illegal cast` rows **67 → 0**. The 57 remaining fails are genuine
+  lost)**; `illegal cast` rows **67 → 1** (corrected in #3547: the residual 1,
+  `S15.3.4.2_A16.js`, is PRE-EXISTING with an identical signature on the
+  pre-fix baseline — a distinct `__module_init` toString.call mechanism;
+  family-attributable elimination is 66. The original "0" was read from an
+  accidentally head-truncated sweep capture). The 57 remaining fails are genuine
   Test262Error oracle verdicts from a DIFFERENT defect (`"" + fn` yields
   `[object Object]` / callback-shim source instead of NativeFunction syntax)
   → filed as **#3540**, out of scope here.
--- a/plan/issues/3546-module-toplevel-closure-reassign-stale-global.md
+++ b/plan/issues/3546-module-toplevel-closure-reassign-stale-global.md
@@ -1,0 +1,70 @@
+---
+id: 3546
+title: "codegen: module TOP-LEVEL closure reassignment writes only the __module_init local shadow — cross-function calls read the stale first closure from the global"
+status: ready
+sprint: Backlog
+created: 2026-07-23
+priority: medium
+horizon: s
+feasibility: medium
+reasoning_effort: high
+task_type: bugfix
+area: codegen
+language_feature: closures
+goal: correctness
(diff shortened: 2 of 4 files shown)


ttraenkler:
**Branch refreshed to current main (merge of 6f95f985 — includes #3505, #3506 vec-expando-props, #3507) and the zero-producer probe RE-RUN on the refreshed tree**, per the removal's own justification requiring "no producers on the *current* tree":

- probe hits: **0** across the 13-case closure corpus, the 5 dedicated mutual-recursion/nested-fndecl/accessor shapes, the 4 module-reassignment shapes (#3546), the full 80-file `Function/prototype/toString` dir, and the 34-file `class/elements/*-literal-names` dir — all driven through the merged tree including #3506's new `src/codegen/vec-props.ts`.
- sweep tallies unchanged (toString 23/57 with only the pre-existing `S15.3.4.2_A16.js` trap row; class-elements 30/4).
- guard tests `issue-3534-closure-value-representation` (6) + `issue-3024-tostring-closure-funcref` (1) green on the merged tree.

The evidence now matches the merged state, not the pre-merge state.

PR #3511, opened 2026-07-23
fix(#3548): infer under-applied declared-function params NULLABLE — zero-arg call of a string-applied function trapped unconditionally

ttraenkler:
## Root cause (soundness, not async)

`function d(x){...}; d('m'); d();` **traps unconditionally in standalone** on the zero-arg (usually passing) path. `inferParamTypeFromCallSites` skipped absent arguments entirely, so the param was inferred as a NON-NULLABLE native-string `(ref $AnyString)`; the zero-arg site's arity pad has no undefined inhabitant for a non-nullable ref and could only emit `ref.null` + `ref.as_non_null` — a guaranteed null-deref trap. The landmine was even documented in `pushDefaultValue`'s "ref" case ("parameter-padding contexts typically don't reach non-null ref params with null values") — they do: this is the canonical test262 template shape (`$DONE('msg')` on fail paths, bare `$DONE()` on the pass path), which is where the 193-row 'async continuation threw' Promise cluster came from. Nothing about it is async.

## Fix — upstream in inference, NOT at the pad site

A non-nullable ref has no valid filler, so the pad cannot be fixed; the **inference** that produced a non-nullable type for a sometimes-absent param is what is unsound.

1. **`param-return-inference.ts`**: any under-applied call site records `sawUnderApplied`; a non-nullable `ref` inference widens to `ref null N` of the SAME type — **not** externref, keeping the precise type and the fast paths. The zero-arg pad then emits a plain `ref.null` against the same nullable signature the callee compiles with.
2. **`native-strings.ts` / `coercion-engine.ts`**: with the param now nullable, `if (err)` routed the nullable string through `__str_flatten(null)` — a second trap. ToBoolean of a `ref_null` native string now goes through a new null-guarded `__str_truthy` helper (null → falsy, rope-len > 0 otherwise); the non-null `ref` arm keeps its historical flatten path byte-identical.

Rejected alternatives (arbitrated): sentinel/optional-param transform (a workaround over a still-wrong inference); never-narrowing declared-function params (the narrowing is a deliberate hot-path optimisation — the bug is only the under-applied case).

## Gate + measured flips (real runs, not candidate counts)

- **Traps before / passes after**: the 2-line repro traps with `RuntimeError: dereferencing a null pointer` on pre-fix main, passes on this branch. Permanent tests: `tests/issue-3548.test.ts` (repro, truthiness T/F/F, canonical $DONE template completing on the zero-arg PASS path, fully-applied no-regression). Probes also verified object-shape, array, and f64 under-applied params behave correctly.
- **Stride-4 over the 193-row cluster** (49 files, `TEST262_TARGET=standalone` + path filter): **0 → 19 of 49 PASS** (baseline: all 193 rows FAIL on the 2026-07-23 promote). Extrapolated ≈ 75 of 193. Residual 30 all fail with ONE signature — `illegal cast [in __then_fulfill_*]` (21 previously masked behind the arity trap) — the separate then-reaction marshalling defect, routed to #3443.

**Honest sizing**: corpus-wide static candidates ≈ 106 rows — hundreds, not thousands, overwhelmingly the $DONE family. This PR is justified on **soundness** (a documented-but-false assumption emitting a guaranteed trap; optional arguments are ubiquitous in real user code), not on corpus yield — do not read the modest count as 'not worth it'.

## Safety

- IR path: the #3536 signature-parity guard keeps the legacy body on any ABI mismatch, so the widened signature cannot produce an invalid module via IR patching.
- Scoped suites: issue-866/869/default-params/arguments-object/issue-1025 all green. The 3 failures in `tests/issue-3471.test.ts`/`tests/call-arg-type-coercion.test.ts` are **pre-existing on main** (verified identical on the base checkout) — not introduced here.
- Local gates: tsc, prettier, oracle-ratchet (+0), LOC budget (granted in the issue file) all green.

Resumes and lands the fable-3417 WIP (predecessor hit its API limit mid-implementation).

Closes #3548.

### DIFF
--- a/plan/issues/3178-standalone-host-async-machinery-retirement-umbrella.md
+++ b/plan/issues/3178-standalone-host-async-machinery-retirement-umbrella.md
@@ -57,9 +57,12 @@ the canonical `$DONE('msg')`/`$DONE()` template shape is just where the
 corpus exercises under-applied+string-applied functions). Fixed in child
 **#3548** (done): under-applied call sites make a non-nullable ref param
 inference NULLABLE + a null-guarded `__str_truthy` ToBoolean for nullable
-strings. Measured stride-4: 2 → 20 of 49 PASS. **Residual 29 = the
-`illegal cast [in __then_fulfill_*]` then-reaction-wrapper sub-family** —
-a distinct marshalling defect, the next open head here (also #3443).
+strings. Measured stride-4: 0 → 19 of 49 PASS (baseline: all cluster rows
+FAIL on the 2026-07-23 promote; ≈75 of 193 extrapolated). **Residual 30
+all now fail as `illegal cast [in __then_fulfill_*]`** — the
+then-reaction-wrapper sub-family (only 9 showed it originally; 21 were
+masked behind the arity trap) — a distinct marshalling defect, the next
+open head here (also #3443).
 
 ## Decomposition 2026-07-17 (fable-3178) — child issues #3386–#3391
 
--- a/plan/issues/3178-standalone-host-async-machinery-retirement-umbrella.md
+++ b/plan/issues/3178-standalone-host-async-machinery-retirement-umbrella.md
@@ -51,6 +51,16 @@ PASS on a stride-4 cluster sample; 3 residuals =
 `language/arguments-object/*async-gen*` (`Cannot access property on null or
 undefined`) — distinct root cause, still open here.
 
+Fourth slice, same day: the ~193-row `async continuation threw` null-deref
+cluster collapsed to an ARITY-FILL soundness bug (nothing async about it —
+the canonical `$DONE('msg')`/`$DONE()` template shape is just where the
+corpus exercises under-applied+string-applied functions). Fixed in child
+**#3548** (done): under-applied call sites make a non-nullable ref param
+inference NULLABLE + a null-guarded `__str_truthy` ToBoolean for nullable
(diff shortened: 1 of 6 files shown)


ttraenkler:
Verification requested by the lead before self-merge (interaction check vs the #3503-suspect main breakage):

Ran `tests/issue-3471.test.ts` + `tests/call-arg-type-coercion.test.ts` on this branch AND on the base main checkout. **Byte-identical results**: same 3 failing tests, same signatures (incl. the exact `Compiling function #4:"check" failed: call[0] expected type f64, found call of type externref @+350` offset), same 3 failed / 10 passed of 13. This PR is behavior-neutral on that surface — the failures are pre-existing on main, not introduced or altered here.

Scope argument for why this PR cannot alter narrowing decisions for fully-applied params: the only inference change sets `sawUnderApplied` when a call site has `arguments.length <= paramIndex`; when every call site supplies the param the returned type is bit-identical to before. The widening itself maps only `ref → ref_null` of the same typeIdx — it can never produce, remove, or alter an f64/i32/externref narrowing decision (the #3471 guard class). The passing #3471 guards (recursive kernel still f64, host-only entrypoint still f64, polymorphic comparator stays externref) plus the `fully-applied string params keep the precise non-null path` test in `tests/issue-3548.test.ts` pin this empirically.

PR #3550, opened 2026-07-24
fix(#3565): restore 3 IR designed demote-to-legacy contracts broken by #3341/#3519

ttraenkler:
## Summary

Continues the STRICT-IR over-promotion sweep that #680 started. #3341/#3519 made the IR overlay treat any post-claim failure whose `outcome.kind === "invariant"` as a **HARD compile error** (`formatIrPathFallbackDiagnostic`) — the correct fix for catching real invalid-Wasm emission, but it over-reached. **Three sites are DESIGNED demote-to-legacy points** whose own code comments document a "clean throw → legacy" / "demotes the function to legacy" contract; a plain `throw new Error` (or an untyped verify error) at those sites was classified as the generic `unexpected-internal-throw` / `verifier-failure` invariant → hard, **failing the compile of valid programs that have a working legacy body**. #680 fixed the two most-visible casualties (generator host-import ref + `.next()` method-call throw) but left these three, silently hard for ~7 days (invisible outside required checks, #3008).

## Honest framing — this is a ~0 test262 flip PR

Measured with a two-compile-per-file probe (IR overlay vs `experimentalIR:false`), casualty = IR hard-fails on an `[IR-FALLBACK]` invariant while pure legacy compiles the same file fine:

- **standalone** (1347 files @ stride 40): **0** casualties — #680 already captured that value.
- **host/default** (5322 files @ stride 10): **3** casualty files — TypedArray element-store, extern element-access, and one duplicate-fn-decl test262 file hitting the verify #1798 return-type gate.

Its value is **forward-looking contract restoration + #3008 invisible-regression closure**, NOT a pass bump: as IR coverage grows, more functions will hit these three demote paths and hard-fail where they must gracefully fall back to legacy. Restoring the documented behaviour now — before it bites wider — is the point.

## Fix (SURGICAL — no blanket `invariant`→`unsupported`)

The purpose of #3341/#3519 making `invariant` hard is to catch invalid-Wasm emission. So this types **only** the three documented-demote sites distinctly; every generic invariant stays a hard error:

- `outcomes.ts` — 3 new `IrUnsupportedCode`s (`element-store-unsupported`, `element-access-unsupported`, `return-type-legacy-coupling`); widen the `unsupported` failure `stage` to admit `"verify"`.
- `from-ast.ts` — `lowerElementStore` TypedArray-view + `lowerElementAccess` slice-12 terminal throws → `IrUnsupportedError` (→ warning → legacy). The sibling *internal* throws (`produced no value`, `unexpected IrType`) stay plain `Error` → hard. `lowerElementStore`'s other documented-demote siblings (packed-vec, non-vec, optional-store) are **knowingly left hard pending measurement** (they don't fire on the corpus; under-converting is the safe risk posture — see issue file).
- `verify.ts` — tag the #1798 return/early.return arity+type errors with `demote: true`; every other verify error (SSA, dominance, branch/instr type, block-id) stays hard.
- `integration.ts` — `verifyIntegrationFailure` routes `demote`-tagged verify errors to `unsupported`; post-hygiene/post-inline re-verifies untouched (still hard).

## Masking guard (the critical constraint)

`tests/issue-3565.test.ts` (5 subtests, folded into the required guard suite #3552): each of the 3 sites demotes (compiles), the #1798 gate error carries `demote:true`, and — critically — a **GENERIC invariant** (a real block-id/type desync; an injected build throw) STILL hard-fails, so #3341's invalid-Wasm-catching purpose is preserved (no masking).

## Validation

- Guard suite green (incl. #680/#3519/#3551 siblings), `issue-3519-ir-outcomes` 29/29 (injected verify failure stays `invariant`), `issue-1924` 11/11, `check:ir-fallbacks` OK, typecheck/lint/format clean.
- Post-fix probe: host casualties 2 → **0**; standalone remains 0. Re-validated on the merged state (origin/main incl. #3562).

**Note:** the real regression gates for a broad IR-path change (test262 regressions, standalone floor/net, catastrophic guard) are `merge_group`-only DESIGNED no-ops at PR level. The residual risk not covered by the CE→legacy reasoning is a caller-ABI/parity cascade (#3536/#3551 class); `equivalence-gate` + `linear-tests` cover the same-module dimension at PR level, and the cross-module dimension is validated in the `merge_group` re-run.

https://claude.ai/code/session_01Tr2Qx6KzQVhnXcGu167zeZ


ttraenkler:
Updated: added a **4th designed-demote site** (compound-assign with a non-f64 RHS, from-ast.ts) found by a bounded standalone-test audit — it flips `tests/issue-2079` (silently red on main) green (13/13). Same surgical pattern (distinct `compound-assign-unsupported` code). Also granted a `loc-budget-allow` in the issue frontmatter for the (mostly-comment) god-file growth the LOC ratchet flagged — that was the only failing check on the prior run; all equivalence-shards / linear-tests / equivalence-gate / sanitizers passed.

PR #3563, opened 2026-07-24
fix(#3024): pad missing method args in iterator next/return dispatcher (8 CE-eliminations)

ttraenkler:
## #3024 — iterator `next`/`return` dispatcher missing-arg pad

Clears the **8-file `__call_next`/`__call_return` arity sub-cluster** (`not enough arguments on the stack for call (need 2, got 1)`) of the #3024 invalid-Wasm residual. Continues the #3024 umbrella (sibling PR #3558 cleared the 25-file module-global slot-desync family).

### Root cause

`emitMethodDispatch` (src/codegen/index.ts) generates the module-level iterator-protocol dispatchers `__call_next`/`__call_return` (`(externref) -> externref`). For each user struct with a `<struct>_next`/`<struct>_return` method it emitted `local.get; ref.cast; call <method>` — passing ONLY the receiver. A user iterator method with a formal parameter (`next(value)` / `return(value)`) has an EXTRA wasm param (receiver + value), so the call was one argument short → invalid Wasm. Parameterless `next()`/`return()` were fine.

### Fix

The dispatcher structurally represents a protocol call with NO value argument (its own signature is `(externref) -> externref`), so pad each param beyond the receiver with the "missing trailing arg" default: real host `undefined` for an externref (untyped-JS) value param when `__get_undefined` is already imported (else `ref.null.extern`, byte-identical standalone); f64 sNaN sentinel / typed zero otherwise. **Byte-inert for parameterless iterator methods** (the common case: `extraParams = []`).

### Measured (honest)

- Repros: `next(v)`, `return(v)` → VALID (were invalid); `next()` control unchanged. Same on `--target standalone`.
- Full 587-candidate re-harvest on THIS branch (base `origin/main`, so #3558's 25 module-global files still counted): **97 → 89 invalid-Wasm, exactly these 8, 0 new signatures**. (Both PRs landed: main → 64.)
- **This is a CE-elimination slice, NOT a pass gain** (0 new passes): the 8 files now `fail` on DISTINCT feature gaps — Iterator.zip null-deref, AsyncFromSyncIterator promise handling — the actual unimplemented proposal semantics. Same shape as the prior #3024 packed-array slice. The fix is a general correctness bug (any user iterator with `next(value)`/`return(value)`), not just these files.
- Adjacent suites green (45 tests): issue-2038 (native `{next()}` carrier — directly exercises the dispatcher), generators, issue-2162-iterators, issue-2157, issue-2571-native-method-generators.

New `tests/issue-3024-iterator-dispatch-arity.test.ts` (3 tests).

### Deferred (roll forward, distinct root causes)

`super.x` in a static method (`C_method`) and `DisposableStack.move` (`__closure_9`) — separate arity mechanisms, not the dispatcher.

https://claude.ai/code/session_01Tr2Qx6KzQVhnXcGu167zeZ


ttraenkler:
## Conflict resolved — merged `origin/main` (rebase-adopted by dev-rebase-3563)

The authoring agent was gone and this PR had gone `DIRTY`. Adopted and merged (never rebased). Now `mergeable: MERGEABLE`.

**Not superseded — verified before investing.** The TaskList reconciler flagged #3024 as already having a merged PR, but it matches on the *umbrella* issue number and #3024 has ~15 merged slices. The merged sibling it saw is #3558 (module-global slot-desync, `global.set expected ...`, 25 files) — a **different error signature** from this slice (`not enough arguments on the stack`, 8 files), so no double-counting. Primary-source proof this slice is still needed: `git show origin/main:src/codegen/index.ts` — `emitMethodDispatch` still emits `local.get; ref.cast; call` with no arity pad, and `extraParams` does not exist on main.

### Conflict resolution

Exactly **one** conflicted file: `plan/issues/3024-invalid-wasm-default-lane-emitter-residual.md` — a planning artifact, resolved as a **union**, not `--theirs`. Both sides appended distinct hand-written sections to the same umbrella issue file; because main is merged *into* the branch, `--theirs` would have meant main's version and **silently deleted this PR's own "Landed: iterator ..." deliverable section**. Kept main's content first (chronological), then this slice's, and annotated main's "Still open" list to record that the `not enough arguments on the stack for call` row is now split (iterator half cleared here; `super`/`C_method` + `DisposableStack.move` still open).

Frontmatter kept main's version — it is a strict superset (carries #3558's extra `loc-budget-allow` entries plus the `collectDeclarations` `func-budget-allow`).

`src/codegen/index.ts` **auto-merged with no conflict** across the 54 intervening main commits.

### Post-merge revalidation (semantic, not just "it merged clean")

A textually clean auto-merge is not evidence of semantic soundness, so the dispatcher was re-exercised on the merged tree:

- `tests/issue-3024-iterator-dispatch-arity.test.ts` — **3/3 pass**, incl. the `next(v)`/`return(v)` repro and the parameterless byte-inert control.
- Adjacent iterator/generator suites — **45/45 pass**: `issue-2038` (native `{next()}` carrier, directly exercises the dispatcher), `generators`, `issue-2162-iterators`, `issue-2157`, `issue-2571-native-method-generators`.
- Gates green locally: `typecheck`, `check:loc-budget` (+43 in `index.ts`, granted), `check:dead-exports`, `check:func-budget`, `check:speculative-rollback`, `check:issue-ids:against-main`.

### Provenance of the harvest numbers (not re-measured)

The **97 → 89 / main → 64** figures in the PR body were measured **pre-merge**, on the branch based at `def8f82` — i.e. *before* #3558 landed, so its 25 files were still counted in that run. They have **not** been re-measured post-merge (a 587-candidate harvest is near-full-test262, out of scope for a dev branch). Post-merge evidence is the test + repro set above. Flagging this explicitly so the old numbers aren't read as post-merge measurements.

PR #3579, opened 2026-07-24
fix(#3591): root-cause 4 silent generator regressions (bisected to #3032 W6); fold both suites into the required guard set

ttraenkler:
## STEP 0 — independent verification on clean `origin/main` @ `7652f033774194`

| suite | result |
|---|---|
| `tests/issue-3164.test.ts` | **3 failed** / 13 |
| `tests/issue-3386.test.ts` | **1 failed** / 17 |
| `tests/issue-2961.test.ts` | **11 PASSED** — the flagged "#2961 ×4" is **STALE** |

The `#2961 ×4` report does not reproduce at all, and that suite is *already* in
the required guard set (added by #3561). No action needed there — a stale
**flag**, not a stale test.

## Determination: (B) THE PRODUCT IS WRONG — ×4. Nothing re-pinned.

All four are real regressions the tests correctly caught. `git bisect` over the
2,937-commit range `a5220f56..7652f033` (12 steps, automated probe):

```
last good  8bc6e1c3ccea74   both suites 30/30 PASS
first bad  1fbb1810bd0713   both suites 26/30 — THE SAME 4 FAIL
           feat(#3032): W6 — host-lane generator declarations route native
           (lazy §27.5 + next(v) two-way); GenState brands;
           sentinel-undefined reads (#3356)   — merged 2026-07-19
```

Both suites **predate** the culprit (`issue-3164` @ `a5220f56`, 2026-07-12;
`issue-3386` @ `3fa9b754`, 2026-07-18) and both were green at its parent — so
these are not never-pinned expectations, they are broken behaviour. The 4
failures at the culprit are identical to the 4 still failing on `main` today.

## Root cause (instrumented, not inferred)

`compileDeclarations` compiles the module-init body **twice** —
`declarations.ts:2312` (discovery) and `:2438` (final) — with top-level function
bodies compiled **between** them. A module-scope generator **function
expression** is lifted afresh on each pass, registering twice with two different
state-struct types for the *same* AST node:

```
closureName=__closure_0 selfTypeIdx=57  nodePos=7   →  stateTypeIdx=61
  .next() dispatch emitted here, registered=[["__closure_0",61]]
closureName=__closure_1 selfTypeIdx=108 nodePos=7   →  stateTypeIdx=111
```

`buildNativeGeneratorDispatch` emits an **inline `ref.test` ladder** frozen at
function-compile time, so it tests only pass 1's dead type `61`; the surviving
pass-2 closure builds `struct.new 111` and falls through to the #1344
`emitBrandCheckTypeError` arm → `TypeError: Generator.prototype.next requires
that 'this' be a Generator` (the 60-char message pinned the site).

**`for-of` survives** because its GENSTATE arm is filled at **finalize**
(`fillNativeIteratorLateArms`), when every registration is visible. That
asymmetry is the entire bug — measured across 8 shapes (decl vs fn-expr ×
binding vs direct × `.next()` vs `for-of`).

## Why this is filed, not fixed here

Reusing pass 1's info is unsound as-is: the lifted **self** struct type also
differs per pass (`57` vs `108`) and the state struct's `__self` field is typed
`ref_null <selfTypeIdx>` → pass 2 would emit invalid Wasm. The real fix is
either late-filling the resume-method dispatch (mirroring what already works for
`for-of`) or memoizing the lifted closure per AST node — a genuine refactor of
closure lifting × two-pass module init.

Filed as **#3584** with the full diagnosis, the 8-shape table, the bisect log and
two concrete fix candidates. The 4 cases are `it.skip`'d **with an inline
pointer** — per "file it properly and leave the test skipped with a pointer" —
never silently re-pinned.

## The durable half — closing the class

These rotted invisibly for 5 days because **neither suite is in the
required-checks set**: the #3008 per-PR gate runs only PR-**touched** root tests
(#3356 touched neither), and the test262 shard matrix is `merge_group`-only and
runs no root tests. So #3356 landed fully green while breaking 4 assertions.

Both suites are now in **`tests/guard-suite.json`** (#3552 — runs on every PR,
merge group and push), with `broken_by` provenance recorded. Same class of fix as
#3561 / #3562 / #3565.

## Validation

- guard suite: **11 files, 83 passed / 4 skipped, 33s** on the merged state
  (budget ~2 min)
- `issue-3164` + `issue-3386` + `issue-2961`: **37 passed, 4 skipped, 0 failed**
- `typecheck`, `lint`, `prettier`, `check:dead-exports`, `check:func-budget`,
  `check:loc-budget`, `check:issue-ids:against-main` — all clean

No `src/` changes; tests + manifest + issue file only.


ttraenkler:
**Renumbered #3584 → #3591 by the PR-queue shepherd** (the authoring session was unreachable, and this PR's `quality` check was red).

**Why:** `plan/issues/3584-native-gen-fnexpr-next-dispatch-stale-statetype.md` collided with `plan/issues/3584-auto-enqueue-blind-to-workflow-touching-prs.md`, which landed on `main` via PR #3577. The `check:issue-ids:against-main` gate (#2531) failed here with:

```
#3584: this branch adds plan/issues/3584-native-gen-fnexpr-next-dispatch-stale-statetype.md
       but origin/main already has plan/issues/3584-auto-enqueue-blind-to-workflow-touching-prs.md
```

**Tie-break:** that id was reserved on `origin/issue-assignments` at `2026-07-24T22:05:41Z`, ~29 min before this PR was opened, and #3577 was already queued while this PR was `BEHIND` — so per the established rule (the queued PR keeps the id) this branch renumbers.

**What changed — mechanical only:**
- `git mv` to `plan/issues/3591-native-gen-fnexpr-next-dispatch-stale-statetype.md`, `id: 3591`, heading updated
- `#3584` → `#3591` in `tests/issue-3164.test.ts`, `tests/issue-3386.test.ts`, `tests/guard-suite.json`, and the `plan/issues/3584-*.md` pointer
- new id `3591` from `claim-issue.mjs --allocate` (not hand-picked)
- merged `upstream/main` (this PR was `BEHIND`)

**Deliberately NOT touched:** the bare `3584` values in `tests/test262-slow-tests.json` are timing numbers, and `8429653584` is an artifact id — neither is an issue reference. **No test expectation, assertion or source behaviour was changed.**

**Verified:** `check:issue-ids:against-main` now passes, and `tests/issue-3164.test.ts` + `tests/issue-3386.test.ts` run **26 passed / 4 skipped** — identical to before (the 4 skips are the regressions this issue documents).

Also added a `## Related` cross-link to **#3586** (`s += yield` compound-assign not claimed by the native generator, landed in PR #3578) — same native-generator territory as this shape-gate work.

Nothing else on the branch was modified.

PR #3581, opened 2026-07-24
fix(#3024): track the real receiver offset in super method calls from a static method

ttraenkler:
## #3024 — `super.m(arg)` inside a **static** method emitted a call one argument short

`compileSuperMethodCallCore` (`src/codegen/expressions/new-super.ts`) pushes the receiver only when a `this` local exists, but then assumed a receiver had **always** been pushed:

```ts
const selfIdx = fctx.localMap.get("this");
if (selfIdx !== undefined) { fctx.body.push({ op: "local.get", index: selfIdx }); }
...
const superParamCount = paramTypes ? paramTypes.length - 1 : expr.arguments.length;  // <- hardcoded
```

A **static** method has no `this` local, and the parent's compiled static method has **no receiver param** either — so the hardcoded `-1` was wrong. The first real argument was mis-binned as an "extra" arg and **dropped**, and the pad loop started at `args.length + 1` (past the end) so it padded nothing.

The arithmetic reproduces each observed error **exactly**:

| shape | paramTypes | superParamCount | pushed | result |
|---|---|---|---|---|
| `Base_g(v)` | 1 | 0 | 0 | `need 1, got 0` |
| `Base_g(a,b)` | 2 | 1 | 1 | `need 2, got 1` |
| 0-arg | 0 | −1 | 0 | already VALID |

### Fix

Derive the offset from what was **actually** pushed — `selfOffset = pushedSelf ? 1 : 0` — and use it for the param count, the type hints, and the pad-loop start. **Byte-inert for instance super calls** (`pushedSelf` true ⇒ `selfOffset` 1 ⇒ identical to before).

This is a **general correctness bug, not a 2-test fix**: any `super.m(arg)` in a static method mis-compiled.

### Measured — by VALUE, not just validity

- `super.g(42)` → VALID, returns **42** (was invalid Wasm)
- `super.g(3,4)` → **34**, chosen so it pins argument **order**, not just count
- `super.g(1,2,3)` → 123; 2-level inheritance chain `super.g(10)` → 11
- **Before/after on the same 11-case suite**, by temporarily restoring `new-super.ts` from `origin/main`: **stock main 5 pass / 6 fail → this branch 11 pass / 0 fail**

**Corpus footprint (syntactic, with denominators):** of **53,009** test262 `.js` files scanned (excl. `_FIXTURE`), **296** mention `super.`/`super[` anywhere, and **12** have it inside a `static` member body — e.g. `class/super/in-static-methods.js`, `in-static-getter.js`, `in-static-setter.js`, `static-init-super-property.js`, `staging/sm/class/superPropStatics.js`. This is the count of files hitting the *shape*; it is **not** a claim that 12 tests flip to pass (several are the getter/field shape this PR deliberately does not fix).

### Deliberately NOT fixed here — and why that matters

The static **property-read** half is a **distinct root cause**, filed as **#3589** (added in this PR).

I initially also padded the static *getter* path. It made the Wasm valid — and I nearly shipped on that basis. Checking **values** showed why not:

```wat
(func $Base_get_x (param (ref null 1)) (result f64))  ;; static getter is compiled INSTANCE-shaped
(func $C_m   ref.null 1 / ref.as_non_null / call 2)   ;; the naive pad
```

`ref.as_non_null` on a null ref **traps**. That trades a loud compile-time validation error for a **guaranteed runtime trap** — strictly worse. Reverted.

Relatedly, `super.<plain static field>` in a static method emits `f64.const 0` — **valid Wasm that silently returns the wrong answer**, live on `main` today and untouched by this PR. Both symptoms need the **class modelled as the receiver** (#3589).

Both are pinned as **KNOWN-OPEN assertions** in the new test (asserting `0`, and asserting the invalid-Wasm error) so a real fix flips them **loudly** instead of silently.

### Tests

New `tests/issue-3024-static-super-arity.test.ts` — **11 tests, all passing**: 4 fixed shapes, 5 controls (instance super method/getter, static setter, 0-arg static super, instance pad-loop path), 2 known-open.

Gates green locally: `typecheck`, `check:loc-budget` (+24 in `new-super.ts`, granted in #3589's frontmatter), `check:dead-exports`, `check:func-budget`, `check:speculative-rollback`, `check:issue-ids:against-main`.


ttraenkler:
**Auto-park diagnosed and cleared — this was a duplicate issue id, not a test262/quality regression.**

Per the park-handling rules I read the cited failure before touching the label. `merge_group` run 30132038409, job `quality`, step *Issue integrity + link gate (#1616)*:

```
--check FAILED: 1 duplicate IDs
```

**Cause:** this PR added `plan/issues/3589-static-super-property-read-class-receiver.md`, colliding with `plan/issues/3589-assert-harness-null-deref-unmasked-by-3563.md`, which landed on `main` via PR #3582. Note the collision was invisible at PR level — `check:issue-ids:against-main` was green here because #3582 had not merged yet when this PR's checks ran. It only surfaced in the `merge_group` on the merged state, which is exactly why auto-park (#2547) exists.

**Tie-break:** 3589 was reserved on `origin/issue-assignments` at `2026-07-24T22:30:26Z`, ~5 min before this PR was opened, and that issue is already merged to `main` — so this branch renumbers.

**Fix pushed — mechanical only:**
- `git mv` → `plan/issues/3594-static-super-property-read-class-receiver.md`, `id: 3594`, heading updated
- fresh id `3594` via `claim-issue.mjs --allocate` (not hand-picked)
- verified **no other file in the repo** referenced `#3589` for this issue
- merged `upstream/main`

**No source, test, or expectation was touched** — the `fix(#3024)` receiver-offset work on this branch is untouched.

**Verified locally:** `check-issue-ids --check` → *no duplicate issue IDs (3150 issues)*, and `--against-main` → *no branch-introduced issue file collides*.

Removing the `hold` so `auto-enqueue` can re-sweep. I am **not** re-enqueueing manually.

PR #3583, opened 2026-07-24
fix(#3592): stop silently dropping a top-level `throw` in the host/standalone lanes

ttraenkler:
Part of #3592 (RC1). `collectDeclarations` had a `ThrowStatement` arm gated on `ctx.wasi` (#2968), so outside WASI a bare top-level `throw` was collected into **nothing**: it emitted no code, `__module_init` ran to completion, and a module whose only statement is `throw new Test262Error(...)` scored **pass**. A silent wrong answer.

> **Issue id note:** this work was originally filed as #3585. `claim-issue.mjs --allocate` reserved that id, but another lane landed `plan/issues/3585-standalone-mapget-call-result-eq-false.md` on `main` while this was in flight, so the `quality` → "Issue-ID fresh-claim gate (#2531)" job correctly rejected it. Renumbered to **#3592** per `reference_cross_session_issue_id_collision_renumber_loser`. The branch name still says `3585`; the issue, tests and source refs all say `3592`.

### The measurement corrected the report twice

1. **It is not standalone-specific.** The **JS-host lane drops it too** — both lanes run a throw-only module to completion (measured through the real `assembleOriginalHarness` prefix + `deferTopLevelInit` + `__module_init` after `setExports`).
2. **It does not inflate the floor.** A TS-parser scan of the whole test262 tree finds exactly **40** non-`_FIXTURE` files with a top-level `ThrowStatement` (of 19,202 that mention `throw`), and almost all are `negative:` tests currently scoring **FAIL** ("expected Test262Error") *because* the throw never happens.

### Exhaustive A/B — a census, not a sample

Those 40 files **are** the complete exposed population. Local-vs-local, same runner (`runTest262File`), same process, only the collection toggled:

| lane           |   n | pass→pass | **fail→pass** | fail→fail | CE→CE | **pass→fail** |
| -------------- | --: | --------: | ------------: | --------: | ----: | ------------: |
| **standalone** |  40 |        26 |         **5** |         2 |     7 |         **0** |
| **host**       |  40 |        26 |         **5** |         3 |     6 |         **0** |

Same five gainers in both lanes: `language/module-code/eval-self-abrupt.js` and `language/line-terminators/comment-single-{cr,lf,ls,ps}.js`.

**Net +5 per lane, zero regressions.** The one standalone signature change (`eval-rqstd-abrupt.js`: `expected TypeError` → `uncaught Wasm-GC exception (non-stringifiable payload)`) lands in an **existing** `STANDALONE_ROOT_CAUSE_BUCKETS` entry, so the #3439 hard-0 unclassified gate is not at risk. The two host-lane signature changes carry no equivalent risk: `--max-unclassified-root-causes` is only wired to the standalone root-cause map, and there is no non-standalone bucket set with an unclassified gate.

### Deliberately NOT in this PR

The issue file records a **second, much larger** root cause found on the way: an under-applied call through `__apply_closure` never happens, which makes **every 2-arg `assert.sameValue`/`assert.throws`/`assert.notSameValue` vacuous in the standalone lane** (`assert.sameValue(1,2,"m")` throws correctly; `assert.sameValue(1,2)` does not). That fix is complete and verified on branch `issue-3585-apply-closure-arity`, and is **held back on purpose**: it flips **15 of the 100** previously-passing tests in a 200-file random sample — an honest-floor de-inflation that must be landed per the F1 recipe (park = measurement, honest-flips vs invalid-Wasm, cluster-route new signatures, `ORACLE_VERSION`), and it has a known blocker (a missing formal supplied as the undefined sentinel traps on a callee param that lowered to a concrete WasmGC ref).

### Test

`tests/issue-3592-toplevel-throw.test.ts` — 6 cases across host + standalone: a bare top-level throw reaches the host, a throw after other statements reaches the host, and a throw-free module still runs to completion.


ttraenkler:
**Park resolved: bounded `trap-growth-allow` declared** (tech-lead go-ahead; no source change on your branch, frontmatter only).

## Diagnosis

Auto-parked on the #3189 ratchet — one gate, nothing else:

```
GATE FAIL: trap category "unreachable" grew 2 → 3 (+1)
  Newly trapping: test/language/module-code/top-level-await/await-dynamic-import-rejection.js
=== Net: +16 pass (30918 → 30934) ===
=== Host stable-path fine-gate net: +39 (40 improvements − 1 regressions) ===
null_deref 159→159, illegal_cast 79→79, oob 59→59
```

## Why this needs a declaration, when #3563's near-identical park did not

I checked the **authoritative baseline jsonl**, not a local run:

| PR | file | baseline status | resolution |
|---|---|---|---|
| #3563 |  |  | **#3595** excludes it — merged with no declaration |
| **#3583** |  | **** () | **needs #3596** |

#3595 excludes baselines that **could not testify** (never instantiated:  /  / absent / identical ). Your file's baseline is  — the module **did** instantiate and returned a real verdict — so it is correctly *outside* that exclusion. Fixing the dropped top-level `throw` simply lets the test run past where it previously stopped, into a **pre-existing latent** unreachable trap. , flavour only; **the test has never passed.**

## What I added

```yaml
trap-growth-allow:
  count: 1
  reason: "#3596 reclassification: … baseline status is `fail` (negative_test_fail) …"
  tests:
    - test/language/module-code/top-level-await/await-dynamic-import-rejection.js
```

in `plan/issues/3592-…md`. Bounded to +1, names the single affected test, and is **machine-checked** by `evaluateTrapReclassification` (#3596, landed in #3589): named + not-previously-passing + no undeclared growth. Had that file been `pass` on the baseline the declaration would be **refused**, not honoured.

Also merged `upstream/main` (which carries #3595 and #3596). Removing the `hold`; **not** re-enqueueing manually.

If the re-run shows a *different* newly-trapping file or a second category, the declaration will correctly fail as incomplete rather than silently cover it — ping me and I'll re-measure.

ttraenkler:
**Park resolved: bounded `trap-growth-allow` declared** (tech-lead go-ahead; no source change on your branch, frontmatter only).

*(Reposting — my previous comment was mangled by shell expansion. This is the intended text.)*

## Diagnosis

Auto-parked on the #3189 ratchet — one gate, nothing else:

```
GATE FAIL: trap category "unreachable" grew 2 → 3 (+1)
  Newly trapping: test/language/module-code/top-level-await/await-dynamic-import-rejection.js
=== Net: +16 pass (30918 → 30934) ===
=== Host stable-path fine-gate net: +39 (40 improvements − 1 regressions) ===
null_deref 159→159, illegal_cast 79→79, oob 59→59
```

## Why this needs a declaration, when #3563's near-identical park did not

I checked the **authoritative baseline jsonl**, not a local run:

| PR | file | baseline status | resolution |
|---|---|---|---|
| #3563 | `Iterator/zip/iterables-iteration.js` | `compile_error` | **#3595** excludes it — merged with no declaration |
| **#3583** | `await-dynamic-import-rejection.js` | **`fail`** (`negative_test_fail`) | **needs #3596** |

#3595 excludes baselines that **could not testify** (never instantiated: `compile_error` / `compile_timeout` / absent / identical `wasm_sha`). Your file's baseline is `fail` — the module **did** instantiate and returned a real verdict — so it is correctly *outside* that exclusion. Fixing the dropped top-level `throw` simply lets the test run past where it previously stopped, into a **pre-existing latent** unreachable trap. `fail → fail`, flavour only; **the test has never passed.**

## What I added

```yaml
trap-growth-allow:
  count: 1
  reason: "#3596 reclassification: ... baseline status is `fail` (negative_test_fail) ..."
  tests:
    - test/language/module-code/top-level-await/await-dynamic-import-rejection.js
```

in `plan/issues/3592-...md`. Bounded to +1, names the single affected test, and is **machine-checked** by `evaluateTrapReclassification` (#3596, landed in #3589): named + not-previously-passing + no undeclared growth. Had that file been `pass` on the baseline the declaration would be **refused**, not honoured.

Also merged `upstream/main` (which carries #3595 and #3596). Removing the `hold`; **not** re-enqueueing manually.

If the re-run shows a *different* newly-trapping file or a second category, the declaration will correctly fail as incomplete rather than silently cover it — ping me and I'll re-measure.

PR #3585, opened 2026-07-24
feat(#3597): make the auto-park bot step-aware — name the failing step, don't park on infra-only failures

ttraenkler:
## The incident this fixes

On **2026-07-24** the merge queue produced **two** parks whose comments were
textually identical — `Failed checks: - check for test262 regressions`, no run
URL, no step name — for **opposite** situations:

| PR | what actually happened | correct action |
|---|---|---|
| **#3566** | shard-artifact download **403'd**; the verdict step never ran. Merged cleanly once unparked | should **NOT** have parked |
| **#3563** | the verdict step ran and caught a real uncatchable-trap regression | park was **correct** |

Each cost a full manual investigation to tell apart. A wrong park is not cheap
either: `enqueue-green-prs.mjs` **skips** held PRs, so it **strands** until a
human notices.

The Actions jobs API already returns `steps[]` with a per-step `conclusion` —
the script was discarding it (`--jq '.jobs[] | {name, conclusion}'`). Job-level
`failure` is identical for "the artifact download died" and "the regression
verdict fired".

## What changed

**1. Report the failing step and the URL.** `fetchJobs` now selects
`{name, conclusion, html_url, steps}`; the park comment renders
`- <job> — failing step: <step> ([job log](<url>))` plus a `Run: <url>` line.
This alone would have made the two parks above distinguishable at a glance.

**2. Don't park when only infra failed.** `classifyRun` gains `infraOnly` /
`unclassifiable` / `shouldPark`. When **every** failed step across **every**
failed job is a recognised setup/infra step (`Set up job`, `Checkout`, `Post *`,
`Set up node|pnpm|…`, `Download/Upload … artifact(s)`, container init/stop), the
verdict never ran → **do not park**. The run stays red, so the queue ejects and
`auto-enqueue` re-adds it — the correct response to a transient infra failure.

## Directionality — the load-bearing invariant

Wrong in the **permissive** direction lets a real regression into `main`; wrong
in the **strict** direction costs one label removal. So **park is the default**
and we skip only on *positive* evidence:

- a failed job whose failing step can't be identified (`steps` absent/empty) is
  `unclassifiable` → **parks**;
- any single non-infra failed step anywhere → **parks**;
- `INFRA_STEP_PATTERNS` is deliberately **tight** — widening it makes the bot
  park *less*, which is the dangerous direction.

The existing #2547 cancellation invariant (0 failed jobs ⇒ queue rebuild ⇒ never
park) is untouched and still covered.

## Tests

- `--self-check`: **10 → 23** pure-logic checks, no network.
- `tests/issue-3590-auto-park-step-aware.test.ts`: **30 cases, 3 ms** — both
  motivating shapes asserted explicitly (**#3566 must NOT park**, **#3563 MUST
  park**), every conservative-default path (no steps / empty steps / mixed
  infra+verdict / infra+unclassifiable), the cancellation invariant, and the
  comment rendering.
- Registered in **`tests/guard-suite.json`** so a later edit to the bot cannot
  silently un-do the classification.

## Follow-up (recorded in the issue, not done here)

**Retry the artifact download** in `test262-sharded.yml` — that removes the
#3566 failure at its source rather than classifying it after the fact. It is a
workflow-level change with its own validation needs; this PR makes the failure
*legible and non-parking*, which is the part that was costing investigations.

## Validation

guard suite **12 files / 87 passed / 20s**; `typecheck`, `lint`,
`check:dead-exports`, `check:func-budget`, `check:issue-ids:against-main` all
clean. No compiler `src/` changes.


ttraenkler:
Retitled `feat(#3590)` → `feat(#3597)` (title only — **nothing on the branch was touched**).

`#3590` on `main` is a different issue (`padMissingArg` emits `ref.null` + `ref.as_non_null`, an unconditional trap landmine). This PR's own issue file is `plan/issues/3597-auto-park-step-aware.md`, so the title now matches it. Cosmetic, but a wrong id in the title corrupts the audit trail once these land.

Unrelated FYI, since it concerns this PR's id: `claim-issue.mjs --allocate` handed **3597** to a second PR at `23:27:46Z`, ~12 min after this one was opened — your open-PR file wasn't seen by the allocator's scan. **This PR keeps 3597**; the other one renumbered itself to 3598. No action needed from you. Tracked in #3598 as evidence that the open-PR scan is unreliable inside the allocator, not just missing from the CI gate.

PR #3589, opened 2026-07-24
fix(#3596): per-PR machine-checked valve for the #3189 uncatchable-trap ratchet

ttraenkler:
Unblocks the class of PR that the #3189 ratchet parks incorrectly, **without** widening it for anything else.

## The gap

The ratchet cannot distinguish a **regression** (`pass → trap`) from a **reclassification** (`fail → fail`, flavour changed because a fix made a module compile far enough to reach a **pre-existing latent trap**). It parks the second as if it were the first:

| PR | net | ratchet delta | outcome |
|---|---|---|---|
| #3563 | **+11 pass** | `null_deref` 159 → 160 (+1) | parked |
| #3583 | **+16 pass** | `unreachable` 2 → 3 (+1) | parked |

Both were flavour changes on tests that **had never passed**. Both blocked, same evening, no way through.

Neither existing lever fit:
- **`TRAP_RATCHET_TOLERANCE`** (#3202) is a **repo-level** Actions variable — while open it blinds the ratchet for *every* queued PR, and the repo has a prior leak incident (hence `trap-tolerance-staleness-alert.yml`).
- **`trap-growth-allow:`** (#3370) was the right shape, but its read sat inside `if (rebaseMode)` (`scripts/diff-test262.ts` L1853), so it was **inert for ordinary same-oracle PRs**.

## The fix — extend the existing mechanism, machine-check the claim

`trap-growth-allow:` is now read in both modes. Rebase mode is unchanged. On a non-rebase PR it is honoured **only if verified**:

```yaml
trap-growth-allow:
  count: 1
  reason: "pre-existing assert-harness null-deref, unmasked not caused (#3593)"
  tests:
    - test/built-ins/Iterator/zip/iterables-iteration.js
```

`evaluateTrapReclassification` (pure, unit-tested) requires all three:

1. **Named** — a bare `count:` is refused outside a re-baseline; an uncheckable claim isn't a declaration.
2. **Not previously passing** — every named test must have a baseline row that is **not `pass`**. **`pass → trap` still hard-fails.** This is what stops the valve becoming an escape hatch. A missing baseline row is also refused (proves nothing).
3. **Complete** — every file causing the growth must be named, so `count: 1` can't quietly excuse unrelated growth in another category.

All prior containment preserved: per-category, positive integer, mandatory reason, declared in the granting issue's frontmatter, change-set scoped (an allowance on `main` grants nothing later), ceiling-not-blank-cheque, declarations don't sum. **`TRAP_RATCHET_TOLERANCE` was not touched.**

## Verification

- `tests/issue-3596-trap-growth-allow-nonrebase.test.ts` — **11 new tests**, including the load-bearing one: a named test that was `pass` on the baseline is refused *even when the ceiling would cover it*.
- `tests/issue-3303.test.ts` — **41 pass**. Two CLI companions added (named-but-passing refused; named `fail → trap` honoured).
- `tests/issue-3189.test.ts`, `tests/issue-2098.test.ts` — pass, unchanged.
- Typecheck clean; biome clean.

### One intentionally-changed existing test — flagging explicitly

`issue-3303`'s *"keeps trap-growth-allow inert without an oracle bump (#3370)"*. The **property** it protected is unchanged and still asserted: a bare declaration grants nothing on a same-oracle PR, still exit status 1. Only the **mechanism** changed — the declaration is now read and then *refused for being uncheckable* rather than never read — so the message assertion was updated and the test renamed. Rationale is recorded in #3596 rather than left implicit in a diff.

## Also in this PR

- **#3589 → `wont-fix`**, superseded by **#3593**, which proves pre-existence more directly (its minimized file traps identically *with and without* #3563's change). #3589 is retained for its `assert` L76 localisation and ruled-out-padding evidence.
- **#3597 filed** — `check:issue-ids` scans `main` but not open PRs, while `claim-issue.mjs --allocate` scans both. That asymmetry is why the #3581/#3582 collision was invisible at PR level and surfaced as a merge_group auto-park. Evidence from both of today's collisions included.

## Tooling warning worth carrying

Plain `grep` returns **nothing** on `scripts/diff-test262.ts` — treated as binary despite clean UTF-8 — so `grep -n "trap" scripts/diff-test262.ts` exits 1 and reads as a confident "not present". **Use `grep -a`.** This produced a wrong conclusion about the valve's existence once during this investigation.


ttraenkler:
**Update: the second issue file renumbered #3597 → #3598** (no change to the valve itself).

PR #3585 already had `plan/issues/3597-auto-park-step-aware.md` — opened `23:15:13Z`, ~12 min before `claim-issue.mjs --allocate` handed me 3597 at `23:27:46Z`. #3585 has priority, so this PR's file renumbers. Verified 3598 is free on `main` **and** unused across all 9 open PRs before taking it.

Worth noting because it sharpens the issue's own argument: **`--allocate` should have prevented this and didn't.** #3585 was already open when the id was handed out, yet its added issue file wasn't seen by the open-PR scan. So that scan is not just *missing from the CI gate* (the headline problem) — it's **also unreliable inside the allocator**. It's recorded in the issue as Collision C, and it's why the proposed fix enforces at the gate: verification at merge-decision time beats reservation at allocation time.

That makes three id collisions in ~90 minutes (3584, 3589, 3597), which is the evidence base for #3598.

ttraenkler:
**Park diagnosed and resolved — duplicate issue id, already fixed on the branch before the park was even read.**

Read the cited run first ([job 89620561377](https://github.com/loopdive/js2/actions/runs/30136239094/job/89620561377)): `quality` → *Issue integrity + link gate (#1616)* → `--check FAILED: 1 duplicate IDs`. No test262/quality regression involved.

**Cause:** this PR's second issue file was `plan/issues/3597-issue-id-gate-should-check-open-prs.md`. PR #3585 (`plan/issues/3597-auto-park-step-aware.md`, opened 12 min *before* `--allocate` handed me that id) has since merged, so 3597 landed on `main` and my file collided.

**Already fixed** — I renumbered to **#3598** at `87d30ed18`, which was pushed after the merge-group run had started, so the park reflects a stale head. Now re-merged `upstream/main` (which carries #3585's `3597-auto-park-step-aware.md`) and re-verified on the current head:

```
✓ No duplicate issue IDs found (3156 issues, mode=workspace)
✓ --against-main OK: no branch-introduced issue file collides with an id on upstream/main
```

Removing the `hold` so `auto-enqueue` can re-sweep. **Not** re-enqueueing manually.

Side note, since this PR files the issue about it: this is **collision #4** in ~2 hours, and it is the second one that reached the merge queue rather than failing at PR level. That is precisely the asymmetry #3598 proposes to close — `check:issue-ids` compares against `main` only, so a collision between two *open* PRs is invisible until one merges.

Also: the new step-aware park comment from #3597 named the failing step directly (`Issue integrity + link gate`), which made this diagnosis immediate rather than a job-log hunt. Working well.

PR #3590, opened 2026-07-25
fix(#3607): stage test262-standalone-current.json in promote-baseline — give the standalone rate a correct in-repo source

ttraenkler:
## The bug is semantic, not staleness

The statusline reported standalone conformance as **56.1 %** against a real
current figure of **~63 %**.

The only standalone summary committed to `main` was
`benchmarks/results/test262-standalone-highwater.json` — a **high-water mark**
maintained by `scripts/check-standalone-highwater.mjs` for the #2097 floor gate.
It is a *best-ever* floor that only ratchets **up**, deliberately never down. It
is not a current pass rate, so reading it as one reads a different number by
construction.

The landing page never hit this: `run-pages-build.mjs` fetches
`test262-standalone-current.jsonl` **fresh from the baselines repo**. Any
in-repo consumer had no correct source at all.

## Root cause — the data was already there

`promote-baseline` **already writes** the right file, in "Promote merged
artifacts to stable baseline":

```yaml
cp shard-artifacts/test262-standalone-report-merged.json \
   benchmarks/results/test262-standalone-current.json
```

…but `stage_files()` — which decides what lands in the atomic main-repo commit —
staged `test262-current.json`, `test262-report.json`, the two `public/` reports
and `test262-standalone-highwater.json`, and **not**
`test262-standalone-current.json`. So it was produced on every push to `main`
and then discarded with the runner. Only the `git add` was missing.

## Fix — one line

```yaml
git add -f benchmarks/results/test262-standalone-current.json 2>/dev/null || true
```

`PROMOTE_FILES` is derived from `git diff --cached --name-only`, so the
snapshot → hard-checkout → `reapply_promote_files` path picks the new file up
automatically. No second change needed.

The file is a copy of `test262-standalone-report-merged.json` — the **same shape
as the host `test262-current.json`** (`official_summary.pass` / `.total`,
`oracle_version`, `baseline_sha`, `summary`, `categories`) — so consumers read
the standalone lane exactly the way they read the host lane.

## Verification

- `benchmarks/results/test262-standalone-current.json` is **absent** from
  `origin/main` today — `git ls-tree origin/main benchmarks/results/` lists only
  `test262-standalone-highwater.json`.
- The host twin carries the expected keys:
  `official_summary {total: 43102, pass: 30364, …}`, `oracle_version: 10`.
- Workflow YAML re-parses clean.

**End-to-end proof is the first `promote-baseline` run on `main` after this
merges** — the file should appear in that `[skip ci]` baseline commit. Worth a
glance then, since CI here cannot exercise the promote path.

## Consumer note (statusline — wired separately, no change here)

Two properties to preserve on that side, recorded in the issue:

- **`git show` costs ~13 s on this repo.** Any lookup of a possibly-absent path
  must stay behind a cheap `[ -f … ]` stat guard — an unguarded `git show` blew
  the statusline's timeout entirely (exit 124, blank bars). The in-repo comment
  claiming `git show` "never blocks" is **false**.
- Falling back to the high-water file when the current file is absent is correct
  for the transition window: the first post-merge `promote-baseline` run is what
  creates the file.


ttraenkler:
**`quality` failure diagnosed and fixed: duplicate issue id.** Renumbered **#3598 → #3607** (mechanical only).

**Cause:** this PR added `plan/issues/3598-standalone-current-summary-never-committed.md`, colliding with `plan/issues/3598-issue-id-gate-should-check-open-prs.md`, which landed on `main` via PR #3589. The merged PR keeps the id.

**What changed:** file renamed, `id:` frontmatter, heading, and a short note recording the renumber. Fresh id from `claim-issue.mjs --allocate`, then **independently verified** free on `main` and across every open PR before use — `--allocate` alone has proven unreliable (it has handed out ids that already-open PRs were holding). No other file referenced this id. **`.github/workflows/test262-sharded.yml` and all behaviour untouched.** Also merged `upstream/main`.

Both gates now pass locally: `--check` (3162 issues, no duplicates) and `--against-main`.

---

## ⚠️ Deliberately NOT enqueueing this yet — please leave it un-enqueued for now

An **intentional-negative** baseline change is landing right now (issue #3592, branch `issue-3585-apply-closure-arity`): it dispatches `__apply_closure` at `max(argc, declaredArity)`, converting silently-vacuous standalone passes into honest failures. **The standalone conformance number will drop, deliberately, on the lead's explicit authorization.**

This PR modifies `promote-baseline` — *the exact cascade we need to observe reconciling* once that floor moves. Landing an instrument change while the measurement is being taken is how you get a result nobody can explain afterwards. So this sits until the floor has settled and the arithmetic is confirmed; then this fix makes the *next* cycle correct.

Convenient accident: because this PR touches `.github/workflows/`, the auto-enqueue App token can never pick it up anyway (the #3584 permission hole working in our favour for once), so making it CLEAN cannot disturb the quiet window. **It will need a deliberate enqueue when the time comes** — it cannot self-serve.

For the record, the pre-landing floor is captured so the post-landing arithmetic can be reconciled against a recorded before rather than a reconstructed one: host **30397/43103** (oracle v11), standalone high-water **25453** `host_free_pass` / **24172** official, baselines-repo head `27c5c79706d8`.

This is the **7th** duplicate-id collision in ~5 hours — tracked, with the proposed gate fix, in #3598.

PR #3593, opened 2026-07-25
Export FyiSourceExecutor/runTest for external test262 integrations (#3598)

ttraenkler:
## Summary
test262.fyi's js2wasm integration calls `js2-test262` once per test file, matching every other engine's `run.js` contract in that shared runner — meaning every single test pays the full cost of loading js2's TypeScript-based compiler from scratch, plus a second nested Node process fork internally (~2.3-2.8s of pure fixed overhead per test, dwarfing actual compile+execute time). `run-test262-fyi.mjs`'s own `main()` (full-suite CI mode) already avoids exactly this by reusing a small pool of `FyiSourceExecutor`s across the whole corpus — that reuse path just wasn't reachable from outside the package.

- `FyiSourceExecutor` and `runTest` are now exported, reachable via a new `./test262-fyi` package.json exports subpath (previously unreachable via `import` at all — only the `js2-test262` bin invocation worked).
- `executeTestFile({ ..., executor })` accepts an optional pre-existing executor. Omitted: identical prior one-shot behavior (fresh executor, shutdown in `finally`). Provided: the caller's executor is reused and left running — caller owns its lifecycle.

**A real bug this surfaced**: `FyiSourceExecutor`'s default `workerPath` resolved to a `scripts/` path that only exists in this monorepo, not in the published package — `new FyiSourceExecutor()` with no explicit `workerPath` threw `Cannot find module '.../scripts/test262-worker.mjs'` for any external caller. Fixed by resolving it lazily next to the module's own location (same fix pattern `js2-test262`'s CLI entry point already used internally for this exact bundled-vs-unbundled problem — `workerPathForCli()` is now redundant and removed).

## Verification
Built the package fresh, packed it (`npm pack`), installed the tarball into a clean directory (a real `npm install` of what would actually get published, not `npm link`), and drove it directly:
- `new FyiSourceExecutor()` now correctly resolves `dist/test262-worker.js` (previously threw immediately).
- 6 sequential `executeTestFile({ ..., executor })` calls against the same reused executor: 6966ms (cold) → 1872 → 1454 → 1106 → 997 → 1260ms steady state, vs ~2.3-2.8s *every single call* through the one-shot CLI.
- The existing one-shot `js2-test262` bin invocation (no executor override) still exits 0 on a real passing test — default behavior unchanged.

## Test plan
- [x] `tests/issue-3598-fyi-source-executor-reuse.test.ts` — 3/3 passing (worker-path resolves to a real file, caller-provided executor is not auto-shutdown, omitted executor preserves original one-shot contract)
- [x] `tsc --noEmit` clean
- [x] `biome lint` clean (`.mjs` files aren't in this repo's biome scope, only `.ts` — my test file passes)
- [x] `prettier --check` clean (verified manually; pre-push hook's own local check hit its 90s timeout and deferred to CI)
- [x] Real `npm pack` + `npm install` verification (see above) — not just unit tests against the unbundled source
- [ ] Consumer-side wiring (test262-fyi/data's `engines/js2wasm/server.mjs`+`client.mjs`) ships separately in that repo, not part of this PR

### DIFF
--- a/.claude/statusline-command.sh
+++ b/.claude/statusline-command.sh
@@ -472,6 +472,21 @@ elif [ -f "$report" ]; then
     pass=$(jq -r '.summary.pass // .pass // 0' "$_local_t262" 2>/dev/null)
     total=$(jq -r '.summary.total // .total // 1' "$_local_t262" 2>/dev/null)
   fi
+  # A locally-run host report that is genuinely NEWER than the committed CI
+  # baseline takes precedence — mirrors the standalone lane below. CI data
+  # (test262-current.json, refreshed by promote-baseline on every push to main)
+  # stays the default; a fresh local `pnpm run test:262` overrides it only while
+  # it is actually newer. Both read .summary.* — the official standard+annexB
+  # scope — so the denominator stays consistent with the CI numbers either way.
+  _local_t262_ci="/workspace/benchmarks/results/test262-current.json"
+  if [ -f "$report" ] && { [ ! -f "$_local_t262_ci" ] || [ "$report" -nt "$_local_t262_ci" ]; }; then
+    _lp=$(jq -r '.summary.pass // empty' "$report" 2>/dev/null)
+    _lt=$(jq -r '.summary.total // empty' "$report" 2>/dev/null)
+    if [ -n "$_lp" ] && [ -n "$_lt" ] && [ "$_lt" -gt 0 ] 2>/dev/null; then
+      pass="$_lp"
+      total="$_lt"
+    fi
+  fi
   pass_pct=$(awk "BEGIN {printf \"%.1f\", $pass * 100 / $total}")
   free_mb=$(free -m | awk '/Mem/{print $7}')
   free_g=$(awk "BEGIN {printf \"%.0f\", $free_mb / 1024}")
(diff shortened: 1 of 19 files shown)


ttraenkler:
**Park diagnosed and resolved — duplicate issue id, not a test262/quality regression.**

Read the cited run first ([job 89625244586](https://github.com/loopdive/js2/actions/runs/30137847668/job/89625244586)): `quality` → *Issue integrity + link gate (#1616)* → `--check FAILED: 1 duplicate IDs`.

**Cause:** this PR added `plan/issues/3598-fyi-source-executor-reuse.md`, colliding with `plan/issues/3598-issue-id-gate-should-check-open-prs.md`, which landed on `main` via PR #3589 at `01:07:37Z`. The merged PR keeps the id, so this branch renumbers.

## Renumbered #3598 → #3599 (mechanical only)

Fresh id via `claim-issue.mjs --allocate`, **then independently verified** free on `main` and unused across every open PR before taking it — `--allocate` alone has proven unreliable tonight (it handed out an id that an already-open PR was holding).

Scoped strictly to the 5 files this PR touches:

- `plan/issues/3598-…md` → `plan/issues/3599-…md` (+ `id: 3599`, heading)
- `tests/issue-3598-….test.ts` → `tests/issue-3599-….test.ts` (+ `#3598` refs and the `js2-issue-3598-` tmpdir prefix)
- `scripts/test262-fyi-cli.mjs`, `scripts/run-test262-fyi.mjs` — comment refs
- `CHANGELOG.md` — entry ref

**Deliberately not touched:** `plan/agent-context/dev-guard-tests.md` mentions `3598` as a *PR* number and belongs to `main`, not this PR. **No test expectation, assertion, or behaviour changed.**

Also merged `upstream/main`. `check-issue-ids --check` and `--against-main` both pass.

**Note on the branch name:** it is still `issue-3598-fyi-source-executor-reuse`. Renaming it would break this PR's association, so I left it — only the contents renumbered.

**Local test note:** `tests/issue-3599-…test.ts` fails locally with `Test262 worker exited before ready`, which is the missing `scripts/compiler-bundle.mjs` / `dist/index.js` build artifacts in my worktree — pre-existing and environmental, not the rename. CI builds those.

This is the **5th** id collision in ~3 hours, and the 3rd to surface only in the merge queue rather than at PR level. That asymmetry is tracked in #3598 (the issue that just landed, ironically the other half of this collision).

Removing the `hold`; **not** re-enqueueing manually.

PR #3597, opened 2026-07-25
docs(issues): file #3599 + #3604 from the porffor#262 benchmark run

ttraenkler:
Files two issues found while benchmarking the polymorphism benchmark from
[CanadaHonk/porffor#262](https://github.com/CanadaHonk/porffor/issues/262#issuecomment-5072076826)
with js2. Nothing was posted upstream. Issue files only — no source changes.

## #3599 — class-typed expando write silently dropped (`priority: high`)

On the default `--target gc` build, assigning a property not declared on a class
to a **statically class-typed** receiver compiles with a warning only and then
**discards the store**. Valid Wasm, no trap, wrong value.

| case | V8/Node | js2 |
| --- | --- | --- |
| class-typed instance, undeclared write | `"X"` | `undefined` ❌ |
| same instance held as `any` | `"X"` | `"X"` ✅ |
| object literal held as `any` | `"X"` | `"X"` ✅ |
| `"extra" in p` (class-typed) | `true` | `false` ❌ |

The dynamic path already works — only the nominal class receiver fails to route
to it. Root cause identified: TS **2339/2551** sit in `DOWNGRADE_DIAG_CODES`
(`src/compiler/import-manifest.ts:336-341`) on the rationale *"dynamic property
access"*, which holds for an `any` receiver but not for a closed-struct class
receiver.

This is what invalidated the benchmark: its whole point is adding properties to
class instances to force shape transitions, so all 7 FLAG variants collapsed to
the same program and the flat js2 timing curve was an artifact, not a result.
Porffor at latest main is bit-exact with V8 here; js2 is not.

## #3604 — `console.time`/`timeEnd` unimplemented, with a position-dependent fallback (`priority: medium`)

Missing from `IR_CONSOLE_METHODS` (`src/ir/from-ast.ts:4547`). The feature gap is
the lesser half; the failure mode is the real bug:

- **module scope** → hard compile error (fine, honest)
- **inside a function** → compiles clean, `console` is emitted as a constant
  `ref.null extern`, and it throws a host-glue `TypeError` at runtime — which
  **no caller-supplied `deps.console` can rescue**, since the null is baked into
  the module.

Acceptance criteria require the failure to be position-independent *even if the
methods stay unimplemented*.

## Validation

- `check:issue-ids:against-main` — ✓ no collision with an id on `origin/main`
- `node scripts/update-issues.mjs --check` — ✓ exit 0
- Both ids reserved via `claim-issue.mjs --allocate` (3599, 3604 — not hand-picked)


ttraenkler:
**Park diagnosed and fixed: duplicate issue id.** Renumbered **#3599 → #3608** (mechanical only).

Read the cited run first ([job 89633046316](https://github.com/loopdive/js2/actions/runs/30140630178/job/89633046316)): `quality` → *Issue integrity + link gate (#1616)* → duplicate id. Not a test262/quality regression.

**Cause:** this PR added `plan/issues/3599-classtyped-expando-property-write-silently-dropped.md`, colliding with `plan/issues/3599-fyi-source-executor-reuse.md` on `main`. The merged issue keeps the id.

**What changed:** that one file renamed, `id:` frontmatter, heading, plus a note recording the renumber. Fresh id from `claim-issue.mjs --allocate`, then **independently verified** free on `main` and across every open PR before use. **Your sibling issue `#3604` is unaffected** — I checked it separately and it does not collide, so it keeps its id. No other file referenced `#3599`; no behaviour touched. Also merged `upstream/main`.

Both gates pass locally: `--check` (3163 issues, no duplicates) and `--against-main`.

---

## ⚠️ Leaving the `hold` ON deliberately — this is not an oversight

An **intentional-negative** baseline change is landing right now (issue #3592): it fixes `__apply_closure` under-application so that calls with fewer args than declared formals actually invoke the callee. Today they silently don't, which means `assert.sameValue(a, b)` can "pass" **without ever running** — RC2 measured ~15% of sampled standalone passes as vacuous. Fixing it converts fake passes into honest failures, so **the standalone number will drop deliberately**, on the lead's explicit authorization.

While that lands there is a window where the promoted baseline floor is briefly stale, and any `merge_group` run starting inside it can misread the intentional delta as *"this PR caused a regression."* On 2026-07-02 that falsely park-held four unrelated, independently-verified PRs in ~30 minutes.

So this PR is **correct and ready, but intentionally parked** until the floor settles. I'll remove the `hold` myself once `promote-baseline` has completed and the refreshed floor is visible — nothing further is needed from you.

If it *does* get park-held again afterwards, the four-point stale-floor signature (byte-identical regression-bucket hash across unrelated PRs being the tell) distinguishes a false positive from a real regression before anyone acts on it.

This is the **6th of 7** duplicate-id collisions in ~5 hours — tracked, with the proposed gate fix, in #3598.

ttraenkler:
**Releasing the `hold` — the window I was holding for never opened.**

I retained this hold pending an intentional-negative baseline landing (#3592 RC2, `__apply_closure` under-application). That landing **did not happen**:

- the `issue-3585-apply-closure-arity` branch has been **deleted from upstream** (it existed at `bdc25987b` earlier tonight);
- **no PR was ever opened** for it;
- the floor is **byte-for-byte unchanged** — official `30397/43103`, `host_free_pass 4298`, oracle v11, and the standalone high-water still `25453` / `24172 official` at sha `3634d5ab7` generated `2026-07-18`. Identical to the snapshot I took before holding.

No delta landed, so there is no stale-floor hazard to protect against, and continuing to hold a fixed, CLEAN PR would just strand it — the exact failure mode that makes forgotten `hold` labels expensive.

This PR's original park (duplicate id `3599`) is fixed: renumbered to **#3608**, `--check` and `--against-main` both pass, `upstream/main` merged, all required checks green. Sibling issue `#3604` untouched and uncolliding.

Removing `hold` so `auto-enqueue` can sweep it. **Not** re-enqueueing manually.

PR #3620, opened 2026-07-25
plan(#3627): spec goal-aware sprint model — schedule a goal, expand to actionable members

ttraenkler:
Architect spec (no implementation) answering the stakeholder's question: _"can a sprint reference a goal, and should a goal be an issue itself or separate?"_

**Answer: separate — and `goal:` already IS the parent link asked for** (3,056 of 3,178 issues carry it). What's missing is the other direction: a goal can't be *scheduled*. This specs that.

Every decision is backed by a measurement over the real corpus on `origin/main` @ `07c8d239`, not an estimate.

## The decision that matters — expansion scope

| goal | members | net-new, **actionable-only** | net-new, **all members** |
|---|---:|---:|---:|
| `spec-completeness` | 371 | **24** | **364** |
| `standalone-mode` | 338 | 17 | 305 |

All-members expansion of one goal is a **15× blow-up** on an already-168-item TaskList (→ 532, ~85 % of it already `done`). Two goals would take it past 800. Spec picks **actionable-only**, governed by: *`sprint:` is the axis a goal speaks to; `status:` is the axis an issue speaks to.*

No cap — measured max net-new is 24, so a cap of 25 would bind on nothing, and truncation would contradict "all its issues will be worked on in the priority given."

## Corrections to the scoping premise, on the record

- **Container drift is real but ~3× on a small sample**: 8/37 = 21.6 % of open container issues are flagged `merged-but-open` vs 27/371 = 7.3 % of leaves.
- **The cited example is half wrong: #2860 IS flagged, #3029 is NOT.**
- **The mechanism is not the lifecycle** — it's `reconcile-tasklist.mjs:228` scanning merged-PR titles for every `#NNNN`. Confirmed by primary source: PR #3501 is titled `fix(#3535): ... (#2860 F3)` — it implements #3535 and merely *mentions* #2860. **This bug survives the design**, so the spec scopes it out explicitly rather than letting "goals stay separate" be misread as fixing it.
- **`umbrella:` is never a goal ref** — all 142 values are numeric issue ids, i.e. the same relation as `parent:`. Migration is a mechanical 3-phase rename, not a semantic merge.

## Other findings that change the design

- **512 dangling `goal:` refs across 63 names.** #2860 itself is `goal: standalone` — no such goal file. Without an `aliases:` field, scheduling `standalone-mode` silently misses 101 issues. Gate is specced as **baseline-then-ratchet** (like `ir-fallback-baseline.json`), since a hard gate can't land against 512 violations.
- **Goal files have no frontmatter at all today** — this introduces one, with deliberately distinct keys (`goal:`/`state:`, not `id:`/`status:`) so a goal file is structurally unmistakable for an issue file.
- **The `--issue` fast-path gap would have made the feature look broken**: the `post-file-edit` hook passes a `plan/goals/*.md` path, which matches no issue file, so editing a goal to schedule it would silently sync nothing.
- **Hard non-goal**: never materialize a task for the goal itself — `targetIssueId()` keys on `#\d+`, so a goal-task could never be completed by the reconciler.
- **Freeze exclusion is already structural** (`listIssues()` globs `plan/issues/` only) — pinned with a comment plus a regression test rather than new mechanism.

Landing the eventual implementation must be a **behavioural no-op**: no goal carries `sprint: current` in the spec.

https://claude.ai/code/session_01TXSXz9G2eZrfNeX3satN5X

### DIFF
--- a/plan/issues/3627-goal-aware-sprint-model.md
+++ b/plan/issues/3627-goal-aware-sprint-model.md
@@ -1,0 +1,497 @@
+---
+id: 3627
+title: "Goal-aware sprint model: schedule a goal into the rolling window and expand it to its actionable members"
+status: ready
+created: 2026-07-25
+updated: 2026-07-25
+priority: high
+horizon: m
+feasibility: medium
+reasoning_effort: high
+task_type: infrastructure
+area: planning
+language_feature: n/a
+goal: maintainability
+sprint: current
+---
+
+# Goal-aware sprint model
+
+**Stakeholder ask:** _"Change the sprint model to allow referencing goals in
+addition to issues. If a goal is added to a sprint, all its issues will be
+worked on in the priority given. Issues in the sprint that are entailed by a
+goal should reference it as a parent. I wonder if a goal should be an issue
+itself or separate?"_
+
+**Answer to the literal question, up front: a goal stays SEPARATE from an
+issue, and `goal:` already _is_ the parent link being asked for — 3,056 of
+3,178 issues (96 %) carry it today. No new parent field is needed.** What is
+missing is the other direction: a goal cannot currently be _scheduled_. This
+issue adds that.
+
+---
+
+## Measured baseline (2026-07-25, `origin/main` @ `07c8d239`)
+
+Every number below was measured over the real corpus, not estimated.
+
+| Measurement                                         |                                 Value |
+| --------------------------------------------------- | -------------------------------------: |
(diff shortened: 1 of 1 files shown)


ttraenkler:
Superseded in part by #3622. This PR landed rounds 1-2 of the #3627 spec. The container-drift framing in the description was refined again by a third stakeholder round (definition of done with kinds, plus carry-over as the default) — see #3622 for the retraction and the corrected model.

PR #3687, opened 2026-07-26
fix(eslint): advance real Linter graph compilation

ttraenkler:
## Summary

Merges current `main` into the ESLint compilation branch, clears **both merge-queue park-holds**, and fixes three merge-integration defects. It also **records a regression this PR does not fix**: after the merge the full ESLint graph no longer compiles. See "Known regression" below — please read it before acting on the label.

## The park is cleared (this was the blocker)

`auto-park-bot` held this PR twice. The failing gate was the **#3189 uncatchable-trap ratchet**: `null_deref` grew 153 → 157, naming five test262 files.

**It was never a pass regression.** Measured against the fetched baseline, all five were *already failing*:

| Test | Baseline status | Baseline error |
| --- | --- | --- |
| `Array/prototype/methods-called-as-functions` | `fail` | `concat Expected a TypeError … no exception` |
| `Error/prototype/toString/called-as-function` | `fail` | `Expected a TypeError … no exception thrown` |
| `RegExp/prototype/toString/called-as-function` | `fail` | `Expected a TypeError … no exception thrown` |
| `expressions/call/tco-non-eval-function` | `fail` | `Expected SameValue(«0», «1») to be true` |
| `expressions/tagged-template/tco-call` | `fail` | `Maximum call stack size exceeded` |

They were *reclassified* from a catchable failure into an uncatchable trap, and the ratchet counts trap categories regardless of prior pass state. The rest of that run was strongly positive: **net +1 pass, host stable-path fine-gate net +41** (42 improvements − 1 non-timeout regression), aggregate compile time **−8.1 %**. The other 40 "regressions" were `compile_timeout`, 39 classified `ct_flake`.

**Root cause — one defect, two emit sites.** `emitNullCheckThrow` downgrades itself while a guarded-cast backup is live: a null then means "the cast failed", so it throws only if the pre-cast value was also null and otherwise falls through (#789). That is sound only when a later arm handles the wrong-struct-type case.

- `call-identifier.ts` — the compensating arm is the #1712 host-call fallback, gated on `calleeMayBeHostCallable`. Cast runs, fallback gated off, nothing catches it, `struct.get` dereferences null. `var g = f; … g(n - 1)` in a **function scope** is exactly that shape.
- `string-ops.ts` — no null check at all; the guarded cast could null and a bare `ref.as_non_null` trapped. That is ``getF()`${n-1}` `` in `tagged-template/tco-call.js`.

Both replaced paths trapped unconditionally, so no previously-passing program can observe the difference. Re-run through `runTest262File`: all five are now a plain `FAIL` with **no** `dereferencing a null pointer`.

## Merge of current main

374 commits behind, DIRTY. 14 hunks / 5 files. The load-bearing call: main had extracted `registerModuleGlobal` into `module-global-registration.ts` **keyed by bare name**, which collapses same-named module globals across packages — precisely the defect this branch fixes (`ms/index.js`'s `var s` vs esquery's lexical `s`). The branch's declaration-keyed identity, cross-file suffixed globals, and same-file redeclaration aliasing were ported *into* the extracted module, preserving main's `programAbiGlobals` observation. Full rationale is in the merge commit message.

## Known regression — the ESLint graph does not compile after this merge

Measured, local-vs-local:

| Stage | pre-merge `489f96dd` | post-merge (this head) |
| --- | --- | --- |
| compile | **success**, 615.9 s | **FAILS** |
| validate | true | not reached |
| instantiate | true | not reached |
| `verify()` | fails — `deprecate is not a function` (#3657 host seam) | not reached |

Main's #3520/#3521 structural program-ABI registry requires one allocator per declaration named exactly `__mod_<name>`; this branch requires N globals per name across files with suffixed spellings. Those invariants contradict, and the ESLint graph is the only workload large enough to exercise it.

Four hard errors surfaced in sequence. **Three are fixed here**: an inherited alias of a builtin parent (`LazyLoadingRuleMap extends Map`) handing a host **import** handle to the registry (measured handle 615 vs `numImportFuncs` 650), and two TDZ/value display-name desyncs. The fourth is **open** and moved to a different subsystem (an IR-binding draft collision plus an IR-fallback hard error), so the site-by-site approach was stopped deliberately rather than continued blind.

That conflict is now **#3798**, which owns the architectural decision. #1400 stays open, blocked on #3798 then #3657.

**`tests/stress/eslint-tier1.test.ts` is in no required check** — verified against the workflow definitions (`equivalence-gate` covers only `tests/equivalence/`; the #3008 changed-test gate is root-only `^tests/[^/]+\.test\.ts$`; nothing references `tests/stress`). So this PR can go **fully green with the graph broken**. The `hold` label is therefore left ON deliberately: this is a green-but-known-regressed PR, and removing the label is a judgment call for a maintainer, not an automatic consequence of green CI.

## Verification

- All 5 parked test262 files re-run via `runTest262File` — no `null_deref`.
- `tests/equivalence/`: **12 files / 32 tests** fail on this branch; the identical **12 / 32** fail on clean `origin/main`, and re-running main's 12 on this branch reproduces all of them. Not branch-caused.
- `tests/issue-3672`, `issue-3164`, `issue-3782`, `issue-3654`, `issue-3655`: 44 passed / 3 skipped.
- `typecheck`, `biome lint`, `format:check`, `loc-budget`, `func-budget`, `oracle-ratchet`, `ir-fallbacks`, `check:issues`, `done-status-integrity`, `stack-balance` — all green, run individually (the `quality` lane fail-fasts and would otherwise mask later gates).

Refs #1400 #3654 #3655 #3656 #3657 #3672 #3798


ttraenkler:
Closing this PR **unmerged**. It was mined for slices instead — stakeholder decision, after it sat `DIRTY` behind a `github-actions[bot]` park-hold (a held PR is skipped by the `auto-enqueue` backstop, so it could not recover on its own) with ~12.9k lines of divergence behind `main`.

Everything below was **re-measured against `origin/main` @ `e4187572`**, not inherited from this PR's own description.

---

## 1. Two headline fixes here are NOT needed — do not re-add them from this diff

This is the strongest evidence the PR was a source of slices rather than a merge candidate. Both were load-bearing in the description above; neither survives contact with current `main`.

### (a) Uncatchable trap → catchable TypeError — **self-inflicted**

`src/codegen/expressions/call-identifier.ts` + `src/codegen/string-ops.ts`, dropping `__lastGuardedCastBackup` so a failed guarded cast throws instead of trapping. This is the fix for **the very failure that parked this PR** (the #3189 ratchet, `null_deref` 153 → 157, five named test262 files).

Probed on `main`:

| shape | result on `main` |
|---|---|
| `` getF()`${n-1}` `` (the `tagged-template/tco-call.js` shape) | throws a **catchable** `tag is not a function` — no null-pointer trap |
| `var g = f; … g(n - 1)` in a function scope (the `tco-non-eval-function` shape) | returns the **correct** value |

The trap reclassification was caused by *this branch's own* module-global identity rework. There is no defect on `main` for these hunks to fix. **They die with the branch.**

### (b) `collectNodeBuiltinImports` — **superseded**

`main` already has `collectGraphNodeBuiltinImports` (`src/compiler/node-builtin-import-collector.ts`), wired at `compiler.ts:1608` for the multi-file graph. In *this* branch's head the function is **imported into `compiler.ts` and never called** — it is already dead code here. `tests/issue-3654.test.ts` on `main` proves `node:path` resolves, imports as `env.__node_path`, and executes through `compileProject`.

### Also already landed by other routes

**#3655** static JSON require (PR #3867, different implementation) · **#3672** resolved-graph heap budget · **#3743** `emitWatOnlyFunctions` · `fix(resolver): restore ESLint package graph resolution` (`c9507546`) and `fix(resolver): scope JSDoc imports to comments` (`d3cb7b9e`) — both were already in this PR's merge base, so the three-dot diff over-reports them.

---

## 2. Filed slices — PR #3924

| id | carries |
|---|---|
| **#3798** | The architectural conflict that actually stopped this PR: declaration-keyed module globals vs the #3520/#3521 structural program-ABI registry. Reuses the id reserved for it on `issue-assignments` 2026-07-30. Records that `function-identity.ts`, `module-global-registration.ts` and the `call-identifier.ts` identity rework are **one cluster and NOT independently landable**, and cites this branch's ref as prior art. |
| **#3930** | `src/cjs-rewrite.ts`'s nested/residual `require()` linking. Measured on `main`: a factory-invoked `require` contributes no graph edge — the file is absent, `success: true`, empty errors, runtime throw; Node gives 1000. Carries this PR's laziness heuristics *and* a bounded-eagerness counter-criterion. |
| **#3932** | The two-line `if (!projectResolutions)` guard in `analyzeMultiSource`. Measured on `main`: a JSDoc-type-only module emits both its name **and** its body constant into the WAT. |
| **#3933** | This PR's own finding that `tests/stress/` is in no required check — re-verified on `main`. This PR recorded a broken ESLint graph **and** could still go fully green; that is the argument. |

Also from this mining: **#3559** gained a rebuttal (its proposed out-of-range safety gate is regression-safe but **insufficient** — a 14-line repro emits valid Wasm and returns 3 where Node returns 103), **#3653** was closed as already-satisfied with all six criteria measured, and **#1400** carries the full landed-elsewhere table plus the updated blocker chain.

---

## 3. Not yet filed — id pending

**Compile-phase profiling** — `src/compile-profile.ts` (66 lines, `JS2WASM_PROFILE_COMPILE=1`, `__JS2_COMPILE_PROFILE__` JSONL per phase with elapsed + RSS) plus its `compiler.ts` wiring across `multi.preprocess` / `multi.checker` / `pipeline.codegen` / `pipeline.binary` / `pipeline.wat` / `pipeline.artifacts` / `multi.optimize` / `multi.total`. Purely additive, opt-in, horizon `s`. Nothing equivalent exists on `main` — `main` measures only whole-compile elapsed at the probe level, which is why the 615.9 s figure in this description has no phase breakdown.

**It has no id yet.** `claim-issue.mjs --allocate` is heavily contended (~10 min per call, one call lost a race and retried) and the fourth call was cut off. The id is deliberately **not hand-picked** — a lane that hand-picked "max + 1" collided earlier today. It will be filed when an id can be allocated properly.

Also preserved for #3798 rather than filed separately: `src/codegen/diagnose-global-collisions.ts` (198 lines, opt-in via `J2W_DIAG_GLOBAL_COLLISION=1`) — its positive control does not reproduce on `main` with a synthetic fixture, so it belongs with the work that needs it.

---

The branch ref `codex/1400-eslint-e2e` (`561c933af16651e49f50556b8128967892ce529e`) is retained; #3798, #3930, #3932 and #3559 all cite specific files on it.

ttraenkler:
Follow-up to the close comment, section 3: the compile-phase profiling slice now has an id — **#3946**, filed on PR #3924.

It was left un-filed deliberately rather than hand-picked. #3880's allocator fix landed on `main` in the meantime; the retry reserved #3946 cleanly and self-verified on `issue-assignments` with `pr_scan=ok`. All five slices from this PR are now filed, and nothing from it is left un-owned.

PR #3518, opened 2026-07-23
chore(quality): refresh godfile-profile baseline — re-green check:godfiles on clean main

ttraenkler:
`pnpm run check:godfiles` exits 1 on a clean `origin/main` checkout — 6 stale-profile regressions: `compileCallExpression` +57, `tryEmitInlineDynamicCall` +172, `generateMultiModule` +156, `planIrOverlay` +283, `fillExternArrayLikeStructArms` +69, and NEW `emitStandaloneArrayConstructor` (175 LOC).

**Decision: refresh the profile, not fix the functions — and here is why.**

- The growth landed on main through merged, reviewed PRs. `check:godfiles` is wired into neither CI nor husky (it's a package.json script only), so the ratchet had no say at merge time — the "detection without enforcement" failure mode.
- `planIrOverlay` +283 is the codex IR lane's active overlay planner (#3518/#3520 territory); the calls.ts growth belongs to the god-file decomposition track (#3111/#3282/#3400) which already owns the shrink direction. Retroactively demanding six function-shrink refactors is real refactoring work, not hygiene, and would collide with that in-flight work.
- A gate that is red on a clean checkout trains everyone to ignore it. Resetting the ratchet restores its signal value.

Regenerated with `node scripts/profile-godfiles.mjs --update` (44 tracked functions; `--check` now exits 0). Baseline was last refreshed at #3305.

**Follow-up decision for the lead** (not taken unilaterally here): wire `check:godfiles` into the `quality` job so future growth is caught at PR time instead of accruing silently. Left out of this PR because changing required-check behavior mid-session risks reddening in-flight PRs that touch the tracked files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01XvU8vk2ntmbYbHoewNrMDb

PR #3796, opened 2026-07-29
feat(perf): optimize Acorn standalone dynamic Wasm

ttraenkler:
## Summary

- add a true standalone runtime-dynamic Acorn lane where a seed enters after compilation while the test loop and checksum remain inside zero-import Wasm
- keep compile-time-static Acorn separate and use generic zero-import staged evaluation to emit a 20,874-byte IR residual; there is no result cache
- lower homogeneous numeric switches directly when the discriminant has a proven `f64` local/parameter slot
- make the typed receiver parameter the complete carrier for bare `this`, allowing twin-exclusive trampolines to omit the ambient receiver frame; guarded and generic fallbacks retain it
- omit direct-call argc frames when `arguments`, overapplication, and parameter defaults cannot observe the supplied arity
- put the representation-disjoint native RegExp `.test` brand ahead of user-struct and field dispatch
- preserve the exact dynamic driver, refresh the npm compatibility page artifacts, and document the remaining IR blockers

## Results

Same pinned `acorn@8.16.0`, Node 24.4.1 / V8 13.6, arm64 macOS, two warmups and nine measured rounds:

| lane | Wasm | Node | outcome |
| --- | ---: | ---: | ---: |
| JS host · runtime dynamic | 1,231,720.08 µs/op | 4,201.58 µs/op | Node 293.16× faster |
| standalone · runtime dynamic | 50,113.99 µs/op | 4,424.38 µs/op | Node 11.33× faster |
| standalone · compile-time static | 0.02692 µs/op | 4,576.68 µs/op | Wasm residual 169,991× faster |

The sound name-preserving profile run is 48,970.21 µs/op versus Node at 4,406.14 µs/op, an 11.11× remaining gap. It is 9.6% below the previous 54,169.95 µs named baseline. The stripped runtime-dynamic module is 1,765,609 bytes, has zero imports, parses every runtime input, and returns checksum 422.

The largest exclusive profile buckets are generic `__extern_get` (11.89%), RegExp test/engine work attributed to `__call_m_test_1` (8.33%), `__closure_378__typed_this` (3.23%), `__closure_543` (2.63%), `ToPrimitive` (2.55%), and GC (2.38%). This is internal legacy WasmGC representation and dispatch—not Wasm-to-host callbacks. V8 is the engine host; the standalone module imports nothing.

The runtime-dynamic parser still emits zero IR functions. The exact call-graph-closed audit considers 43 reachable functions: 19 body-shape refusals, 17 unresolved parameter types, 3 unsupported logical values, 2 dynamic RegExp constructors, 1 constructor-identity case, and 1 derivative call-graph-closure refusal. The direct-backend changes in this PR are semantic/performance parity requirements for that migration, coordinated with the active IR task.

The static row is intentionally separate: its zero-import stage evaluates the exact compile-time-known operation, then emits a generic result-preserving IR residual. It does not claim that a new runtime input parses faster than Node.

## Validation

- `pnpm run typecheck`
- 81 focused Vitest cases covering strict switch behavior, numeric-local ABI inference, direct calls, arity/defaults, receiver proofs, guarded fallback, bare `this`, and native RegExp dispatch
- `pnpm run check:loc-budget`
- `pnpm run check:func-budget`
- changed-file Biome lint and Prettier checks
- full `pnpm run benchmark:acorn`: 3,507/3,518 compatibility cases pass, unchanged at 99.69%
- full `pnpm run generate:npm-compat`; benchmark/public JSON copies are byte-identical
- `git diff --check`

Repository-wide Biome lint remains blocked by the pre-existing `noShadowRestrictedNames` diagnostic in `src/compiler/ground-call-fold.ts`; none of the changed files has a lint error.

## Stack

This PR is intentionally based on #3795 (`codex/3778-clsx-cookie-perf`) so the clsx/cookie changes are not duplicated. Retarget to `main` after #3795 lands. It remains draft because runtime-dynamic Acorn is not yet faster than Node.

Progresses #3780.


ttraenkler:
## IR migration handoff

The active IR-migration session audited the exact `fa8bfd5462192e` runtime-dynamic standalone Acorn driver with `trackIrOutcomes` and `JS2WASM_IR_SHAPE_DIAG=1`.

IR considers **43 reachable functions and emits 0**:

- 19 body-shape rejected
- 17 parameter type not resolvable
- 3 logical value unsupported
- 2 RegExp constructor unsupported
- 1 constructor resolution unsupported
- 1 derivative call-graph closure

The 19 body-shape failures split into 7 `var` declarations, 5 module identifiers not in scope, 3 mutated parameters without slot representation, and one each for a tail `ForStatement`, `FunctionExpression`, truthy property-access condition, and `ForInStatement`.

### Ownership

- **#2949 is the critical path:** all 17 parameter failures, the 3 logical values, 3 mutated dynamic parameters, truthy property condition, and much of the dynamic module/constructor carrier work.
- **#2952:** `ForInStatement` and remaining loop/multi-exit shape.
- **#3214:** first-class `FunctionExpression` in `pushComment`.
- **#3520:** source-qualified constructor/callable identity, including `new TokenType(...)`.
- **#2858:** call-graph closure is already done; this occurrence is derivative and should clear with its callee.
- The two dynamic RegExp/module-binding cases have no active residual owner despite #3529 being complete.
- The seven `var` selector gates and five module-data-binding/not-in-scope cases also need a new child slice under #3518 or assignment through #3583, probably after #2949 provides the carrier.

### Integration plan

1. Land #3795, retarget this PR to `main`, then land #3796 as the direct-backend performance baseline. Acorn-wide IR claimability should not block it.
2. Preserve this PR's exact dynamic driver and measurements; keep production IR changes out of this stack.
3. Implement #2949 first, then #2952.
4. After the carrier exists, parallelize the Acorn `var`/module-binding slice, #3214, focused dynamic RegExp residual, and the next #3520 continuation.
5. Re-run the exact driver after every slice. Selection is call-graph closed, so clearing only the wrapper or one hot export cannot move dynamic Acorn onto IR.

Overlap is clean: #3796 owns generic direct-codegen work in `typed-this.ts`, `closures.ts`, `context/types.ts`, `param-return-inference.ts`, and `statements/control-flow.ts`. Current #3799 touches none of those exact files and can merge independently. Near-term IR work should remain under `src/ir/*`; any unavoidable overlap should rebase on #3796 rather than duplicating its changes.

This preserves #3796's direct optimizations as semantic and performance parity requirements for the future IR path.

PR #3871, opened 2026-07-31
fix(#3872): consult per-property [[Writable]] on all four assignment shapes

ttraenkler:
> **Stacks on #3869**, which carries `plan/issues/3872-*.md`. Merge after it lands.

## What

`Object.defineProperty(o,"p",{writable:false}); o.p = 20` neither threw nor left the value alone. `ctx.definedPropertyFlags` mirrors the descriptor attributes with `PROP_FLAG_WRITABLE`; `Object.defineProperty` wrote it and **nothing on any assignment path read it**.

## Four shapes, four sites, one predicate

Acceptance names dot / computed / compound — which turns out to be **four** distinct lowering sites that share no funnel:

| shape | site |
| --- | --- |
| `o.p = v` | `assignment.ts::compilePropertyAssignment` |
| `o.p %= v` | `operator-assignment.ts::compilePropertyCompoundAssignment` |
| `o[k] = v` | `assignment.ts::compileElementAssignment` |
| `o[k] %= v` | `operator-assignment.ts::compileElementCompoundAssignment` |

All four consult one exported predicate, `isNonWritableDataProperty`, so there is a single source of truth rather than four copies that drift. It takes the **receiver expression** rather than a `PropertyAccessExpression` — otherwise the element sites needed an `as unknown as` cast that would have compiled and been unsound.

## Plus the standalone mirror gap

`definedPropertyFlags` was written **only** in the `useStruct` arm of `compileObjectDefineProperty`. Standalone compiles `const o: any = {}` to a native `$Object`, which takes the **externref** arm and recorded nothing — so no compile-time consult could fire there at all. Instrumented: host `{"o@41:p": 14}` vs standalone `[]`.

A runtime consult can't substitute: standalone's `__extern_set_strict` is **deliberately aliased to the non-throwing native `__extern_set`** (#2017) because the native runtime has no TypeError bridge. It can suppress the store but never raise — so the throw must be compile-time.

Recording it in the externref arm closes the gap. **The redefine hazard I flagged earlier does not apply**: the two arms are mutually exclusive per call, so a define can never observe its own record; across calls a prior record *is* a genuine redefine. Recording at the `object-ops.ts:1146` chokepoint instead **would** be unsafe, for the reason the `definePropertyReceiverKeys` comment gives. And it records only when `writable` is stated **explicitly** — omitted attributes leave the bit clear, which is right for a fresh define but wrong for a redefine of an existing writable property, and the externref arm has no `isKnownExistingField` to tell them apart.

## Measured — one harness throughout

| | stock main | dot | +compound | +mirror | +computed |
| host | **7/13** | 9/13 | 12/13 | 12/13 | **12/13** |
| standalone | 10/13 | 10/13 | 10/13 | 13/13 | **13/13** |

The one remaining host failure (`accessor setter still runs`) is **pre-existing on stock main**, confirmed by A/B.

Computed probes **5/5 both lanes**. `tests/issue-3872.test.ts` **22/22**. Regression sweep over 8 adjacent descriptor/property files **68/68**.

Real corpus rows via `runTest262File`:

| row | host | standalone |
| --- | --- | --- |
| `compound-assignment/11.13.2-25-s.js` | FAIL → **PASS** | FAIL → **PASS** |
| `assignment/11.13.1-1-s.js` | FAIL → **PASS** | FAIL → **PASS** |
| `types/reference/8.7.2-3-s.js` | PASS | FAIL → **PASS** |
| `compound-assignment/11.13.2-54-s.js` | PASS | FAIL (unchanged) |

`11.13.2-54-s` has **no `writable:false` in source** — a frozen/sealed variant, a different mechanism, correctly untouched. It is the concrete reason the sizing was quoted as `≤24` rather than `24`.

## Status stays `in-progress` — one acceptance clause is not met

Acceptance asks for sloppy-mode no-op on **all** forms. Sloppy is covered for the two simple forms; **sloppy-mode compound is not**, and is deliberately left rather than approximated: it needs the *computed* value (`GetValue ∘ op ∘ RHS`) as the expression result while suppressing only the store, and the lowering fuses those three. Returning the bare RHS instead — the #2667 mapped-arguments shortcut — is right for a simple assignment and **wrong** for a compound one. The corpus is `onlyStrict`, so the strict arms cover it.

## Near-miss recorded

`issue-2580-m3-bacc`'s `forEach over plain data array-like` (expected 60, got 0) failed during this work and looked like a regression from the mirror change. It is **pre-existing on stock `origin/main`** — confirmed by reverting all three touched files and re-running. It contains no `defineProperty` call at all.


ttraenkler:
**Clarification on a commit-message reference** (posting as a comment rather than amending — this PR is queued, and a push would eject it to the back).

The commit `7b0a96fe` says `definedPropertyFlags` is now *"byte-identical to pre-#3872 main"*. That is ambiguous and would send a reader to the wrong place: **PR #3872** is `docs(#3878): pin the release-pending root cause to the line`, unrelated to this change. **Issue #3872** is this PR's issue. PR numbers and issue ids share one sequence in this repo, so the bare `#3872` is not disambiguating.

Read it as: *`definedPropertyFlags` is byte-identical to its state before the changes in this PR* — i.e. the map is untouched relative to `main`, so the read-path / `getOwnPropertyDescriptor` blast radius that caused the merged-state park is removed **by construction**, not by argument.

Caught by `shepherd-ci-fix` on review.

ttraenkler:
## `hold` removed — provenance

Removing the auto-park added at **05:36:54** after run 4 (merge_group `pr-3871-6f6455c9`, head `2654bc0c`). Recording the order explicitly so this does not read as a label cleared on vibes — the auto-park rules require diagnosis **before** removal, and that is the order followed:

1. **Cited run read.** Run 4 failed with `other = 28`, **27 of 27 core survivors, zero cleared**. That matched the branch I had registered *before* the run: core intact ⇒ the snapshot fix (`2654bc0c`) changed nothing measurable and the cause lay elsewhere.
2. **Root cause identified.** `isNonWritableDataProperty` fell back to `ctx.definedPropertyFlags`, which is **approximate about writability**: `applyDescriptorFlags` leaves the WRITABLE bit clear when a descriptor merely *omits* `writable` — correct for a fresh define, wrong for a redefine, where omitted means "keep existing". Its historical consumers (gOPD reporting, redefine validation) tolerated that. Making it decide whether a **write** is legal did not. `mapped-arguments-nonconfigurable-4.js` is the clearest case: its descriptor never mentions `writable`, yet the consult suppressed the following write.
3. **Fix verified on the branch** (`6124fa3c`). The consult now reads only an **explicit** `writable: false`, recorded from **all three** `defineProperty` lowering arms. **10 corpus paths sampled across all three affected groups** (`defineProperties` 4, `defineProperty` 4, mapped-`arguments` 4/4 — complete), host and standalone, each A/B'd against stock `origin/main`.
4. **All 7 required checks green** on `6124fa3c`, `quality` genuinely run at 7m39s.

**Head is `6124fa3c`.** Any enqueue pin on `2654bc0c` is dead and must not be reused. The merge_group verdict — not PR-level green — governs re-entry here; run 4 tested the *previous* head and says nothing about this one.

**Registered prediction, stated before the run:** all 27 core rows clear. Falsifiable; if any survive I want their paths.

PR #3875, opened 2026-07-31
docs(#3884): file the ~20-regression blind spot in the test262 gate

ttraenkler:
Files #3884. Measured entirely from CI merge_group report artifacts against a fixed baseline — no local-harness measurement contributes to any claim.

## The defect, as a consequence

**A PR introducing up to ~20 genuine semantic regressions and zero genuine improvements PASSES the test262 regression gate.** This is not cosmetic inflation of a reported number: the gate systematically **under-detects**, and the blind spot's size is set by a property of the promoted baseline rather than by anything about the PR under test.

## Mechanism

**1. The baseline hands every candidate ~20 free "improvements."** It carries ~120 rows recorded `compile_timeout`, of which ~20 are load artifacts for tests that reliably pass. Holding the baseline **fixed** and varying the candidate across four independent merge_group runs of #3871:

| run | baseline `compile_timeout` rows | of which **pass** in candidate | genuine `other`→pass |
| --- | ---: | ---: | ---: |
| 1 | 120 | 21 | **0** |
| 2 | 120 | 21 | **0** |
| 3 | 120 | 19 | **0** |
| 4 | 120 | 21 | **0** |

**19 of those paths are identical in all four runs.** Not one run produced a single genuine `other`→pass improvement. The reconstruction reproduces the gate's own `Raw host improvements before canary quarantine` line exactly (23/23/21).

**2. The credit defeats BOTH hard gates.** Per #3457 the step has two independent hard-fail conditions — the net gate (`improvements − regressions < 0`) and the ratio gate, whose breach is **downgraded to a warning when `net ≥ 0`**. For a PR with `R` regressions and zero real improvements, the phantom credit gives `net = 20 − R ≥ 0` for `R ≤ 20`, which passes the net gate **and thereby disarms the ratio gate** — the very check designed to catch one-directional regression. The regression numerator already filters `compile_timeout`; the improvement numerator does not. **That asymmetry is the bug.**

**3. Observed.** #3871 was caught only because it carried **27** — seven past the blind spot. A sibling defect with 15 regressions and no improvements would have merged silently.

## Structural aggravator

`promote-baseline` **overwrites** the baseline on every push to main, so **past gate verdicts cannot be audited**. The control originally planned here (re-running #3867's diff) was impossible because its baseline no longer exists; the fixed-baseline design was substituted and is stronger evidence for this claim than the control would have been.

## Scope

**Host lane only** — that is the gate's own arithmetic (`Host stable-path fine-gate net`), so it is the right lane for the verdict. Whether the standalone baseline shares the artifact is **unmeasured**, and the issue says so rather than implying both.

## Fix direction

Exclude `compile_timeout`→`pass` and `absent`→`pass` from the improvement numerator, symmetric with the existing regression-side `noiseFiltered` filter; re-record baseline timeout rows before promotion; decompose the reported `Improvements` figure by baseline source status.

**Primary acceptance criterion:** a PR carrying N genuine regressions and zero genuine improvements fails the gate for all N ≥ 1.

Docs-only: adds one issue file. Sibling of #3883 (configuration staleness in the same baseline). Id reserved via `scripts/claim-issue.mjs --allocate`.

### DIFF
--- a/plan/issues/3884-phantom-improvement-credits-blind-gate.md
+++ b/plan/issues/3884-phantom-improvement-credits-blind-gate.md
@@ -128,23 +128,30 @@ load-recorded status, that one is configuration.
 - **"The baseline claims `pass` on rows main actually fails."** Raised from a
   local A/B showing `compound-assignment/11.13.2-25-s.js` and
   `assignment/11.13.1-1-s.js` failing on stock main while the baseline says
-  `pass`. **Not a baseline defect** — but the residual is unexplained and is
-  tracked here rather than closed. CI evidence: both rows read `pass` in the CI
-  baseline **and** `pass` in the CI candidate merged reports for runs 1 and 4.
-  Baseline and candidates agree, so the baseline is not claiming a pass the
-  pipeline disagrees with.
+  `pass`. **RULED OUT — root cause confirmed: a stale local `origin/main`.** The
+  A/B ran against a local ref pinned at `a1f72e93` while main was `0694da8f`.
+  Re-run against current main, host **passes** both rows, agreeing with the CI
+  baseline and with the CI candidate merged reports for runs 1 and 4. Baseline,
+  CI candidate and current main all agree on host. No pipeline fault.
 
-  The local harness has since been **validated** — it reproduced a known CI host
-  failure (`6-a-161` on `2654bc0c`) with the exact assertion and values — so
-  "permissive local harness" is *not* the explanation. The live hypothesis is
-  therefore that the local checkout's `origin/main` was **stale**, i.e. the A/B
-  measured an older main that genuinely failed those rows. That trap recurred
-  **four times in this single session**: plain `git fetch origin main` repeatedly
-  left `refs/remotes/origin/main` behind, and each time the stale ref produced a
-  confident, wrong answer. Anyone re-testing must first run
-  `git fetch origin '+refs/heads/main:refs/remotes/origin/main'` and verify
-  `git rev-parse origin/main` against
-  `gh api repos/loopdive/js2/commits/main --jq .sha`.
+  Note the local harness had already been **validated** — it reproduced a known
+  CI host failure (`6-a-161` on `2654bc0c`) with the exact assertion and values
+  — so "permissive local harness" was never the explanation. The input commit
+  was simply wrong. Both rows are also **standalone-only** failures, invisible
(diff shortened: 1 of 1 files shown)


ttraenkler:
## Retraction of an adjacent claim, so nobody chases it

Adding this here rather than as a new id, since it is a residual of the same investigation.

**I claimed the host baseline over-reports `pass` on rows main actually fails. That claim is WITHDRAWN — it was my stale local `origin/main`, not a baseline defect.**

What I reported: baseline says `pass` for `compound-assignment/11.13.2-25-s.js` and `assignment/11.13.1-1-s.js`, while stock main FAILs them on host — presented as a third baseline-integrity defect alongside the `compile_timeout` blind spot this issue tracks.

What was actually true: my `refs/remotes/origin/main` was stale at `a1f72e93`. After `git fetch origin '+refs/heads/main:refs/remotes/origin/main'` it moved to `0694da8f0605c40f4a27f9d8381888a08aeb68c8`, verified equal to `gh api repos/loopdive/js2/commits/main --jq .sha`. Re-running the same A/B against current main:

| path | CI baseline | current main (host) | current main (standalone) |
|---|---|---|---|
| `compound-assignment/11.13.2-25-s.js` | pass | **pass** | FAIL |
| `assignment/11.13.1-1-s.js` | pass | **pass** | FAIL |

Baseline, CI candidate reports, and current main **all agree on host**. No baseline defect on these rows.

**Why it is worth recording rather than deleting:** the corrected reading gives a mundane, defect-free explanation for a result that otherwise looks like a measurement bug. Those rows fail only in the **standalone** lane; the fine gate measures the **host stable path**; so a standalone-only change registers zero host improvements **by design**. If someone later sees "real fix, zero improvements counted" they should check lane before suspecting the baseline.

**Does not affect this issue's finding.** The `compile_timeout` numerator asymmetry — regression side filters `noiseFiltered` / `r.to !== "compile_timeout"`, improvement side does not — is a genuine signed-noise bug in the gate's own arithmetic and stands entirely independent of the above.

**Trap worth noting for the next reader:** a plain `git fetch origin main` can silently leave `refs/remotes/origin/main` behind. It does not fail; it hands you a coherent A/B that answers a different question. Use the explicit `+refs/heads/main:refs/remotes/origin/main` refspec and verify against the API before trusting any main-vs-branch comparison.

PR #3880, opened 2026-07-31
docs(#3879,#3880): umbrella false-STOP, and claim-wedge frequency evidence

ttraenkler:
Both onto existing issues rather than new ids.

## #3879 — a fourth false-STOP class: umbrella citation treated as ownership

**An umbrella issue cites every one of its children by design — that is what an umbrella *is*.** The overlap check doesn't know that, so each citation becomes `ACTIVE overlap — Another agent may already own this work as a slice`. Consequence: **every child of every active umbrella is permanently un-startable.**

Measured on one real gating run. Hunting for a single L/XL `standalone` task, umbrella **#2860** alone produced a BLOCKER on three candidates:

| issue | claim record | blocked by |
|---|---|---|
| #2865 | hard claim (`fable-2865`) | genuinely owned |
| #2872 | hard claim (`dev-std-2`) | genuinely owned |
| **#2916** | **does not exist** | **#2860 citation only** |

#2916 had **no claimant** and was still STOP — and was only dispatched by a human overriding the gate. *A gate that needs routine override to release unowned work is the failure mode.*

The absence was verified **with a control**, per the silent-empty rule: the claim ref holds 20+ neighbouring files (2900, 2903, 2906, 2910–2914, 2920–2932…), so a missing `2916.json` is a real absence rather than an unfetched or broken ref.

**Proposed rule:** skip the citation-overlap check when the referencing issue is an umbrella. An umbrella citation carries no ownership information; only the claim ref and open PRs do.

Also records the **cost asymmetry**: a false CLEAR wastes one agent's cycle, but a false STOP can strand a task indefinitely — each successive agent hits the same STOP and moves on, and nothing in the system notices work that is never started.

## #3880 — frequency evidence

A single anecdote is a flake; this is a rate. One agent, one session: **four wedges plus two slow successes**, each stall 5–15 minutes, with three other agents observed in the same state concurrently (`3661`, `3672`, `3655`) — so it's fleet-wide contention, not one bad checkout.

**Concrete cost:** #3420 was implemented and merged with **no claim record ever taken**. The protection this tool exists to provide was simply absent for that task; the only thing preventing duplicate work was a human noticing.

The shared `refs/claim-issue/base` mirror is implicated — every agent fetches into the *same* ref, so N concurrent claims contend on one lock, which is failure #2 in the table surfacing directly.

Also notes a sharp edge on the sanctioned workaround: *"proceed if the record says `released`"* is safe only while the **reader** path works. A wedge producing a stale read would let two agents both see `released` and both start.

Docs-only: two `plan/issues/*.md` files, no source change.

### DIFF
--- a/plan/issues/3879-pre-dispatch-gate-false-stops.md
+++ b/plan/issues/3879-pre-dispatch-gate-false-stops.md
@@ -55,15 +55,59 @@ subject carries a conventional-commit prefix **other than `docs(`** — e.g.
 escalate. #3688 was dispatched as live work while `8b4d74f1 perf(#3688): …` was
 already an ancestor of main, with 18 test pins.
 
+## 4. An UMBRELLA citing its own children is treated as ownership
+
+**An umbrella issue cites every one of its children by design — that is what an
+umbrella *is*.** The overlap check does not know that, so it reports each citation
+as `ACTIVE overlap — … Another agent may already own this work as a slice.`
+
+Consequence: **every child of every active umbrella is permanently un-startable.**
+Not a policy anyone chose, and it scales with the number of umbrellas.
+
+### Measured on a single real gating run (2026-07-31)
+
+Looking for one L/XL `standalone` task, seven candidates were gated. Umbrella
+**#2860** (`standalone-vs-js-host-test262-gap-umbrella`, `in-progress`) alone
+produced a BLOCKER on **three** of them:
+
+| issue | claim record on `origin/issue-assignments` | blocked by |
+| --- | --- | --- |
+| #2865 | hard claim (`fable-2865`) | — genuinely owned |
+| #2872 | hard claim (`dev-std-2`) | — genuinely owned |
+| **#2916** | **does not exist** | **#2860 citation only** |
+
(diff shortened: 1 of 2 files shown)


ttraenkler:
**Scope reduced after checking against main — dropped a duplicated section.**

My first commit added a `## Frequency evidence` table to #3880. Main's own `## Measured, 2026-07-30/31` section **already carries that data** (four agents, ~50 minutes, the `--allocate` wedges, the silent-success releases, #3420's out-of-sync record) — my table restated the same facts in a different shape. That is duplication, not evidence, so it is removed in `777b09e1`.

**What remains in this PR:**

- **#3879** — the umbrella false-STOP class, which main does **not** cover. Verified by grep against current main before keeping it.
- **#3880** — one point only: the sanctioned workaround (*"proceed if the record says `released`"*) is safe **only while the reader path is sound**. It assumes a wedge can lose a *write* but never return a stale *read*, and nothing guarantees that. Under that failure mode the workaround is itself a duplicate-dispatch mechanism rather than merely an absence of protection.

Net diff on #3880 is now 9 insertions / 29 deletions — the deletions are my own duplicate section coming back out.

PR #3883, opened 2026-07-31
docs(#2916,#3880): mark 2916's claim stale; file the release-wedge evidence

ttraenkler:
Docs-only. Two `plan/issues/*.md` files, no source change.

## #2916 — its claim record is stale, and the issue is available

`origin/issue-assignments:2916.json` reads `assignee: ttraenkler/dev-es5-coercion`, `status: in-progress`. **That is wrong.** The work was handed off deliberately; the release could not be executed.

Three attempts on 2026-07-31:
1. **Node crash** mid-run — the caller nearly recorded success, because the shell's `EXIT=0` was `tail`'s status from a pipe rather than the script's
2. **exit 1** — `cannot lock ref 'refs/claim-issue/base': is at 9619a610… but expected c5faa81c…`
3. **wedged >560 s** after manually force-updating the mirror ref, then killed

**The record was deliberately NOT hand-edited.** Rewriting a shared ref that other lanes read trades a bookkeeping problem for a corruption risk. Instead the *issue file* — where a gating human or agent actually looks — now carries a `Claim status: STALE` banner stating the issue is unclaimed and available, with a pointer to PR #3881 for the groundwork.

Nothing is in flight and nothing is half-implemented: the stopping point was chosen *before* the Slice B/C substrate, because a stranded `OrdinaryHasInstance` tri-state is how a wrong `true` ships.

## #3880 — fifth occurrence, and the argument for priority

Three firsts in this one episode:

- **First failure observed on `--release`.** Prior evidence covered `claim` and `--allocate`; this makes it a **third entry point**, so the fault is the shared-ref layer rather than any one code path.
- **First verbatim capture of the lock error**, naming `refs/claim-issue/base` — the shared mirror every agent fetches into — as the contended resource.
- **First case where a manual mirror force-update did not rescue the retry.** That workaround had worked before; here the retry wedged anyway.

**The priority argument is stronger than the count.** This issue's own acceptance criterion says *"a failed release/allocate can never be mistaken for success by its caller."* Occurrence 1 was exactly that, and was caught only because the caller read the record back instead of trusting the exit code. Had it not been, #2916 would now sit falsely claimed by a departed agent.

So: **the tool built to prevent stranding nearly caused one, then wedged trying to undo it** — and the mitigation was to bypass it entirely. While the claim ref cannot be corrected, it is not authoritative for anything.

## Caller-side hazard, recorded alongside

`cmd | tail -2; echo $?` reports the **pipe's** last stage, so a crashed script reads as success. Use `cmd > file 2>&1; echo $?`, `${PIPESTATUS[0]}`, or run bare — and verify the **effect** (read the record back), not the code.

Two agents hit this same trap within an hour on 2026-07-31, one of whom had it written in their own memory at the time. Vigilance demonstrably did not prevent it, so the rule is stated mechanically rather than as advice.

### DIFF
--- a/plan/issues/2916-standalone-native-instanceof-and-isprototypeof.md
+++ b/plan/issues/2916-standalone-native-instanceof-and-isprototypeof.md
@@ -21,6 +21,30 @@ origin: "2026-07-01 — sr-tail2 escalation: leaky-PASS conversion cluster, subs
 
 # #2916 — Standalone native `instanceof` operator + `isPrototypeOf` residual
 
+> ## ⚠ Claim status: the `issue-assignments` record for #2916 is STALE — this issue is UNCLAIMED and AVAILABLE
+>
+> `origin/issue-assignments:2916.json` reads
+> `assignee: ttraenkler/dev-es5-coercion`, `status: in-progress`. **That is
+> wrong.** The work was handed off deliberately; the release simply could not be
+> executed.
+>
+> Release failed **three times** on 2026-07-31:
+> 1. Node crash mid-run (the caller nearly read `tail`'s exit status as success)
+> 2. exit 1 — `cannot lock ref 'refs/claim-issue/base': is at 9619a610… but expected c5faa81c…`
+> 3. wedged >560 s after force-updating the mirror ref by hand, then killed
+>
+> This is **#3880** (`claim-issue.mjs` ref-lock wedge). The record was **not**
+> hand-edited: rewriting a shared ref that other lanes read trades a bookkeeping
+> problem for a corruption risk. This note is the durable correction — the issue
+> file is where a gating human or agent looks; the claim ref is advisory and was
+> known-unreliable that day.
+>
+> **Take this issue freely.** Groundwork (measured scope, counted baseline,
+> tiered acceptance, confirmed #2580 M3 dependency, both leak sites) landed in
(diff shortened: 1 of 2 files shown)


ttraenkler:
Closing as superseded — content is carried in two places, with one correction.

- The **+46 lines on `plan/issues/3880-*.md`** are superseded by **#3901** (`fix(#3880): make claim-issue.mjs honest in both directions`, +456/-37 on the same file plus the tooling fix itself).
- The **+24 lines on `plan/issues/2916-*.md`** are carried — **corrected** — in **#3900**.

The correction matters: this PR states that #2916's `issue-assignments` record is stale at `status: in-progress`. Read back from the ref, it says:

```json
{ "id": "2916", "status": "released", "released_at": "2026-07-31T08:55:03Z" }
```

The release **did** land — one of the attempts reported as failed had already written the record. Separately, `pre-dispatch-gate.mjs` tested `assignee` alone and ignored `status`, so a *released* record still printed `CLAIMED by …`, which corroborated the wrong story (fixed in #3901).

#3900 keeps the durable half — take #2916 freely, groundwork in #3881, nothing half-implemented, and the rationale for stopping before the Slice B/C substrate — and records the general shape as a trap: *when tooling cannot fix a record, agents write prose explaining the record is wrong, and the prose outlives the problem.*

Nothing here is lost. Closed rather than rebased because it is DIRTY on exactly the file #3901 rewrites.

PR #3889, opened 2026-07-31
fix(#3893): admit whole-param-default generators to the native standalone path

ttraenkler:
## What

`isNativeGeneratorExpressionShape` bailed on `param.initializer`, so **every** `function*({x} = {…})` fell back to the eager host-buffer lowering and emitted `__create_generator` / `__gen_*` / `__get_caught_exception`. This is a **one-line relaxation in the no-JS-host lane** — the machinery it needs already exists.

**#3386 (`done`) admitted binding-PATTERN params to this same predicate and deliberately left defaults out.** The closure trampoline's argc/default emission (`emitParamDefaultCheckInline`, `closures.ts`) already runs in the lifted body — which for a native generator **is the factory**, and that is exactly where the spec wants it: `[[Call]]` runs FunctionDeclarationInstantiation (§10.2.11, where parameter defaults evaluate) and only *then* EvaluateGeneratorBody (§27.5) creates the generator object.

`questionToken` / `dotDotDotToken` still bail (no arity model, no state-struct slot). **The host lane keeps the old bail explicitly**, so nothing host-side moves.

## Attribution — a controlled experiment, not an inference

Same tree, same commit, standalone lane, **one-token** difference between arms:

| arm | source | host-import names emitted |
|---|---|---|
| CONTROL | `function*({x:y}) {…}` | `__gen_resume___closure_` *(native resume fn's own name)* |
| CONTROL2 | `function*(y) {…}` | `__gen_resume___closure_` *(same)* |
| **SUBJECT** | `function*({x:y} = {x:23}) {…}` | **`__gen_create_buffer, __gen_push_f, __create_generator, __gen_next, __get_caught_exception`** |

The subject's import set is **exactly** the dominant standalone leak cluster's (`EXN + GEN_SYNC`, 1,271 records) — which also shows the `__get_caught_exception` leak is **generator-lowering collateral, not a separate exception defect**.

## Verification

**Kill-switched.** With the bail restored, all five assertions in `tests/issue-3893.test.ts` fail with `Import #0 "env": module is not an object or function` — the module cannot instantiate host-free. Restored: 5/5 pass. A test never seen failing is not a test.

**Verified by value, not just by import-freedom.** Standalone, instantiated with **no imports**: default applied when omitted (23) · skipped when supplied (7) · identifier-param default (9) · and the ordering assertion — with a side-effecting default, after `g()` and before any `next()`, `calls === 1` and `bodyRan === 0`; the body runs only on `next()`. (#3032 is the precedent for the body-runs-early violation.)

**Real test262, standalone lane, `runTest262File`.** All four subjects were `compile_error` on the baseline:

| file (`language/expressions/generators/dstr/`) | after |
|---|---|
| `dflt-obj-ptrn-prop-id.js` | pass |
| `dflt-obj-ptrn-id-init-fn-name-fn.js` | pass |
| `dflt-ary-ptrn-elem-id-init-fn-name-arrow.js` | pass |
| `dflt-obj-ptrn-rest-getter.js` | **fail** — advanced `compile_error` → *honest* fail (object-rest getter count), **not** a regression |
| CONTROL `obj-ptrn-prop-id.js` (baseline pass) | pass ✓ |
| CONTROL `ary-ptrn-elem-id-init-fn-name-arrow.js` (baseline pass) | pass ✓ |

**3 of 4 flip; the fourth becomes an honest fail.** Leaking is not flipping.

Adjacent suites green: **8 files / 61 tests** (`generators`, `generator-iife`, `generator-method-destructuring`, `generator-yield-contexts`, `issue-1665-standalone-generator-forof`, `issue-2864-d4-catch-across-yield`, `issue-2571-native-method-generators`, `issue-2581-objlit-method-generators`).

## Sizing — with denominators

Standalone baseline 2026-07-31 (48,088 records, baselines `6cd657e6` / main `ff6dd114`; control: 3,617 carry the leak field). Of the 1,907-record `iterator_protocol` class, **603** are the whole-param-default family, **523 of which (86.7 %) already pass on the HOST lane** — the known-achievable set. The remaining **1,304 have no bail attribution yet** (class/dstr 384, class/elements 354) and are deliberately **not** claimed here.

**Do not read 603 (or 523) as a pass delta** — every one is `compile_error` today, so the prize is host-free *instantiation*, after which each test still has to pass everything else it asserts.

*Denominator note:* two `-dflt` regexes were circulating. `/-dflt/` yields **497** and silently drops the **106** files whose basename *begins* with `dflt-` — including the file this probe is built on. 497 + 106 = 603; this PR quotes **603 / 523**.

## Coordination

`src/codegen/generators-native.ts` is #2864's territory, so the overlap was checked before touching it: all five `issue-2864-*` fork branches are **ancestors of `origin/main` (0 commits ahead)**, **no open PR touches the file**, main last touched it 2026-07-25 (#3620), and the merged #2864 regions are `buildNativeGeneratorPlan` (D4) and `yield*`-delegation (D2/R1) — **not** this predicate or the trampoline argc path.

Net LOC +0 (no budget allowance needed).


ttraenkler:
## ⚠️ Correction to this PR's own numbers — two claims retracted

The **code is unchanged and still correct**; this corrects what the description claims. Posting as a comment rather than pushing, because **this PR is currently in the merge queue** and a push would eject it to the back. The follow-up commit is ready (`64aefcef`) and will land as a small docs PR immediately after this one merges.

### 1. RETRACTED — "3 of 4 test262 files flip to pass"

A/B on the two branches (pre-fix `issue-3628-es3-editions` vs post-fix `issue-3893-gen-param-default`), same six files, gives **byte-identical output: 3 pass and 1 fail on both sides.** They already passed.

`runTest262File(..., "standalone")` **does not enforce host-import-freedom** — it supplies the host imports, so a leaking module still runs and still scores `pass`. That status could not have returned the other answer, so it was never evidence. (The baseline's `compile_error` rows come from the CI worker, which *does* apply the host-free gate — which is why the population is real even though this runner cannot see it.)

**The two instruments that do gate on imports are unaffected** and are what actually support the fix: the bare standalone compile + import-set scan, and `tests/issue-3893.test.ts`, which instantiates with **no import object** — the reason the kill-switch works at all.

### 2. SCOPE narrowed — function expressions only, ~98 rows, not 603/523

`isNativeGeneratorExpressionShape` is consulted **only** for `ts.isFunctionExpression(decl)`. Class and object-literal generators are `MethodDeclaration`s behind a different gate, so this fix cannot move them. Import-set A/B, same two branches:

| shape | pre-fix | post-fix |
|---|---|---|
| `function*({x} = {…})` | leaks | **native** ← this fix |
| `{ *m({x} = {…}) }` objlit | leaks | leaks (unchanged) |
| `class { *m({x} = {…}) }` | native | native (never leaked for this reason) |
| `class { *#m(…) }` private | leaks | leaks (a different defect) |

By path family the 603 splits: **`generators/dstr` 98** (fn-expr — demonstrated) · `class/dstr` + `class/elements` **388** (`MethodDeclaration`; the sync private-generator defect is being attributed separately) · `object/*` **102** (still leaks, unowned) · other **15** (not enumerated).

**603 / 523 remains the honest size of the whole-param-default *symptom*** across three declaration shapes with three different gates — it is **not** the size of this fix. Since other agents are sizing the residue against these numbers, the distinction is load-bearing.

PR #3907, opened 2026-07-31
fix(#3635): close the last retention gap; the "985k artifacts" premise is falsified

ttraenkler:
Closes #3635. **The headline collapsed under measurement** — read this before acting on the number.

## What was claimed vs what is true

| #3635 claim | verdict |
| --- | --- |
| ~985k artifacts | **Real, and larger** — `total_count` is now **1,017,559** |
| storage exhaustion | **Not supported** — 99.1 % are already **expired**; live ≈ **4.9 GB** |
| naive scale ≈ 1.2 TB | the issue rightly warned against quoting this; truth is **~250× smaller** |
| likely cause of the artifact 403s | **Not supported** — uploads work; 403s not recurring |
| Fix 1+2: short retention on heavy uploads | **Already implemented** (26/27 steps) |
| Fix 3: bulk-delete the backlog | **Would reclaim ~0 bytes — do not run** |

## `total_count` is a metadata row count, not storage

An artifact GitHub has **expired** keeps its row, with `size_in_bytes` still populated — which is exactly how a naive sum manufactures a fake terabyte — while occupying **zero** storage.

Binary search for the expired/live boundary (artifacts are newest-first, so expiry is a suffix):

| page | expired / sampled | oldest |
| --- | --- | --- |
| 1 | 0 / 100 | 2026-07-31 |
| 288 | 98 / 100 | 2026-07-26 |
| **289** | **100 / 100** | 2026-07-26 |
| 5088 | 100 / 100 | 2026-06-19 |

**Everything before ~2026-07-26 is already expired — a ~5-day horizon, not 90.**

- **LIVE ≈ 8,800 (0.9 %) ≈ 4.9 GB**
- **EXPIRED ≈ 1,008,000 (99.1 %) — zero storage**

## Storage is not exhausted, and the 403s stopped

At **14:42:23Z** the newest artifact was created at **14:41:52Z — 31 seconds earlier**. A quota-exhausted repo cannot upload. And **14 of the last 15** `test262-sharded` runs are `success` (15th `cancelled`, none failed), **including every `merge_group` run** — the ones that move artifacts heavily. Storage has only *grown* since the 2026-07-24/25 403s, so the suspected causal link is unsupported. Consequence for **#3634**: it should be judged on its own merits, not deprioritised as a symptom of this.

## The proposed fix was already in place

26 of 27 `upload-artifact` steps already declared `retention-days`, and `test262-sharded.yml` — the 114-per-run matrix #3635 blames — has all 7 at **1–3 days**. That is *why* the observed horizon is 5 days.

**The one real gap, now closed:** `vacuity-canary.yml` was the single upload with no `retention-days`, inheriting the 90-day default, on a **scheduled** workflow — the shape that accumulates.

**Honest sizing: it leaks nothing today.** That workflow has run **once** (2026-07-27) and produced **zero** `vacuity-*` artifacts. This is hygiene against a future leak, not a storage win. Claiming otherwise would repeat #3635's original error at smaller scale.

## Verification method — and the broken instrument I caught

Grep line counts are not sufficient: a *comment* mentioning `retention-days` inflates them (27 uploads vs 28 matches reads as "covered"). So I used a step-aware checker.

**Its first version reported 0/27.** `uses:` and `with:` are **sibling keys at the same indent**, so it ended each step's scan before ever reaching `with:`. I caught it only because grep had already proven `test262-sharded.yml` *has* `retention-days` — without that positive control, "0/27" was entirely believable and would have produced a confidently wrong PR. Fixed, then **kill-switched**: reverting the `vacuity-canary` change makes it report **26/27, naming `vacuity-canary.yml:89`**.

## Not done here

- **Bulk delete — recommending against it**, with reasons above. Destructive and outward-facing; not a dev's call, and the measurement says it buys nothing.
- **Lowering the repo default from 90 days** (`maximum_allowed_days: 90`) — a repo *setting* needing org admin. Every upload already overrides it, so it changes nothing today; its value is that the *next* forgotten upload fails safe. Flagged for the maintainer.
- **The actual bill** is still unread: `orgs/loopdive/settings/billing/shared-storage` now returns **410 "endpoint has been moved"** *and* needs `admin:org`. ~4.9 GB is a measurement of this repo's live artifacts, not a reading of the bill, and 41 other private repos share the account quota.

## Note on this PR

It is **fork-head and touches `.github/workflows/`** — precisely the cell measured in #3584 that the app token cannot auto-enqueue. If it sits green and unqueued, that is the known #3584 behaviour, not a new fault, and #3906 is the follow-up that would fix it.


ttraenkler:
**One-shot manual enqueue performed — do NOT repeat.**

Enqueued at **2026-07-31T17:25:40Z**, position 5, `isInMergeQueue=true`, `expectedHeadOid` pinned to `61ded11d9fc2`. Authorized by the tech lead as the documented #3584 case.

Preconditions checked immediately before, at the moment of action:
- **not already in the queue** (`isInMergeQueue: false`)
- `mergeStateStatus: CLEAN`, not a draft, no `hold`/`do-not-merge`/`wip`
- user token (`ttraenkler`), not the Actions `GITHUB_TOKEN` — the latter suppresses the `merge_group` event

**This PR is a live end-to-end confirmation of #3584 on its exact predicted cell** (fork-head *and* workflow-touching). Token-relative divergence was observed directly: the app token reported `BLOCKED` while the admin PAT reported `CLEAN` for the same PR at the same moment. The classifier fired correctly at 17:18:32Z with `SUSPECTED PERMANENT (green-16m-still-blocked)` and applied `needs-manual-enqueue` — surfacing at 16 minutes what would previously have stalled silently and indefinitely.

**The `needs-manual-enqueue` label is now satisfied.** I have deliberately left it in place rather than removing it, since I did not apply it. Anyone sweeping: this label no longer requires action — **check the queue first** (as the protocol requires) and you will find the PR already queued. A second enqueue would rebuild the merge group and cancel the in-flight `merge_group` run.

Known limitation, already recorded in #3584 — do not chase it: a later sweep may log `#3907 skip (UNKNOWN)`. The app's `mergeStateStatus` oscillates `BLOCKED`/`UNKNOWN`, and the classifier fires only on `BLOCKED` by design, because `UNKNOWN` means GitHub is still computing and firing on it would manufacture a false positive.

PR #3914, opened 2026-07-31
perf(#3901): call-free, exactly pre-sized split; direct-splice replace

ttraenkler:
## Description

Closes #3901. Files #3913 for the remaining gap.

| benchmark | base | new | speedup | ratio before→after | criterion | |
|---|---|---|---|---|---|---|
| `string/split` | 3.407 ms | 1.782 ms | **1.91×** | 5.46× → 3.02× | ≥2× | at the bar |
| `string/replace` | 0.232 ms | 0.128 ms | **1.82×** | 3.21× → 1.79× | ≥1.5× | ✅ |
| `mixed/csv-parse` | 1.942 ms | 1.574 ms | **1.23×** | 2.92× → 2.25× | ≥1.8× | ❌ |

**Leading with the miss**: `csv-parse` is the headline benchmark and it did not clear its bar. `split`'s 1.91× ranged 1.83–2.15× across campaigns — inside the noise against a ≥2× bar, so it is called "at the bar", not "met".

Interleaved same-process A/B, median of 25, **6 paired rounds** swapping codegen files between rounds (file copies, not `git stash`). Harness sanity check: measured base `replace` ratio 3.21× vs published 3.28×, base `csv-parse` 2.92× vs published 2.66×.

### The issue's allocation-bound premise was wrong for `split`

**Allocations per `split(",")` call: 10 before, 10 after — unchanged, and that is the finding.** `__str_substring` was **already** an O(1) slice view (`NativeString` is `{len, off, data}`), so the count was already minimal and **split copies zero characters**. There is no slice-view follow-up to file: one exists and is already used.

The real cost was per-call/per-char overhead, the same family #3899 identified:

- `__str_split` called `__str_indexOf` **and** `__str_substring` once per piece — **17 helper calls per 8-piece split** — each re-running the flatten preamble (`ref.test` + guarded `__str_flatten`) and re-loading len/off/data behind a `ref.cast` per `struct.get`
- `__str_indexOf`'s scan is a nested block/loop whose inner `j >= nLen` test costs ~12 ops/char even for a 1-char separator
- `__str_substring` re-clamped both bounds (4 `select`s + a swap `if`) although split's bounds are always in range
- the call site emitted a **second, redundant `__str_flatten`** on both receiver and separator

### Where allocations did move

The capacity-8-then-double growth path is deleted — it fired on csv's 11-line outer split every iteration. **Reference-slot volume per csv iteration: 104 → 41.** `replace` went **11 → 2** allocations, via a `$`-free early-out in `__str_getSubstitution` and a direct 3-way splice, gated on `newLen < 64` so the ConsString rope path for large strings is untouched.

### Coordination note for #3899

This rewrites `__str_split` to scan inline with a single-code-unit fast path, **removing its calls to `__str_indexOf` entirely**. #3899's first-char skip therefore no longer affects `split` or `csv-parse` — it still helps `replace`/`replaceAll`. Flagged so the two PRs do not double-count. No file overlap: this is `native-strings-rewrite.ts` + the new `native-strings-split.ts`; #3899 is `native-strings-search.ts`.

### Why csv-parse could not close — filed as #3913

After the fix it is allocation/GC-bound, not scan-bound. Its inner splits are **66%** of the benchmark (1.646 ms full vs 0.556 ms with the inner split removed), and **50 of the 63 allocations per iteration are the `cols` array — dead except for `.length`**. No split-side work removes an allocation the program asks for.

#747 cannot elide it: it is marked `status: done` but **only Phase 1 landed, and Phase 1 is explicitly inference-only and inert** — it writes classifications and never mutates IR, with no consumer. Separately, `split` is lowered on the legacy AST path so its result is not even an IR alloc site.

### Budget gates: split, not suppressed

`check:loc-budget` and `check:func-budget` both blocked the commit. Resolved by **actually splitting the units**, not by taking a `loc-budget-allow:` allowance: `__str_split` moved to a new `src/codegen/native-strings-split.ts`, and `emitStrReplaceHelpers` split into three builders behind a thin ordered facade. Verified code-motion-only — compiled csv-parse module **byte-identical at 23,116 B** before and after.

### Verification

`tests/issue-3901.test.ts` — **29 cases, all passing**, every expectation cross-checked against V8, including **ConsString rope receivers and separators** which specifically cover the removed eager flatten. Re-run independently by the coordinator. `tests/equivalence/{string-methods,regexp-methods}` 64 passed. `tsc`, `biome lint`, `check:oracle-ratchet`, `check:ir-fallbacks`, `check:loc-budget`, `check:func-budget` all clean.

⚠️ **2 pre-existing failures** in the targeted batch (`#2161 B1` plain-object arg, `#2161 B0` undefined element to string param) — confirmed reproducing on unmodified `main` by swapping the codegen files back. Filed as #3918. Full test262 not run per dev policy.

⚠️ Contended 4-core box — ratios trustworthy, absolutes inflated.

## CLA

---


ttraenkler:
**`hold` removed — park confirmed spurious. Diagnosis below, per the auto-park protocol.**

The bot's own footnote flagged the possibility ("if it is a setup/infra step rather than a verdict step, the verdict never ran"), and checking the cited run confirms it.

**`test262 js-host shard 10/66`** ([job log](https://github.com/loopdive/js2/actions/runs/30651327837/job/91225064631)) died in `actions/setup-node`, **1.6 seconds in**:

```
Attempting to download 25...
##[end-action id=__self.__actions_setup-node;outcome=failure;conclusion=failure;duration_ms=1635]
```

Corroborating evidence that no test ran: the shard's artifact upload reported `No files were found with the provided path: …test262-results-…-mgchunk10.jsonl`. There is no verdict of any kind in that job — not a pass, not a fail.

**`merge shard reports`** failed at *"Fail if required test262 shards did not succeed"*, which is downstream of the above. It is reporting the missing shard, not an independent regression.

So neither cited check produced a conformance verdict. There is no test262 or quality regression here to fix.

**Not an isolated incident.** The identical Node-25 manifest failure hit `cross-backend-parity` on #3917 earlier in this batch; a re-run of that job passed with no code change. `actions/setup-node@v6` is intermittently failing to resolve Node 25 from the manifest, and the direct-download fallback is not saving it. That is worth its own infrastructure issue — it will keep parking unrelated PRs, and each park costs a diagnosis cycle. **Pinning a Node version that is reliably in the manifest, or adding a retry around the setup step, would end it.**

Per the protocol I am removing the label **once**, on a confirmed-infra determination, and not re-enqueueing by hand — the server-side auto-enqueue workflow owns that. If this PR parks again on a *verdict* step rather than a setup step, that is a different finding and I will treat it as a real regression.

For completeness, this PR's own state is unchanged and green on its merits: `tests/issue-3901.test.ts` 29/29, independently re-run, with `tsc`, `biome`, `check:oracle-ratchet`, `check:ir-fallbacks` and both budget gates clean.

---

PR #3916, opened 2026-07-31
fix(#3904): give the dom/* benchmarks a real host-call lane, and stop the harness swallowing failed strategies

ttraenkler:
## Description

Closes #3904.

All four `dom/*` benchmarks published a **JS-only bar**. Verified against `history.json`: across all **170 recorded runs**, zero contain a wasm lane for any `dom/*` benchmark. There was no prior parity to regress from — the lane has been dead for the entire recorded history.

### The verbatim error, identical for all four

```
[host-call skipped (runtime): Cannot read properties of undefined (reading 'createElement')]
```

Note `(runtime)`, not a bare `skipped:` — the modules compiled and instantiated fine and trapped on the first warmup call, which ruled out the compile-failure hypothesis immediately.

**Root cause, traced not guessed**: each module imports `env.global_document`, a `declared_global` intent that `resolveImport` (`src/runtime.ts:15126`) keys on the global's **own name**. `benchmarks/suites/dom.ts` passed `deps: { Document, Element }` — the extern *classes*, which satisfy `Document_createElement`/`Element_appendChild` but say nothing about the `document` global. No `document` key, Node has no ambient `document`, so it fell through to the `() => {}` stub and the module got `undefined` as its Document handle.

The wiring the original author intended — `extraEnv: { __get_document: () => mockDoc }` — **was never read by the harness anywhere**. A dead field, set four times.

### Outcome: fixed, not removed

A wiring bug, not a Node-harness incompatibility, so removing four working benchmarks would have been wrong. Adding `document: mockDoc` to `deps` makes all four run:

| benchmark | js | host-call | vs js |
|---|---|---|---|
| dom/create-elements | 0.125 ms | 1.083 ms | 8.7× |
| dom/modify-text | 0.130 ms | 1.180 ms | 9.1× |
| dom/read-attributes | 0.179 ms | 0.424 ms | 2.4× |
| dom/set-attributes | 0.459 ms | 1.065 ms | 2.3× |

Slower throughout, as expected for pure host interop — the goal was an honest bar, not a win.

⚠️ **These numbers do not predict browser behaviour, and the page should not be read as if they do.** `benchmarks/suites/dom.ts` defines `MockDocument`/`MockElement` as plain JS classes where `createElement` is `new MockElement(tag)` and `setAttribute` is one property store. When the wrapped work is ~one allocation, the measurement is almost entirely boundary crossing. The spread is the tell: the two cheapest operations sit at 8.7×/9.1× while `set-attributes`, whose JS baseline costs 0.459 ms, drops to 2.3× as the fixed cost amortises. In a real browser `createElement` costs microseconds of engine and layout work that would dominate the crossing. #3903 carries the boundary-cost labelling for the published page.

### The second half: failed strategies are now recorded, not dropped

`benchmarks/harness.ts` downgraded any strategy failure to a skip in three places — setup, calibration, mid-loop — each writing to **stderr** and returning `null`. The row never reached `latest.json`, so the chart omitted the bar. A reader could not distinguish "not applicable" from "crashed".

A failed strategy is now a `status: "failed"` row with `error` and `failedPhase` (`setup`/`warmup`/`calibration`/`mid-loop`) plus zeroed timings, alongside an exported `isMeasured()` guard. A `skip`-listed strategy still produces **no row** — so **absent = not applicable, `failed` = broken**.

Consumers updated: `report.ts` (`winner()` no longer lets a zero-median failed lane win; new `## Failed strategies` table; `buildHistory()` excludes failed rows), `scripts/benchmark-lifecycle.mjs` (`validateInternalSuite` ran `finitePositive(medianMs)` on every row and would have rejected the new rows outright — now exempts them but demands a non-empty `error`, and rejects a failed `js` row), and `website/components/perf-benchmark-chart.js` (renders a failed lane as a bar labelled `failed` rather than hiding the whole chart).

**This is what made a 26-benchmark gap measurable.** The linear lane produces results for only 2 of 28 benchmarks; until failed lanes record themselves, nobody could tell how many of the missing 26 were deliberate skips versus silent failures. #3908's inventory answered it: **4 skips, 22 real failures.** The absent-vs-failed ambiguity made that gap structurally invisible, not merely inconvenient.

### Verification

`tests/issue-3904.test.ts` — **11 cases**, re-run independently by the coordinator: all four lanes compiling and running end-to-end, `dom/read-attributes` returning 1000 through the host boundary, and the failed-vs-skipped row taxonomy. `tests/benchmark-lifecycle.test.ts` 20/20. `tsc --noEmit`, prettier, biome lint clean.

`benchmarks/results/latest.json` is **not** regenerated — it is a whole-suite artifact and a `--suite dom` run truncates it to four benchmarks. Two exploratory runs clobbered it and it was restored from git both times (verified clean). `benchmark-refresh.yml` regenerates it on `main`.

### Follow-up filed

#3911 — `benchmarks/{dom,arrays,strings,mixed}.ts` are byte-identical dead copies of `benchmarks/suites/*` that nothing imports, and `benchmarks/dom.ts` still holds the buggy pre-fix version. Deleted in the docs PR.

## CLA

---


ttraenkler:
**Blocked on #3921 — not fixable in this PR.**

`measure-and-gate` fails with:

```
benchmarks/results/latest.json[3].medianMs must be a positive number
Exit code 2 = usage or invalid artifact error
```

**This PR's own change is what trips it, and the change is correct.** `benchmark-refresh.yml` deliberately validates a PR's *candidate* snapshot using the *baseline's* copy of `benchmark-lifecycle.mjs` on `pull_request` (the `lifecycle=` selection at `:295-298`) so a PR cannot weaken its own gate. Sound design — but it means an artifact **format** change can never go green in the same PR that teaches the validator about it.

This PR stops the harness silently dropping failed strategies and records them as placeholder rows (zero timings + `status: "failed"` + `error`). The validator update shipped here is correct and tested — but CI runs `main`'s copy, which predates the format.

Diagnosed by reproduction, and three plausible causes were ruled out first:

- the failed-row exemption **is** present and correct on this branch
- the committed `latest.json` is clean (76 rows, no non-positive medians)
- packaging does **not** strip fields — artifacts are copied whole

Regenerating the strings suite here emits 10 failed rows, every one correctly carrying `status` and a real error message. **This branch's validator passes on that file; `main`'s does not.**

**#3921** lands the validator tolerance on `main` alone. Safe ahead of this PR: no row on `main` today carries `status`, so behaviour there is unchanged. Once it merges, this PR should go green on a re-run **with no changes to this branch**.

Worth noting for reviewers, since it argues for the change rather than against it: this swallowing is why a **26-benchmark gap in the linear backend was structurally invisible** — nobody could distinguish "not applicable" from "crashed". #3908's inventory, once the recording existed, found 4 deliberate skips and **22 real failures**.

---

PR #3918, opened 2026-07-31
fix(#3908): give linear `find`'s accumulator the element's ValType

ttraenkler:
## Description

Closes #3908. Files #3922, #3923, #3924, #3925 from the inventory.

The linear-memory lane of `array/find` emitted a module that failed validation at instantiation:

```
WebAssembly.instantiate(): Compiling function #50:"run" failed:
local.set[0] expected type i32, found local.get of type f64
```

### Traced cause — and the issue's hypothesis was wrong

`compileArrayHOF` (the inline `filter`/`map`/`some`/`find`/`flatMap` expansion) computes an `elemType` for the callback's element local — `f64` for `number[]`/`boolean[]`, `i32` pointer otherwise — then allocated `find`'s result accumulator with a **hard-coded `{ kind: "i32" }`**, ignoring it. The WAT makes it plain: `(local $__hof_result_10 i32)` sitting between `(local $x f64)` and `(local $found f64)`.

It mismatched at **both** ends, not one:
- **store side** — the found-branch emits `local.get <elem:f64>` / `local.set <result:i32>`, the reported error
- **read side** — the accumulator is pushed as the call's value and the caller stores it into an f64 local, because `inferExprType` resolves `arr.find(...)` → `number | undefined` → non-nullable `number` → f64

The "not found" sentinel was the same defect on the init path (`i32.const 0` into an f64 slot).

**The predicate-returns-boolean hypothesis in the issue is dead** — the predicate result is consumed immediately by `emitTruthyCoercion` and never hits a slot. The reused slot is the **element**, not the boolean.

Fix: the accumulator takes `elemType`; the sentinel is `f64.const 0` / `i32.const 0` to match. The sentinel **value** is unchanged, so the existing `found !== undefined` → `f64.ne 0` lowering keeps working. `array/find`'s linear lane now measures 5.6 ms and returns the correct `500000`.

### The 26-lane inventory — the more valuable half

Each linear lane compiled, instantiated, and driven for `warmup + iterations` calls. A 1-shot call **under-reports**, since several only trap once the arena fills — that methodology choice is what made the traps diagnosable.

| bucket | count |
|---|---|
| Working | **3** (was 2) |
| Deliberate skip | **4** |
| Compile error — unimplemented builtin | **16** |
| Runtime trap — `memory access out of bounds` | **5** |

**Of the 26 previously-absent lanes: 4 are deliberate skips, 22 are real failures.** This was unmeasurable before #3904 made failed lanes record themselves.

- **Skips (4)** — all `dom/*`. Legitimate: no host/DOM boundary in the linear lane.
- **Compile errors (16)** — every one is `Unsupported method call` / `Unsupported Array method`, i.e. a **missing builtin**, not a miscompile. Filed as #3922 (7 String) and #3923 (7 Array).
- **Runtime traps (5)** — a controlled experiment shows **4 of the 5 are one root cause**. Recompiled with `allocator: "arena-reset"` and rewinding between calls, all four pass with correct values: `string/split` (trap@4/55 → `80000`), `array/map-filter` (28/55 → `3334`), `mixed/csv-parse` (6/25 → `30000`), `mixed/sieve` (7/25 → `9592`). The bump arena is never reclaimed and the harness invokes `run()` repeatedly. Filed as #3924.
- `string/concat-short` **still traps at 0/55 even with reset** — distinct defect, ~1.5 GB of quadratic intermediates *within* one call. Filed as #3925.

None is another instance of the f64→i32 slot mismatch: `filter`/`map`/`flatMap` accumulate an array pointer and `some` a boolean, both already matching. That shape looks unique to `find`.

### Verification

`tests/issue-3908.test.ts` — **6 cases**, re-run independently by the coordinator: the repro, the not-found sentinel, a **fractional** element (guards against "fixing" this by truncating into i32 rather than widening the slot), a `boolean[]` element, a `string[]` no-regression guard, and a structural WAT assertion that `__hof_result_*` is declared with the element's type. **5 of 6 fail against the pre-fix compiler**, verified by swapping in `git show HEAD:src/codegen-linear/index.ts`; the `string[]` case passes both ways by design.

Gates clean: `tsc --noEmit`, prettier, biome lint, `check:oracle-ratchet` (+0 checker usage — reused the existing `elemType`, added no `checker.*` calls), `check:loc-budget` (−1, by collapsing the three-way `if` into one `resultType` expression), `check:func-budget`. All linear test files pass (26 files, ~220 tests).

⚠️ One **pre-existing** failure unrelated to this change: `tests/issue-3497-linear-jsdoc-landing-signatures.test.ts` has one failing case on the base branch, failing identically with and without the fix.

Includes #3904's commit (`4da552ac`) as a merge, since the inventory depends on its failed-lane recording.

## CLA

---


ttraenkler:
**Blocked on #3921 — not fixable in this PR.**

`measure-and-gate` fails with:

```
benchmarks/results/latest.json[3].medianMs must be a positive number
Exit code 2 = usage or invalid artifact error
```

This is a **bootstrap problem, not a defect in this branch.** `benchmark-refresh.yml` deliberately validates a PR's *candidate* snapshot using the *baseline's* copy of `benchmark-lifecycle.mjs` on `pull_request` (the `lifecycle=` selection at `:295-298`), so a PR cannot weaken its own gate. The consequence is that an artifact **format** change cannot go green in the same PR that teaches the validator about it.

This branch merges #3904's commit — needed because the 26-lane inventory depends on its failed-lane recording — so it inherits the new row shape: a failed strategy is recorded as a placeholder row with zero timings plus `status: "failed"` and an `error`, instead of being dropped. `main`'s validator predates that and rejects it.

Diagnosed by reproduction rather than inference:

- Regenerating the strings suite emits 10 such rows, every one correctly carrying `status: "failed"` and a real error string (e.g. `string/concat-short` / `linear-memory` / `memory access out of bounds`).
- **The candidate's own validator passes on that exact file; `main`'s does not.**
- Ruled out: the exemption *is* present and correct on the candidate, the committed `latest.json` is clean, and packaging does not strip fields.

**#3921** lands the validator tolerance on `main` by itself. It is safe ahead of the harness change — no row on `main` today carries `status`, so every row still takes the existing `finitePositive` path. Once #3921 merges, this PR should go green on a re-run **with no changes to this branch**.

No action requested from reviewers; recording the blocker so a red check here is not mistaken for a defect in the linear-backend fix. The `array/find` change itself is unaffected — `tests/issue-3908.test.ts` is 6/6, with 5 of the 6 failing against the pre-fix compiler.

---

PR #3933, opened 2026-07-31
perf(#3921): share one zero-length vec backing store — 8,922 fewer allocations per acorn parse

ttraenkler:
First code change off the census (#3928). Intermediate result — the anomaly it came from is **not** solved.

## What

`[]` lowers to `struct.new $__vec_T (i32.const 0, array.new_default $arr 0)`. That zero-length array is **dead on arrival** whenever the array is ever pushed to: `push` grows on `capacity < length + argc`, which from capacity 0 is unconditionally true, so the first push allocates a real store and drops this one. One immutable singleton per element type replaces it.

Sharing is sound because a zero-length array has no observable contents — no slot can be read (element reads are gated by the vec's `length`, which is 0), no slot can be written (writes go through the grow path, and `capacity 0 < length + 1` always trips it), and `array.len` answers 0 either way. The invariant it rests on — **the grow path must be the only producer of a writable backing store** — is stated in the module header rather than left implicit, and the regression test pins the aliasing case that breaks if it stops holding.

## Measured, and smaller than I predicted

Instrumented standalone acorn, checksum 422 both sides:

| | allocations/parse |
| --- | ---: |
| before | 647,346 |
| after | **638,424** |
| delta | **−8,922 (−1.4%)** |

The vec backing-store counter drops 36,283 → 27,361.

**I predicted ~31,414 removed and was wrong by ~72%.** Only literals with `prealloc == 0` reach this path — #1001's counted-push-loop preallocation already sizes many of them, and `[a, b]` literals never touch the empty path at all — so the eligible population was ~9 K, not ~31 K. Recording that because the estimate came from the census and the census did not support it.

## What this does not fix

123,337 backing stores against 31,414 vec headers is ~3.9 per vec; this takes it to ~3.6. The remaining reallocation is most likely the `max((len + argc) * 2, 4)` growth curve from a cold start — `0 → 4 → 10 → 22` is three allocations to reach 20 elements. That's a bigger fish than this change and is the next thing to read.

## Correctness

38/38 array equivalence tests (push-pop, prototype-methods, zero-arg-methods, of-structs, bounds-elimination). The new test pins non-aliasing of two independently-pushed empties against the `JS2WASM_SHARED_EMPTY_VEC=0` control **and** against plain JS.

Note for review: the test's `delete process.env.X` carries a `biome-ignore` rather than the suggested `= undefined` autofix — assigning to `process.env` coerces to the string `"undefined"`, which `!== "0"` reads as *enabled*, so the control would have run instrumented and the comparison would have been vacuous. Second time this trap appeared in this issue's tests.

---
_Generated by [Claude Code](https://claude.ai/code/session_016zVUkJ2adCdrLoET5AkoUL)_


ttraenkler:
Park diagnosed and fixed; `hold` removed. Recording the analysis here because removing a bot park-hold without one is exactly what the protocol warns against.

## Root cause — index bookkeeping, not the aliasing premise

`ctx.sharedEmptyVecGlobals` cached an **absolute** global index. `addStringConstantGlobal` inserts each new string literal as an **import** global at index `numImportGlobals` — *below* every module-defined global — then calls `fixupModuleGlobalIndices` to shift everything above it. That shifter correctly updates already-emitted `global.get` instructions, and separately updates a **hand-maintained list of cached indices**. This map was not on that list, so the next `[]` of the same element type reused a stale index naming an unrelated global.

That explains the otherwise odd split in the failure:

| stale index lands on | result | count |
| --- | --- | ---: |
| an i32/f64 global | fails validation outright | 400 `wasm_compile` |
| an externref global | coercion layer repairs it with `any.convert_extern` + `ref.cast`, so it **validates** and traps at run time | +3,634 `illegal_cast` |

Verbatim from the WAT of a failing fixture — index 6 is `$fromCodePoint`, not `$__empty_arr_1`:

```wat
i32.const 0
global.get 6                 ;; stale
any.convert_extern
ref.cast null (ref null 1)   ;; traps
struct.new 2
```

**This is the fourth occurrence of the same bug**: `newTargetGlobalIdx` (#2023), `holeGlobalIdx` (#2001), `genEagerFlagGlobalIdx` (#3032) each landed the identical one-line shift after the same class of failure — each with a comment sitting directly above the line this cache should have been added to. The durable fix is deferred resolution at finalize (as `recordInModuleInitFlagRead` does for `__in_module_init`), not a fifth entry in a list that grows by regression. Noted in #3921 rather than done here.

## The aliasing premise held

A stress fixture forcing every empty array in a program onto one shared singleton — `length=` grow, `fill`, `splice`, `copyWithin`, `reverse`, `sort`, spread, `concat` — gave identical results with the optimization off, on, and on+fixed.

## Measured A/B

| sample | OFF | ON (as merged) | ON + fix |
| --- | --- | --- | --- |
| 14 files, `RegExp/property-escapes` + `unicodeSets` | 14 pass | **0 pass** | 14 pass |
| 20 files, `class/dstr` | 13 pass | 12 pass | 13 pass |
| `for-of/dstr/var-ary-ptrn-elem-id-init-exhausted` | pass | **fail** | pass |

## Why CI was green the first time

Two independent reasons, both measured, both now fixed in the test:

1. `target: "standalone"` forces `nativeStrings`, and `addStringConstantGlobal` early-returns without registering an import global — `numImportGlobals` stays 0 and the shifter never fires. **Every test in the original file was standalone**, so the bug was invisible by construction.
2. The fixtures (`var a = [1]; a.pop()`) never reached the modified branch at all — small functions compile through the IR front end, which lowers `[]` itself. Instrumenting that exact body produced zero events and no `__empty_arr` global in the WAT.

The new regression test pins the JS-host lane with a real string-constant import and two empty literals of one element type, and was **verified to fail with the one-loop fix removed** (`struct.new[1] expected type (ref null 1), found global.get of type i32`) — the paired control this PR lacked originally.

---

PR #3944, opened 2026-08-01
fix(#2928): link a refusal runtime-eval provider in the standalone CI lane

ttraenkler:
## Root cause

E6 (#3691) wired the standalone runtime-eval provider prebuild into
`scripts/run-test262-vitest.sh`. **CI never runs that script.** Every
`test262-sharded.yml` shard job — and every `refresh-baseline.yml` heal shard —
invokes `node node_modules/vitest/dist/cli.js run tests/test262-chunkN.test.ts`
directly, so `scripts/build-runtime-eval-provider.mjs` never executes and the
provider cache is always cold in CI.

The published standalone baseline is a **CI** artifact. Measured against
`.test262-cache/test262-standalone-current.jsonl` (baselines run
`20260801-010858`, 48,088 rows): **361 files**, all `status: fail`, die at

```
WebAssembly.instantiate(): Import #0 "js2wasm:runtime-eval":
module is not an object or function
```

led by `annexB/language/eval-code/indirect` (104), `annexB/.../direct` (32),
`built-ins/Function/prototype/apply` (22) and `/call` (22), plus the whole
`built-ins/Function/S15.3.2.1_*` / `length/*` Function-constructor family.

## Why that costs more than the eval assertions

`js2wasm:runtime-eval` is a **module-level** import. A file that merely
*mentions* dynamic `new Function` / indirect eval carries it and cannot
instantiate, so it loses **every** assertion it has — not only the eval ones.
That matters because the standalone `new Function` route already performs
§20.2.1.1.1 argument ToString **AOT at the call site**
(`emitStandaloneDynamicFunctionRuntime` → `compileAndEmitToString`), so a
throwing `toString` throws before the provider is ever consulted. Verified with
a compiled probe: such a module's *only* import is
`js2wasm:runtime-eval::__runtime_new_function`.

## The change

The real Acorn+interpreter provider is **2,447,002 bytes / 151 s to compile** —
not affordable in each of the 36 standalone shards. So this adds a second tier:

- **Refusal provider** — a js2wasm-compiled, **zero-import** core-Wasm module
  with the *same* `[ok, value]` envelope ABI and the same two exports, and **no
  capability**: both entry points return `[false, TypeError]`. **53,152 bytes,
  2.5 s to compile.** No JS-host capability is injected; it is core Wasm from
  the compiler under test, just an empty one.
- `scripts/test262-worker.mjs` links it whenever the interpreter tier is absent,
  so eval-mentioning modules instantiate and only the dynamic-code call itself
  throws the typed, catchable **#2960 Tier-3 TypeError** that direct eval
  already reports.
- CI builds it per standalone shard (`--refusal-only`) in `test262-sharded.yml`
  (both the push and `merge_group` matrices) and `refresh-baseline.yml`.
- **`TEST262_FULL_RUNTIME_EVAL=1` is now the opt-in** for the interpreter tier,
  in both the runner and the worker. Previously a local sweep silently used any
  cached interpreter provider while CI could not, so the two lanes disagreed by
  exactly the interpreter's yield.

The build **canary-verifies before caching**, and the canary is a *cross-module*
positive control: a throwaway user module takes the dynamic route and must
report a catchable `TypeError`, with an explicit assertion that it really does
import the provider so the canary cannot pass vacuously.

## Measurement

Scope `built-ins/Function/`, `--official-scope-only`, same machine, same HEAD,
`COMPILER_POOL_SIZE=2 TEST262_WORKERS=2 TEST262_IT_TIMEOUT_MS=600000`.
**Row count floored at 515 in both arms** (509 + 6 annexB) — E6 finding 1's
silent row loss did not recur.

| Arm                                                    | pass    | fail | CE |
| ------------------------------------------------------ | ------: | ---: | -: |
| A — control, `TEST262_DISABLE_RUNTIME_EVAL_PROVIDER=1`  |     174 |  308 | 33 |
| B — refusal provider linked                             | **179** |  303 | 33 |

**+5 flips, 0 regressions, 0 other status changes.** Attribution is by
*removal*: arm A is the kill-switch arm.

**Three of the five are genuine; two are coincidental — please read the number
this way.**

| file                                                      | why it flips                                            | honest?          |
| --------------------------------------------------------- | -------------------------------------------------------- | ---------------- |
| `Function/S15.3.2.1_A1_T1.js`                             | body `toString` throws `7`; coercion is AOT              | genuine          |
| `Function/S15.3.2.1_A3_T1.js`                             | param `toString` throws first — coercion-order assertion | genuine          |
| `Function/prototype/toString/built-in-function-object.js` | intrinsics walk, `visited.length !== 0`                  | genuine          |
| `Function/prototype/apply/S15.3.4.3_A8_T5.js`             | wants a `TypeError` from `new (Function("…").apply)()`   | **coincidental** |
| `Function/prototype/call/S15.3.4.4_A7_T5.js`              | same shape via `.call`                                   | **coincidental** |

The two coincidental ones are *confirmed* coincidental by arm C below: with the
real interpreter linked, `Function("…")` succeeds and both correctly **fail**.
Honest net for this scope: **+3 real, +2 right-answer-for-the-wrong-reason.**
The refusal error type is deliberately not tuned away from `TypeError` — that is
the already-shipped #2960 contract direct eval throws today, and picking a
non-matching class purely to dodge coincidental matches would be tuning the
instrument to the corpus.

The 103 in-scope link failures go to **0**, replaced by *diagnostic* failures
naming the real blocker: 63 `TypeError: dynamic code evaluation is not
supported…`, 16 harness-wrapped variants, 10 SyntaxError-detection cases. About
**90 of the 103 genuinely need the interpreter**; the refusal tier banks the few
that never did and makes the rest legible.

**No regression vector.** Only already-failing files carry the import, so the
tier cannot demote a pass. One theorised exception is recorded in the issue: the
worker scores a `negative: {phase: runtime}` test as a pass when *instantiate*
throws a matching error, so a runtime-negative test expecting `TypeError` could
have been passing vacuously off the missing import. Zero such regressions
measured in this scope; flagged for the full-lane run.

**No shard-budget risk.** Slowest `exec_ms` in arm B was 1,094 ms; the refusal
provider returns immediately. Per-shard cost is one ~2.5 s prebuild.

### Arm C (interpreter tier) — PARTIAL and load-confounded, recorded as such

A third arm with `TEST262_FULL_RUNTIME_EVAL=1` was **killed at chunk 8/16**
(262 of 515 files). Restricting all three arms to exactly those 262:

| Arm (262-file restriction)    | pass   | fail | CE | compile_timeout |
| ------------------------------ | -----: | ---: | -: | --------------: |
| A — control                    |     83 |  163 | 16 |               0 |
| B — refusal                    |     84 |  162 | 16 |               0 |
| C — full interpreter (partial) | **93** |  123 | 15 |          **31** |

Arm C ran at a 1-min load average of 19–28 against arm B's 7–17, so its 31
`compile_timeout` rows are **confounded by machine contention and are not
attributed to the interpreter** (several are `prototype/toString/
async-generator-*`, which has no dynamic-code path). Contention cannot
manufacture a pass, so the **17 flips to pass over arm B stand as a lower bound**
on the interpreter tier's yield. It is worth finishing — and worth not smuggling
in on a confounded measurement.

## Still owed (recorded in the issue, not claimed here)

1. A clean, uncontended arm C to separate genuine interpreter hangs from
   pool-queue contention.
2. The 151 s / 2.4 MB interpreter compile — #2527 on-demand packaging, or a
   cheaper compile.
3. The E6 finding-2 hang family (~100 annexB `function-in-if` files).

#2928's ≥30-file acceptance box stays **unchecked**. This PR delivers the floor
beneath it.

## Test

`tests/issue-2928-refusal-provider.test.ts` — asserts the unlinked failure is
really present (control), that the refusal module has **zero imports**, that the
previously-lost file now yields its **real** value (`7`, not merely "did not
crash"), and that a call which does reach the provider surfaces a catchable
`TypeError`.


ttraenkler:
One more caveat on arm C, added to the issue file in b3be403a9: it was measured at `608cd95e6`, which **predates** #3933 (shared zero-length vec backing store — 8,922 fewer allocations per acorn parse) and #3940/#3951 (numeric keys all hashing to bucket 0, O(n) → O(1)). Those are exactly the acorn/collection hot paths the interpreter runs on, so arm C's 31 `compile_timeout` rows are **stale as well as load-confounded**. Any re-measurement of the interpreter tier has to sit at a head that includes them.

This does not touch the landed change: arms A and B are same-head, same-machine, and the refusal provider does not execute anything.

PR #3952, opened 2026-08-01
[PROBE][DO NOT MERGE] #3934 — does a skipped required check mask a red same-named run?

ttraenkler:
**[DO NOT MERGE] — measurement probe, will be closed by its author within the hour.**

Measuring the open question surfaced by #3934: **does a `skipped` check run of a required context satisfy branch protection while a *same-named* run is RED?**

This matters because `test262-pr-stub.yml` and `test262-sharded.yml` both publish the same three context names on a src-touching PR (verified on `61d58e077c94`: stub `skipped` in run 30677373432, real `success` in run 30677373444). If `skipped` satisfies regardless of the twin's conclusion, the stub can mask a red authoritative gate. Check-run history cannot answer this — it records what was *published*, not which run the ruleset *evaluated*, and `mergeStateStatus` is not retained historically.

Two commits, read `mergeStateStatus` after each:

1. **CONTROL** — reddens `quality`, a required context with **no same-name twin**. Must report `BLOCKED`. Without this the treatment reading cannot be distinguished from a broken probe.
2. **TREATMENT** — reddens only `merge shard reports`, which **does** have a `skipped` twin from the stub. `BLOCKED` ⇒ no masking. `UNSTABLE`/`CLEAN` ⇒ masking is real and the required-check gate has a hole.

Labelled `hold`. Nothing here is proposed for main.


ttraenkler:
Closed unmerged. The measurement **could not be taken safely from an agent**, and the obstruction is structural rather than bad luck — recording it so the next person does not re-derive it.

**The reading needed** was `mergeStateStatus` on a PR whose only red required check has a `skipped` same-named twin: `BLOCKED` ⇒ no masking, `UNSTABLE`/`CLEAN` ⇒ masking is real.

**Why it is unreadable here:**

1. `BEHIND` **outranks** `BLOCKED`/`UNSTABLE` in `mergeStateStatus`, and the ruleset sets `strict_required_status_checks_policy: true` — so a PR goes `BEHIND` the moment `main` advances, which tonight is several times an hour.
2. This PR carries `hold`, which it must, so a deliberately-red DO-NOT-MERGE PR can never be auto-enqueued. But `auto-refresh-prs` **excludes `hold`** — so the one automation that clears `BEHIND` is exactly the one a probe is opted out of. It stayed `BEHIND` across 30 consecutive polls.
3. Manually re-merging `main` costs ~10–15 min per cycle in this container (`git fetch` over this repo's 6,145 refs alone runs 2–5 min — the same cost centre #3934 removed from the stub). That is **longer than the interval between merges into `main`**, so the readable window closes before the cycle completes. Not a race lost sometimes; one that cannot be won.
4. Dropping `hold` so `auto-refresh-prs` keeps the PR current is only safe **if masking is not real** — because if it is, the PR reads `CLEAN` and `auto-enqueue` (grace 0) merges a deliberately-broken PR within about a minute. That is circular: the safety of the method depends on the answer it is trying to produce.

**The question therefore stays UNMEASURED and should be labelled that way**, not assumed benign. Taking it needs a human-held kill switch: a maintainer watching, the queue paused, or `auto-enqueue` disabled for the duration.

**What this probe did establish** (filed as #3970, PR #3955): on its first control round, a PR touching only `plan/issues/` took the stub's **silent** arm, because `detect` diffs `base.sha..head` and a stale branch drags `main`'s own `src/**` commits into the changed set. The stub's relevance filter and GitHub's native `paths:` filter disagree on stale branches — benign today only because the divergent arm publishes `skipped` rather than SUCCESS.

Also confirmed in passing, and now documented in `docs/ci-policy.md` §1 via #3934: `auto-refresh-prs` skipping `hold`/draft PRs is not a footnote — it is why a held PR rots against a moving `main` indefinitely.

PR #3957, opened 2026-08-01
docs(#3963): file the setup-node@v6 Node-25 manifest flake that parks unrelated PRs

ttraenkler:
## Description

Files `plan/issues/3963-setup-node-25-manifest-flake.md`. Docs-only — one new issue file, no `src/`, `tests/`, `scripts/`, `.github/` or `benchmarks/` changes.

`actions/setup-node@v6` intermittently fails to resolve **Node 25** from the version manifest; the direct-download fallback does not save it. The step dies in ~1.6s and the job never runs a test:

```
Attempting to download 25...
##[end-action id=__self.__actions_setup-node;outcome=failure;conclusion=failure;duration_ms=1635]
```

Observed twice on 2026-07-31, on two different workflows, both on PRs whose code was fine:

| PR | check | outcome |
| --- | --- | --- |
| #3917 | `cross-backend-parity` | re-run passed with **no code change** |
| #3914 | `test262 js-host shard 10/66` | **auto-parked** with a `hold` label |

The issue records why this is worse than an ordinary flake: when it hits a test262 shard in the `merge_group`, `auto-park` (#2547) correctly `hold`-labels the PR, and per the auto-park rules that label must not be removed without diagnosing the cited run. So each occurrence costs a human-grade diagnosis cycle, and the held PR strands until someone does it — the auto-enqueue backstop skips held PRs. The #3914 hold was cleared only after confirming against the run that the verdict never ran.

Fix options recorded in the issue: pin a Node version reliably present in the manifest (likely sufficient on its own), retry the setup step, or both. Acceptance criteria require the fix reach **every** workflow that sets up Node — the two observed failures were in different workflows, so the exposure is repo-wide.

### Not grouped onto the open docs PR

The standing rule is to push docs-only commits onto an already-open docs PR. The open one (#3955) has its head on the `ttraenkler/js2` fork, which this container has no remote or credentials for — `origin` here is `loopdive/js2` only. Grouping was not possible, so this opens separately. Happy to have it folded into #3955 by anyone who can push to that fork.

## CLA

---

### DIFF
--- a/plan/issues/3963-setup-node-25-manifest-flake.md
+++ b/plan/issues/3963-setup-node-25-manifest-flake.md
@@ -1,0 +1,104 @@
+---
+id: 3963
+title: "CI: actions/setup-node@v6 intermittently fails to resolve Node 25 from the manifest — parks unrelated PRs and costs a diagnosis cycle each time"
+status: ready
+created: 2026-07-31
+updated: 2026-07-31
+priority: high
+feasibility: easy
+reasoning_effort: medium
+task_type: infrastructure
+area: ci
+language_feature: n/a
+goal: n/a
+sprint: current
+horizon: s
+es_edition: n/a
+related: [2547, 3597]
+---
+
+# #3963 — `setup-node@v6` Node-25 manifest flake parks unrelated PRs
+
+## Status: open — observed twice in one session, on two different workflows
+
+## Problem
+
+`actions/setup-node@v6` intermittently fails to resolve **Node 25** from the
+version manifest, and the direct-download fallback does not save it. The step
+fails in ~1.6 seconds and the job dies before running anything:
+
+```
+Attempting to download 25...
+Not found in manifest. Falling back to download directly from Node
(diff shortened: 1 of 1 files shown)


ttraenkler:
Closing as superseded by #3958.

This PR filed #3963 with a diagnosis that turned out to be wrong in a load-bearing way. It said `setup-node@v6` *intermittently* fails to resolve Node 25 from the manifest, and recommended pinning a full `25.x.y` as the likely fix. Reading the manifest before applying that fix showed:

```
majors present: 26, 24, 22, 20, 18, 16, 14, 13, 12, 10, 8, 6
entries matching 25.x: 0
```

Node 25 is not in `actions/node-versions` at all, so the lookup failed *deterministically on every job* — all 27 pins were on the direct-download fallback, and the recommended `25.7.0` pin was not in the manifest either. #3958 carries this commit plus the correction and the repo-wide fix, so merging both would land the wrong version of the file.

---

PR #3958, opened 2026-08-01
fix(#3963): Node 25 is absent from the setup-node manifest — move the 7 baseline-free workflows to 24

ttraenkler:
## Description

Every CI job in this repo is running on `setup-node`'s direct-download fallback, and nobody knew.

### The originally-filed diagnosis was wrong in a load-bearing way

#3963 was filed as *"`setup-node` **intermittently** fails to resolve Node 25 from the manifest."* Reading the manifest settles it:

```
$ curl -sS https://raw.githubusercontent.com/actions/node-versions/main/versions-manifest.json
majors present: 26, 24, 22, 20, 18, 16, 14, 13, 12, 10, 8, 6
total entries:  363
entries matching 25.x: 0
```

**Node 25 is not in `actions/node-versions` at all** — not one build, at any patch level. The lookup does not fail intermittently; it fails **deterministically on every job**, and the `Not found in manifest` line in the incident logs is the steady state rather than the anomaly. All 27 pins fetch Node from `nodejs.org` on every run instead of taking a tool-cache hit. The intermittency lives in **that download** — an unconditional third-party network dependency, invisible until the day it flakes.

**The fix the issue proposed would not have worked**: it recommended pinning a full `25.x.y`, but `25.7.0` — the exact version two workflows already pin — is not in the manifest either. Pinning harder inside an absent major does not move a job off the fallback. That is why the manifest got read rather than trusted: the fix required knowing *which* `25.x` to pin, and the intended answer turned out not to exist.

### Scope: 7 workflows, deliberately not all 27 pins

| workflow | from | to |
|---|---|---|
| `cross-backend-parity.yml` | `25` | `"24"` |
| `cla-check.yml` | `25` | `"24"` |
| `diff-test.yml` | `25` | `"24"` |
| `native-messaging-smoke.yml` | `25` | `"24"` |
| `porffor-direct-ab.yml` ×2 | `25` | `"24"` |
| `porffor-source-canary.yml` | `25` | `"24"` |
| `vacuity-canary.yml` | `25` | `"24"` |

`cross-backend-parity` is one of the two workflows that actually parked a PR (#3917), so a real observed incident is fixed. Corroboration that 24 takes the tool cache, from a runner log: `Found in cache @ /opt/hostedtoolcache/node/24.18.0/x64`.

### Why not all of them — conformance results are Node-version-bound

The first version of this PR moved *every* pin. The `merge_group` re-validation auto-parked it, correctly:

```
pass           31086 → 31035    -51
compile_error    652 →  1829  +1177
skip            1278 →   108  -1170
```

`skip` −1170 and `compile_error` +1177 are mirror images — ~1170 previously-skipped tests were suddenly compiled, the quarantine list wall-to-wall `Temporal/…: skip → compile_error`. That is a **test-selection change, not a codegen regression**, and this PR carries no compiler source at all.

**The attribution took a wrong turn worth recording.** My first hypothesis was a fail-open in `classifyTestScope` — `getTest262RelativePath(...) ?? ""` disables all three path-based skip rules when `filePath` is missing. That fail-open is real (and deserves its own issue) but was **not firing**: both call sites `readFileSync(filePath)` immediately, so `filePath` cannot be undefined.

What settled it was **#3964** — an unrelated PR that passed the *same* `check for test262 regressions` gate in `merge_group` in the same window, **on Node 25**, because it merged before this change. Same gate, same window: one clean, one showing a 1170-test selection flip. That points at the Node move, not at baseline drift, and reverses my earlier "likely spurious" read.

Kept on 25, deliberately:

| kept on 25 | why |
|---|---|
| `setup-node-pnpm` composite default | used by exactly **two** files — `ci.yml` and `test262-sharded.yml`, both of which run test262 |
| `test262-sharded.yml` | produces the committed baseline and every PR/`merge_group` verdict |
| `refresh-baseline`, `test262-canary`, `test262-differential`, `test262-cache-prune` | regenerate, compare, or cache test262 results |
| `baseline-floor-staleness-alert`, `baseline-summary-sync`, `deploy-pages`, `issue-tests` | read or publish baseline-derived data |
| `benchmark-refresh`, `landing-four-lane-backend` | committed **benchmark** baselines — same principle in its own domain: the JS lane measures V8, so moving the Node major silently moves the published numbers |

The earlier revision of this PR argued the benchmark-baseline shift was an acceptable deliberate cost. **That is withdrawn.** Both benchmark workflows and all four of their guards (`tests/issue-3498-…`, `tests/benchmark-lifecycle.test.ts`, `docs/ci-policy.md` §6, `docs/benchmarks/landing-four-lane-backend.md`) are byte-identical to the merge base again.

**Correction:** an earlier revision described the composite action as "the shared choke point — covers `test262-sharded` and 11 other workflows". That is wrong; only **two** files reference it. The 12 was a count of matching grep *lines*, not files.

## Still open — not fixed by this PR

The workflow that parked #3914 is **not** fixed. `test262-sharded.yml` keeps the absent-25 pin and its per-run `nodejs.org` download, so that park can recur there. Closing it requires **regenerating `test262-current.jsonl` under Node 24 first**, otherwise every future PR compares Node-24 results against a Node-25 baseline. #3963's acceptance criterion 2 is marked **not met** rather than quietly restated to match what shipped.

Also still open, unchanged from the original writeup: `auto-park` could decline to park when the failing step is a known setup step rather than a verdict step — it already names the failing step (#3597). That would remove the manual diagnosis cycle for this whole class, but narrowing the park deserves its own judgement.

## Verification

- Diff is **8 pin sites across 7 workflows** plus the issue file — no workflow that runs test262 or produces a benchmark baseline is touched, so the `merge_group` failure cannot recur from this diff.
- All 34 workflow files plus the composite action re-parse as valid YAML.
- The eight baseline-critical files confirmed **byte-identical to the merge base** by explicit per-file check.
- `update-issues.mjs --check`, `check-issue-spec-coverage`, `biome lint`, `check:oracle-ratchet`, `check:loc-budget`, `check:func-budget` all exit 0.

## Supersedes #3957

#3957 filed the issue with the incorrect diagnosis and has been closed; this PR carries the file plus the correction.

## CLA


ttraenkler:
**Park diagnosed, cause identified, scope narrowed. Requesting the `hold` be cleared.**

Per the auto-park rules I did not touch the label before diagnosing the cited run ([30704023142](https://github.com/loopdive/js2/actions/runs/30704023142)). The failing step was `Fail on regressions` — a **verdict** step, so the #3597 spurious-park footnote did not apply and this had to be treated as real.

### What the verdict actually said

```
pass           31086 → 31035    -51
compile_error    652 →  1829  +1177
skip            1278 →   108  -1170
```

`skip` −1170 and `compile_error` +1177 are mirror images: ~1170 previously-skipped tests were suddenly compiled, with the quarantine list wall-to-wall `Temporal/…: skip → compile_error`. That is a **test-selection change, not a codegen regression** — and this PR contains no compiler source whatsoever. The headline "27 regressions / net −51" is downstream of ~1170 tests that should never have run.

### Attribution, including a hypothesis I got wrong

My first theory was a fail-open in `classifyTestScope` — `getTest262RelativePath(...) ?? ""` silently disables all three path-based skip rules. That fail-open is real and worth its own issue, but it was **not firing here**: both call sites `readFileSync(filePath)` immediately, so `filePath` cannot be undefined and `relPath` cannot be `""`.

What settled it was **#3964** — an unrelated PR that passed the *same* `check for test262 regressions` gate in `merge_group` in the same window, **on Node 25**, because it merged before this change. Same gate, same window: one clean, one showing a 1170-test selection flip. That points at the Node move and reverses my earlier "likely spurious drift" read.

**Conclusion: test262 conformance results are Node-version-bound.** Moving the shards to Node 24 requires regenerating `test262-current.jsonl` under Node 24 first, or every PR compares Node-24 results against a Node-25 baseline.

### What changed in this push

Narrowed to **seven workflows**, none of which produce or consume a committed baseline: `cla-check`, `cross-backend-parity`, `diff-test`, `native-messaging-smoke`, `porffor-direct-ab` (×2), `porffor-source-canary`, `vacuity-canary`. `cross-backend-parity` is one of the two workflows that actually parked a PR (#3917), so a real observed incident is still fixed.

Kept on 25, deliberately: the `setup-node-pnpm` composite default (used by exactly two files — `ci.yml` and `test262-sharded.yml`, both of which run test262), every test262-producing/consuming workflow, and — applying the same lesson to its own domain — both **benchmark** workflows, since the JS lane measures V8 and moving the Node major silently moves published numbers. My earlier "deliberate baseline shift" is withdrawn; those two and all four of their guards are byte-identical to the merge base again.

**Still open, recorded not hidden:** the workflow that parked #3914 is *not* fixed. It keeps the absent-25 pin and its per-run `nodejs.org` download, so that park can recur. The issue's acceptance criterion 2 is marked **not met** rather than quietly restated.

The `merge_group` failure cannot recur from this diff — it no longer touches any workflow that runs test262.

---

PR #3964, opened 2026-08-01
fix(#3912): fast mode can stringify a number — nine sites conflated "native strings" with "no JS host"

ttraenkler:
## Description

`fast: true` — the whole gc-native lane — could not stringify a number. Six of nine number→string operations **trapped at runtime** on a module that compiled and instantiated cleanly.

### Root cause

`ctx.nativeStrings` and "there is no JS host" were treated as the same condition. They are not, and `fast` is the one reachable config where they come apart: **native strings with a live JS host**. Every gate that asked `wasi || standalone` when it meant *"are strings native here?"*, or asked `nativeStrings` when it meant *"is there no host here?"*, had a wrong answer in exactly that cell.

Each site was internally consistent, which is why inspecting any one of them made it look correct — the bug lived *between* them. A shared `usesNativeNumberFormat(ctx)` predicate now answers the question once so the import gates cannot re-diverge. It spells out all three conditions rather than relying on `wasi`/`standalone` implying `nativeStrings`, because that implication is an `options?.nativeStrings ?? …` default a caller can switch off.

### Why a differential probe rather than code reading

The prescribed gate change **alone fixed 8 of the 9 headline operations and broke three that had been working.** Reading the ~25 `number_toString` consumers would not reliably have caught that. A 52-case `fast`-vs-`host` probe did — every value bound to a **variable** (constant folding hides this whole area) and every case returning a **number** (so a wrong string representation can't be confused with export-boundary marshalling).

```
main:  36 of 52 cases differ from host
now:    3 of 52  (only ONE is a `fast` case)
regressions: zero — everything that passed on main still passes
```

### The nine sites

1–3. `import-collector.ts` — the `number_toString` / `number_toString_radix` / `emitNativeNumberFormat` gates. The headline fix.
4. `string-ops.ts` template numeric spans — unbox unconditionally. `__str_from_extern` marshals a *genuine host string* and silently yields **empty** for a native box; that is why `` `v${3}` `` evaluated to `"v"`.
5. `call-receiver-method.ts` `unwrapToNative` — `(n).toString()` reported a bare externref for what was an `$AnyString`.
6. `call-identifier.ts` `String(<number>)` — same; the two spellings of one operation now agree on their result type.
7. `call-identifier.ts` parseInt/parseFloat **argument** — a native string reached a host import via `coerceType(…, externref)`, which only *widens* the GC ref, so V8 threw `Cannot convert object to primitive value`. Broken on main for a plain literal too: `parseInt("42")` trapped.
8. `import-collector.ts` `__str_to_number` — a **pure-Wasm helper name** was requested as a JS-host import that `src/runtime.ts` does not provide, so `Number("42")` returned `NaN` across the whole gc-native lane.
9. `calls.ts` `tryEmitJsonStringifyPrimitive` — the two arms disagreed on type under `fast`, so `JSON.stringify(<number>)` emitted an **invalid module**.

Plus both JSON string boundaries in `call-namespace-static.ts`. Both directions are emitted at the *producing* call site because the distinction is **provenance, not ValType**: a native box and a host string are both `externref`, and only the site that made the value knows which it is.

### The same conflation, one level down — found by CI

The first version of this PR broke `strictNoHostImports`. `create-context.ts` derives `nativeStrings` from **five** options:

```ts
nativeStrings = nativeStrings ?? (fast || wasi || standalone || strictNoHostImports || utf8Storage)
```

so `usesNativeNumberFormat` is true in six lanes, two of which the probe never exercised. `compile(needs-host.ts, {strictNoHostImports:true})` went from `success: true` to a hard codegen error.

The cause was this PR's own thesis, recurring: `__str_to_extern` / `__str_from_extern` live in the *native* string subsystem but are inherently **host-dependent** — there is no pure-Wasm way to manufacture a JS string, so they copy UTF-16 code units through linear memory via `env.__str_from_mem` / `__str_to_mem` / `__str_extern_len`. It surfaced as an opaque `absoluteFuncIndex: unresolved call target` rather than a clean refusal because `ensureNativeStringExternBridge` **bakes** those funcidxs into compiled helper bodies *before* the strict gate drops the imports.

Fix: a named predicate in `native-strings.ts` — `hostStringBridgeUsable(ctx) = !wasi && !standalone && !strictNoHostImports`. The two boundary marshals consult it and **decline**, so the caller keeps its pre-#3912 lowering in hostless lanes.

### Does `fast` now import `env.__str_*`?

It already did. Measured on pristine main, `` `v${n}` `` and `console.log(s)` under `fast` imported all three. This PR routes two more shapes (`JSON.stringify`, `parseInt`) onto that existing bridge; it does not introduce the category. The alternative main used for `parseInt` was `extern.convert_any`, which hands the host an opaque WasmGC struct — precisely the `Cannot convert object to primitive value` / `NaN` failure being fixed. Now pinned by an explicit test so the tradeoff is visible rather than implied by the "no `env.number_*`" assertion alone.

## Verification

Six lanes, `JSON.stringify(obj)`:

| lane | success | `env.__str_*` | `env.number_*` |
|---|---|---|---|
| host | ✅ | 0 | 0 |
| fast | ✅ | 3 | 0 |
| strictNoHostImports | ✅ | **0** | 0 |
| utf8Storage | ✅ | 3 | 0 |
| standalone | ✅ | **0** | 0 |
| wasi | ✅ | **0** | 0 |

Hostless lanes request no bridge at all; only host-bearing lanes use it.

- `tests/issue-3912-fast-number-stringify.test.ts` — **31/31**.
- `tests/issue-3912-native-string-lanes.test.ts` — **6/6**, new; compiles 7 affected shapes in every `nativeStrings`-implying lane and asserts a hostless lane requests no `__str_*`. Confirmed **non-vacuous**: with the two guards neutralised, both it and the CI test fail.
- `tests/host-import-allowlist-gate.test.ts` — **11/11** (the gate that caught the regression).
- **Equivalence: zero regressions, verified against main rather than asserted.** The branch run showed 32 failures, 8 of them `coercion/arithmetic-add` string concatenation — squarely in the area this touches, so they warranted checking rather than trusting the probe. Re-running those files on `main`: the same 8 fail, and the other 24 match by name. All 32 pre-existing.
- Gates green: `coercion-sites` (net `__str_to_number` **−2**), `check:func-budget`, `check:ir-fallbacks`, `check:oracle-ratchet`, `check:loc-budget`, `check:issue-spec-coverage`.
- #3902's temporary `coercion-sites-allow` for `src/codegen/array-methods.ts` is **removed**, as #3902 instructed, together with the dead `number_toString` host-import detection probe it covered.
- #3917 closed as fixed by #3907, **re-measured, not assumed**, and its permanent repro now cited so the #2093 done-flip gate can see it.

## Remaining, traced, not fixed

- `JSON.stringify({a: 42})` under `fast` returns `"{}"`. **Not new** — byte-identical to main, invisible only because reading the result trapped first. Cause is site nine's twin in `struct-field-exports.ts` (`if (ctx.nativeStrings) return;`), whose own comment asserts "there is no JS host". Flipping it makes the body emit under native strings, where the string-constant globals it reads do not exist (`global index out of range`). Pinned by a test.
- `console.log(s)` under `strictNoHostImports` hits the bridge unguarded and fails with the identical error; `` `v${n}` `` under strict fails on an unrelated `env.__to_bigint` drop. Both verified failing on pristine main.

## Not verified locally

Full test262 (CI owns it — watch `built-ins/Number`, `built-ins/JSON`, `Array/prototype/join`) and `wasmtime` execution of `--target wasi`.

## CLA


ttraenkler:
**Blocker, being fixed on branch — `quality` red on `tests/host-import-allowlist-gate.test.ts:82`.**

Recording the diagnosis so it's on the record rather than only in an agent's context.

`expect(result.success).toBe(true)` fails for `needs-host.ts` compiled with `{ strictNoHostImports: true }`. Reproduced locally and diffed against `main`:

| lane | main | this branch |
| --- | --- | --- |
| `{strictNoHostImports:true}` | success **true**, no `__str_*` imports | success **false** — hard codegen error |
| `{fast:true}` | success true, no `__str_*` imports | success true, but now imports `env.__str_from_mem`, `env.__str_to_mem`, `env.__str_extern_len` |
| `{}` (default host) | unchanged | unchanged |

```
Codegen error: absoluteFuncIndex: unresolved call target (funcIdx=undefined) baked into a
compiled function body … the strict gate dropped these host import(s), one of which was baked
into a compiled helper body rather than degrading cleanly: env.eval, …, env.__str_from_mem,
env.__str_to_mem, env.__str_extern_len
```

**Root cause.** The new `usesNativeNumberFormat` predicate is right in intent but reaches two lanes the PR's probe never exercised. `src/codegen/context/create-context.ts:47-49`:

```ts
const nativeStrings =
  options?.nativeStrings ??
  !!(options?.fast || options?.wasi || options?.standalone || strictNoHostImports || options?.utf8Storage);
```

`strictNoHostImports` **and** `utf8Storage` both imply `nativeStrings`, so `wasi || standalone || nativeStrings` now covers two lanes that previously took the host path. Strings genuinely *are* native there, so the predicate isn't wrong — the problem is what the native path pulls in. `ensureNativeStringHelpers` requests `__str_from_mem` / `__str_to_mem` / `__str_extern_len` as **host `env` imports**. Under `fast` that works because a JS host exists; under `strictNoHostImports` there deliberately is none, so the strict gate drops them and one is already baked into a compiled helper body.

That is the *same* conflation this PR set out to fix — "native strings" vs "there is a JS host" — resurfacing one level down, in the string-memory bridge rather than the number formatter. Worth noting the PR's own gate test asserts fast imports no `env.number_*` formatter, which is true, but fast now imports host string-memory bridges instead; whether that trade is intended needs to be stated and pinned, not left implicit.

The probe that validated this work compared `fast` against `host` only. Extending it to treat `strictNoHostImports` and `utf8Storage` as first-class lanes is the gap that let this through.

Unaffected by this: the equivalence result was verified independently against `main` — all 32 failures on this branch are pre-existing (same 24 names plus the 8 `coercion/arithmetic-add` ones, which fail identically on `main`). This PR stays draft until the allowlist gate is green.

---

ttraenkler:
**Fixed and verified. Also correcting one claim I made above.**

### Correction to my diagnosis

I wrote that `fast` "now imports `env.__str_from_mem` / `__str_to_mem` / `__str_extern_len`" where main did not. That's true only for the program I probed (`JSON.stringify(obj)`). Measured on pristine main, `` `v${n}` `` and `console.log(s)` under `fast` **already** imported all three. So this PR does not introduce the host-`__str_*` category under `fast` — it routes two more shapes (`JSON.stringify`, `parseInt`) onto a bridge that was already in use. My "before/after" framing overstated it by generalising from a single program.

The alternative main used for `parseInt` was `extern.convert_any`, which hands the host an opaque WasmGC struct — that's the `Cannot convert object to primitive value` / `NaN` failure this PR fixes. So the trade is real but favourable, and it's now pinned by an explicit test rather than left implicit behind the "no `env.number_*`" assertion.

### The fix

`hostStringBridgeUsable(ctx) = !wasi && !standalone && !strictNoHostImports` in `native-strings.ts`. The two boundary marshals consult it and **decline**, so the caller keeps its pre-#3912 lowering in hostless lanes.

The root problem was the same conflation this PR set out to fix, one level down: `__str_to_extern` / `__str_from_extern` live in the *native* string subsystem but are inherently **host-dependent** — there is no pure-Wasm way to manufacture a JS string, so they copy UTF-16 units through linear memory. And it surfaced as an opaque `unresolved call target` rather than a clean refusal because `ensureNativeStringExternBridge` **bakes** those funcidxs into compiled helper bodies before the strict gate drops the imports.

### Verified independently, six lanes, `JSON.stringify(obj)`

| lane | success | `env.__str_*` | `env.number_*` |
|---|---|---|---|
| host | ✅ | 0 | 0 |
| fast | ✅ | 3 | 0 |
| strictNoHostImports | ✅ | **0** | 0 |
| utf8Storage | ✅ | 3 | 0 |
| standalone | ✅ | **0** | 0 |
| wasi | ✅ | **0** | 0 |

Exactly the intended shape — hostless lanes request no bridge at all; only host-bearing lanes use it.

- `tests/host-import-allowlist-gate.test.ts` — **11/11 green** (was the failing gate).
- `tests/issue-3912-fast-number-stringify.test.ts` — 31/31.
- `tests/issue-3912-native-string-lanes.test.ts` — 6/6, new; compiles 7 shapes in every `nativeStrings`-implying lane. Confirmed **non-vacuous**: with the two guards neutralised, both it and the original CI test fail.

### Pre-existing, not fixed

`console.log(s)` under `strictNoHostImports` hits the bridge unguarded and fails with the identical error, and `` `v${n}` `` under strict fails on an unrelated `env.__to_bigint` drop. Both verified failing on pristine main, so they're out of scope here rather than newly broken.

---

PR #4008, opened 2026-08-02
fix(react): pass the full scored upstream suite in Wasm

ttraenkler:
## Summary

- preserve symbol-valued struct fields as identity-stable symbols through direct reads, dynamic reads, host boxing/unboxing, and standalone Wasm
- keep polymorphic React helpers and mixed-representation locals on sound dynamic carriers
- retain all arguments for high-arity dynamic host calls
- fix the related frozen-field, ordinary-object constructor, class-heritage TDZ, and dynamic constructor-static semantics exposed by React
- raise the React npm-compat result and regression floor from 39/55 to 55/55

## Root cause

React's `$$typeof` element discriminator used the compiler's generic i32 field carrier without preserving its symbol brand. Dynamic and host-boundary reads therefore surfaced a number instead of the original symbol. React then rejected its own elements. The newly passing upstream clusters exposed several independent representation and host-dispatch gaps behind that first failure; this PR fixes each generically and locks them down with focused tests.

## Validation

- `pnpm run dogfood:react-upstream-suite` — 55/55 scored tests pass; 272 of 273 upstream tests are admitted, and 209 Jest/DOM-infrastructure cases remain explicitly unscored
- 67/67 focused and adjacent symbol/class/runtime tests pass
- typecheck, lint, format check, IR fallback, LOC/function budget, oracle, coercion, issue-id, and issue-spec gates pass
- npm-compat catalog/report tests pass and the canonical/public report copies match

The local pre-push hook's single #3765 WAT-shape assertion still fails identically on clean current `origin/main`; the React branch does not alter that path.

Closes #3961

### DIFF
--- a/benchmarks/results/npm-compat.json
+++ b/benchmarks/results/npm-compat.json
@@ -604,9 +604,9 @@
       },
       "tests": {
         "kind": "upstream-suite",
-        "passed": 39,
+        "passed": 55,
         "total": 55,
-        "passRatePct": 70.91,
+        "passRatePct": 100,
         "admitted": 272,
         "upstreamTestsSeen": 273,
         "harnessIncompatible": 209,
--- a/plan/issues/3961-symbol-valued-struct-field-loses-identity.md
+++ b/plan/issues/3961-symbol-valued-struct-field-loses-identity.md
@@ -1,10 +1,11 @@
 ---
 id: 3961
 title: "A symbol stored in a struct field reads back as a bare integer — React.Children sees zero children"
-status: ready
+status: done
 sprint: current
 created: 2026-08-01
-updated: 2026-08-01
+updated: 2026-08-02
+completed: 2026-08-02
 priority: high
 horizon: m
 feasibility: hard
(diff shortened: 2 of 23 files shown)


ttraenkler:
**The park comment above is out of date — the failure it names is fixed. The current blocker is a different gate and needs one line.**

I'm posting this because auto-park re-ran on this PR at 10:21:04Z, concluded `success`, and added no new comment: the `hold` label was already present from 04:59Z, so it treated the PR as already parked. The result is that the only visible explanation still describes the original floor failure, while the actual blocker is only in the run log.

### The floor regression is resolved

Your own `f5fbe574 fix(inference): scope dynamic conflicts to recursive calls` fixed it. In run `30743086710` (merge_group, base `2c767617`) the `#2097` high-water gate **passes**:

```
[standalone-highwater] current pass=26422, mark=26176 (floor=26126, delta=+246)
```

That is +1560 over the 24862 this PR was parked at, and +20 above main's own 26402. The PR now measures net +17 pass (31237 → 31254), with the host stable-path fine-gate at +60 (62 improvements − 2 regressions).

**So the ~1500-file `"X should be an own property"` regression implied by the park comment no longer exists.** Please don't spend time chasing it.

### What actually fails now

```
GATE FAIL: trap category "null_deref" grew 153 → 154 (+1)
Now trapping: test/language/types/object/S8.6.2_A5_T3.js (baseline: fail)
```

That file's baseline status is **`fail`** — it was not passing before, it only changed failure *mode* from fail to trap. Per the contract in `scripts/check-baseline-trap-growth.ts`, `pass ⇒ genuine regression (no valve applies)`, but a **fail→fail reclassification** is covered by a named, change-scoped `trap-growth-allow:`.

### Suggested remedy

1. Add a bounded `trap-growth-allow:` naming `language/types/object/S8.6.2_A5_T3.js`.
2. ⚠ Put it in an issue file **this PR itself touches**. A grant written into another issue's file does not resolve in CI — that has bitten us before.
3. Push, let CI go green, and remove the `hold` only after that, in that order. The label is currently correct for a new reason, so clearing it first would just re-admit a PR the queue will reject.

Diagnosis by the PR-queue shepherd; gate text, label state, and park-comment contents re-verified before posting. Tracked on our side as task #68.

PR #4012, opened 2026-08-02
docs(#4083): record the loud→quiet trade that #4082 makes visible

ttraenkler:
One issue file, no code. Filed as **my condition for shipping PR #4011 (#4082)**, which is sound and should merge.

## What #4082 does well

Population 53 → mechanism 12 → reachable 12 → **9 flips**; kill-switch 12/12 fail on revert; regression control 500 → 496 with all 4 non-passes failing identically on revert (**0 attributable**). Its 9 flips assert a genuine `TypeError` on a non-RegExp receiver, confirmed by probe (`caught === 2`) rather than assumed — real, not vacuous.

## The residual it makes visible

```js
var re = /a/;
re.borrowed = RegExp.prototype.test;
re.borrowed("banana");
```

| | before #4082 | after #4082 |
| --- | --- | --- |
| outcome | **crashes** (invalid Wasm) | **validates, answers `null`** |

**No test moves pass→fail** — the file was broken before and after. What degrades is *diagnosability*: a loud failure becomes a quiet wrong value. Same class as #4075 (a refusal erased before it is reported), #4071 (enumeration silently empty), #4064 (silent infinite recursion).

A loud→quiet trade is acceptable when it is **tracked**, and unacceptable when it is absorbed into a green PR. Hence this file.

## Not caused by #4082 — checked, not assumed

A trace shows the #4082 arm selecting `f64.convert_i32_s ; call $__box_number`, which **cannot** yield null. The null comes from the outer dispatch's own `ref.null.extern` on a **#3992 exact-identity receiver-matching gap** that the crash had been masking. #3992's own doc block already names `null` as that gap's symptom.

## ⚠ Explicit prohibition carried in the issue

**Do not add a test asserting the current `null`.** #4082's tests deliberately do not pin the value. Pinning it would freeze the bug into exactly the permanent false green that #3603 removed ~989 of.

## Provenance

The authoring agent surfaced this itself rather than waving it through, and asked to be overruled if the call went the other way. It did not — #4082 ships — but the agent stood down before it could file this, so the lead filed it.

Docs-only: touches no `src/`, `tests/`, `scripts/`, `.github/` or `benchmarks/`. No other docs-only PR was open.

### DIFF
--- a/plan/issues/4083-transferred-regexp-receiver-returns-null-instead-of-crashing.md
+++ b/plan/issues/4083-transferred-regexp-receiver-returns-null-instead-of-crashing.md
@@ -1,0 +1,87 @@
+---
+id: 4083
+title: "A transferred `RegExp.prototype.test` on a RegExp receiver now returns `null` instead of crashing — a #3992 receiver-matching gap that the crash was masking"
+status: ready
+sprint: current
+created: 2026-08-02
+updated: 2026-08-02
+priority: high
+horizon: m
+feasibility: medium
+reasoning_effort: high
+task_type: bug
+area: standalone
+language_feature: n/a
+goal: standalone-mode
+related: [3992, 4082]
+---
+
+# Transferred RegExp receiver answers `null` instead of crashing
+
+Filed by the lead 2026-08-02 as a **required condition** of shipping #4082
+(PR #4011). The authoring agent surfaced this residual itself rather than
+letting it pass, and explicitly asked to be overruled if the call went the other
+way. It did not — #4082 ships — but the trade is recorded here rather than
+absorbed.
+
+## Repro (standalone)
+
+```js
+var re = /a/;
+re.borrowed = RegExp.prototype.test;
+re.borrowed("banana");
+```
+
+| | before #4082 | after #4082 |
+| --- | --- | --- |
+| outcome | **crashes** (invalid Wasm) | **validates, answers `null`** |
+
+## This is a LOUD failure becoming a QUIET wrong answer
+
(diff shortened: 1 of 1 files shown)


ttraenkler:
Closing — **superseded by PR #4011, which files the same defect as the same id, better placed.**

This is my duplicate, not theirs. I set filing the residual as a condition of shipping #4082, then filed it myself when the authoring agent appeared to have stood down. It had not — my own message resumed it — so we both wrote `plan/issues/4083-*.md`, with different slugs, in two open PRs.

**#4011's copy is the right one and should land:**

- It lives in the **code PR**, which is where the protocol puts an issue edit that belongs to a code change — splitting it into a docs PR would have orphaned it.
- It is **integrated**: #4082 cites it, the PR body cites it, and its fourth test names #4083 as the thing that upgrades that test from "validates" to a real assertion.
- Its acceptance criteria require `=== true` / `=== false` **by identity**, so a boxed number `1` cannot satisfy them either — it closes the failure mode next door, which my version did not.

No id is burned: 4083 stays with #4011. Nothing here is lost.

For the record, the coordination error was mine in a second way too — I filed #4081 from the same agent's *mid-investigation handoff* while it was still working that mechanism, which left a dispatchable issue duplicating a fix already in flight. #4011 resolves that by marking #4081 `done` with a do-not-dispatch banner pointing at #4082, and by correcting the lead I had propagated from the handoff (`closed-method-dispatch.ts` emits no `call_ref` at all; the real arm is `buildTransferredNativeProtoCallInstrs` in `src/codegen/closures/transferred-native-proto.ts`).

PR #4013, opened 2026-08-02
feat(eval): route standalone eval and Function through bytecode interpreter

ttraenkler:
## Summary

- route standalone direct `eval`, indirect `eval`, `Function(...)`, and `new Function(...)` through the shared zero-import Acorn/bytecode-interpreter provider
- reify caller lexical/variable environments, synchronize eval-visible globals, and preserve callable identity, receiver state, and catchable exceptions across separately compiled Wasm modules
- implement non-strict EvalDeclarationInstantiation preflight across caller lexical environments, including atomic collision errors and Annex B block-function cancellation
- keep literal direct eval honest by applying caller lexical/default-parameter collision rules before its AOT fast path
- document the measured Test262 acceptance result and resumable follow-up order in issues #2928 and #2929

## Why

The parser and interpreter could execute an arithmetic canary, but production runtime eval still lacked a complete module boundary: caller bindings were not live, provider-owned closures could not safely return to or call into AOT code, exceptions used private tags, and global updates were not synchronized. This checkpoint closes those routing and bridge seams while preserving the existing Acorn parser ABI and zero-import provider contract.

The follow-on collision slice closes the largest standard-eval correctness cluster: declaration preflight now walks intervening declarative environments before allocating persistent bindings, while eligible Annex B block functions are cancelled rather than overwriting a caller lexical binding. The literal AOT path applies the same collision rules instead of accidentally bypassing early errors.

## Acceptance evidence

Same-worktree standalone `language/eval-code` refusal-provider/full-interpreter A/B measurement:

| Route | Full interpreter |
| --- | ---: |
| Standard direct | 108/286 |
| Standard indirect | 60/61 |
| Annex B direct | 185/309 |
| Annex B indirect | 120/160 |
| **Total** | **473/816 (58.0%)** |

The refusal-provider control passed **158/816**. File-by-file comparison shows **315 fail-to-pass transitions, zero pass-to-fail regressions, and no other status changes**, satisfying #2928's interpreter-attributable acceptance gate.

Declaration-collision follow-on, using the exact maintained Test262 filter `declare-arguments|var-env-lower-lex`:

| Run | Pass | Runtime fail | Compile error |
| --- | ---: | ---: | ---: |
| Pre-slice full interpreter (`20260803-015311`) | 52/198 | 102 | 44 |
| Collision candidate (`20260803-034357`) | **153/198** | **1** | 44 |

That is **101 fail-to-pass transitions, zero pass-to-fail regressions, and 44 unchanged compile errors**. The sole remaining runtime failure is the parameter-environment closure case where a default-parameter arrow resolves a later body `arguments` function instead of the eval-created parameter-environment value.

## Validation

- merged the latest `origin/main`/remote PR update before publication
- `pnpm run typecheck`
- focused interpreter/environment/provider gate: **167/167**
- post-merge collision + real zero-import Acorn package gate: **75/75**
- Annex B `typeof`/block-function regression gate: **18/18**
- `pnpm run check:dead-exports`
- changed-root commit gate, plus oracle, LOC, function-budget, and coercion ratchets
- pre-push: typecheck, lint, Prettier, oracle/coercion ratchets, numeric-local IR parity, and issue integrity
- real Acorn runtime-provider and self-compiled interpreter canaries remain green and host-free

Generated Test262 reports remain local and are intentionally excluded from this PR.

## Resumable follow-up

This is the runtime-routing MVP, not full eval conformance. The next high-value work is:

1. fix the one remaining collision-slice runtime failure by preserving the parameter-environment closure across later body declarations
2. close the 44 unchanged compile errors in that slice: 36 invalid-Wasm method/generator cases and 8 host-import leaks
3. continue Annex B outer-binding initialization/update coverage, then `with` object-environment records and Proxy MOP convergence under #2929


ttraenkler:
**This PR is a draft, and draft status is opting it out of branch maintenance — it is now 57 commits behind `main` and diverging.**

Flagging it because the mechanism isn't obvious and nothing will surface it on its own:

- **`auto-refresh-prs` skips drafts.** It will never rebase this branch, so the gap only grows.
- **`auto-enqueue.yml` skips drafts too.** Even at `mergeStateStatus: CLEAN`, a draft is never enqueued — so this cannot land in its current state regardless of how green it is.

Measured just now against `upstream/main`:

```
head 237c92cb   ahead 5   behind 57   status: diverged
```

Last push was 2026-08-02T06:22:14Z. The change is 30 files / +1375 in direct-eval capture and strict declaration environments — a high-conflict area, so the longer it diverges the more expensive the eventual catch-up merge gets.

No action needed from you if this is deliberately parked. But if it's meant to land, the two useful steps are to merge `origin/main` in now rather than later, and to mark it ready for review so branch maintenance and auto-enqueue start applying. If it's *not* meant to land soon, closing it and re-opening from a fresh branch later would cost less than carrying the divergence.

Raised as part of an open-PR sweep; posted with the project lead's go-ahead. Not a review comment on the change itself.

ttraenkler:
**Update to my earlier note — this PR has gone from `CLEAN` to `DIRTY`, and the price of the options I listed has changed.**

When I commented earlier it was `CLEAN`, 57 commits behind, and "merge `main` in and mark ready" was cheap. That is no longer true.

**Now:** `DIRTY`, **79 commits behind** (57 → 79 in ~75 minutes), 30 files / +1375.

**What caused it, specifically.** PR #4027 merged at 12:12:44Z. It overlaps this branch on exactly two files:

```
src/codegen/expressions/calls.ts
src/codegen/statements/nested-declarations.ts
```

`nested-declarations.ts` is the file #4027 changed most heavily (**+1/−62**, Annex B cancel-site work). This PR is direct-eval capture — the same codegen neighbourhood. Nothing about this branch changed; its environment did.

**Why that matters more than an ordinary conflict:** `auto-refresh-prs` skips `DIRTY` **as well as** drafts, so there is now no automated path back to current — even if it were un-drafted. Resolution has to be manual, in two actively-changing codegen files, and every further merge into that neighbourhood raises the price. **The cost of leaving it parked is no longer zero; it is a growing conflict-resolution bill.**

For what it is worth, this is the failure mode the earlier comment predicted, arriving on schedule: draft status opted the branch out of the one mechanism that would have kept it current, and the consequence stayed invisible until an unrelated PR happened to land in the same files.

Still entirely your call — no action is being taken on this PR from our side. But the three options are no longer equally priced, and "decide later" now has a real and increasing cost attached. If it is meant to land, resolving now is materially cheaper than resolving after the next codegen merge. If it is not, closing it and re-branching later would cost less than carrying the divergence.

Posted as part of an open-PR sweep, with the project lead's go-ahead. Not a review of the change itself.

ttraenkler:
**Triage notes on the park, posted because two load-bearing facts are not visible from the conversation.**

**1. The park comment above is the FIRST failure only. The re-enqueued run failed identically and added no comment** (auto-park doesn't append when `hold` is already present). Fresh run for the current head:

- 2nd merge_group: https://github.com/loopdive/js2/actions/runs/30780676800 — job `91586115137`, step `Standalone pass-count high-water floor (#2097)`, `failure`. Floor read `current 26814, mark 27046, delta −232` (first run −220; the delta difference is main advancing underneath the unchanged branch, not the PR worsening).

**2. ⚠ A `trap-growth-allow` CANNOT clear this park — don't reach for the #4008 playbook.** Of the six newly-trapping files in the ratchet failure, **five have baseline `pass`**:

```
test/annexB/language/function-code/if-decl-else-decl-a-func-existing-var-update.js   (baseline: pass)
test/annexB/language/function-code/if-decl-else-decl-b-func-existing-var-update.js   (baseline: pass)
test/annexB/language/function-code/if-decl-else-stmt-func-existing-var-update.js     (baseline: pass)
test/annexB/language/function-code/if-decl-no-else-func-existing-var-update.js       (baseline: pass)
test/annexB/language/function-code/if-stmt-else-decl-func-existing-var-update.js     (baseline: pass)
test/language/expressions/tagged-template/tco-call.js                                (baseline: fail — this one IS valve-eligible)
```

Per the gate's contract in `scripts/check-baseline-trap-growth.ts`: `pass ⇒ genuine regression (no valve applies)`. The five must be fixed in code; only the `tco-call.js` fail→trap reclassification could take a named allowance.

**3. Clustering hint — hypothesis, not a finding (rows unread):** all five are `annexB/language/function-code/*-func-existing-var-update.js` — Annex B B.3.3 function-in-block with an existing var binding, i.e. declaration-environment semantics. Two pointers that may shorten the path:

- Your diff touches declaration environments directly (strict declaration environments, eval capture) — the B.3.3 web-compat var-update path is the likeliest collision surface.
- **`src/codegen/annexb-cancel.ts` is NEW on main as of 2026-08-02** (PR #4027) and owns exactly this territory (B.3.3 cancel-site collection, cited in the same file family). If your branch predates or partially merges around it, the interaction between your declaration-environment changes and that module's cancel-site answers is worth checking before anything else — the five failing files are its home turf.

Floor context: the high-water mark now ratchets on every merge (fixed 2026-08-02), so you're being judged against a live bar (27046) that rises as main advances — re-measure against the current mark before concluding a fix worked.

Posted with the project lead's go-ahead; factual state only, not a review of the approach.

PR #4014, opened 2026-08-02
refactor(ir): seal prepared components before lowering

ttraenkler:
## Summary

- derive each R2 component's source, import, runtime, intrinsic, global, and type dependencies before any owned body is lowered
- publish exact scoped Program ABI views and reject post-seal locator, alias, type, or provider drift
- carry backend-neutral string carrier, literal, length, concat, equality, `charAt`, and `charCodeAt` dependencies in final IR
- bind prepared string intents to exact host imports or stable native helper handles, including a distinct `__str_concat_owned` path
- record the remaining static string-builder presize and counted-loop aggregation optimizations as explicit retirement blockers

This advances #3521's prepare-before-emit inversion. Prepared production bodies consume sealed symbolic references instead of discovering numeric Wasm indices while lowering, while JS type inference remains independent of the selected backend.

## Validation

- focused Program ABI, prepared-component, routing, string-contract, and owned-append matrix: 96/96
- expanded numeric-local and host/native string matrix: 238/238
- cross-backend differential: 29/29
- required broad #3521 suite: 104/106; only the two parent-recorded inline-tail-call and imported-overload failures remain
- full equivalence: 1,605 passing / 38 failing; the same six parent-recorded failures remain outside the 36-entry baseline
- typecheck, lint, Prettier, fallback ratchet, adoption, optimization ledger, oracle/coercion ratchets, LOC budget, function budget, and pre-push issue integrity: passing
- hybrid readiness: READY, 31/37 IR-emitted, 6 typed Unsupported, 0 invariants
- strict IR-only: NOT READY, because 37/37 bounded lane units still emit a legacy body and six units remain unsupported

Production disassembly resolves the prepared builder body's numeric call index and proves it targets `__str_concat_owned`; the existing geometric-growth and string-hash runtime cases remain green.

## Remaining retirement work

String iteration and oversized native literals still need symbolic preparation. The direct string-builder presize proof and counted literal-loop `repeat(N)` transform must be migrated before those handlers are deletable. Dynamic/object/ref-cell/closure/union/vector layouts, iterator/generator/exception/async providers, pass-derived callables, and the explicit emission transaction remain later R2 dependencies.


ttraenkler:
Superseded by #4025, which contains this preparation stack plus the current component-local/static-method continuation and is now in the merge queue.

PR #4017, opened 2026-08-02
fix(#4055): standalone hasOwnProperty consults the closure own-property bag

ttraenkler:
Closes #4055 (LEVER 1, function-carrier stratum). Instance **#7 of #4080**.

## The gap

#3468 gave `__extern_get` / `__extern_set` / `__extern_method_call` a fallback for a
receiver that is not a `$Object`: an identity-keyed side table mapping the carrier to a
`$Object` "bag" of its own properties. **`__hasOwnProperty` was never wired to it** and
still bailed with `0` on `ref.test $Object`, so a function denied a property it had just
stored *through the same substrate*:

```js
var f = function () {};
f.enumerable = true;
f["enumerable"];                 // true   — __extern_get reads the bag
f.hasOwnProperty("enumerable");  // false  — the gap
```

Not merely a wrong boolean. `__obj_define_from_desc`'s ToPropertyDescriptor (§6.2.5.6)
gates **every** descriptor field on `HasProperty` before reading it, so a **Function
descriptor carrier** — the dominant test262 spelling, `var descObj = function(){};
descObj.enumerable = true;` — produced an EMPTY descriptor and CompletePropertyDescriptor
filled in `undefined` plus all-false attributes. Silently: no refusal, wrong content.

## Read this before pattern-matching it to #4047's reverted arm

The arm #4047 measured at +6 and reverted resolved a **`Properties` map** through the bag.
Enumerating a map requires a **complete own-key source**; the bag is not one — `props.p = v`
lands in the bag while `Object.defineProperty(props,"p",…)` lands in the separate #3251
overlay (Array) or nowhere (Function) — so it enumerated empty, defined nothing and
returned normally.

`hasOwnProperty(k)` is a **fixed-key presence query**. No key source is needed, and the bag
is precisely where `__extern_set` put the write, so presence and read agree by
construction. `Object.defineProperty(fun,"p",…)` still lands nowhere and this arm still
answers `false` for it: **no new inconsistency**. `tests/issue-4047.test.ts`,
`issue-3957`, `issue-3468-closure-own-props`, `issue-3537`, `issue-3251`, `issue-1906`,
`issue-3984`, `issue-4032` all pass (55/55 post-merge on the first four).

## Measurement

Baselines force-refetched before sizing; standalone rows timestamped 2026-08-02 07:26.
Back-to-back same-box arms from one script, 180 s compile timeout, row counts floored
(`rows == expected`, else exit 9), row SETS identity-checked, read from the JSONL — never
a vitest reporter tally. Attribution by **kill-switch removal**.

Instrument validated first: the BASELINE arm agreed with the published standalone baseline
on **203 / 203** rows.

| stratum | denominator | n | base pass | treat pass | fail→PASS | pass→FAIL |
|---|---|--:|--:|--:|--:|--:|
| **T** failing ≤ES5 descriptor-area files with a FUNCTION descriptor carrier | **census** — all 78 | 78 | 0 | 14 | **+14** | 0 |
| **W** failing with an expando descriptor build on a NON-function carrier | 40 of 98 | 40 | 0 | 0 | 0 | 0 |
| **N** currently-PASSING files asserting a **FALSE** `hasOwnProperty` | 45 of 206 | 45 | 45 | 45 | 0 | **0** |
| **C** currently-PASSING descriptor-area files | 40 of 1,775 | 40 | 40 | 40 | 0 | **0** |

Stratum **N** exists because this change is **bidirectional** — it flips `hasOwnProperty`
false→true, so the failure mode to detect is OVER-reporting. Zero moved. Both arms
produced **0 `compile_error`** rows (load 11.98 → 10.52), so there is no contention flake
to re-run solo.

Flip ratio **14 / 78 = 17.9 %** on a CENSUS, not a sample — no extrapolation.

### Vacuity audit of all 14

Every flip carries a POSITIVE assertion; none is satisfiable by a silent no-op:
`obj.property === "Function"` / `"functionGetProperty"` (a **getter that must fire** in
`-3-218`), `accessed !== true` (for-in must SEE it), `result1 === true`,
`beforeWrite && afterWrite` (`writable` honoured), `beforeDeleted === true`
(`configurable` honoured), and `data === "overrideData"` (a **setter that must fire**).
`15.2.3.6-3-248` asserts `obj.hasOwnProperty("property")` — not circular: `obj` is a plain
`{}`, the carrier is `funObj`, and its companion value assertion is independent.

## Deliberately NOT shipped

- **The ARRAY half.** A vec arm was written, **measured unreachable, and removed rather
  than shipped as decoration**: `fillVecHasOwnHelpers` **unshifts** a prologue into
  `__hasOwnProperty` that answers from `__vec_gopd` and `return`s for every vec receiver.
  Probe: `a=[1,2,3]; a.q=5` → `a.q === 5`, `a.hasOwnProperty("0") === true`,
  `a.hasOwnProperty("9") === false`, but `a.hasOwnProperty("q") === false`. That is the
  #3251-overlay-vs-#3537-bag split filed as **#4010**. Pinned by the last test so the
  boundary is a decision, not an oversight. **This fixes a symptom, not the substrate.**
- **`isOpenDescriptorShape`** — a second, real defect found and root-caused on this issue
  (`d.enumerable` reads `undefined` while `hasOwnProperty` says true), measured at
  **+0 flips / 0 regressions** over 145 files (35-file census + 40 + 30 + 40 control,
  instrument validated 145/145) and **left in place**. It is inert because the test262
  shapes here use exotic carriers, not widened plain objects. Recorded in the issue with
  the reason, which is worth more than the change would have been.
- **#4062 is NOT subsumed** — its repro is byte-identical base vs treatment. **#4061**
  (descriptor-ARGUMENT *validation*) is untouched; this is about *reading* well-formed
  descriptors.

## Shape of the diff

The fix is a new subsystem module, `src/codegen/carrier-bag-hasown.ts`. What lands in the
god-file is the irreducible splice — one import, a rationale comment, one call, one entry
in the locals list: **+4 LOC / +3 in `ensureObjectRuntime`**, granted in the issue
frontmatter. Moving the bail body (including its terminal `i32.const 0; return`) into the
module is what took it from +17/+16 down to +4/+3.

Coordinated with `L-enum` (#4071): disjoint — it owns *which helpers receive* the
closed-struct arms, this owns a new arm in `emitHasOwn`; neither touches the other's
function.


ttraenkler:
## Park resolved — rescoped, not shrunk

The auto-park was a **real merged-baseline regression**, diagnosed before the label came off.

**What it was.** v1 widened `__hasOwnProperty` / `__object_hasOwn` themselves. Artifact diff against #4016 on the identical base `49cf55a4545`: **-684 host-free passes**, 713 files lost, of which **682 (95.7%) are `name.js`/`length.js`** and **696 fail "descriptor should be configurable"**. `propertyHelper.js` reaches `Object.prototype.hasOwnProperty` on every `built-ins/**/{name,length}.js` test — a ~700-file population **disjoint from every stratum v1 sampled**.

**The arm was not subtly wrong.** Its answers were the ones asked for. It was wired at the most GENERAL point that could express the fix, and generality there *is* blast radius.

**v2.** The widening now lives in `__desc_has_own`, a native **only ToPropertyDescriptor calls**. It calls `__hasOwnProperty` **first** and consults the bag only if that answered false — it can add a `true` the helper declined, never override one it gave. `Object.prototype.hasOwnProperty`, `Object.hasOwn` and `propertyIsEnumerable` are byte-identical to before.

| check | result |
| --- | --- |
| function-carrier flip **census**, all 78 | **14 pass** — every flip survives the rescope |
| park repro (`Math/ceil/length`, `Array/prototype/fill/name`) | 2/2 pass (v1: 0/2) |
| **`built-ins/**/{name,length}.js`**, 972 files | **971/972 agree with baseline, 0 baseline-pass → FAIL** |
| unit + 8 neighbour guard suites | 100/100 |

Stratum re-run **after** merging current `main` (incl. #4071): identical.

**The harness-only trigger was never isolated — recorded on purpose.** Three candidate mechanisms were measured and none reproduced outside the full harness: `gOPD(Math.ceil,"length").configurable` is `true` on **both** arms; `isConfigurable`'s real body already answers wrong on **both** arms (pre-existing); `verifyProperty`'s vacuity gate is `true` on **both** arms. When a mechanism resists isolation, narrowing the change until it is out of scope beats chasing it.

**Why v1's controls could not see it** — not "missed it", *incapable*, two ways, both now closed: (1) `runTest262File` does not apply the #2961 host-import refusal (`standaloneHostImportError` is called only from `runSyntheticTest262File`), so no runner-based control can see `host_import_leak_class`; (2) the affected population was disjoint from all three strata sampled. A corpus-wide 220-file sweep run during diagnosis did catch it independently (7 pass→FAIL, all `name`/`length`, 3.2% vs the artifact's 2.6%) — the sweep *axis* was sound, only the *stratum* was wrong.

Both are adopted as standing controls: the `name`/`length` stratum, and an imports check with a positive control in every standalone control from the start.

`hold` removed after diagnosis + verification, per the auto-park rules. Not enqueuing — `auto-enqueue.yml` owns that.

PR #3365, opened 2026-07-18
ci(#3431): consolidate test262 shards for merge_group validation (~60min -> ~20-30min)

ttraenkler:
## Summary

The merge queue can build up to `max_entries_to_build: 5` groups concurrently, each running test262-sharded.yml's full 57-shard x 2-target = 114-job matrix — up to 570 concurrent jobs. Real merge_group run telemetry (2026-07-18, `gh run view`) shows this job-count contention, not fixed per-job setup overhead (measured ~30-45s/job), is what stretches merge_group validation past 60 minutes and drains the queue at ~1 PR/hour.

This PR splits `merge_group` into its own **consolidated** shard job (`test262-shard-mg`), separate from the existing `test262-shard` job. `pull_request`/`push`/`workflow_dispatch` are completely unchanged — they keep using `test262-shard`'s original 114-job matrix.

- `scripts/gen-test262-mg-matrix.mjs` computes a 59-job matrix: **40 js-host + 19 standalone** shards (different counts per target, because they have very different per-shard runtimes at the same split — measured js-host avg 13.6min/max 15.9min vs. standalone avg 5.8min/max 7.0min at the current 57-way split). Scaled estimates for the new counts (~19.4min/22.7min host, ~17.4min/21.0min standalone) stay under the existing 25-min shard timeout with headroom — **no timeout change needed**.
- Every matrix cell runs `tests/test262-chunk-dynamic.test.ts` (new), a dynamic entry point that reads `TEST262_CHUNK_INDEX`/`TEST262_CHUNK_TOTAL` from the env and calls the existing `runTest262Chunk(idx, total)` — this is a pure function of `(chunkIndex, totalChunks)` that re-derives and bin-packs the full corpus, so it's behaviorally equivalent to the static per-chunk files (already proven by `tests/test262-local-shard*.test.ts` running the same corpus at a different chunk count for local dev).
- `merge-report`'s aggregation already globs shard artifacts by pattern rather than hardcoding a count — no changes needed there beyond `needs:`/`SHARDS_RAN` accepting success from either shard job (only one runs per event).

Total merge_group job count: 114 -> 59 (48% reduction).

Full evidence, invariant analysis (corpus coverage, RAM scaling), and rollback notes are in `plan/issues/3431-consolidate-mergegroup-test262-shards.md`.

## Test plan

- [x] `pnpm run typecheck` clean
[...]

PR #4027, opened 2026-08-02
rescue(#4091): land the /workspace staged work that was blocking the shared checkout

ttraenkler:
**I did not author this code and I have not measured it.** Read the "not verified" section before treating any of it as sound. It is opened so the work is reviewable rather than rotting — not because it is known-good.

## Why

**1,726 lines sat staged in the shared `/workspace` checkout** — no branch, no PR, no issue. Invisible to the merge queue, to `auto-enqueue`, and to every dispatcher. It also blocked `sync-workspace-main.sh`, so `/workspace` was **189 commits behind**, which meant `sync-current-tasklist.mjs` could not see a single issue filed today.

Pinned first as `refs/rescued/workspace-staged-20260802` (`5f77eeaa`) so it could not be lost, then restored here.

## Restored as a PATCH, not a file copy

The pin's base is 189 commits old. A whole-file copy onto a moved base **silently reverts whatever landed underneath**, and the diffstat is the only cheap signal — that near-miss happened to another lane earlier today (161/102 where +50/−10 was expected).

Verified: **778 insertions / 88 deletions across 8 files**, byte-identical to the pin's `src/` + `tests/` portion. `tsc --noEmit` clean. The patch applied with **no conflicts** despite the gap.

## Docs deliberately excluded

The pin also carried 8 `plan/issues/` files under `3178/3973/3974/3975/3977/3980/3982/3983`. **Those ids are now different issues on main**, and the content already landed under **4020–4025** via PR #3994 — verified by title match (pinned `3983-apply-drops-receiver` is identical to main's `4025-apply-drops-receiver-no-this-binding`). Re-adding them would be a duplicate-id collision *and* wrong content.

## What it contains

`src/codegen/annexb-cancel.ts` (new, 376 lines) · `src/codegen/named-this-call.ts` (new) · changes to `expressions/calls.ts`, `expressions/identifiers.ts`, `js-errors.ts`, `statements/nested-declarations.ts` · two test files. Subject matter maps to **#4023** (Annex B B.3.3 hoisting on early error) and **#4025** (`.apply(thisArg)` drops the receiver).

## ⚠ NOT verified

- **No test262 measurement.** Not stale — *absent*.
- No kill-switch attribution, no regression control, no funnel.
- Nobody has confirmed the two new modules do what their names say.

**The CI gates — especially the `merge_group` standalone floor (#2097) — are the first real evidence about this.** Two PRs parked on that floor today for −684 and −1500 host-free passes, both invisible to PR-level checks. If this parks, that is the system working correctly.

## Two allowances, ONE growth

Stacking allowances is a smell, and a lane correctly refused a second one on #4089 today. **That is not what this is.** Both gates report the *same 8 lines*:

| gate | reading |
|---|---|
| `check:loc-budget` | `calls.ts`: 8444 → 8452 **(+8)** |
| `check:func-budget` | `calls.ts::compileCallExpression`: 1765 → 1773 **(+8)** |

A file-size view and a function-size view of one hunk. Both are **provisional** and share one retirement condition in #4091: **if nobody validates this with measurements, close and revert rather than keep the allowances.** The subsystem-module remedy was not attempted because restructuring unmeasured code I did not author would be worse than the +8.


ttraenkler:
**The park is correct. I verified it against the job logs rather than the park comment, and the regression is real — this must not land as-is.**

Recording it here because I opened this PR and the issue file predicted exactly this outcome. From #4091, written before CI ever ran:

> The CI gates — especially the `merge_group` standalone floor (#2097) — are the first real evidence about this change. […] **If this parks, that is the system working.**

It parked. The system worked.

### Measured (run `30743793429`, both gates)

| gate | improvements | regressions | net | tolerance | bucket signature |
| --- | ---: | ---: | ---: | --- | --- |
| Standalone regression guard (#1897) | 107 | 143 *(7 compile_timeout flake excluded)* | **−36** | −15 | `3b09e6051d855a32` |
| Host stable-path fine-gate | 146 | 152 | **−6** | 0 | `b7e705d343fd7b50` |

Host coarse net **−46** (31237 → 31191). The floor gate (#2097) **passed** at `pass=26383, +207` — the floor is not the problem, the net-regression gates are. Both rate it `LIKELY-REAL REGRESSION (baseline content-current, #2562)`, so baseline drift is ruled out by the gates themselves, and no other open PR carries either signature.

### This is not one bug — it is partially-complete work

The shape says so: it **both fixes and breaks a great deal** (146 up / 152 down host; 107 up / 143 down standalone). A single defect does not produce that.

The diff says the same thing more precisely: `statements/nested-declarations.ts` is **+1/−62** while a new `annexb-cancel.ts` is **+376**. That is logic lifted out of one module into another — and an incomplete lift produces exactly this signature, because the new module covers most but not all of what the deleted inline code did.

**So whoever picks this up should expect to split the working half from the incomplete half, not hunt one root cause.** The ~250 combined improvements are real and worth recovering; they are simply being sold at a loss right now.

### Standing caveats

- ⚠ **Do not remove the `hold`.** The park is accurate and the PR would fail the queue again.
- ⚠ **Nothing here was authored or measured by the rescuing session.** #4091 says so explicitly: no test262 measurement, no kill-switch attribution, no regression control. These gate numbers are the first real evidence this change has ever had.
- The branch name reads `chore-rescue-…`, but the diff is **6 `src/codegen/` files**. It is a codegen change, and should be reviewed as one.

### If nobody takes it

#4091 pre-committed the answer, and I'd rather honour it than quietly let the PR linger:

> if this work is not validated by an owner who can measure it, **close the PR and revert rather than keeping the allowance.** […] Either land it properly with numbers, or close this and let the pin be the record.

The pin `refs/rescued/workspace-staged-20260802` (`5f77eeaa`, 16 files / 1,726 lines) is intact and independent of this PR, so **closing costs nothing** — the work stays recoverable either way. The rescue's actual objective (unblocking the shared checkout and making 1,726 invisible lines reviewable) is already achieved and is not contingent on this landing.

ttraenkler:
Park resolved — removing `hold` per the bot comment's own instruction, after diagnosing the cited run.

**Diagnosed run `30743793429` (both cited jobs) before touching the label.** Standalone guard `-36` (107 imp / 143 reg, sig `3b09e6051d855a32`) and host fine-gate `-6` (146 / 152, sig `b7e705d343fd7b50`).

**One defect, not a broad regression.** Every one of the 152 host and 143 standalone non-timeout regressions carries the same error text:

```
Internal error compiling expression: Invalid value used as weak map key
```

`compileIdentifierCore` calls `collectAnnexBCancelSites(id.getSourceFile())`, and `getSourceFile()` returns **undefined for a synthesized identifier**. The new module memoized on that key, so `WeakMap.set(undefined, …)` threw, and `compileExpressionBody`'s speculative catch turned the throw into a **whole-file compile_error**. The trigger is script-goal top-level `this`, which `expressions.ts` (#3365) lowers via a fresh `ts.factory.createIdentifier("globalThis")` — so the damage landed on ordinary code (`statements/with` 27, `compound-assignment` 11, `Object/defineProperty` 10), none of it Annex B. Corpus-wide occurrences: **11 baseline → 677 host / 679 standalone candidate**.

Fix is `if (!sf) return NO_SITES;` at the new module's entry point — the narrowest site that produces the whole effect.

**Attribution, full populations (not samples):**

| lane | population | base `upstream/main` | guard OFF | guard ON |
| --- | ---: | ---: | ---: | ---: |
| host | 20-file control | 20/20 pass | — | — |
| host | all 152 regressed | (baseline: pass) | 0/152 | **152/152 pass** |
| standalone | all 143 regressed | (baseline: pass) | 0/143 | **143/143 pass** |

Baseline force-refreshed; re-diffing the parked run's own artifacts reproduces CI's numbers. **Caveat:** the local sweep uses `runTest262File`, which is *not* the CI path (no #2961 host-import refusal) — sound for recovering a compile-time throw, not for certifying standalone conformance. The `merge_group` re-run remains the certifying measurement.

Worth noting from the parked run itself: **zero `pass → fail` regressions in either lane.** The Annex B semantic change produced no false-positive ReferenceErrors; 100% of the damage was the crash.

Both halves of the change earn their place, so nothing was split off: 96 of the improvements are the `annexB/*-skip-early-err-*` family, and the `fail → pass` set includes `10.4.3-1-{69,88}{-s,gs}.js`, whose body is literally `f.apply(o)`. (Most of the *other* raw improvements are `compile_timeout → pass` flake recoveries — CI counted 53 — so do not read the headline number as that many fixes.)

Not enqueuing; `auto-enqueue.yml` owns that. Full measurement in `plan/issues/4091-rescue-workspace-staged-annexb-cancel-named-this-call.md`.

ttraenkler:
**Correction to my earlier comment — my "this is not one bug, it is partially-complete work" diagnosis was WRONG, and refuted by measurement.**

I wrote above that the shape "both fixes and breaks heavily" meant an incomplete module lift needing a split. It was **one defect**.

`compileIdentifierCore` passed `id.getSourceFile()` into the new `annexb-cancel.ts`; for a **synthesized** identifier that returns `undefined`, and the module memoized on it. `WeakMap.set(undefined, …)` throws, and `compileExpressionBody`'s speculative catch turns that into a **whole-file compile_error**. Trigger is script-goal top-level `this`, which is lowered by re-entering `compileIdentifier` with a fresh synthetic `globalThis` (#3365).

The tell was visible in the artifacts the whole time: **all 152 host and all 143 standalone non-timeout regressions carry the identical error text.** And the regressed buckets — `statements/with` (27), `compound-assignment` (11), `Object/defineProperty` (10) — are not Annex B at all, which should have made me suspicious of my own reading.

Fixed at the narrowest site, inside the module this PR adds (`bef66a37`): `if (!sf) return NO_SITES;`. Attribution on full populations, not samples — guard off **0/152** and **0/143**; guard on **152/152** and **143/143**.

### Where the reasoning failed, since that is the reusable part

The split hypothesis was inferred from **summary counts** (146↑/152↓ host, 107↑/143↓ standalone) via "both directions ⇒ cannot be one bug." That inference is unsound. A single defect in a **cache on a hot path** produces exactly that mixed profile, because improving and regressing cases both consult it. The counts were correct; they simply could not distinguish the two hypotheses, and I treated them as if they could.

> **A regression summary tells you the SIZE of a problem, never its SHAPE. The rows do.**

Reading the regressed files and diffing their error texts took minutes and collapsed the hypothesis space immediately.

**Nothing else in my earlier comment changes:** the park was correct, the regression was real, and the `#2097` floor was not the failing gate. Only the mechanism was misdiagnosed — and it was labelled a hypothesis rather than a finding when handed to the owner, which is why the measurement was run instead of a nonexistent split being hunted.

Both halves of the change are retained; neither was dropped. The `merge_group` re-validation remains the certifying measurement.

PR #4057, opened 2026-08-02
docs(#4098): re-measure, map the mechanism, escalate — population 124, blocked on #4010

ttraenkler:
Docs-only. No source changes. #4053 (the previous docs PR) merged, so this opens a new one rather than riding it.

Measured before designing, per #4098's own instruction. Baselines `baseline_sha 6660c1158`, 2026-08-02T19:39:50Z — 43,505 standalone / 43,484 host official rows, **0 corpus files unopenable** (floored). The baseline **includes #3976 slice 1** (`be6cceef6`, verified by ancestry), so the residual is measured, not inferred.

## Both filed claims confirmed — one of them is stronger than filed

**Unanimity reproduces exactly:** the `C.prototype` residual is **60**, all **60/60** in the single descriptor bucket. No second bucket.

**"Do not add 60 + 276" understates it — the 60 is a strict SUBSET.** Instance population **124**, union **124**, overlap **60**. So the number to size by is **124** (not 60, not 276, not 336), and fixing instance fields **subsumes #3976's entire residual**. The filed 276 was the pre-slice-1 figure.

Funnel, with #3976's stage figures in brackets: 43,505 official → 5,503 fail-SA∧pass-host → **1,285** include propertyHelper [1,810] → **170** own-property/descriptor buckets [1,136] → **124** class area [826]. The shrink is slice 1 draining the bucket.

## The deciding measurement came back the expensive way

#4098 said to check the writable/configurable ratio **before** committing to a slice. Descriptor literal readable in **124/124** (floored):

`writable` **100.0%** · `configurable` **100.0%** · `enumerable` **100.0%** · `writable ∧ configurable` **100.0%**

Identical to #3976. **A presence-and-descriptor-only fix flips ZERO of 124.** The delete path is not optional — `isConfigurable` (`propertyHelper.js:138`) performs a real `delete obj[name]` and requires `hasOwnProperty` to become false.

## Mechanism map — and a correction to the title

⚠ **The first probe was wrong, and how it was wrong is the finding.** Using a **literal** property name it reported "everything works except delete" — that measured the **static fast path**. `verifyProperty` receives the name as an *argument*, so the corpus takes the **dynamic-name** path. Isolated 2×2 (instance scope × literal-vs-variable name): **only the name matters**; instance scope is irrelevant.

Dynamic-name behaviour for a public instance field, standalone, host-free (0 imports):

| operation | standalone | spec |
| --- | :---: | :---: |
| `hasOwnProperty(o, name)` | ✓ | ✓ |
| dynamic read `o[name]` | ✓ | ✓ |
| `getOwnPropertyDescriptor(o, name)` | **✗ undefined** | ✓ |
| `Object.keys(o)` includes it | **✗** | ✓ |
| `delete o[name]` removes it | **✗** | ✓ |

So the title is imprecise: instance fields **are** already own properties for `hasOwnProperty` and dynamic reads — the closed-struct arms work. Three specific gaps, which is the filing's candidate 1 confirmed and pinned:

1. **no `getOwnPropertyDescriptor` arm** — why the error reports *all four* attributes wrong: the descriptor is `undefined`, not wrong-valued;
2. **`Object.keys` omits the field** — the **opposite direction** from #4099 (which fails to *exclude* non-enumerables), so **#4099 alone does not fix this**; adjacent, not the same;
3. **`delete` is a no-op** on a struct field.

## Why escalated rather than built

(1) and (2) are tractable beside the existing arms. **(3) is not** — a WasmGC struct field cannot be removed, so `configurable: true` (asserted by **100%** of the 124) needs a **per-instance store with tombstones**: the filing's candidate 2, and precisely **#4010**'s substrate, interacting with **#4086**'s arm screening. That is the escalation boundary I was given.

Building (1)+(2) alone is the worst option available: **zero flips**, plus a `getOwnPropertyDescriptor` reporting `configurable: true` — a lie the next `delete` disproves. That trades a clean absence for a confident wrong answer.

Set `status: blocked`, `blocked_by: [4010]`. Land the unified per-receiver own-property side table first; then all three become one coherent slice against a store that can represent a tombstone. Populations are not flip ceilings — expect #3976's discount on re-run.


ttraenkler:
Grouped a second docs-only commit onto this PR per the docs-only rule (no other docs PR was open): **#4010 regrounding**.

Adds a **receiver × operation capability matrix** (7 receivers × 7 ops, one module per cell). Control validated — `$Object` is 7/7 ok, matching #3991's independent 9/9 on the same substrate.

**The delete column needed its own control, and the first pass was wrong in two ways at once:** array/function report the property **STILL PRESENT** after `delete` (the probe read "ok" only because `hasOwnProperty` is itself ABSENT there, so it answered false regardless), and Date/RegExp/Error/class-instance deletes are **vacuous** — the write never landed. That is the #4065 family biting the instrument itself: *"is q an own property of o?"* derived two ways, disagreeing, with no home for the invariant — which is independent evidence for #4010's own design goal.

Also records that **no "narrower ready-to-take prerequisite" exists** for #4010 (`related: []`; the only narrower issues, #4006/#4007, are the ones it says explicitly not to fund separately), and proposes S1/S2/S3 slicing with S1 as the only provably non-displacing slice.

The PR title still reads #4098 only — `gh pr edit --title` fails against this repo with the projects-classic GraphQL deprecation error.

PR #4063, opened 2026-08-03
fix(#4010): S2 — tombstones, so `delete` is real on a non-$Object receiver

ttraenkler:
## Read this first: acceptance here is NOT pass-count

Per #4010's slice plan, agreed with the tech lead:

> Accept on the **delete-column cells going STILL-PRESENT → genuinely-absent**,
> verified **both** by value read **and** by the store's own record.

Tombstones alone flip few test262 files, because currently-invisible properties
fail *earlier* in `propertyHelper`. Deletability and visibility only pay out
**together**, and visibility is S3. **A low corpus flip count is the expected
outcome of this slice, not a failure** — see the numbers at the bottom.

## The defect

`__delete_property`'s non-`$Object` arm returned **1 (success) without deleting
anything** — a loud claim of success covering a silent wrong answer. Measured on
current main, standalone:

```js
const a = [1,2,3]; a.q = 12; delete a.q;   // returns TRUE
a.q                                        // => 12   (STILL PRESENT)

function f(){}  f.p = 12;  delete f.p;     // returns TRUE
f.p                                        // => 12   (STILL PRESENT)
```

Both values live in a per-receiver `$Object` **bag** — `vec-props.ts` (#3537)
for arrays, `closure-props.ts` (#3468) for functions — that the delete native
had never heard of.

## The fix

A bag **is** an ordinary `$Object`, so all of OrdinaryDelete already exists for
it (§10.1.10 configurability preflight, `FLAG_TOMBSTONE`, count/tombstone
bookkeeping). The new `__carrier_bag_delete` finds the right bag and delegates.
**No delete semantics are re-implemented** — which is why a non-configurable
entry still refuses correctly, byte-identically to the `$Object` control.

**The half the design sketch did not predict.** The sketch flagged as a *risk*
that a companion-only tombstone *might* leave `__extern_get` answering from the
bag. It does, always — `__obj_find` **skips** tombstoned entries, so tombstoning
the companion simply restores the fall-through:

```js
a.q = 12;
Object.defineProperty(a, "q", { writable: true });   // now in BOTH tables
delete a.q;                                          // companion tombstoned
a.q                                                  // => 12   (measured)
```

So the vec arm now also **shadows the bag** after a successful companion delete.

**Tri-state on purpose**: `-1` not handled / `0` refused / `1` deleted.
Collapsing `-1` into `1` reproduces the original defect exactly — "I could not
see anything" becoming indistinguishable from "there was nothing". It is also
what makes the change **strictly additive**: the arm fires only for a key the
bag *demonstrably holds*, so every receiver/key the old code answered for keeps
its answer bit-for-bit.

## Scope, and why the −684 stratum cannot fire

Per #4010's ordering law (*visibility cannot ship before deletability*), **no
own-property visibility surface moves**: `hasOwnProperty` / `__object_hasOwn` /
`__vec_gopd` / `Object.keys` reach is unchanged, pinned by SCOPE PIN tests
(S1′'s two are untouched; three more added).

The **#2896 builtin-fn arm still runs FIRST and returns**, so `delete fn.name` /
`delete fn.length` never reach the new code — `built-ins/**/{name,length}.js`,
the ~700-file population that cost #4055 v1 **−684** host-free passes, is
untouched. Pinned by a committed regression guard **and** measured (below).

## Acceptance evidence — two independent derivations per cell

Every case is precondition-gated (a failed precondition returns a distinct code,
so it fails loudly instead of measuring nothing) and every compiled module is
asserted to have **zero imports**.

| cell | derivation | before | after |
| --- | --- | --- | --- |
| array named expando | value read | **STILL PRESENT** | genuinely absent |
| array named expando | **bag's own record** — attribute-only redefine re-seeds from the bag via S1′, never touching the read lane | bag still held `12` | bag empty |
| array, key in BOTH tables | value read | **STILL PRESENT** | genuinely absent |
| array, companion-only key | value read + gOPD | already correct | unchanged |
| function expando | value read | **STILL PRESENT** | genuinely absent |
| function expando | **closure bag's own record** — `__desc_has_own` via #4055's function-as-descriptor path, a different consumer entirely | bag still held `42` | bag empty |
| `$Object` **CONTROL** | value read · hasOwn · non-configurable refusal | 3/3 correct | 3/3 correct |

One derivation was never enough here: the delete's own **return value already
answered `true` on both arms while the value survived**.

## Corpus numbers (low, as predicted — with denominators)

- **74 / 74 files identical**, base vs. this branch, over the population most
  likely to be affected: test262 files performing a member `delete` whose
  receiver is statically an array or function (74 of the 797 member-delete files
  in the 43,488-file corpus). Zero regressions, zero gains.
- **`built-ins/**/{name,length}.js` stratum (1,240 files)**: see the comment
  below for the base-vs-branch diff.
- **Equivalence gate: no new regressions** (32 failing / 1,619 passing / 36
  known-failures in baseline; 4 pre-existing baseline failures now pass, all
  unrelated — i32 peephole, `Math.pow`, Symbol — so the shared baseline is
  deliberately **not** ratcheted here).
- **gc/host mode is byte-identical** (same sha256 on a mixed probe module
  exercising array/function/object deletes and for-in). Only standalone (+166 B)
  and wasi (+133 B) change.
- 5 pre-existing failures in adjacent delete tests (#1821, #2130 ×3, #2992)
  reproduce **identically on base** — not caused here.

## Found while measuring, deliberately NOT fixed

**`Reflect.deleteProperty` throws for every non-`$Object` receiver**, including
plain success cases. Verified **identical on base and after** — pre-existing and
independent of S2. The `delete` operator is unaffected. Fixing it would move a
surface this slice promised not to move; it wants its own issue.

## Structure

`src/codegen/vec-overlay.ts` **shrank** — its `__delete_property` arm moved into
`vec-bag-seed.ts`, which now owns both directions of the overlay↔bag seam
(seeding a value in, taking a property out). `object-runtime.ts` shrank by 19
lines. No `loc-budget-allow` grant was needed or taken.

Closes the S2 slice of #4010; unblocks #4098. `status:` stays `in-progress`
because S3 remains.

### DIFF
--- a/plan/issues/4010-m2-disjoint-receiver-side-tables.md
+++ b/plan/issues/4010-m2-disjoint-receiver-side-tables.md
@@ -1,10 +1,11 @@
 ---
 id: 4010
 title: "M2 — own properties on a non-$Object receiver live in TWO DISJOINT side tables that clobber each other; unify them"
-status: ready
+status: in-progress
 sprint: current
 created: 2026-08-01
-updated: 2026-08-01
+updated: 2026-08-03
+assignee: ttraenkler/dev-4010-s2
 priority: high
 horizon: l
 feasibility: medium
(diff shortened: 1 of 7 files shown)


ttraenkler:
## Corpus numbers — from the `merge_group` re-validation (the certifying gate)

The PR body promised these. The full-corpus measurement on the **merged state**
(run [30779916547](https://github.com/loopdive/js2/actions/runs/30779916547),
baseline content-current, 0 test262-relevant commits behind) supersedes any
local sweep:

```
Standalone guard: diff-test262 gate PASS (exit 0, authoritative — #3303):
  improvements=13, wasm-change regressions=0
  (compile_timeout flake=12, excluded), net=13.

Catastrophic guard: PASS — 7 raw wasm-change regressions vs baseline.
[standalone-highwater] FRESH: pass=27039, mark=27046, floor=26996.
```

**+13 improvements, 0 standalone wasm-change regressions, net +13** over
43,487 files. Low and positive — exactly what the slice plan predicted for
tombstones without visibility ("tombstones alone flip little, because
currently-invisible properties fail *earlier* in `propertyHelper`"). The payout
is S3's.

Critically for #4010's ordering law: **0 regressions** means the
`built-ins/**/{name,length}.js` stratum — the ~700-file population that cost
#4055 v1 **−684** — is untouched, as the design intended (the #2896 builtin-fn
arm runs first and returns, so `delete fn.name` / `delete fn.length` never reach
the new code).

### On the local sweeps, and why they are not quoted as the headline

Two local instruments were built and **both were refuted before being trusted**,
which is why the CI number is the one above:

1. A single long-lived process over the 797 member-delete files went from **0
   spurious throws in its first 20 files to 663 of 797 overall** — state
   accumulation, exactly `reference_long_single_process_sweep_overcounts`.
2. Rebuilt as one process per file, it reported **17 `compile_error` on the same
   24 files** the in-process path called **12 pass / 2 fail**. Two instruments
   disagreeing means neither is evidence.

What *did* survive, run in the regime where the instrument was observed clean
(20-file batches, fresh process each) and measured on **both** trees:

- **74 / 74 files identical**, base vs. branch, over the population most likely
  to be affected — files whose member-`delete` receiver is statically an array
  or function (74 of the 797 member-delete files). Zero regressions, zero gains.
- The `{name,length}` stratum sweep was still running when the queue merged; the
  `merge_group` covers it in full, so it is redundant.

PR #4064, opened 2026-08-03
perf(#4122): an unresolvable assignment is not a cross-domain one — method axis 3.6x, plus #4121/#4123

ttraenkler:
## Description

Started as an investigation into what is still ≥10x slower than node. Contains one fix, two issues, and one bug found on the way.

Closes #4122. Files #4121 and #4123.

---

## The fix — #4122

A `git bisect` of the cross-engine `method` axis landed on one line:

| commit | `method` js2/node |
|---|---:|
| `82c3ea98d` (parent) | 1.56 |
| **`9a91db042`** (first bad) | **5.26** |
| `d369562d7` (main) | 5.42 |

0.82 ms → 2.92 ms in a single commit (PR #4008, issue #3961). The defect is what `bindingHasMixedAssignmentCarrier` asks:

```ts
if (assignedTag === "mixed" || carrierDomain(assignedTag) !== initialDomain)
```

`"mixed"` is the oracle's answer for **cannot resolve**, not for *proven cross-domain*. Absence of evidence counted as evidence of mixing, so every numeric accumulator fed by a dynamically-dispatched call was demoted to a boxed carrier:

```js
var s = 0; for (…) { s = s + p.inc(); }
```

— the most common accumulator shape in ordinary JavaScript, paying ~300k `__box_number` allocations per bench call. The same file already read `"mixed"` the *other* way for an initializer, so the two sites disagreed about the same unknown.

**This is not a revert.** #3961's hazard — an i32 boolean slot destroying a later string assignment by coercing it to truthiness — is real, and three new tests pin it still boxing. An unresolvable assignment now gets a second question: does the whole-program fixpoint prove every definition of the slot numeric? Only if that also fails does the conservative demotion stand.

### It took three parts, not one

Consulting the fixpoint was the easy part. It couldn't prove the accumulator at all until two gaps closed — each verified by querying the verdict directly before and after:

1. **Self-reference.** `s = s + f()` is unprovable by a plain least fixpoint: proving `s` numeric requires `s` numeric. The grounded set now assumes the slot while judging its own definitions — the induction `withSelf` already gives the property path — then **re-checks groundedness with the assumption withdrawn**. So `var s = s + 1` (self-justifying, NaN in JS) is still rejected, and a mutual cycle `var a = b; var b = a` still cannot enter, because the assumption covers a slot's own name and never its partner's.

2. **`numericFunctions` on a non-`this` receiver.** It was consulted only for `this.m()`, though the verdict is whole-program and NAME-keyed — equally true of `p.inc()`, since a single non-numeric `inc` anywhere removes the name for both. Widened to bare-identifier receivers only, so member chains (`a.b.inc()`) and call results (`f().inc()`) keep the old answer.

   **This one genuinely widens a trust boundary** and is the part most worth reviewing: an `inc` that is not statically visible (a host/builtin method behind an opaque receiver) cannot demote the name. That exposure already existed for `this.m()`, whose receiver is equally unconstrained at runtime, and it is the same name-keying trade the file documents throughout — but it is a widening, not a no-op.

### Measured

Same container, interleaved, `chk=45000150000` on every leg:

| axis | before | after | node |
|---|---:|---:|---|
| method | 2.85–3.01 ms | **0.813 ms** | 0.49–0.60 |
| tokenizer (unchanged) | 0.253 | 0.257 | 0.17 |

**3.6x**, i.e. ~1.35–1.67x vs node from 4.7–6.2x. `$s` externref → f64, box 3 → 1, unbox 2 → **0**. Behind `JS2WASM_MIXED_CARRIER_NUMERIC=0`.

---

## #4121 — where the residual 10x actually is

The **micro-benchmark axes are at or near parity**; the ≥10x gaps are entirely in real npm packages (acorn 11.5x, cookie 10.6x, clsx 10.3x standalone runtime-dynamic).

Two published numbers don't mean what they look like, and both are recorded:
- the **`compile-time static` lane is not a comparison** — it reports acorn parsing a 226 KB bundle in 0.246 µs, which is constant folding, not execution;
- the **JS-host lane** (cookie 570x) is a different problem: **736 wasm→host calls** for one `parseCookie` on a 38-character header, identically on the second call.

Reduced to the smallest program that shows the standalone case:

| source | `$i` slot | box | unbox |
|---|---|---:|---:|
| `let i = 0; i = i + 1;` | **f64** | 0 | 0 |
| `let i = 0; i = s.indexOf(";");` | **externref** | 2 | 1 |

**A falsified hypothesis is recorded so nobody rebuilds it:** a box/unbox peephole finds nothing — of 59 box sites in the standalone cookie module, **0** are immediately re-unboxed. The round-trip is real but goes *through a local slot*, which is dataflow, not adjacency.

The issue targets the **IR**, not a fifth AST-side consumer: `IrType.union`, `box`/`unbox` instructions, `propagate.ts` and the pass pipeline all already exist, so the missing piece is the propagation between them — and the existing lowering would then emit nothing. It also records the blocking caveat that an IR pass only applies where the IR *claims* the function, which makes an IR-coverage pre-flight the first acceptance criterion and may block the issue on #2855.

---

## #4123 — a silent miscompile, found on the way

While building a guard test for #4122 I hit a wrong answer on **unmodified main**, verified by stashing all local changes:

```js
function K(){} K.prototype.k = function(){ return 3; };
function f(o){ return o.k(); }
export function main(){ return f(new K()); }   // → null, JS says 3
```

A prototype method on a **parameter** receiver returns `null`/`0`; the identical call with a *local* `new K()` receiver is correct. No trap, no diagnostic. Filed as #4123 with a six-row table and the receiver-flow starting point; being worked separately.

---

## Testing

- `tests/issue-4122-mixed-carrier-numeric.test.ts` — 13 tests. Both directions: the accumulator unboxes, and every genuinely cross-domain binding (`bool→string`, `num→string`, `num→bool`, and an unresolvable slot with *no* numeric proof) still boxes. Plus the self-reference cases: `var s = s + 1` and a mutual cycle both rejected.
- 28 tests green including the #3765 suite (no regression in the neighbouring analysis).
- `tsc --noEmit` clean. Gates pass: `check:issues`, `check:loc-budget`, `check:func-budget`, `check:oracle-ratchet`, `check:coercion-sites`, `check:any-box-sites`, `check:issue-spec-coverage`, `check:done-status-integrity`, `check:test-vacuity-shapes`, `lint`.
- `check:godfiles` fails — **verified pre-existing on untouched main** (`src/codegen/index.ts` mega-functions), not from this change.
- `loc-budget-allow` granted for `src/codegen/context/types.ts`: the verdict must reach `bindingHasMixedAssignmentCarrier`, which runs before any carrier type exists, so there is no subsystem module upstream of it to hold the field.

**Stated plainly:** my local full equivalence run had not finished when this PR was opened, so I am *not* claiming "no equivalence regressions" — CI's `equivalence-gate` is the authority. I will post the local result when it lands.

Issue ids came from `claim-issue.mjs --allocate`; the open-PR scan degraded (no `gh` auth in this container), so I checked the two open PRs via the API instead — #4062 claims #4118, no collision.

## CLA

### DIFF
--- a/plan/issues/4121-generic-carrier-unboxing.md
+++ b/plan/issues/4121-generic-carrier-unboxing.md
@@ -113,6 +113,23 @@ Case 4 in the table above (`var i = s.indexOf(";")`, declared `any`, so it *is* 
 candidate) still fails, so there is at least a second, independent gap in
 proving a string receiver for an unannotated parameter. Both need pinning.
 
+### The demotion has since been pinned to one line — see #4122
+
+An independent bisect of the `method` axis landed on the mechanism behind the
+table above: `bindingHasMixedAssignmentCarrier`
+(`src/codegen/analysis/mixed-assignment-carrier.ts`, wired at
+`src/codegen/statements/variables.ts:150`) demotes a binding to `externref` when
+`oracle.staticJsTypeOf` of any assignment is `"mixed"` — which is the oracle's
+answer for **unresolvable**, not for **proven cross-domain**. Absence of
+evidence is read as evidence of mixing.
+
+#4122 covers that narrow fix (a measured 3.5x regression on the `method` axis).
+It is the immediate first slice of this issue and should land independently.
+This issue remains the general problem: even once `"mixed"` is read correctly,
+the verdict is still consulted **per carrier kind**, so a numeric value flowing
+local → argument → parameter → return → field is re-boxed at every hop that has
+no bespoke pass yet.
+
 ## The structural problem: one analysis, four bespoke consumers
 
 `analyzeNumericPropertyNames` computes one whole-program verdict. It has been
--- a/plan/issues/4121-generic-carrier-unboxing.md
+++ b/plan/issues/4121-generic-carrier-unboxing.md
@@ -1,0 +1,223 @@
+---
+id: 4121
+title: "perf: generic carrier unboxing — one `any`-typed definition boxes an entire numeric local, and every carrier needs its own bespoke pass"
+status: ready
+sprint: current
+created: 2026-08-03
+updated: 2026-08-03
+priority: high
+horizon: l
+feasibility: hard
+reasoning_effort: high
+task_type: performance
+area: codegen
+language_feature: value-representation
+goal: performance
+related: [684, 3683, 3754, 3765, 4118, 2773, 1624]
+origin: "measured on upstream/main d369562d7, 2026-08-03, investigating the residual 10x vs node on real npm packages"
+---
+
+# #4121 — generic carrier unboxing
+
+## Where the remaining gap actually is
+
+Measured on `d369562d7`, all legs same container, checksums matching.
+
+The **micro-benchmark axes are at or near parity** — the boxing on those paths
+has already been eliminated by #3683 / #3754 / #3765:
+
+| axis      | node | js2  |        |
+| --------- | ---: | ---: | ------ |
+| numeric   | 2.42 | 2.40 | parity |
+| prop      | 1.14 | 1.08 | parity |
+| alloc     | 0.24 | 0.24 | parity |
+| tokenizer | 0.18 | 0.26 | 1.4x   |
+| string    | 0.15 | 0.45 | 3.1x   |
+| method    | 0.60 | 3.0  | 4.9x   |
+
+The ≥10x gaps are **entirely in real npm packages**:
+
+| lane                                  | vs node |
+| ------------------------------------- | ------: |
+| cookie · JS host                      |    570x |
+| acorn · JS host                       |    459x |
+| acorn · standalone · runtime-dynamic  |   11.5x |
+| cookie · standalone · runtime-dynamic |   10.6x |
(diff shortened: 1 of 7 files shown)


ttraenkler:
Local equivalence result, as promised in the PR body.

```
Test Files  12 failed | 202 passed (214)
     Tests  32 failed | 1619 passed | 3 todo (1654)
```

**No regressions.** The failing set is byte-identical to the baseline captured before this change — same 32 tests, same names, `diff` empty. Those 32 were separately A/B-verified by re-running all 12 failing files with the fix disabled: identical failing sets both ways, so none of them are attributable here.

Full output captured to file rather than piped through `tail`, and compared as a *set* rather than as a count — a matching count alone would not be evidence.

One thing I cannot fully explain and would rather state than omit: this run reports 214 files / 1654 tests against the baseline's 213 / 1646. All 8 extra tests pass, so the delta cannot hide a regression (a regression would surface as a new *failing* name, and there are none), but I have not tracked down where the extra file comes from.


---

PR #4068, opened 2026-08-03
fix(eval): close declaration collision compile errors

ttraenkler:
## Summary

- recover object-literal method closure fields from direct-eval-reified `externref` bindings before typed struct reads
- distinguish bare body bindings named `arguments` from executable uses of the implicit arguments object, keeping sync and async generators on native standalone paths
- record the measured EvalDeclarationInstantiation collision-cohort handoff in #2929

## Root cause

Direct eval reifies observable AOT bindings into shared `externref` cells. The object-literal method-value path still emitted a bare `struct.get`, producing invalid Wasm for reified receivers. After that was fixed, the remaining generator cases were refused because the generator admission gate counted `let arguments;` and `var arguments;` declaration names as arguments-object reads.

## Measured impact

This PR is stacked on #4013 while its merge-group checks finish; the repository retarget workflow will move it to `main` after #4013 merges.

The maintained standalone runtime-eval collision cohort moves from 154/198 with 44 compile errors to **198/198 (100%)**. Relative to the pre-slice 52/198 baseline, all 102 runtime failures and all 44 compile errors become passes, with zero pass-to-fail transitions.

## Validation

- Test262 collision cohort: 198/198
- focused eval tests: 29/29
- changed-root pre-commit gate: 121/121, including the real zero-import Acorn provider and interpreter/link canaries
- targeted native-generator slice: 60 passed, 3 skipped; its one unrelated mixed async-generator warning assertion reproduces on the exact stacked baseline
- typecheck, lint, formatting, oracle/coercion ratchets, LOC/function budgets, issue integrity, and dead-export audit pass


ttraenkler:
Closing as **superseded by `3ba70a5d7`** (landed via #4013).

This branch's fix commit `ad576a4dc` "fix(eval): close declaration collision compile errors" was re-applied onto the #4013 branch and landed on `main` as `3ba70a5d7` — same title, author, timestamp (Aug 3 05:07:58 +0200), same 8 files. Comparing both patches restricted to `src/codegen`, only blob hashes and hunk line offsets differ; the source changes are identical.

Verification:

- `git merge origin/main` into `codex/2929-eval-method-generator-ce` merges cleanly and produces a tree **byte-identical to `origin/main`** (`c3a33ca70` both sides). The PR contributes zero over `main`.
- `main` is in fact *ahead* of this head in `src/interp` (~153 lines this branch lacks).
- #4013 is merged (`b65d2f5ae`); base branch `codex/2929-direct-eval-capture` (tip `95a184b`) is an ancestor of `origin/main`.

Retargeting to `main` was attempted and refused by the API (`Changing a pull request's base branch is not permitted for this session type`), so this PR still points at the now-merged stack branch — which is why it reads `DIRTY`. Since the content has landed either way, closing rather than retargeting.

#4077 is unaffected: its base is already `main`, so it is no longer stacked on this PR.

---

PR #4074, opened 2026-08-03
fix(codegen): 12 compiler defects found compiling ESLint; 2 emit blockers remain (#4001, #4133, #4134 + 9 more)

ttraenkler:
## Description

Twelve distinct compiler defects found and fixed while trying to compile ESLint. Most are general — any minified npm bundle triggers several — rather than ESLint-specific. ESLint itself does **not** compile: two out-of-frame emit blockers remain, and the reason they remain is documented below rather than papered over.

&gt; This description has been corrected three times. See the [merge correction](https://github.com/loopdive/js2/pull/4074#issuecomment-5167592599), the [`null_deref` retraction](https://github.com/loopdive/js2/pull/4074#issuecomment-5169158622), and the [park resolution](https://github.com/loopdive/js2/pull/4074#issuecomment-5173456344).

### Status: unparked, awaiting CI

| | |
| --- | --- |
| `hold` label | **removed** — the park was a measurement artifact, fixed on `main` by #4141 |
| ESLint Tier 1a | **2** out-of-frame functions, no binary (out of scope here) |
| Merged with `main` | yes — `686d4cf3`, which carries #4141 |

This is a **working checkpoint**, not the finished ESLint goal. ESLint compiling end-to-end is deliberately not a merge condition; the point is to land twelve verified fixes before they drift further.

### The `null_deref` park — resolved

Three queue attempts parked this PR on a `null_deref` growth of ~+1,204, essentially all `built-ins/Temporal/**`. It was not a regression:

The baseline heal step in `test262-sharded.yml` ran **without** `TEST262_INCLUDE_PROPOSALS`, rewriting ~1,229 real Temporal traps as `skip`, while the never-healed candidate kept them as traps. Same files, two different scope filters — all of them charged to the candidate as new trap growth. #4088 was hit by the same phantom. `evaluateTrapCategoryGrowth` also had no exclusion for a baseline `skip`, so #3595's baseline-can't-testify rule never fired on them.

**#4141 fixed both halves on `main`** and names this PR by number.

Measured locally on `built-ins/Temporal/Duration` (540 tests):

| | compile_error | null_deref |
| --- | ---: | ---: |
| pristine `main` | 165 | 1 |
| this branch | 0 | 166 |

`main`'s 165 compile errors are `local index out of range — 5`, the exact defect this PR fixes; those files pull in `temporalHelpers.js` (1,285 lines, closures returned from functions), the shape #4134 hits. With this PR they compile and then null-deref on `Temporal` being unimplemented. Can't-compile → compiles-then-reports-missing-feature is strictly better; the ratchet bucketed both as "trap". Isolating #4018 and #4001 each changed the count by zero.

### ESLint pipeline movement

| Stage | Before | After |
| --- | --- | --- |
| Module resolution / front-end | ✅ | ✅ |
| Codegen | hard aborts | ✅ zero aborts |
| Binary emit | ❌ 14 out-of-frame functions | ❌ **2** |

The two survivors are `__fnctor_MurmurHash3_new` (worst index 24, frame 15) and `assertASTDidntChange` (worst 51, frame 4). An earlier revision claimed 1 and called it "the last blocker" — wrong on both counts.

### Fixes

- **#4001** — quadratic module-init over the source graph (`ModuleInitMode`). ESLint's 146-source graph never finished; it now completes. Module init was 51% of compile time.
- **#4018** — ambient `.d.ts` declarations produced a TDZ global (minimatch: 0 → 119,213 bytes).
- **#4019** — cycle guard for self-referential types.
- **#4028** — nested functions as imported direct-call targets.
- **#4029** — `importObject` dropped `string_constants`.
- **#4030** — internal compiler errors now carry a source location.
- **#4033** — duplicated callable role ordinal in the program ABI.
- **#4037** — multi-source `$ObjVecArr` reservation parity.
- **#4038** — JSDoc nameless parameter.
- **#4133** — cross-module function-name collisions: top-level (`funcMap` per-source rebinding) and nested (a nested declaration's bare name scoped to its enclosing function).
- **#4134** — out-of-frame local references: the out-of-frame capture guard.
- **#4027** — **credit correction:** `main` landed this independently during the drift. The merge takes main's version; the test file here is a regression guard on behaviour main already has, not evidence of anything this PR adds.

### Two were silent wrong answers, not crashes

They compile and validate cleanly — only the computed value detects them:

- A nested function calling a capturing sibling returned `2` where node gives `101` (#4134).
- A factory whose nested `equal` lost to another module's `equal` returned `0` where node gives `306` (#4133).

### The out-of-frame guard is now main's shared helper

`main` landed `captureSourceSlot` (`src/codegen/closures/capture-source-slot.ts`) independently. It is behaviourally identical to the inline guard this branch carried at both call sites — prefer the mapped slot when the name is one of this lifted function's own capture params, or when `outerLocalIdx` lands outside the frame entirely (where the historical read is provably invalid Wasm, so there is no behaviour to preserve). The merge takes the helper and drops the duplication.

A third site (`funcref-as-closure.ts`, the closure-**value** materialisation path) was tried and **measured ineffective** — `localMap` is empty there, so the guard falls through to the same index. That negative result is written up rather than left implied.

### The unresolved tension (#4133), stated plainly

Neither end is currently safe, and this PR picks the less-bad one:

| out-of-scope nested binding | consequence |
| --- | --- |
| suppress it | call falls to `ref.null.extern` → `null_deref` +1200 |
| let it through | wrong-frame capture index → module fails to validate |

The branch lets it through: an invalid module fails loudly. A correct fix makes the right callee *reachable* — the existing #2029 family-A promotion to `capturedGlobals` is the mechanism — which changes mutation semantics and needs a spec. That is also what the two remaining ESLint blockers need. Full write-up in `plan/issues/4134-*.md`.

### Verification

- New tests for #4001, #4018, #4027, #4033, #4038, #4133 (×2), #4134 — each checked against the unfixed base, all fail there with the stated values, so none is vacuous.
- On the merged tree: `tsc --noEmit` clean; 15/15 across `issue-4001`, `issue-4133` (×2), `issue-4134`.
- The 157 test files that fail anywhere in the suite, run on both sides: identical failure sets (510 failed / 1233 passed each), zero regressions.
- `uri.all.js`: 8 out-of-frame functions and no binary → **0 and a valid 131 KB module**.
- lodash `_createWrap` / `partial`: valid.
- `issue-165 &gt; IIFE with fewer arguments` fails, but fails identically on a pristine `origin/main` worktree — checked, not assumed.
- **No clean full-suite local run.** `--shard` does not partition as documented (one shard ran 1h36m) and the unsharded suite OOMs in a ~512 MB vitest worker. test262 and the full suite are CI's.

### Also true, and left alone

- **`check:godfiles` fails** — identically on pristine `main`, including entries this branch never touches. Not silenced from here.
- **`node-version: 25`** in the test262 shard matrix is a live spurious-park source whenever the runner manifest lags (it parked this PR once). Worth pinning; not this PR's change.

### Size-gate allowances

`loc-budget-allow:` / `func-budget-allow:` in `plan/issues/4134-*.md`, granted after verifying both gates pass on a pristine `main` worktree so the growth is genuinely this branch's. It is mostly env-gated instrumentation; extracting the remaining `index.ts` reporters is noted as follow-up.

### New instrumentation (env-gated, inert by default)

`JS2WASM_COMPILE_PROFILE`, `JS2WASM_CHECK_FRAMES`, `JS2WASM_FRAME_OPS`, `JS2WASM_FRAME_STAGES`, `JS2WASM_FRAME_TRAP`, `JS2WASM_TRACE_SLOT`.

## CLA


ttraenkler:
## Correction after merging `main` (449 commits of drift)

The PR description above was written before this branch caught up with `main`. Two claims in it no longer hold, and one part of the work turned out to be duplicate. Correcting rather than quietly editing, because the headline number changed.

### 1. ESLint no longer emits a binary on the merged tree

Before the merge: **7,705,310 bytes, 0 out-of-frame functions.**
After the merge: **no binary, 1 out-of-frame function** —

```
[js2:frames] position 1290 '__fnctor_MurmurHash3_new' frame=15 (2 params + 13 locals) worst local index=24
```

`__fnctor_*` is a construct `main` added during the drift; this branch never touches it. It is **not** caused by my re-added guard — measured directly by disabling that guard and re-running the full graph:

| | out-of-frame functions |
| --- | --- |
| merged tree, my guard ON | **1** (`__fnctor_MurmurHash3_new`) |
| merged tree, my guard OFF | **6** (the functor one **plus** 5 `__closure_N`) |

So the guard removes 5 of 6 and the functor breach is independent. The pipeline claim in the description should read **binary emit ❌ (1 function)**, not ✅, until that one is fixed.

### 2. Part of #4075 landed on `main` independently

`transitiveSiblingCaptures` and `liftedCaptureNames` are already on `main`, refactored into `preRegisterCapturingSibling`. That commit of mine is now duplicate work and the conflicting hunks take `main`'s version.

`main`'s call-site fix is **narrower** than mine, though: it prefers the mapped slot only for a lifted capture param, dropping the out-of-frame case. Without that case `uri.all.js` regressed straight back to 4 out-of-frame closures. Re-added as a second disjunct, restricted to out-of-frame so every in-frame call stays byte-identical.

### 3. Two issue ids collided with `main`

`main` had assigned 4045 and 4075 to unrelated issues. Renumbered to **#4133 / #4134** (issue files, test filenames, in-code references). The allocator's open-PR scan is degraded here (`gh` unauthenticated in this container), so I ran that scan by hand via the API — the five other open PRs top out at #4132 — and reserved with `--allow-unscanned`. Flagging it because the reservation carries `pr_scan=degraded`.

### Still true

- `uri.all.js`: 0 breaches, valid 131 KB module.
- The cross-module collision repro: 1306, node-correct (0 before).
- The three transitive-capture fixtures: 101 / 202 / 602.
- lodash `_createWrap` / `partial`: valid.
- `issue-165 > IIFE with fewer arguments` fails, but fails identically on a pristine `origin/main` worktree — checked, not assumed.

### CI

`quality` failed twice and is fixed forward: the dead-export gate (my #4033 assertion is referenced only from its test — an audit blind spot, baseline refreshed as the gate instructs), then the LOC and function budget gates (allowances granted in frontmatter; both gates verified passing on pristine `main` first, so the growth really is mine).

`check:godfiles` also fails here — but it fails **identically on pristine `main`**, including entries this branch never touches. Left alone deliberately; silencing it from this PR would claim a fix that isn't one.

---

ttraenkler:
## The `null_deref` park is very likely this PR — fix is ready but cannot be pushed while queued

**For the other five PRs in that batch** (#4001, #4029, #4033, #4045, #4075): you are probably parked on my defect, not yours. Details so you can decide whether to wait or re-attribute.

### The gate

```
GATE FAIL: trap category "null_deref" grew 156 → 1357 (+1201) — uncatchable-trap ratchet (#3189)
```

Net conformance on the same run was **+18** (55 improvements vs 37 regressions), so this is not broad breakage — it is one trap category exploding, almost entirely `Temporal/**`.

### Why it points here

#4133 in this PR makes a nested function declaration's bare name invisible outside its enclosing function. The first cut implemented that by setting `isLocallyShadowed`, which suppresses **both** the closure path and the `funcMap` path. When the name then had no other binding, the call fell through to the graceful `ref.null.extern` fallback — so a call that previously reached *some* function now yields `null`, and the next use traps. That is precisely a `null_deref` explosion.

I am **not** claiming proof. The group is batched and the failure is attributed to the group as a whole. What I can say is that this branch contained a mechanism producing exactly this signature, and I have removed it.

### The fix (commit ready, unpushed)

The suppression now only ever *prefers* a better binding:

- out-of-scope name → closure/local paths run **first** (this is what finds the correctly-scoped callee);
- if neither resolves it → fall through to the historical `funcMap` path rather than suppressing it.

Reaching the wrong same-named function is a wrong answer; turning a resolvable call into an uncatchable trap is worse — and neither licenses the other.

Re-verified locally after the narrowing: cross-module collision repro node-correct (1306), `uri.all.js` 0 breaches / valid, lodash `_createWrap` valid, the #4134 fixtures 101/202, all 8 new test files green.

### Why it is not pushed yet

The branch is back in the merge queue, and GitHub refuses updates to a queued branch:

```
remote: A pull request for this branch has been added to a merge queue.
remote: Branches that are queued for merging cannot be updated.
```

Head is still `2eebc317` — i.e. **the queue is currently re-running the version without the fix**, so it should fail the same gate again.

I am deliberately **not** dequeuing: that rebuilds the in-flight group and cancels the runs for the five PRs batched with this one. I will push the moment it is ejected. If someone wants it sooner, dequeue #4074 alone and I will push immediately.

---

ttraenkler:
Fix pushed — head is now `76a7652d`, and I have removed the `hold` label on this PR only.

Per the auto-park bot's own instruction ("fix the failure and remove the `hold` label"; "re-enqueue them ONE AT A TIME to attribute the failure"), this PR goes back in alone so the next `merge_group` run attributes the `null_deref` growth cleanly. I am not touching the `hold` on the other five PRs from that batch — if this one now passes, that is your evidence the regression was mine and yours can be released.

I did not enqueue anything; `auto-enqueue.yml` picks it up once the PR is `CLEAN`.

---

ttraenkler:
## This park is SPURIOUS — infra, not a regression. Confirmed against the run.

The bot's own footnote applies here ("if it is a setup/infra step rather than a verdict step, the verdict never ran"). Checked the cited job:

```
with:
  node-version: 25
Attempting to download 25...
##[end-action id=__self.__actions_setup-node;outcome=failure;conclusion=failure;duration_ms=1739]
```

`actions/setup-node@v6` could not fetch Node 25 — a transient distribution/manifest failure. The shard then ran **zero** tests, which the artifact step confirms:

```
##[warning]No files were found with the provided path:
  benchmarks/results/test262-standalone-results-...-mgchunk8.jsonl
```

`merge shard reports — Fail if required test262 shards did not succeed` is a pure cascade from that missing report, not an independent finding.

### The signal that matters

`check for test262 regressions — Fail on regressions` is **not** in this failure list. It was the failing gate on the previous park (`null_deref` 156 → 1357). So the fix in `76a7652d` did what it was meant to: the regression verdict now passes, and what remains is a Node-download flake.

### Action taken

Removed `hold` from **this PR only**, on a confirmed-flake determination (per rule (c): flake/collateral may be re-admitted; a real regression may not). This is the single re-admission for this flake — I am not looping.

### For the other six in the batch

**#4001, #4029, #4033, #4045, #4075, #4133** — you were parked by the same grouped failure, and it is the same Node-25 setup flake. Your holds are yours to remove; I have not touched them. Worth confirming against your own cited runs first, but if the failing step is `Setup Node and pnpm (cached)`, there is no verdict behind it.

Also worth someone's attention independently: `node-version: 25` in the test262 shard matrix is a live flake source whenever the runner manifest lags. Pinning to a version that is reliably in the manifest would stop this recurring — not something I am changing from this PR.

---

ttraenkler:
## Correction: my `null_deref` fix did NOT resolve this, and I said otherwise on bad evidence

**Retracting** the claim in [my earlier comment](https://github.com/loopdive/js2/pull/4074#issuecomment-5168676828) that the fix in `76a7652d` resolved the `null_deref` growth. I inferred that from the absence of `check for test262 regressions` in the *flake* run's failure list — but in that run the verdict step never executed at all (Node 25 failed to download). Absence there was not evidence of a pass, and I should not have read it as one.

The verdict has now actually run:

| Park | My PR's code | `null_deref` |
| --- | --- | --- |
| 1st | `2eebc317` (no fix) | 156 → 1357 (**+1201**) |
| 2nd | — | infra flake, verdict never ran |
| 3rd | `76a7652d` (with fix) | 156 → 1360 (**+1204**) |

### What that changes

A delta of **+3 across two materially different states of this PR** is not what a causal fix looks like. Combined with both parks being 6–7 PR batches over the same membership, the evidence now points **away** from this PR being the cause. My original attribution was a hypothesis that fit the signature; it has not survived测 measurement.

I am **not** removing the `hold` this time. This is a verdict-step failure on a real regression, and rule (c) is explicit that a real regression must be fixed before re-admission — removing it now would be re-admitting on nothing more than a hunch that it is someone else's.

### What is actually needed

Attribution requires this PR to go through the queue **alone**. All three attempts were batched, so no run has yet isolated any single member. Someone with queue control should either enqueue one member at a time, or hold the other six while this one goes through. I can't do that from here without cancelling in-flight runs for the other PRs, which I've deliberately avoided throughout.

The `null_deref` growth is almost entirely `Temporal/**` with `baseline: skip` — i.e. tests that previously didn't run reaching a trap instead. That shape suggests something changed which global/undefined resolution does, and it is worth checking against the other batch members' diffs, not just mine.

---

ttraenkler:
## The `null_deref` park was a CI measurement artifact — attributed, fixed, and merged. No solo queue run is needed.

Your description lists "a solo (unbatched) queue run to attribute the regression" as what this PR needs. That blocker is gone: **#4141** diagnosed the phantom, names this PR by number, and its fix is on `main`.

### Root cause

`scripts/heal-poison-rows.ts` re-runs poison rows, and for `built-ins/Temporal/**` that hits `classifyTestScope` → **proposal** → skipped unless `TEST262_INCLUDE_PROPOSALS=1`. The **shard** jobs set it; the two **heal** steps did not. Healing runs on the *baseline* path only, so ~1,229 Temporal rows the shard had measured as real traps were rewritten to `status: "skip"` in the baseline while the candidate kept them as traps. `diff-test262.ts` did not exclude `skip` from the "baseline can't testify" set, so the whole asymmetry landed as trap growth on whoever was in the group.

The tell is a healed row carrying `compile_ms: 4827` while claiming to be skipped.

### Why this is your PR's exoneration, not a hunch

- The delta is **byte-identical** on two unrelated branches in different lanes — this PR and #4088 — so it is a property of the comparison, not of either diff: `skip 1322 → 115 (−1207)`, `null_deref 156 → 1360 (+1204)`, zero `pass → trap`.
- A codegen change **cannot un-skip 1,200 tests**. Skipping is a pre-compile scope decision; no compiler diff can reach it.
- Your own +3-across-two-states observation was right, and the reason is now mechanical rather than inferential.
- **#4088, the co-victim of the identical phantom, has since merged** (21:42).

Verified on `main`: `TEST262_INCLUDE_PROPOSALS` is now set on both heal steps (`test262-sharded.yml` L2356 and L3201), and `skip` is excluded from the trap-growth comparison.

Your retraction of the earlier "the fix resolved it" claim was the correct call on the evidence you had — the verdict step genuinely had not run in that flake.

### Possibly relevant to a remaining blocker

#4088 also landed `2ef595b7`, *"fnctor twin sibling captures from the declaring frame's dead slots"*, cited against `__fnctor_Parser_new`. One of your two survivors is `__fnctor_MurmurHash3_new` — the same shape. Worth merging current `main` and re-measuring the Tier 1a frame check before treating that one as outstanding; I have not verified it applies.

The `hold` is left in place — your description says this PR is not ready to merge on its own terms (2 out-of-frame emit blockers), and that is your call, not the park's.

---

ttraenkler:
## Attributed locally with measurements. The `null_deref` growth is this PR — and it is the fix *working*.

Ran `built-ins/Temporal/Duration` (540 tests) three ways in this container, same filter, same worker count. Correcting my [earlier retraction](https://github.com/loopdive/js2/pull/4074#issuecomment-5169158622), which guessed from the batch delta that this PR was not the cause:

| source tree | `null_deref` | `compile_error` | pass |
| --- | --- | --- | --- |
| this branch | **166** | 0 | 56 |
| this branch minus #4018 | 166 | 0 | 56 |
| this branch minus #4001 | 166 | 0 | 56 |
| pristine `main` src | 1 | **165** | 56 |

Everything else is identical across all four (129 `Temporal is not defined`, 33, 31, …).

### What main's 165 `compile_error`s actually are

```
Binary emit error: RangeError: Codegen error:
  local index out of range — 5 (valid: [0, 5)) at function '__anon_6_ch…'
```

That is **exactly the #4134 out-of-frame defect this PR fixes.** So the causal chain is:

1. On `main` these 165 files emit invalid Wasm and never instantiate.
2. This PR fixes the out-of-frame emit, so they compile.
3. They then reach a **different, pre-existing** gap — Temporal is not implemented — which surfaces as an uncatchable null-deref in `__module_init` rather than a catchable `ReferenceError`.

Dropping a commit cannot remove this. Reverting the fix just restores the compile error, which is why isolating #4018 and #4001 changed nothing.

### The gate's own rationale already covers this — it just doesn't fire

`scripts/diff-test262.ts` (#3595), verbatim:

> `compile_error` is the SAME class of baseline-can't-testify: an invalid-Wasm module never instantiated, so `__module_init` never ran and never had the opportunity to trap. A later trap on that file is therefore *unknown*, not *introduced*.

That is this situation precisely. It does not fire because the exclusion tests the **published baseline** row, which for these files says `skip` — Temporal is a proposal, excluded from the baseline run, while the `merge_group` runs with `TEST262_INCLUDE_PROPOSALS=1`. So the baseline never ran them *at all*, which is a stronger "cannot testify" than `compile_error`, yet `skip` is not in the exclusion:

```ts
if (base?.status === "compile_timeout" || base?.status === "compile_error") {
  unknownBaselineTimeouts[row.error_category].push(file);
  continue;
}
```

### What I am NOT doing

I am not adding `skip` to that condition from this PR. Editing a CI safety gate inside the PR the gate is blocking is self-exempting on its face, even when the argument is right — it needs to be a separate, reviewable change. I am also not declaring a `trap-growth-allow` for 1,204 named tests: the #3596 valve is for bounded, enumerated fail→trap reclassifications, and using it at that scale would be laundering a category-wide gate gap through a per-test escape hatch.

### Options, for whoever owns the queue

1. **Separate one-line PR** adding `skip` to the #3595 exclusion, with this measurement as the justification. Cleanest; unblocks this PR and every future one that makes a skipped-baseline file compile.
2. **Re-baseline with proposals included** so `skip` rows stop appearing where the merge_group runs them.
3. **Accept and valve** — `trap-growth-allow` with count 1204. Mechanical, but I'd argue against it for the reason above.

Net conformance on the parked run was **+18** (55 improvements vs 37 regressions); the only failing gate is this trap ratchet.

---

ttraenkler:
## Park-hold resolved by #4141 — removing the `hold`

Per the auto-park rules, diagnosing the cited run before touching the label. The finding is that this PR's park was a **measurement artifact, not a regression** — and #4141 landed the fix on `main` independently, naming this PR by number.

### What the gate saw

`null_deref` growth of ~+1,204, essentially all `built-ins/Temporal/**`.

### What was actually happening

The baseline heal step in `test262-sharded.yml` ran **without** `TEST262_INCLUDE_PROPOSALS`, so it rewrote ~1,229 real Temporal traps as `skip`. The candidate side was never healed and kept them as traps. Same files, two different scope filters — so every one of them was charged to the candidate as newly-introduced trap growth. #4088 was hit by the same phantom.

`evaluateTrapCategoryGrowth` also had no exclusion for a baseline `skip`, so #3595's baseline-can't-testify rule (which covers `compile_error`/`compile_timeout`) never fired on them.

### That this PR is an improvement, not a regression

Measured locally on `built-ins/Temporal/Duration` (540 tests):

| | compile_error | null_deref |
|---|---:|---:|
| pristine `main` | 165 | 1 |
| this branch | 0 | 166 |

`main`'s 165 compile errors are `local index out of range — 5` — precisely the defect this PR fixes. Those files pull in `temporalHelpers.js` (1,285 lines, closures returned from functions), which is the exact shape the #4134 nested-capture bug hits. With this PR they compile and then null-deref for the honest reason: `Temporal` is not implemented. Can't-compile → compiles-then-reports-missing-feature is strictly better; the ratchet just bucketed both as "trap".

Isolating #4018 and #4001 each changed the count by zero, confirming it is not removable by dropping a commit.

### Fix, and what changed here

#4141 fixed both halves on `main`: the healer asymmetry in the workflow, and the missing baseline-`skip` exclusion in `evaluateTrapCategoryGrowth`. I had a duplicate one-liner in flight and discarded it.

`main` is now merged in (`a454d6ed`), which also resolves two conflicts in favour of main's new shared helper `captureSourceSlot` (`src/codegen/closures/capture-source-slot.ts`) — behaviourally identical to the inline out-of-frame guard this branch carried at both sites, so the duplication is gone.

Verified on the merged tree: `tsc --noEmit` clean; 15/15 across `issue-4001`, `issue-4133` (×2), `issue-4134`.

### Still open, deliberately

Two ESLint binary-emit blockers remain (`__fnctor_MurmurHash3_new`, `assertASTDidntChange`). They need capture **reachability** — promotion to `capturedGlobals`, the #2029 family-A mechanism — not re-indexing, and are out of scope here. This PR is the working checkpoint, not the finished ESLint goal.

---

PR #4091, opened 2026-08-03
fix(#4010): S3 — own-property visibility over the carrier bags; the -684 isolated

ttraenkler:
Closes #4010 — the final slice (S1′ #4058 · S2 #4063 · **S3, this PR**).

Routes `hasOwnProperty` / `Object.hasOwn` / `in` / `getOwnPropertyDescriptor` /
`Object.keys` / for-in / `getOwnPropertyNames` / `propertyIsEnumerable` through
the per-receiver own-property store for **array** (#3537 bag) and **function**
(#3468 closure bag) receivers.

## ⛔ The −684 was never the query widening — the mechanism is isolated here

`carrier-bag-hasown.ts` records that #4055 measured three candidate mechanisms
for the −684 and **none reproduced outside the full harness assembly**. S3 had to
widen the same surface, so it had to be found. It reproduces in six lines,
standalone, host-free, on the pre-S3 tree:

```js
const f = Array.prototype.push;
f.name = "unlikelyValue";   // REFUSED by the read path (#2896 meta arm wins)
f.name;                     // => "push"           — looks correct
delete f.name;              // #2896 arm clears the meta bit
f.name;                     // => "unlikelyValue"  ← THE BAG KEPT IT
```

`__extern_set` had no builtin-fn arm, so a write to a builtin's **non-writable**
`name`/`length` was deposited in the closure bag and sat there invisibly,
shadowed by the read path. `propertyHelper.js` performs exactly that write, in
exactly the wrong order:

| `verifyProperty` line | what runs |
| --- | --- |
| 113 → `isWritable` | `obj[name] = "unlikelyValue"` → **pollutes the bag** |
| 120 → `isConfigurable` | `delete obj[name]; return !__hasOwnProperty(obj, name)` |

A bag-aware `hasOwnProperty` then answers `true`, `isConfigurable` returns
`false`, and every file asserting `configurable: true` fails with *"descriptor
should be configurable"* — the 696. **The pollution was already on `main`;
#4055 v1 only made it observable**, and any future widening of any own-property
query would have hit it again.

Fixed at the **source**, not the query: §10.1.9 OrdinarySet over an existing
non-writable own data property is a no-op, so the non-`$Object` set arm now
refuses a key the #2896 metadata still owns. The refusal is scoped to **live**
metadata — after `delete fn.name` an assignment lands again, as the spec
requires. Three committed cases pin all three states.

## Mandatory stratum control — `built-ins/**/{name,length}.js`

Run explicitly, per #4010's non-negotiable S3 gate. Results are posted as a
follow-up comment on this PR; **the PR carries `hold` until they are in**, so
`auto-enqueue` cannot take it early.

## What shipped

`src/codegen/carrier-bag-visibility.ts` — four natives over both bags (LOOKUP,
never ensure):

| native | answer | "not handled" is |
| --- | --- | --- |
| `__carrier_bag_of(o)` | the bag, `ref.test $Object`-screened | null |
| `__carrier_bag_has(o,k)` | `__obj_find(bag,k) != null` | 0 |
| `__carrier_bag_gopd(o,k)` | `__getOwnPropertyDescriptor(bag,k)` | **null, kept distinct from "absent"** |
| `__carrier_bag_push_keys(o,vec,nonEnum)` | `__obj_ordered{,_all}(bag)` keys | 0 |

Wired at **seven sites, each one a place whose previous answer was the literal
"absent" constant**. Composition is #4055 v2's throughout — existing helper
first, bag only on a miss — so this can add a `true` and never override one. The
#2896 arms still run first and return, so `{name,length}` on a builtin never
reaches the bag. Tombstones are free: `__obj_find` / `__obj_ordered` /
`__obj_ordered_all` all skip them.

### The one place S2 and S3 could have silently disagreed

A bag-aware `__vec_gopd` lets a **bag-only** key reach S2's companion-delete arm,
where `__delete_property(comp, k)` answers **1 (absent ⇒ success)** while
`bagShadow`'s own refusal is discarded — a loud success over a possibly-surviving
value, i.e. precisely the defect S2 existed to remove. `vec-bag-seed.ts` now
routes a key the companion does not own back to the tri-state bag delete. S2's
own "key held by BOTH tables" case still passes.

## Matrix cells flipped (24/24 probe, standalone, zero imports asserted per module)

| receiver | read | hasOwn | `in` | gOPD | keys | delete | defineProp→read |
| --- | --- | --- | --- | --- | --- | --- | --- |
| **`$Object` (CONTROL)** | ok | ok | ok | ok | ok | ok | ok |
| array | ok | ABSENT→**ok** | ABSENT→**ok** | ABSENT→**ok** | ABSENT→**ok** | ok (S2) | ABSENT→**ok** |
| function | ok | ABSENT→**ok** | ABSENT→**ok** | ABSENT→**ok** | ABSENT→**ok** | ok (S2) | ok |
| Date / RegExp / Error / class instance | unchanged — **no bag exists**, `__carrier_bag_of` answers null |

Screens that did **not** move, each with a committed guard:
`Object.keys(new Date(0))` and `Object.keys(/ab/)` stay `[]` (#4071's −5);
builtin `name`/`length` reflection byte-unchanged; `Object.defineProperty` on a
function still lands nowhere **and gOPD still reports it absent** — no new lie;
#4055's function-as-descriptor path unchanged.

`tests/issue-4010.test.ts`: **47 cases**. S1′/S2's **five SCOPE PINs are flipped
deliberately** — that is what they existed to force — and kept, not deleted.

## No `loc-budget-allow` / `func-budget-allow` taken

The first cut grew three god-files and the gate refused it. Restructured instead:
`fillVecHasOwnHelpers` **moved** out of `vec-overlay.ts` into `vec-bag-seed.ts`
(which already owns both directions of the overlay↔bag seam), the builtin-fn
refusal sited in `vec-props.ts`, every god-file call site reduced to one line
calling a composite builder with the prose in the subsystem module. Final:
`object-runtime.ts` −3, `vec-overlay.ts` −58, `object-runtime-descriptors.ts` −1
against baseline; both budget gates green.

## Instrument note for anyone measuring standalone locally

`runTest262File` does **not** supply the separately compiled
`js2wasm:runtime-eval` provider the CI worker links, so any standalone module
carrying those imports dies at `WebAssembly.instantiate`. Measured trigger:
`propertyHelper.js` in the assembly (prefix alone = 0 imports; prefix +
propertyHelper = 2). A pilot of the stratum scored **0 / 11 runnable** before
this was found — the instrument was blind, not the compiler broken. Neutralising
`$262.evalScript`'s `eval` in a *local* copy of `scripts/test262-fyi-runtime.js`
restores it; the repaired instrument agrees with the published CI standalone
baseline **16/16**. That local edit is **not** part of this PR. This is not a
defect on `main`.

## Follow-ups (not in scope here)

- **#4098 is unblocked but not solved**: its 124-file population is *class
  instances*, which have no bag at all (matrix row 7). What this PR gives it is
  the worked shape, the additive wiring points, and the −684 lesson.
- #4006 / #4007 remain deliberately unfunded symptoms of this substrate.
- #4129 (`Reflect.deleteProperty` throws for every non-`$Object` receiver) is
  pre-existing and untouched.


ttraenkler:
`hold` set by the author: #4010's S3 gate requires the entire `built-ins/**/{name,length}.js` stratum to be run explicitly **before this PR goes near the queue**. The run is in flight; I will post the numbers and remove the label myself. This is an author hold, not an `auto-park-bot` regression flag.

ttraenkler:
## ✅ Mandatory stratum control — `built-ins/**/{name,length}.js` — **729/729, zero regressions**

Run explicitly before this PR goes near the queue, per #4010's non-negotiable S3 gate. Removing `hold` on the strength of this.

### Population — complete, not sampled

The glob is **1,240** files; **729** of them pass in the published CI standalone baseline. A file that already fails cannot regress, so those 729 are the **complete at-risk set** — the ~700-file population that cost #4055 v1 −684.

| | files |
| --- | ---: |
| `built-ins/**/{name,length}.js` (whole glob) | 1,240 |
| …passing in the CI standalone baseline ⇒ **at risk** | **729** |
| rows measured in the post arm (floor check) | **729 / 729** |
| pass | 724 |
| compile **timeout** (60–74 s vs a 60 s limit) | 5 |
| **regressions** | **0** |

All five timeouts re-run **SOLO**: **5/5 pass in 9–17 s**, versus 60–74 s under three-shard contention at load 13–16. Load artifacts of the instrument — none carries an assertion signature, let alone *"descriptor should be configurable"*.

**Be precise about the arms.** The base arm is the **published CI standalone baseline** (generated 13:19Z the same day, from the commit this branch forked from), **not** a fresh local 729-file base sweep — that would have cost another ~3 h of wall clock for files that provably did not move. What makes it sound: the post arm moved **nowhere**, so there was nothing to attribute; every post-arm non-pass got a fresh same-instrument solo re-run; and the local instrument was calibrated **16/16** pass/not-pass against that same baseline before use. Where a real flip *did* exist — the four gains — a full two-arm same-instrument comparison was run.

## Payout — funnel with denominators

| stage | files |
| --- | ---: |
| standalone baseline rows | 48,619 |
| `status = fail` | 14,859 |
| …own-property / descriptor error signature | 1,111 |
| …unreadable source (floored) | **0** |
| …static array/function-expando + reflection shape | **11** |

Ran **121** files = the **complete** 11-file shape-candidate set ∪ a deterministic stride sample of 112 from the 1,111 bucket.

### 4 gains, 0 regressions

- `built-ins/Object/freeze/15.2.3.9-2-a-7.js`
- `built-ins/Object/freeze/15.2.3.9-2-a-9.js`
- `built-ins/Object/seal/object-seal-p-is-own-property-of-a-function-object-that-uses-object-s-get-own-property.js`
- `built-ins/Object/seal/object-seal-p-is-own-property-of-an-arguments-object-which-implements-its-own-get-own-property.js`

**Attribution by kill-switch REMOVAL**, not narrative: all four re-run with the eight changed codegen files restored to their `609c995ce` contents (file copies — `git stash` is a single **shared** stack across every worktree of this repo and is never safe here) → **0/4 pass**. With S3 → **4/4**.

### The informative half is the 0/110

The random-stride sample of the broad own-property/descriptor bucket moved **not at all** (0/110), while the statically derived at-risk set moved **4/11**. That is #4010's capability matrix speaking: the bucket is dominated by receivers that have **no bag** — class instances, Date, RegExp, Error — i.e. exactly #4098's population, and exactly what S3 does not claim to fix. A low flip count is the expected outcome for substrate work (#4084 precedent, and this issue's own S1′ note), not a shortfall.

### Local suite

`tests/issue-4010.test.ts` — **47/47**, including the five deliberately flipped SCOPE PINs, the three −684 mechanism cases, and the #4071/#2896/#4055 guards. Both budget gates green with **no allowance taken**.

PR #4106, opened 2026-08-04
perf(#4150) + fix(#4139/#4144/#4149): benchmark parity work, and the two acorn-UMD fixes PR #4088 did not carry

ttraenkler:
## Description

Two independent bodies of work, both on this branch.

**Note on the branch.** PR #4088 used this same branch name and is merged — but it merged at `77f080d0`, the *three*-defect state, while its own description listed five. The fnctor-twin fix (`2ef595b7`) and the stack-balance fix (`659c0bf9`) were never on `main`. This is a **new PR**, not a reopening of #4088; it carries those two plus everything since.

---

### Part 1 — #4150 benchmark parity

Three optimizations, each behaviour-preserving, each with a committed correctness test.

| commit | change |
|---|---|
| `5b60fec0` | **Integral fast path in `__fmod`.** Every JS `%` ran a binary long-division loop (~26 f64 iterations for `19998 % 3`). When both operands are whole numbers in i32 range the answer is `i32.rem_s`. The guard `f64.convert_i32_s(i32.trunc_sat_f64_s(v)) == v` is exact, not a heuristic — NaN, ±Inf, fractional and out-of-i32-range all saturate to a finite i32 that fails the compare, so everything the slow path handles specially still reaches it. A trailing `copysign` supplies the two results i32 cannot represent: `-6 % 3` and `-0 % 3` are `-0` (§6.1.6.1.6), not `+0`. `INT_MIN % -1` (which would trap `i32.rem_s`) is excluded. |
| `a41115bc` | **Host-import wrappers.** The DOM gaps were not codegen — the emitted wasm is already one monomorphic import per member with string literals hoisted to wasm globals and exactly one crossing per DOM op. The cost was two stacked JS wrappers each allocating a rest-args array per crossing (7,000 in one `dom/set-attributes` call). Arity-specializes the universal depth guard (arity from the **wasm import signature**, newly on `ImportDescriptor.paramCount` — the host fn's `.length` excludes rest/defaulted params and would silently drop args), takes a direct property access for `extern_class` get/set, and inlines the method shim's ordinary case per arity. |
| `e9667a95` | **Single-pass `__str_split`.** It scanned every character twice (count pass + fill pass). #3901 chose that to size the array exactly; but the vec's field 0 is the *logical* length, separate from backing-array size, so slack is the normal growable-vec representation and free, while rescanning is not. |

`9b2e5424` promotes the two development harnesses into `tests/`. They are **correctness guards, not presence detectors** — they pass on the pre-optimization code too, which is the point: each optimization must be behaviour-preserving, so the test's job is to catch a *broken* one, not a *missing* one. Verified to have teeth by mutation (reporting the split result's capacity as its length, and dropping the doubling, are each caught).

#### On the measurements — please read before quoting a number

Wasm-side medians, one box, same session, back to back:

| benchmark | before | after |
|---|---:|---:|
| `array/map-filter` | 0.767 ms | 0.176–0.204 ms |
| `mixed/csv-parse` | 1.89 ms | 1.23–1.35 ms |
| `dom/set-attributes` | 0.623 ms | 0.399–0.446 ms |
| `dom/read-attributes` | 0.351 ms | 0.245–0.247 ms |
| `dom/modify-text` | 0.293 ms | 0.213–0.225 ms |

**Ratios against node are deliberately omitted.** This machine's absolute numbers run ~1.5–2× the published benchmark environment's, and the node baselines here are bimodal — V8 hoists the loop-invariant work in some runs and not others. `mixed/csv-parse`'s node side swings 0.34–1.39 ms across recent history and gc-native has *beaten* node in 18 of 220 historical runs, so any single-run ratio for it is noise. Compare `benchmarks/results/history.json` medians, not one run. This is the failure mode #4150 §2 and §5 describe.

`string/case-convert` was looked at and **not** fixed: the gc-native lane is already fully native (`fast: true`, zero imports) and `__str_toLowerCase` is a minimal single pass with an inline ASCII branch. Scaling probe puts it at ~64 ns fixed + ~3.9 ns/char, i.e. ~155 ns/call vs V8's ~86 ns on a 23-char string. The gap is two required GC allocations plus per-char work against a bulk-copy intrinsic; WasmGC `(array i16)` offers no multi-element load, so there is no word-at-a-time path. Near the structural floor.

---

### Part 2 — acorn UMD chain (#4139, #4144, #4149)

| commit | defect | fix |
|---|---|---|
| `2ef595b7` | fnctor twin sibling captures sourced from the declaring frame's dead slots (`__fnctor_Parser_new`) — **#4139** | twin prologue casts the standalone `__constructor_identity` param (already the ctor's closure value) and spills capture fields into frame locals — `fnctor-twin-captures.ts` |
| `659c0bf9` | stack-balance walker typed `local.tee` as pass-through; wasm types a tee's result as the **local's declared type** — **#4144** | tee arm pops and pushes `localTypes[idx]` |
| `53f6c6f4` | a property-function written through one alias of an object was unreachable through another — silently `null`, no throw — **#4149** | a `{}`-typed binding was pinned to a zero-field closed struct while the runtime value is a native `$Object`, so the guarded cast nulled every alias; empty object shapes now resolve to `externref`, and a member write whose resolved struct lacks the field routes to the dynamic terminal instead of being dropped |

`#4149` is the CommonJS/UMD wrapper shape (`exports.f = fn` … `m.exports.f()`) and was defect 6 in the #4145 chain. 17 repro variants across both lanes; acorn 8.18 UMD in-factory probes (`getOptions`, `wordsRegexp`, `new Parser`, `nextToken`, `parse`) all work on standalone and the tiny-parse returns 1.

Still open, filed not fixed: the full 233 KB self-parse on the UMD build throws further down the chain, and the gc-lane UMD still fails to compile at `__fnctor_Parser_new` (gc twins have no `__constructor_identity` param — the #4139 gc half).

---

### Validation

- Full equivalence suite in three batches (212 files), plus the split-exercising set and a scoped 54-file re-run after merging `main`. **Every** failure — `arguments-nested-and-loops`, `array-inline-return`, `delete-sentinel`, `logical-conditional-identity`, `tdz-reference-error`, `reflect-api`, `optional-direct-closure-call`, `new-non-constructor`, `issue-1472`, `issue-2099`, and all four `dom-containment` — reproduces with identical counts at the pre-change commit.
- `__fmod`: edge cases (signed zeros both positions, NaN/±Inf both positions, i32 boundaries incl. the `INT_MIN % -1` trap, #2056 ULP-drift repros) + seeded 20k-pair sweep, compared to the host's `%` with `Object.is` so `-0` is a real assertion.
- `__str_split`: checksum over piece count, piece length and first code unit across separators × limits, plus the growth boundaries this change introduces (8/9/17/41 pieces) and a seeded random corpus.
- Full-suite CI is the final gate.

### Unrelated finding, not fixed

`dom/modify-text`'s JS baseline does 10× the writes of its wasm source plus a string concat the wasm side doesn't do, and `dom/read-attributes` tests `!== null` where the wasm tests `.length > 0`. Those two published rows compare different programs, independent of any compiler work.

## CLA

---
_Generated by [Claude Code](https://claude.ai/code/session_01QGvZAfx48LzWpkWsmJvm9H)_


ttraenkler:
Diagnosed the park; pushed `84a1d177`. Removing the `hold`.

**First, the batch attribution in the bot comment is wrong and worth fixing separately.** It reports the group as "(#4139, #4144, #4106)", but neither #4139 nor #4144 is a PR — `GET /pulls/4139` is a 404. Those are **issue** references parsed out of this PR's own title (`fix(#4139/#4144/#4149): …`). So there was no batch: the failure is attributable to this PR alone, and the "re-enqueue them one at a time" advice had nothing to apply to. Filed as a note here rather than a separate issue since it is a bot-parsing bug in the park comment, not a compiler defect.

### What actually failed

Not a pass→fail. The #3189 uncatchable-trap ratchet: `illegal_cast` 48 → 50, on two tests that are **`fail` in the baseline in every arm** — a change of failure *mode*. The run was otherwise strongly positive: net **+217 pass** (31,581 → 31,798), 279 improvements vs 7 regressions, other trap categories flat or down (`null_deref` 1639→1635, `oob` 52→52, `unreachable` 3→3).

### Root cause — measured, and my first two hypotheses were both wrong

I A/B'd with the real runner (`runTest262File`), not the diff:

- **Not `resolveWasmType`.** Reverting `src/codegen/index.ts` alone left the trap in place.
- **Not private-name routing.** Guarding the new fallback with `isPrivateIdentifier` changed nothing, so I dropped that guard rather than ship an untested behaviour change on a hunch.

It is `assignment.ts`, **indirectly, by un-masking**. `set #m(v) { this._v = v; }` writes a field with *no declaration*; the generic struct path used to drop that write outright, so both tests died at their **first** assert (`Test262Error: Expected SameValue(«undefined», «"test262"»)`) and never reached the brand check. #4149 makes the write land, so they now run past it and hit a **pre-existing latent** uncatchable `illegal cast` on foreign-receiver private access — where §7.3.28 PrivateBrandCheck requires a *catchable* TypeError.

| tree | result |
| --- | --- |
| `upstream/main` | `Test262Error` |
| this branch | `illegal cast` |
| this branch, `assignment.ts` reverted | `Test262Error` |

### What I did

Declared it rather than hid it: `trap-growth-allow` in `plan/issues/4149-*.md`, `count: 2` with **both tests named** — the machine-verified #3596 shape, confirmed to parse via the gate's own `parseFrontmatterCountReason`. Filed **#4154** for the real fix.

#4154 should make both files **pass**, not merely stop trapping: each one's remaining assertion is exactly the `assert.throws(TypeError, …)` that a catchable throw satisfies. Landing it retires this declaration.

---

ttraenkler:
**Suspended — this PR needs an owner.** Handover written to `plan/issues/4150-*.md` under `## Suspended Work` (pushed as `27240be1`), with pointers from #4149 and #4154.

State at suspension: head `27240be1`, open, **not merged**, no `hold`, CI re-running after a `main` catch-up merge. I have cancelled my self check-in and unsubscribed from PR activity, so **nothing is watching this** — an armed auto-wake would contradict a suspend.

**The one decision that is not mine to settle:** the merge-queue park was resolved with the `trap-growth-allow` **valve** rather than by fixing #4154 in this PR. Defensible — documented mechanism for baseline-`fail` trap growth, near-identical precedent in #2900 — but it ships an uncatchable trap where a catchable error used to be, on two already-failing tests. Fixing #4154 instead would very likely turn both **FAIL → PASS**. Worth deciding before this lands.

**If this PR is closed without merging, two compiler fixes are lost again.** #4088 merged at `77f080d0` — the three-defect state, not the five its description advertised. The fnctor-twin (`2ef595b7`) and stack-balance tee (`659c0bf9`) fixes have never been on `main`; #4106 is the only thing carrying them.

#4154 is now `sprint: current`. Three other diagnosed-but-unfixed items are recorded in the handover: `string/case-convert` (near its structural floor), the DOM benchmark-definition mismatch, and the `declared_global` caching P4.

---

ttraenkler:
Closing in favor of a replacement PR, per the project lead's decision.

This PR's head lives on the `ttraenkler/js2` fork, which the cloud sessions doing the work cannot push to (the egress proxy denies `receive-pack` there). The completed continuation of this work — the `src/runtime.ts` conflict resolved keeping both recovery arms, the 5 issue links fixed, and #4154 fixed in-PR with the `trap-growth-allow` valve removed — has been sitting unreachable on the upstream copy of the same branch since 2026-08-04, and has now been refreshed against current `main` (the host-import wrapper collision with #4156 resolved in favor of this branch's signature-arity specialization; rationale in merge commit `a100272a7`).

The replacement PR from the upstream branch supersedes this one; nothing from this PR's diff is dropped.

---

PR #4124, opened 2026-08-04
ES5 standalone conformance: annexB B.3.3 (#3980 +95), own-prop MOP (#3979 +29/-1), .apply receiver (#3983), String.prototype borrowing (#3978) + harvest/census

ttraenkler:
# ES5 standalone conformance: annexB B.3.3 (#3980 +95), own-prop MOP (#3979 +29/-1), .apply receiver (#3983), String.prototype borrowing (#3978) + harvest/census

## Summary

One session of ES5-standalone conformance work, from the `ttraenkler` fork. Includes everything merged to the fork in ttraenkler#11 plus the follow-ups in ttraenkler#12.

| Issue | Change | Standalone delta | Status |
| --- | --- | --- | --- |
| #3980 | Annex B B.3.3: skip the web-compat var binding when `var F` would be an Early Error (flat `localMap` leaked the lexical `F` to nested closures) | **+95** (104->199, 312-file family), 0 regressions | measured |
| #3979 | Reflective MOP over #3468/#3537 own-property bags (hasOwn/gOPD/delete/in/keys) + writable-gate in `__closure_prop_set` | **+29/-1** (six `Object.*` descriptor dirs, 2,471 files) | measured; the -1 documented below |
| #3983 | `.apply(thisArg)` silently dropped the receiver (`.call` emits a this-binding thunk, `.apply` emitted a bare call) | **+4** (+4 host) | measured |
| #3978 | Borrowed `String.prototype` receivers (unwired proto members, `charAt`-only transfer dispatch, missing ToPrimitive pre-reduction) | unmeasured | 16/16 unit tests; needs its scoped A/B |
| #3982/#3973/#3974/#3975 | Harvest issues from the 2026-08-01 baselines run | - | docs |
| #3977 | es5 census: published 59% was stale (#3892); measured 66.3%, ceiling 92.4% without eval/with, lever table | - | docs |

## The #3979 fix worth review attention

`__closure_prop_set` stored into the expando bag unconditionally. `verifyProperty` probes `writable:false` by WRITING - the probe planted a stale bag shadow for `name`, masked while the #2896 builtin-fn metadata was live, exposed once `delete fn.name` cleared it: hasOwnProperty answered from the bag and the harness concluded "not configurable" (the 7-test `name.js`/`length.js` family). Fix: gate the store on `__builtinfn_get_meta` - non-null = live builtin name/length = writable:false per s17 = sloppy no-op. Post-delete assignment still creates an ordinary own property; expandos untouched. New unit suite `issue-3979-bfn-name-writable-gate.test.ts`.

**Known accepted -1:** `15.2.3.6-4-594` needs an inherited `Function.prototype` setter dispatched before an own-prop store (OrdinarySet). It passed before only because two bugs cancelled (bag write invisible to hasOwn + bag read). Deferred to #2992/#3251; recorded on the issue.

## Honesty notes

- Early session measurements mislabelled lanes (a positional-arg bug passed `{target}` as `timeoutMs`; every "standalone" number was host). Corrected on the issues and in a comment on ttraenkler#11; #3980's standalone figure is the authoring agent's own measurement.
- The branch base is behind upstream main (upstream moved ~767 commits during the session); conflicts in `object-runtime*.ts` / `array-object-proto.ts` are likely and a merge of upstream main into the branch may be needed.
- #3978 claims NO conformance delta yet - its scoped A/B is the gate before that half merges.

## CLA

---
## CLA


ttraenkler:
Reviewed this against current `main` while checking whether its conflicts touched #4125/#4126. They don't — zero shared paths. But the rebase is not the main problem, and I think this PR should be **harvested rather than rebased**. Evidence below so it can be checked rather than taken on trust.

## Two of the four code changes appear to have already landed on `main`

- **`.apply` receiver (#3983)** — already merged as **#4025, status `done`** (PR #4120, commits `6de45ca` + `a64c5ff`, in `d3d7ec4`). `main`'s `src/codegen/named-this-call.ts` already handles `apply`.
- **annexB B.3.3 (#3980)** — this PR lists `src/codegen/annexb-cancel.ts` as **added**, but that file already exists on `main` at 480 lines, most recently touched by `ce19d43 fix(#4131): Annex B B.3.3.1 step 3 — update an EXISTING var binding`.

These are the two changes carrying the measured deltas (+4 and +95), so the headline numbers may already be banked. Worth confirming case-by-case before discarding — if this branch's version covers cases `main`'s doesn't, that delta is worth porting on its own.

## The issue ids collide with *different* issues on `main`

Same number, unrelated content:

| id | on `main` | in this PR |
| --- | --- | --- |
| 3979 | "Calling a function stored in a MIXED array literal returns null" | standalone reflective MOP over own-prop bags |
| 3980 | "A generator PARAMETER mutated in the body reverts…" | annexB B.3.3 hoisting skipped on early error |

Five more issue files duplicate `main`'s under *different* ids: 3973→#4020, 3974→#4021, 3977→#4022, 3982→#4024, 3983→#4025.

This is the hand-picked-id hazard CLAUDE.md documents — ids chosen on a long-lived fork branch while `main`'s sequence moved underneath. The required `check:issue-ids:against-main` gate rejects this as-is, and the fix is `claim-issue.mjs --allocate`, not renumbering by hand.

## Suggested disposition

A 767-commit merge into `object-runtime*.ts` / `array-object-proto.ts` would mean resolving conflicts against work that has already landed — the worst kind, where both sides look correct.

1. **Drop** the `.apply` and annexB halves after confirming `main` covers the same cases.
2. **Keep #3979 (own-prop MOP).** `src/codegen/own-prop-bag.ts` is genuinely absent from `main` — this is the real new content, and the `+29/-1` is the substantive win. Re-cut it as a fresh branch off current `main` with a freshly allocated id.
3. **Drop the duplicate issue files**; fold anything new in them into the existing #4020/#4021/#4022/#4024/#4025.
4. **Hold #3978** (String.prototype borrowing) — unmeasured by this PR's own description; it needs its scoped A/B first.

Sequencing note: #3979's accepted `-1` (`15.2.3.6-4-594`, an inherited `Function.prototype` setter dispatched before an own-prop store) is deferred to #2992/#3251 — the same epic #4159's implementation plan builds on. If #3979 is re-cut, landing it before the #4159/#4160 shared pre-scan slice avoids two changes in `object-runtime-descriptors.ts` at once.

Not closing this — it belongs to another lane, which may have context that isn't visible here.

---

ttraenkler:
Closing — every slice in this PR has landed on `main` by another route, and the branch is now costing more than it carries.

## Where each slice went

| slice in this PR | landed as |
| --- | --- |
| annexB B.3.3 (`#3980`, claimed +95) | **#4023** (`src/codegen/annexb-cancel.ts`, since extended to 480 lines), plus follow-on **#4131** (B.3.3.1 step 3, existing var binding) |
| own-prop MOP (`#3979`, +29/-1) | **#4161** (merged) |
| `.apply` receiver (`#3983`) | **#4025** (`src/codegen/named-this-call.ts`) |
| String.prototype borrowing (`#3978`, claimed +29) | mostly already on `main` by other routes — the case-conversion family, `indexOf`, `charCodeAt` and `substring` were each probed on current main and pass. The only outstanding piece was `slice`, landed today as **#4132** |

The `#3978` headline number is worth flagging for anyone reading this later: it was measured against a base ~767 commits behind, and re-probing each claim on current `main` found all but one already fixed. A stale delta is not evidence of remaining work.

## Why closing rather than rebasing

- No common ancestry with current `main` in any practical sense — re-deriving each slice against `main` was the only workable path, and that is now done for all four.
- The branch hand-picks issue ids **4163–4171** without reserving any of them on `origin/issue-assignments`. That collided with two PRs today (**#4142**, **#4145**), and clearing the range cost four throwaway reservations. Every new issue filed while this stays open risks the same.
- It also nearly caused a duplicated implementation: a lane was about to re-write the B.3.3 slice from scratch before checking whether it had already been harvested.

Nothing here is lost. If any part is still wanted, it should be re-derived against current `main` and measured there — the four harvests above are the template.

---

PR #4153, opened 2026-08-06
fix(#4180): stop fabricating a descriptor from a struct's internal wasm fields

ttraenkler:
`emitDescriptorStructReify` (`src/codegen/object-ops.ts`, #2372) turns a typed WasmGC descriptor struct into a fresh `$Object` by copying the struct's **wasm fields**, and hands that to `__obj_define_from_desc`. It fires for **any** descriptor argument that compiled to a named struct with ≥ 1 field.

That is correct for the case it was written for — a descriptor object literal (`var d = {value: 1}`) whose closed struct fields *are* its own properties. It is silently wrong for every other struct, because it transcribes the **internal representation** and throws the object's real own properties away.

```js
var arrObj = [];
arrObj.enumerable = true;          // lands in the #3537 vec bag
arrObj.value = 42;
Object.defineProperty(obj, "property", arrObj);
obj.property;                      // => undefined   (spec: 42)
```

The emitted `__module_init` literally does:

```wat
call $__new_plain_object                        ;; descObj
__extern_set(descObj, "length", arrObj.length)
__extern_set(descObj, "data",   arrObj.data)
call $__obj_define_from_desc
```

ToPropertyDescriptor finds no `value`/`enumerable`, CompletePropertyDescriptor fills in `undefined` + all-false. No refusal, no diagnostic. Same for `new Date()` (`{timestamp}`) and subviews (`{length, data, byteOffset}`).

## The fix

`isDescriptorTranscribableStruct` (`src/codegen/property-descriptor-shape.ts`) gates the transcription:

- object-literal structs (`__anon_*`) **always** transcribe — their fields are their own properties by construction, including `{foo: 1}`, which must yield an empty-but-valid descriptor rather than a TypeError;
- any other struct transcribes only if it carries at least one of the six §6.2.5.6 field names (a fnctor instance with `this.value = …` is a real carrier — unchanged);
- everything else passes through as an externref, so the runtime applier runs ToPropertyDescriptor over the actual object.

A **plausible-descriptor** test rather than a builtin-representation denylist: a denylist must track every struct the compiler ever mints (`__vec_*`, `__subview_*`, `__Date`, `__StandaloneRegExp`, error/box structs, …) and fails **open** when it falls behind — the wrong direction for a helper whose failure mode is a silent wrong answer.

Pass-through is safe on both gates it must clear: `__obj_define_from_desc`'s Type check has been `typeof === "object" || "function"` since **#3246** (not a `ref.test $Object`), and `__typeof_object` answers 1 for any non-null non-primitive. It is also strictly safer on a **null** struct ref — the transcription read each field under `ref.as_non_null` and would trap, whereas the applier treats a null descriptor as a lenient empty-descriptor no-op.

## Measured

Instrument: the CI-aligned scoped runner with the `js2wasm:runtime-eval` provider shim (**#4162** — `tests/test262-runner.ts` does not supply the namespace `scripts/test262-worker.mjs` does; without the shim every `propertyHelper.js` test dies at instantiate). 558-file lever list, `--target standalone`.

| | pass |
| --- | ---: |
| base | **92 / 558** |
| this branch | **104 / 558** |

**+12, 0 regressions on the list** — `built-ins/Object/defineProperty/15.2.3.6-3-{34,39,87,92,140,145,166,171,219,224,249,254}.js`, the Array and Date `'Attributes'` carriers.

Instrument responsiveness verified in **both** directions: with the two source files swapped back to their `origin/main` copies those twelve score **0/12**; restored, **12/12**. (File-copy A/B — never `git stash`.)

## What this refutes — the more useful half

The lever was framed as descriptor-**reader** gaps. The reader is fine. The same array whose descriptor read as empty already reports its expando through `Object.getOwnPropertyNames`, `Object.keys` **and** `Object.getOwnPropertyDescriptor`:

```
read=101  gopn=zz  keys=zz  gopd=101/true      own=false  in=false  forin=(empty)
```

The descriptor path never asked them — it had substituted its own answer at **compile time**. The carrier-bag arm in `__desc_has_own` that the reader framing implies would have measured **+0**.

Reusable discriminator for the next wave: make the same value reach the call site as an **externref** instead of a typed struct (`var c = esc([]); c.value = 3; Object.defineProperty(o, "p", c)`) — it works. **If behaviour changes on static representation alone, the defect is in the static lane, not the runtime.**

## Sizing correction

The 558-file list was cut as failures, but **92 already pass** on current main. Real residue: **466**.

## Gates

`check:oracle-ratchet` +0/+0 · `check:func-budget` OK · `check:coercion-sites` OK · `check:ir-fallbacks` OK · `check:loc-budget` +10 on `object-ops.ts`, **granted** with a per-entry reason in the issue frontmatter (the gate must be consulted at the reify decision; all 58 lines of logic went to the subsystem module). `tests/issue-4180.test.ts` — 8 cases, including the "literal still transcribes" guards. Equivalence gate green, with +4 baseline fixes.

## Named next slices (rationale in `plan/agent-context/W5-descriptor-residue.md`)

- **44 files** `-1` "of prototype object" (`RegExp.prototype.value = …`). Nearest substrate is #4160's `proto-index-store.ts`, gated to integer keys and only 2 companions. Widening needs the `protoIndexDirty` **pre-scan** widened too, which turns the substrate on for far more modules — the exact shape that cost #2660 S2 **−40** on the floor. Measure separately.
- `hasOwnProperty` / `in` / `for-in` on a *statically typed* array bypass the runtime helpers that already consult the bag (#4159-adjacent, static-lane routing).
- 13 `[SITE-PROPS-BAG-NOT-AUTHORITATIVE]` refusals — the obvious vec arm is the one #4047 measured at **+6 and reverted**. Not worth re-litigating for 13.
- Error carrier (27) needs #4098 storage; Math/JSON carriers (46) hit a compile-time refusal (#1907 / #1888 S6-b), not a descriptor bug.

---


ttraenkler:
`hold` removed — this park is **spurious**, in exactly the way the bot's own footnote anticipated (#3597).

Both cited failures are the same GitHub Actions platform outage, not a regression:

**`detect test262-relevant changes`** — failing step is `Set up job`, and the log is:

```
Failed to resolve action download info. Error: Service Unavailable
##[error]Internal Server Error occurred while resolving "actions/checkout@v5"
```

The job never checked out. The verdict never ran.

**`merge shard reports`** — "Fail if required test262 shards did not succeed" is **downstream** of the above: the shards could not run because the detect job died at setup. It is reporting the absence of a result, not a bad result.

Corroborating that this is platform-wide rather than specific to this PR, within the same ten minutes: #4153's `cla-check` failed on the step whose only job is to *post a passing status*, and #4155's `equivalence-shard (3)` failed with an identical `Internal Server Error … actions/checkout@v5`. Meanwhile the same CLA step succeeded in the merge_group runs for #4151 and #4152 shortly before.

There is nothing in this diff to fix, so per the auto-park rules — diagnose the cited run first, then distinguish real-regression from flake — this qualifies as a confirmed flake and the label comes off. Leaving it on would strand the PR, since `auto-enqueue` skips held PRs.

Not re-enqueueing manually; `auto-enqueue` will re-take it now that it is `CLEAN` and unheld.

---

PR #4154, opened 2026-08-06
fix(#4181): count non-assign binary drops in the #3623 telemetry; add missing `**=` to the module-init assignment-operator list

ttraenkler:
Two small changes to `src/codegen/declarations.ts`'s module-init collection, plus the #4182 design doc. Stacked on #4151, which has now merged, so the diff is just these two commits.

## 1. Telemetry hole (behaviour unchanged)

A non-assignment `BinaryExpression` statement `continue`s out of the collection loop **before** reaching the #3623 drop classifier. So top-level comma and short-circuit statements (`f(), 1;`, `x && x.foo();`, `a || b;`) were dropped **silently and uncounted** — invisible even to the tooling built specifically to make drops loud.

Now routed through the classifier before being skipped. No behavioural change; the point is that the next person auditing this gets a true count instead of re-deriving it by hand.

Context for why that matters: this was found during an audit of all 20,255 failing standalone-baseline files, prompted by `with` turning out to be the fourth bug of the "silently dropped from `__module_init`" family (#2992 top-level `delete`, #3592 top-level `throw`, #3615 top-level bare read, #4179 top-level `with`). That audit concluded the statement-kind allow-list is now **exhaustive** — `with` was the last gap — and that the remaining expression-shape drops are the deliberately-deferred #3623 population (~40–70 files, none on the ES5 lever). This closes the measurement gap so that conclusion stays checkable.

## 2. A real one-line bug: `**=` was missing

Found while writing the telemetry fix. The module-init assignment-operator list carries all 15 of its siblings (`+=`, `-=`, `*=`, `/=`, `%=`, `&=`, `|=`, `^=`, `<<=`, `>>=`, `>>>=`, `??=`, `||=`, `&&=`, `=`) but **not `**=`**.

The #3623 classifier said *keep*; the collector dropped it. Probed: a top-level `x **= 3` on `x = 2` now yields **8**, previously left it at **2** — a silent wrong answer of exactly the same shape as the four bugs above.

## Gates

`tsc --noEmit` clean; budget gates green with allowances recorded in the issue frontmatter.

---


ttraenkler:
Park diagnosed as **collateral, not a defect in this PR**; `hold` removed.

**The failure is a docs-vs-baseline bookkeeping mismatch, and conformance went UP:**

```
committed: standalone (host-free) test262 conformance: 27,825 / 43,505 (64.0 %)
generated: standalone (host-free) test262 conformance: 27,999 / 43,505 (64.4 %)
```

**Not fixable from this branch.** This PR touches neither `README.md` nor any baseline artifact — its diff is two changes in `declarations.ts`'s module-init collection plus a design doc. Running `pnpm run sync:conformance` against current `main` content produces **`0 updated, 5 unchanged`**, so there is no local change to commit: `main` is *in sync* right now. The +174 exists only in the merged state, which this PR could not have known at push time.

**This is the drop mode `test262-sharded.yml` documents against itself**, in the promote-baseline step:

> *"Sync the conformance numbers baked into the prose docs to match the freshly-promoted `test262-current.json`. This is the ONLY writer of the number on main, so it must update everything derived from it ATOMICALLY in the same commit — otherwise main goes out of sync the instant the baseline bumps, and every in-flight PR that reaches the merge-queue head is dropped by the `quality` gate's `sync:conformance:check` … See #1522 + the 2026-05-29 merge-queue drops of #895/#896."*

Note that writer is **non-fatal on failure**:

```
node scripts/sync-conformance-numbers.mjs || \
  echo "WARN: sync-conformance-numbers failed (non-fatal) — docs may drift"
```

So a single silent `WARN` on any promote leaves the next PR at the queue head to absorb the park. With a lot of conformance-improving work landing tonight (+76, +58, and +106/+46 still in flight), that window is being hit repeatedly.

Per the auto-park rules this is the "confirmed collateral" branch, not a real regression — the merged state is *better*, not worse — so it gets a single re-admission, not a loop. If it parks again on the same step I will escalate rather than re-enqueue.

---

PR #4155, opened 2026-08-06
fix(#4176): per-brand proto-property store — named keys on builtin prototypes live through the chain (+76 on the 219-file ES5 lever)

ttraenkler:
Closes #4176. The W2/#4172 residue slice: after #4145 (+95) the remaining 124 failures on the 219-file prototype-chain lever decompose into ONE dominant mechanism (~62 files, refuting the 15/14/12/8/~75 estimate handed over): **a named key written onto a builtin's `.prototype` is invisible** (`Function.prototype.value = "x"` / `Object.prototype.zzz = 1` / `Array.prototype.enumerable = true` — the §8.10.5 inherited-descriptor-field idiom), plus ~15 nested-descriptor `Object.create({}, {prop: descObj})` files that clause A missed.

## Measured

CI-aligned shimmed instrument (runtime-eval provider shim — **#4162** still unfixed):

| 219-file lever list (`--target standalone`) | pass |
| --- | ---: |
| base = `origin/main` with #4145 merged | 95 |
| this branch | **171** |

**+76, 0 regressions on the list.** Instrument responsiveness verified by base-file swap of `declarations.ts` (probes revert to red). Subsystem unit tests 158/158 (4160 / 2660×4 / 3468 / 3537 / 4055 / 3251×3 / 802×2). Ratchets: oracle +0, coercion-sites net **−1**, loc/func growth granted with per-entry reasons in the issue frontmatter.

## Five independent gaps, each probed before fixing

1. **Top-level `<Builtin>.prototype.<name> = …` compiled to NOTHING.** The module-init collection drops statements with no module-global root identifier, and `Object`/`Function`/… are builtins — measured: the write was simply absent from `__module_init` WAT. Flag-gated keep-arm in `declarations.ts`, mirroring the #1719 / #2660 / #3468 keeps. (This is the *fifth* instance of the module-init drop family, after #2992, #3592, #3615, #4179.)
2. **#4160's store was integer-only and Object/Array-only.** Generalized to a per-brand companion table (one `$Object` slot per `BUILTIN_BRAND_TABLE` entry, lazily minted); named keys admitted — the integer gate protected nothing, since refused keys were silent no-ops on the proto singleton. Reads are now RECEIVER-aware (`__protoidx_brand_off`: vec ⇒ Array, closure ⇒ Function, RegExp/Date/Error structs ⇒ their brands, `$NativeProto` ⇒ own brand, boxed-primitive wrappers ⇒ String/Number/Boolean via the `[[PrimitiveValue]]` slot's box type, default ⇒ Object; chain depth 2 → `Object.prototype`).
3. **The #2372 descriptor-struct reify severed carriers.** A vec/Date/RegExp descriptor argument was copied field-by-field into a fresh `$Object`, losing carrier-bag own fields *and* inherited companion keys. Such carriers now pass through as externref; `__obj_define_from_desc` reads them directly (its §10.1.6 gate has accepted any object since #3246). *(Independently found by the concurrent descriptor lane — see #4153.)*
4. **Clause A one hop deeper** — property values inside an object-literal argument of a builtin `Object.*`/`Reflect.*` call are dynamic consumers (`Object.create({}, {prop: new Con()})`).
5. **Pre-scan + ESM cycle fix** — a new `protoNamedDirty` flag, deliberately separate from `protoIndexDirty` so `String.prototype.foo = …` (the polyfill idiom) reserves the store *without* disabling the HOF hole visit-skip / typed element lanes. Brand table extracted to a dependency-free `builtin-brands.ts`, closing a pre-scan → native-proto ESM cycle that was crashing with a TDZ in `collections-brand.ts`.

## Scope guard

Every runtime emission sits behind the store reservation (`ctx.standalone && (protoIndexDirty || protoNamedDirty)`), so flag-clear modules keep byte-identical bodies (consult builders return `undefined`). The two changes not behind that flag are semantically conservative: the `__extern_has` arm restructure is behaviour-identical when the store is unreserved, and the reify pass-through only affects vec/Date/RegExp descriptor arguments, whose static struct fields are not descriptor field names.

## Note on `tests/equivalence/spec/coercion-arithmetic-add.test.ts`

This lane flagged its 8 failures as a possible queue-wide blocker. They are **pre-existing and already recorded** — all 8 are in `scripts/equivalence-baseline.json`'s `knownFailures` (one per lane × two cases, 22% of that 36-entry baseline). `equivalence-gate` fails only on failures *not* in the baseline, so it stays green. Root cause is filed as **#4178**. Reproduced identically on the base worktree without these changes.

The full local equivalence run OOMs in this container (known limit); scoped runs and subsystem suites were used instead, and CI's `equivalence-gate` is the arbiter.

## Left undone (in the handoff)

`in`-operator static folds on typed receivers; Math/JSON namespace own-props; Error-subclass 3-level chain; for-in over inherited companion keys. The remaining 46+2 are genuinely different mechanisms (toString tags, `Number.prototype` primitive value, filter borrows, `__get_builtin` CEs) — and 4 TypeError files on `defineProperties` 7-6-a-16x/17x may belong to the array `ArraySetLength` lane.

---


ttraenkler:
**Held deliberately to sequence a known conflict — not a park, and not a problem with this PR.**

This branch and **#4153** both independently rewrote the same #2372 reify gate in `src/codegen/object-ops.ts`. Verified rather than assumed:

```
$ git merge-tree --write-tree <4176-head> <4180-head>
CONFLICT (content): Merge conflict in src/codegen/object-ops.ts
```

The queue here is **grouped**, so two conflicting PRs in one batch fail the batch and eject both. Holding this one lets #4153 land first; I will then clear the hold, merge `main` in, and resolve.

**Resolution decided in advance** (recommendation from the lane that found the conflict, and I agree): keep **#4153's `isDescriptorTranscribableStruct` plausible-descriptor predicate** and drop this branch's 3-name skip list. The predicate subsumes the list and **fails closed** — a denylist of builtin struct names must track every struct the compiler ever mints and fails *open* when it falls behind, which is the wrong direction for a helper whose failure mode is a silent wrong answer.

## Independent confirmation of this PR's value while it waits

A separate lane took the descriptor family's 44-file "of prototype object" slice and measured it against this branch:

| tree | pass |
| --- | ---: |
| `origin/main` | **0 / 44** |
| this branch (unmerged) | **43 / 44** |

CI-aligned shimmed instrument, bundle and provider rebuilt per tree, responsiveness confirmed both directions. That lane then wrote **no code**, because this PR is already that slice's fix — so the +76 measured here understates it by a further ~43 on a lever nobody had attributed to it.

The single remaining file is root-caused and filed separately as **#4187** (standalone `hasOwnProperty` constant-folds and ignores a runtime `delete`).

---

PR #4156, opened 2026-08-06
fix(#4178): concat of a boxed-any value returned `null` (then trapped) or folded numerically

ttraenkler:
Two independent **legacy-path** defects, both reached by `"" + <any-valued expr>`. Neither is the IR bail #4178 names, and neither is eval-mode- or ternary-specific. **#4178's own root cause is refuted below** — but its stated acceptance criterion is met.

## (A) `coerceType`'s `ref → ref_null` arm was missing the `$AnyValue` unbox case

It is one of four arms in the same function dispatching on `from.kind`/`to.kind`; the other three (`ref_null→ref_null`, `ref_null→ref`, `ref→ref`) all carry the unbox, including the #1988 "a native string rides in `externval` (field 4), not `refval` (field 3)" split.

`compileAnyBinaryDispatch` returns exactly `{kind:"ref", typeIdx:$AnyValue}`, so every `any`-operand `+`/`-`/`*` result assigned to a **nullable** slot fell through to the generic guarded `ref.cast` below — which tests the **box** against the target type, always fails, and stores `ref.null`. The next reader (`__str_concat`, `.length`) dereferences null and **traps**.

## (B) `tryStaticToNumber` traced `const` initializers for VALUE but not STRING-NESS

Its `+` guard only asked "is this a syntactic string literal?" and "does the checker say `string`?", so for

```ts
const a: any = "1"; const b: any = 2; a + b
```

neither fired (`a` is an identifier; its declared type is `any`) — while the identifier arm at the bottom of the *same function* happily resolved `a` through its `const` initializer to `Number("1") === 1` and folded the whole expression to `f64.const 3`.

`resolvesToStringConstant` now traces the same way via `ctx.oracle.constInitializerOf` (#1930), with the same `const`-only and #1607 self-reference restrictions, so the guard can no longer be weaker than the folder it guards.

## Measured

| instrument | before | after |
| --- | --- | --- |
| `scripts/equivalence-gate.mjs` | 36 known-failures | **12 now PASS, 0 new regressions** |
| `tests/equivalence/spec/coercion-arithmetic-add.test.ts` | 8/20 fail | **20/20 pass** |
| 30 scoped test262-harness probes (standalone, #4162 shim) | 7 trapping + 1 wrong answer | **all 8 correct, 0 moved backwards** |

The 12 rows are **all 8** `coercion/arithmetic-add` any-concat entries (host, host-O, standalone, standalone-O × two cases) — 22% of the 36-entry baseline and #4178's stated acceptance criterion — plus four that came free: `#1197 i32 element specialization … x | 0 collapses`, `Math.pow/min/max with array element args`, and two `Symbol basic support (#471)` rows.

Ratchet post-merge with `node scripts/equivalence-gate.mjs --update` (deliberately not committed; CI refreshes on main).

## What this refutes

**#4178's root cause does not hold for the population that matters.** Every unit in the failing shapes is rejected at IR **SELECT** stage (`body-shape-rejected` / `call-graph-closure`, read with `trackIrOutcomes`) and never reaches `lowerConditional`. Module-init — which is what test262 top-level statements compile to — is always in that class. **Fixing the bail alone would have moved zero measured failures.**

The two behaviours #4178 called "same mechanism, two outcomes" are two mechanisms: the hard `IR-FALLBACK` is the bail; the wrong answer / trap is (A)+(B). And the 8 equivalence rows contain **no ternary** at all — they were tagged #1988 in the test file, which is also not what breaks them.

## Attempted and reverted: the `emitBox` fix

I implemented #4178's proposed fix — box mismatched arms to `dynamic` via `IrFunctionBuilder.emitBox`, each box appended to its own arm's `collectBodyInstrs` buffer (preserving #1820), with a `jsTagOfIrType` refinement. It typechecked and compiled. **Backed out on measured evidence, twice over:**

1. **Correctness regression.** `"" + (true ? someBoolean : "s")` IR-compiles to `"1"` instead of `"true"` — the boolean brand on the arm's `i32` does not survive to `emitBox`, so it boxes tag-2 (number) rather than tag-4. That is a *silent wrong answer* replacing a loud compile error, and precisely the hazard `emitBox`'s own header warns about.
2. **The bail just moves.** With it retired, the same functions immediately hit `local 'g' annotated as string but initializer is dynamic`, then `arg 0 of call to len is dynamic, expected string`. Net-zero IR coverage — one compile error traded for another.

The real slice is **(a)** carry the boolean/symbol brand onto the arm IrType, **(b)** teach annotated-local writes and call-arg positions to accept a `dynamic` producer with an explicit unbox/ToString, **then** **(c)** retire the bail. That is an **L**, not an M. `tests/issue-4178.test.ts` pins the bail's current behaviour so the boundary is visible rather than rediscovered. **#4178 stays open.**

## Two adjacent defects found, not filed (both want ids)

- **`$AnyValue === nativeString` inline answers `false`**: `const g = a + b; g === "12"` is true, but `(a + b) === "12"` is false.
- **Eval-consumer `$AnyValue`/externref mismatch**: `registerReassignedFunctionGlobals` (`index.ts:6006-6027`) widens every top-level binding to `externref`, consumers still expect `ref_null $AnyValue`, nothing coerces, and `stack-balance.ts:1504` papers over it with a blind `any.convert_extern; ref.cast_null`. Traps on a `$BoxedNumber`; mis-tags as tag-5 "string" (the #1888 lie) on an `$AnyValue`. **Fixing the boxer alone converts the wrong answer into a trap — read site first.** This is the real home of the earlier "739 ES5-label standalone failures in eval-consumer modules" figure; it is *not* #4178's population and must not be used to size it.

## Notes

`loc-budget-allow: src/codegen/type-coercion.ts` in #4178's frontmatter, reason recorded: the arm is a `coerceType` branch selected by `from.kind`/`to.kind`, and splitting one four-way dispatch across two files is what let the fourth arm silently diverge in the first place. Oracle ratchet net-neutral.

---


ttraenkler:
**Measurement addendum: the test262 conformance delta is NULL.** Recording this explicitly so the PR body's measured table isn't read as implying a conformance gain — it deliberately claims only equivalence-gate results, and that remains the whole proven yield.

800-file ES5-label standalone A/B (500 baseline-fails + 300 baseline-passes, #4162-shimmed): **0 gained, 0 lost.**

The raw run *looks* like **+25 / −22**, and both numbers are artifact:

- the **+25** (nearly all `Object/defineProperty|defineProperties|create|isFrozen`) is **committed-baseline staleness** — seven PRs landed while this lane was paused;
- **18 of the 22** losses are the lane's own `--refusal-only` provider tier, not the change.

Settled by re-running rather than argued: the 25 apparent gains plus the 4 non-shim losses were re-run on **base** and **patched** — `pass 25/29` both times, **0 per-file differences**.

### Honest split

- **Proven:** `equivalence-gate` — 12 baseline known-failures fixed, 0 new regressions, across all four lanes. All 8 `coercion/arithmetic-add` rows (#4178's stated acceptance criterion), plus #1197's i32 peephole row, the `Math.pow` test262 pattern, and 2 Symbol rows. `coercion-arithmetic-add.test.ts` 8/20 → 20/20.
- **Unproven, not disproven:** conformance yield. The sample is ~24% of the 2,063 ES5-label failures, so a true effect below roughly ±4 tests would not surface.

Also verified clean: `check:ir-fallbacks` OK; `tests/issue-{1988,745,2104,2107}` fail 6, a **byte-identical set on base** — all pre-existing.

Worth generalising from this: **the committed baseline goes stale within hours on an active day, and stale-baseline gains look exactly like real ones.** Any A/B run against it needs the apparent movers re-run on both trees before the number means anything. That is what turned a plausible +25 into a measured 0 here.

Post-merge: `node scripts/equivalence-gate.mjs --update` to ratchet the 12.

---

PR #4158, opened 2026-08-06
feat(#4182): annexB B.3.3.2 global-code — module-scope block-fn live bindings (+46 on the lever)

ttraenkler:
Closes #4182. Measured on the full `annexB/language/global-code` directory (153 files, standalone lane, in-process runner with a fresh #4162 provider built from this branch): **98 → 144, zero regressions** (re-verified post-merge).

Exposure population — every non-strict, non-module test262 file with a **module-scope** block/`if`/`switch`-nested function declaration (246 files, scanned on this tree): **151 → 191, +40 fixes, ZERO pass→fail conversions.** Unlike #4151 and #4179, this change converted no vacuous passes: the affected statements already executed, only their binding *resolution* changed.

## Root cause

A module-scope sloppy `{ function f() {} }` bound **statically** through `ctx.funcMap`:

- pre-evaluation reads saw the compiled function, but B.3.3.2.b says the binding is created as `undefined` at GlobalDeclarationInstantiation;
- the B.3.3.2.c evaluation-point `SetMutableBinding` never happened, so an outer `function f` was never updated by a later block `f` (`*-existing-fn-update` read "outer declaration" forever), and a second same-named block declaration was silently skipped by the `funcMap.has` early-return.

The function-scope Annex B machinery (`fctx.annexBOuterBindings` TDZ locals, #4131's existing-var update) writes `fctx.localMap` **locals** and cannot fire at module scope, where bindings are module **globals**.

## Fix — reuse the #2931 live-binding-global mechanism

New module `src/codegen/annexb-global-live-binding.ts`. Every arm is gated on the normally-empty `ctx.annexBModuleBindings` set, so programs without a module-scope sloppy block function are byte-identical.

1. **Registration** — for each eligible name (Annex B statement position, enclosing var scope is the SourceFile, no intervening lexical binder, no top-level `let`/`const`/`class` per B.3.3.2.a), back the name with a mutable externref module global. Strict/module sources are skipped entirely, confining the pass to script-mode compiles.
2. **Seed split** — a name that is *also* a real top-level `function f` seeds that closure (GDI initializes it); a block-only name seeds the `$undefined` singleton. Without the split, pass 2 of the #2965 two-pass init compile seeded the block function's closure and the pre-evaluation read wrongly observed it.
3. **Evaluation point** — compile *this* declaration as its own Wasm function and `global.set` its closure, ahead of the `funcMap.has` early-return (B.3.3.2.c.v–vi). Idempotent across the two-pass init compile via a per-node WeakMap.
4. **Reads/calls/typeof** — identifier reads route through the module-global arm; identifier calls take the live dynamic-dispatch path (a direct funcMap call would pin the GDI winner forever); `typeof` gets a dedicated arm ahead of both static folds.
5. **Catch-param leak fix** — with no prior local to restore, a `catch (f)` parameter stayed in the flat `localMap` past the catch scope and shadowed the live global forever. Scoped to these names, since the general leak is pre-existing and other paths may lean on it.

**Exclusion**: a name reassigned inside its declaring range keeps today's static path. The spec's block-local lexical binding is distinct from the global var binding (`{ function f() { f = 123; } }` mutates only the block binding — `*-block-scoping` asserts exactly this), and one shared global cannot model the split.

## Three findings the design didn't anticipate

1. The seed must be the `$undefined` singleton **and standalone-only** — the host lane's `undefined` *is* the null extern, and the tag-1 singleton would surface to host helpers as an opaque object.
2. `typeof` needs a `ref.is_null` branch for the host lane's null pre-state.
3. The reassignment exclusion must scan **into nested function bodies** — `*-block-scoping`'s write is inside the block function itself.

## Remaining 9 lever failures, itemized

All out of AOT scope: 8× `*-existing-non-enumerable-global-init` (assertions live inside `$262.evalScript(...)` strings — interpreter-side GDI) and 1× `script-decl-lex-collision` (expects a SyntaxError from a second `$262.evalScript`). Both #4137-adjacent.

## Boundaries honored

- **Function-code (#2200 Phase 2) untouched** — every new arm is gated on module-init compilation or `annexBModuleBindings` membership, so the **−1180** ground is not entered. `issue-2200-annexb-block-fn-hoist`, `issue-2552-annexb-phase2`, `issue-4131`, `issue-3980`, `issue-2931` all pass.
- **Eval-code untouched** (#4137).
- `src/codegen/declarations.ts` touched only in the seed loop — no overlap with #4151/#4181's collection-arm territory.

## Gates

`tsc --noEmit`, `check:oracle-ratchet`, `check:coercion-sites`, `check:ir-fallbacks` clean; loc/func-budget green via per-entry-justified allowances in the issue file. `tests/issue-4182-annexb-global-blockfn.test.ts` 9/9, plus 6 neighbouring annexB suites (58 tests) post-merge.

## Pre-existing failures flagged, not from this change

Verified by swapping `origin/main`'s `src` in on the same instrument, identical counts on both trees: `tests/issue-3633-eval-annexb-bindings.test.ts` (4/4 fail) and 12 equivalence files (32 subtests). Worth a look by whoever owns eval-code.

---


ttraenkler:
Park diagnosed and resolved; `hold` removed.

**The cited run's only hard failure was the #3189 uncatchable-trap ratchet** — `illegal_cast` 44 → 45 (+1). Everything else in that run was positive: net **+56 pass** (32,040 → 32,096), host stable-path fine-gate net **+102** (103 improvements, 1 regression), and every other trap category flat (`null_deref` 1626→1621, `oob` 44→44, `unreachable` 3→3).

**Attribution without a re-enqueue cycle.** The bot correctly flagged this as a batched group (#4137, #4158) and asked for one-at-a-time attribution. That wasn't needed here: **#4137 merged ~15h earlier**, so its content is already in the baseline — and the run itself reports *"0 test262-relevant commits separate the baseline from main HEAD"*. The delta is therefore attributable to this PR.

**It is a reclassification, not a regression.** Both newly-trapping files were `fail` on the baseline:

- `for-await-of/async-func-decl-dstr-obj-id-put-unresolvable-no-strict.js`
- `for-await-of/async-func-decl-dstr-obj-prop-put-unresolvable-no-strict.js`

That is the #3596 *baseline-did-testify* branch (change of failure **mode**), not the #3595 never-instantiated class, and not a conformance regression. Mechanism: binding a module-scope sloppy block function through a live module-global instead of statically through `funcMap` lets the sloppy put-to-unresolvable in these two async destructuring tests execute further than it previously did, reaching a **pre-existing latent** `illegal_cast`. The trap is reached by this change, not introduced by it. Consistent with this PR's own exposure measurement of **zero** pass→fail conversions.

Declared as a bounded `trap-growth-allow` in `plan/issues/4182-…md` (`61f0819026`). **Both** files are named even though the net category growth is +1 — `evaluateTrapReclassification` rejects a bare count and hard-fails on any *undeclared* newly-trapping file, so naming only the net file would not have verified.

No source change; the fix stands as measured (+46 on its own lever, plus an independently measured +16 in `annexB/language/eval-code` that no one had attributed to it).

---

PR #4175, opened 2026-08-07
feat(ir): #743 field↔param mutual fixpoint — field slots as lattice variables in the fnctor-graph satellite

ttraenkler:
Implements the committed Fable spec `## Implementation Plan — field↔param mutual fixpoint (Fable spec, 2026-08-07)` in `plan/issues/743-whole-program-type-flow-analysis.md` (the spec commit rides in this PR).

Field slots become lattice VARIABLES solved together with ctor/method param facts inside the fnctor-graph satellite, so the `this`-field-read bucket the #4166 measurement named as the dominant remainder can converge. Satellite-only — the main `IrUnitTypeMap` is untouched (the #1712 parity hazard).

## What lands

- **Edge (a), the write taxonomy** — plain assign, every numeric compound assign, `++`/`--`, `+=` (string-or-number), logical assigns, name-based `"all"` attribution for untracked receivers, and the poison set (`delete`, `this[k]`, `Object.defineProperty/assign` on `this`, destructuring and for-in/of targets, replaced/runtime-defined prototypes, escaped owners).
- **Edge (b), the reads** — DIRECT `this.<x>` answered from the field facts, returning raw `unknown` (not `dynamic`) so ctor-param↔field cycles start optimistic instead of freezing; nested and bare-`this` reads (`this.pos - this.lineStart`, `new Token(this)`) answered from a per-owner instance ATOM bound to the reserved scope key `"<this>"`.
- **The undefined-read guard** — an ordered ctor walk with if/else intersection and return-freezing gives every read a definiteness snapshot, so a read that could observe `undefined` never carries a numeric fact.
- **Non-convergence returns EMPTY facts**, never a partial state: the atom-mediated reads are not monotone.
- **A real call-forwarding UNSOUNDNESS closed** — `finishNodeAt` in the acorn dist is a tracked top-level function invoked *only* as `finishNodeAt.call(this, …)`. That shape previously produced neither an edge nor a poison, so its params sat at lattice BOTTOM forever and its `node.end = pos` write contributed *nothing* instead of widening. `.call` is now an edge; `.apply`, `.bind` and extracted forms poison; `.constructor` outside a comparison drops every ctor fact.

The satellite is split into four cohesive modules (`fnctor-graph-model`, `fnctor-field-writes`, `fnctor-field-lattice`, `fnctor-method-edges`) rather than granted a god-file allowance — the single file had reached 1,720 LOC, +220 over the ratchet.

The only touch outside the satellite is one rule in `inferExpr` resolving `this` from `"<this>"`. That key is not a legal identifier and nothing but the satellite binds it, so the rule is provably inert for the main fixpoint (a key of `"this"` would NOT be — a TS this-parameter declares a real parameter with that text).

## Result: the mechanism lands, the acorn census does NOT move

- **Census 55 typed / 1 discarded / 40 unknown — unchanged. Zero slots moved.** Binary 874,370 B unchanged; canaries 2,3,4,5; imports `[]`; exactly the 3 pre-existing parity IR-FALLBACKs.
- Flag-off byte-identity vs `origin/main` asserted by sha256 on the standalone acorn binary: `11aa8e23…b84ba89d`, 861,854 B — identical before and after the module split, and again after merging current main.
- Wall A/B not run: the spec pre-registered it at ≥10 movers.

**Why nothing moved, measured write-by-write** (not assumed): every stalled slot in the §9 bucket reads through `Parser.pos`, whose field fact is `dynamic` because of (1) one `"all"`-attributed `err.pos = pos` onto a `SyntaxError` in `raise`, and (2) ~22 `state.pos = <local>` writes in the `regexp_*` methods that the **params-only** scope model cannot type, plus `this.pos = this.nextIndex(…)` (only identifier callees resolve). So the binding constraint is the shared evaluator's value-flow precision, not graph reach and not field↔param mutuality. The issue file carries the full per-slot table and the three ranked next levers.

`JS2WASM_FNCTOR_CTOR_PARAM_TYPES` stays **OFF**.

## Spec deviations (both flagged with evidence in the issue file)

1. §7 states the acorn `.call` sites are "on untracked receivers" — they are not, and the omission was unsound rather than merely imprecise (see above).
2. §3.4's dynamic-key poison must not fire on non-`this` receivers: read literally, `newNode[prop] = node[prop]` (acorn's `copyNode`) sets `poisonAllFields`, which is a whole-module kill switch. Measured — it zeroed *every* acorn field fact on the first run. Scoped to `this[k]`-form writes, where the owner is localizable; pinned by a test.

## Verification

`tsc` · biome lint · prettier · oracle-ratchet · loc-budget (grant: `propagate.ts` +9) · func-budget · dead-exports · coercion-sites · stack-balance · `check:ir-fallbacks` — all exit 0.

New `tests/issue-743-mutual-fixpoint.test.ts` (24 tests) covers both cycle directions, the write taxonomy, definiteness/ordering, the poison set, `.call`/`.apply`/`.bind`/`.constructor` forwarding, an E2E slot assertion with flag-on ≡ flag-off runtime, and flag-off byte-identity. `issue-743-*` 49/49; `issue-4177`, `issue-4155-*`, `issue-2660-*`, `ir-*` green except `ir-scaffold` (1) and four `tests/equivalence/*` cases — all five A/B-confirmed PRE-EXISTING on `origin/main` sources in the same worktree, as are the nine `issue-3520-*` census failures.

## CLA

https://claude.ai/code/session_016zVUkJ2adCdrLoET5AkoUL

---
_Generated by [Claude Code](https://claude.ai/code/session_016zVUkJ2adCdrLoET5AkoUL)_


ttraenkler:
Park cleared — this was **collateral, not a regression from this PR**.

Diagnosed against the cited run ([31147826246](https://github.com/loopdive/js2/actions/runs/31147826246/job/92770960004)) before touching the label. The failing step is `quality` → "Conformance numbers in sync (#1522)", and the log names the exact mismatch:

```
[sync-conformance] DIFFERS  README.md (standalone)
  committed: 27,825 / 43,505 (64.0 %)
  generated: 27,999 / 43,505 (64.4 %)
```

That is a mismatch between `main`'s own `README.md` and `main`'s own `benchmarks/results/test262-current.json` — the standalone line has a second, non-atomic writer, so the two artifacts drifted and **every** PR in the queue failed the gate regardless of its diff. Nine PRs were parked on it. Fixed on `main` as #4178; the standalone line now reads 28,222.

Nothing in this PR's diff touches conformance numbers, and no test262/quality verdict step failed — the sync step aborts `quality` before them.

Resolution: merged current `main` into the branch (`03dabe65cc`, clean, no conflicts) and re-verified in the worktree:

```
node scripts/sync-conformance-numbers.mjs --check   → exit 0, 5 unchanged
```

`hold` removed so `auto-enqueue` can pick it up. Not enqueued by hand.

---

PR #4249, opened 2026-08-08
feat: ES5 standalone conformance — wave 3 (+63 measured test262 flips, 6 work packages)

ttraenkler:
## Goal

Continuation of the ES5-standalone-90% program (waves 1+2 merged as #4234, taking ES5 standalone to 86.79%). This wave lands six packages, each measured by sequential A/B with zero regressions, projecting ≈87.6% (~195 tests to 90%). Spec + outcome tables: `plan/goals/es5-standalone-90.md`.

## Work packages

| Issue | Flips | What |
| --- | --- | --- |
| #4230 | +10 | dynamic descriptor bags for `Object.defineProperties`/`create` — admission reframed as key-source completeness over the bag∪overlay union; indexed-vec and no-store receivers keep refusing loudly |
| #4231 | +1 (+31 tests advance past assertions 1–10) | four `with` scope-chain root causes; the family gate was a Tier-2 fallback write registering a dead local that shadowed later reads |
| #4234 | +15 | StringToNumber applied ×10 per exponent step (94 cumulative roundings) — replaced with a 10^k table (10.26 → 1.00 ulp); Number ctor value constants seeded onto the carrier |
| #4232 | +18 | `new Object(<primitive>)` `.constructor` via initializer tracing; `Object(null)`; `new String` §10.4.3 index exotics; reflective `replace` arm |
| #4233 | +15 | RegExp: static-pattern tracing through variables, exec arities/result shape, construction-time SyntaxError, `env::RegExp_exec` leak routed |
| #4229 | +4 | `arguments.callee` as a real own property (§10.6 13.a); strict-mode callee poison (partial) |

## Review flags

- `chore: refresh godfile profile` — deliberate `--update` for three waves of accumulated finalize-fill arm growth in `object-runtime.ts` (gate is local-only).
- #4229's strict `arguments.callee` poison is lane-independent; its negative case is regression-tested but the agent's sweep was standalone-only — the `equivalence-gate` covers gc.
- #4229 documents a latent pre-existing subtype-cast hazard in `ensureFuncClosureSingleton` (constructible vs plain closure wrapper) — worked around via a cast-free reader, not fixed.
- Issue files carry `loc-budget-allow:`/`coercion-sites-allow:` grants with rationale where gates required them.

## Validation

- Each package: sequential per-file A/B on its bucket, zero pass→fail; 100+ new vitest regression cases across 8 new test files, all green post-integration; `tsc` clean.
- A consolidated standalone run over all wave-3 buckets is running as an independent check; the merge queue's `merge_group` re-validation remains authoritative.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PmwHrtyzm28H9YuH2nYLYy)_


ttraenkler:
**Deliberate `hold` (author) — standalone regressions found by branch-side validation, fixes in flight.**

A consolidated standalone test262 run over the wave's touched buckets (all editions, not just ES5) found **140 pass→fail regressions** vs the post-#4234 baseline that the per-package ES5-scoped A/Bs could not see:

- **105 × descriptor-attribute regressions** ("…descriptor should (not) be configurable") — a wave-3 descriptor path mis-handles attribute defaults/deletion on `defineProperties`/`create` results and builtin function `length` descriptors.
- **33 × `regexp-modifiers` compile errors** — #4233's new construction-time SyntaxError validation over-rejects ES2025 `(?i:…)` inline-modifier syntax that previously compiled and passed.
- 2 singletons (a `RegExp(R)` identity over-fold; one under A/B triage).

Two fix branches are being prepared; the hold comes off once the full 140-file list re-passes sequentially on the branch. The ES5 gains (+63) are unaffected — their regression suites stay green under both fixes.

---

ttraenkler:
**Hold released — all 140 validation regressions fixed and re-verified (twice, sequentially, on the integrated head `6a9bec6a`).**

- **105 descriptor-attribute failures → 105/105 pass.** Single root cause: six boolean builtin statics (`Array.isArray`, `Object.is`, `Object.hasOwn`, `Reflect.has/set`, `Number.is*`) were registered without the `boolean` brand, and the funcref-wrapper registry keys one wrapper per wasm signature — so the unbranded statics won the registration race and every same-signature user predicate boxed `false` as the number `0`; propertyHelper's `isConfigurable()` then failed every `=== false` comparison. Fixed by branding the statics + routing a fourth private copy of the closure-result boxing decision through the #4082 helper. Pinned by `tests/issue-4223-boolean-predicate-brand.test.ts` (fails without the fix).
- **33 `regexp-modifiers` "failures" → measurement artifact**, not a regression: they were diffed under Node 22, whose host RegExp predates ES2025 inline modifiers; under CI's Node 25 they pass on both main and this branch. No code change.
- **1 real RegExp regression → fixed**: the `RegExp(R)` identity fold now requires a statically-provable absence of `Symbol.match`/`constructor` overrides (§22.2.3.1 step 4.b), falling back to the clone path otherwise; +2 pins in `tests/es5-standalone-regexp.test.ts`.
- **1 was a wave-3 *fix*, not a break**: the `interp/emitter: unsupported: regex` refusal string only exists on main; this branch removed it.

Known sharp edge documented in-code rather than glossed: a genuine native-i32-*number* closure sharing a signature with a boolean predicate would still mis-box (structurally identical wasm types can't be `ref.test`-separated); unreachable in today's corpus.

CI is running on `6a9bec6a`; auto-enqueue owns the rest once CLEAN.

---

PR #4252, opened 2026-08-08
perf(#4241): carrier-intrinsic $bag slot on the closure family — the 12.98% registry-scan frame goes to 0.22%

ttraenkler:
Step 1 of #4241: give closure carriers their expando bag **intrinsically**, so
the dynamic-lookup path stops walking a global registry.

> **Branch/issue numbering.** The branch is still named `issue-4237-…` and the
> first two commits say `#4237`: this issue was filed as 4237 and renumbered to
> **4241** mid-flight because another session merged a different
> `plan/issues/4237-*.md` to main the same day. Renaming a pushed branch is not
> worth the churn. Unrelated: PR #4237 is a merged PR, and PR #4241 is the
> fnctor-layout flip this issue's baseline measures against — issue ids and PR
> ids share one sequence.

## The problem, measured

`__closure_bag_lookup` was the acorn standalone profile's **#1 frame at 12.98 %
self** — a 696-char leaf doing a linear `ref.eq` walk of the global
`$ClosurePropEntry` registry. Deterministic call census (5 parses, checksum
2110):

| caller | calls/parse |
| --- | ---: |
| `__closure_prop_get` | 20,850 |
| `__instance_prop_get` | 18,588 |
| `__carrier_bag_of` | 17 |
| **total** | **39,455** |

Those are the dynamic read-MISS arms: a lookup that misses the struct/`$Object`
arms consults the expando bag before answering `undefined`. The registry also
never removes an entry — it grew **75/parse** and pinned every carrier that ever
grew a property, so the aggregate scan cost was quadratic over a session.

## The fix

`$bag` — nullable, mutable `externref` — inserted into the closure header at
`CLOSURE_BAG_FIELD_IDX = 2`, exactly the way #3673 inserted `$arity` at index 1,
with `CLOSURE_CAPTURE_FIELD_BASE` bumped 2 → 3 so every capture / TDZ slot
follows by the constants rule rather than a bare literal. Lookup for a slotted
carrier is `ref.test` + `ref.cast` + `struct.get`; a carrier that never grew a
property reads back the null it was born with. Ensure allocates once and
`struct.set`s the slot, so the bag is reachable only from its carrier.

Both helper **bodies** are swapped in place, so all ~340 call sites and every
consumer (instance-props, `carrier-bag-{hasown,define,delete,visibility}`,
instance-tombstones, object-integrity-carrier, closed-struct-extern-set) inherit
the fix through the helper names with zero call-site edits. The slotted/slotless
split is resolved by **field name** on the type table (the `$shape` idiom), so a
root without a slot routes to the registry rather than being mis-read.

## Results (paired, one session, one box — nothing inherited)

| | base | slot |
| --- | ---: | ---: |
| `__closure_bag_lookup` self | **12.98 %** | **0.22 %** |
| wall, 300 parses | 25,579 ms | 21,508 ms (**−15.9 %**) |
| `__extern_get` self | 7.15 % | 7.24 % (flat — step 2 owns this) |
| `dynamic-lookup` bucket | 12.06 % | 13.34 % |

The `dynamic-lookup` **rise** is the pre-registered denominator caveat firing:
removing ~13 % of total self-time inflates every surviving bucket ~1.16×. And
the 300-parse wall figure **overstates** the per-parse gain, since the removed
cost had a quadratic term.

Wall A/B on the perf lane: slot > base in **both orderings** (ratio 0.1468 →
0.1567 and 0.1336 → 0.1789), but the box was loaded enough that native Node's
own time swung 67 % across runs, so only the **sign** is claimable, not the
magnitude.

## The finding that redirects the next slice

The plan expected this to kill the registry and the leak wholesale. **It does
not.** All 75 `__closure_bag_ensure` calls/parse arrive via
`__instance_prop_set` — user class / `__fnctor_` / `__anon_` instance carriers,
which are not in the closure family. So registry **population** is unchanged;
registry **traffic** is down 53 %, and that 53 % was the expensive half
(closures never had bags, so every closure consult was a full-length miss, while
instance consults mostly hit near the prepended head). Step 1b — the
instance-carrier slot, with its two viable seams — is written up in the issue.

## Validation

- **Cold-tail differential, build-vs-build: bit-identical.** All 64 per-field
  rolling hashes and every presence count agree across 32,506 AST nodes, in both
  `computed` and `copy` read modes.
- Cold-tail copy-vs-computed: the same 5 residuals with the same counts on both
  builds. No new divergence.
- **Equivalence gate: 8/8 shards, 0 regressions** (re-run post-merge).
- 206 tests across the #3468 / #4194 / #4225 / #3920 / #3201 / #4010 / #3673 /
  #2864 / #4122 suites pass.
- New `tests/issue-4241-carrier-bag-slot.test.ts`, 15 cases — pins the behaviour
  **and** the field shift (captures at their new index, `bfnstate`/`bfnid` after
  their 2→3 / 3→4 move, `$arity` still answering `.length`).
- Three pre-existing failures were confirmed identical on the pre-change build
  rather than assumed: #2984's own `KNOWN GAP` case, `add.bind(null,1)`
  invocation trapping, and `g.apply(null,[…])` answering 0.

Two `struct.new` sites were missing the new operand and **both were found by the
failure being loud**, not by audit — `$__bound_fn`'s second allocation site
("need 4, got 3") and the method-closure singletons ("need 3, got 2", surfaced
by the `equality-mixed-types` shard, nine failures from one site). That is the
argument for inserting the field rather than smuggling it in.

Issue status stays `ready`: step 2, the receiver-stamp `br_table`, is still open.


ttraenkler:
Unparking after full diagnosis (recorded in plan/issues/4241-extern-get-receiver-stamp-dispatch.md): the −120 floor breach decomposed as −17 pre-existing main drift + −103 from two promise-executor struct.new sites missing the new $bag operand (115 pass→compile_error, one signature, whole-population check: the only new compile errors in all 48,619 rows). Fix pushed (2001eb425), kill-switch verified. 9 of 141 regressed files are locally unmeasurable (eval provider absent in the container) — the merge_group re-run is the authoritative verdict on those. Single re-enqueue via the normal auto-enqueue path once CLEAN, per park protocol.

PR #4255, opened 2026-08-08
feat(#743): derive types by default — flip the derivation flag family ON

ttraenkler:
Stakeholder decision, 2026-08-08: **"derive types always; consumers arrive later."**

Every "flag verdict: STAYS OFF" in `plan/issues/743-*.md` was decided on a *benefit* criterion — the pass moved no slots, no allocations, no wall time on acorn. That criterion is retired. So the bar for this PR is not "does it pay?" but **"is it free?"**, and the two numbers that answers were the two nobody had produced: this family has never been run against test262, and its compile-time cost had never been quantified at all.

## Flag inventory and disposition

Identified from **code**, not from the issue prose — the records name more levers than exist as gates.

| flag | gates | disposition |
| --- | --- | --- |
| `JS2WASM_FNCTOR_CTOR_PARAM_TYPES` (#4117/#743) | satellite graph + `new`-site **parameter** narrowing in both inference lanes | **flipped ON** |
| `JS2WASM_DTS_ENTRYPOINT_SEEDS` (#743) | `.d.ts` entrypoint parameter seeds | **flipped ON** |
| `JS2WASM_FNCTOR_TYPED_READS` (#4155 P2) | typed `struct.get`/`struct.set` for a struct-typed receiver | **flipped ON** |
| `JS2WASM_FNCTOR_TYPED_BINDINGS` (#2660 S3b) | binding retype to `(ref null $__fnctor_F)` | **flipped ON** |
| `JS2WASM_FNCTOR_CTOR_PARAM_SLOTS` (**new**) | `inferFnctorFieldTypeFromCtorParam` — ctor-param fact → physical field **slot** | **EXCLUDED, opt-in `=1`** |
| `JS2WASM_FNCTOR_TYPED_INSTANCES`, `JS2WASM_FNCTOR_LAYOUT_EMIT` | — | already ON (2026-08-04 / #4241), untouched |

Flip pattern is the #4241 layout-emit idiom verbatim: **unset ⇒ ON**, explicit `0`/`off`/empty ⇒ OFF. The unset-spelling tests invert with the default, as the layout-emit test did.

The #4246 satellite passes (`fnctor-receiver-provenance.ts`, `fnctor-local-bindings.ts`, `fnctor-string-producers.ts`) have **no gate of their own** — they are unreachable except through `JS2WASM_FNCTOR_CTOR_PARAM_TYPES`, so nothing separate to flip.

## The one sub-lever excluded, and why

`inferFnctorFieldTypeFromCtorParam` types a field's physical slot from the **constructor's** write and consults nothing about writes reaching that field elsewhere:

```js
var A = function A(n) { this.tag = n; };
var a = new A(1);
a.tag = "s";
typeof a.tag === "string";   // 1 with the lever off  ·  0 with it on
```

**Both arms have the hole**, probed separately — the `this.f = <param>` arm and the `this.f = this.<y>` field-fact arm. The field-fact arm was expected to be safe (the satellite's field pass joins over reaching writes); it is not, because that pass enumerates `this.<f>` writes inside the owner's own methods and `a.mark = "s"` through an instance *binding* is not in that set.

**The defect class is not new.** The identical wrong answer is reachable on `main` today with every flag off whenever the constructor writes a literal (`this.tag = 1` derives f64 and loses the later string write the same way — verified on `15c3c9375`). What the lever changes is the **population**: it extends the hazard to opaque-parameter constructor writes, the normal shape in real JS and precisely the acorn case it was built for. Against that, the slots it recovers measured **zero** value-level effect (#4246: `$AnyValue` 22,008 of 233,320 allocations per parse, identical flag-on and flag-off) for +124 B. Enlarging a silent-wrong-answer class in exchange for a measured null is the wrong trade, so the **default stays sound** and the lever is opt-in.

The exclusion is deliberately narrow: it does **not** gate the `new`-site *parameter* narrowing. A parameter's writes ARE its call sites, which the existing conflict / under-application / escape rules already enumerate.

**#4250** is filed for the whole-program per-field write-kind verdict that fixes both populations and unblocks the flag — with the `main`-baseline repro, a mechanism sketch (the fixpoint already enumerates reaching writes — the 78→56 counts), and acceptance criteria. Cross-linked from #743.

## Two real defects the flip found (both fixed here)

Invisible while the flag was OFF — nothing exercised the combination.

1. **Presence-tracked fields were narrowed to a machine slot.** `function T(k){ this.keyword=k; if(k>100){this.opt=k;} }` — `$opt` derived `(mut f64)` *while keeping its `$$presence_0` bit*, and `this.type.opt === undefined ? -1 : this.type.opt` returned a boxed opaque object where the flag-off compile returned `-800`. A wrong answer, not a slower one. #3683 S4a and #3753 S1 both already carve out `onlyConditional` fields; this adds the third. It must be a revert *after* the constructor walk, because presence-tracking is not decided when the field type is chosen.
2. **The narrowing fired in the JS-host lane**, where its trust argument ("the module owns every write") does not hold. Now gated to standalone — matching `numericPropertyNames` (standalone-only), `fnctor-typed-bindings.ts` admission rule 1, and every measurement this pass has been given.

A third finding is an **instrument** issue: the #3683 S4a differential suite builds its control lane with `JS2WASM_NUMERIC_FIELDS=0` to reproduce pre-S4a shapes, and #743 reaches some of the same slots by a different proof. With it on, **nine pins failed on their BASELINE assertion** — which reads as "S4a broke" when S4a had not moved. That suite now pins `JS2WASM_FNCTOR_CTOR_PARAM_TYPES=0` in *both* lanes so it keeps measuring one variable.

---

# Evidence

## 1. CI conformance pair — artifact vs artifact

Never against the committed baseline: a dispatch run's own regression gate diffs a config-mismatched baseline and prints thousands of phantom regressions (#4239). Both sides are `test262-merged-report` artifacts from two `test262-sharded.yml` dispatches with **identical default inputs**, differing only in this PR's commits. Both report `oracle_version: 13`.

| | base | tip |
| --- | --- | --- |
| run | [31277942149](https://github.com/loopdive/js2/actions/runs/31277942149) | [31281721914](https://github.com/loopdive/js2/actions/runs/31281721914) |
| sha | `5586d72d6` (docs-only ⇒ compiler ≡ `main`) | `b15fc0bd9` |

**js-host** — 48,619 rows both sides

| | base | tip |
| --- | ---: | ---: |
| pass | 32,173 | **32,202** (+29) |
| fail | 15,633 | 15,634 |
| compile_error | 464 | 466 |
| compile_timeout | 234 | 202 |

240 flips, **240 in the timeout band, 0 non-timeout**: `compile_timeout→pass` 107, `pass→compile_timeout` 78, `compile_timeout→fail` 27, `fail→compile_timeout` 26, `compile_timeout→compile_error` 2. Every single one has `compile_timeout` on one side.

**standalone** — 48,619 rows both sides

| | base | tip |
| --- | ---: | ---: |
| pass | 28,954 | 28,950 (−4) |
| fail | 13,190 | 13,189 |
| compile_error | 6,341 | 6,342 |
| compile_timeout | 19 | 23 |

23 flips: `pass→compile_timeout` 10, `compile_timeout→pass` 5, `fail→compile_timeout` 3, `compile_timeout→fail` 3, `compile_timeout→compile_error` 1 — and **one non-timeout flip, a `fail→pass` gain** (`test/built-ins/Function/15.3.5.4_2-90gs.js`, previously `TypeError: Access to strict function 'caller' is forbidden`). The −4 net is exactly the +4 net additional compile_timeouts.

**Zero non-timeout regressions in either lane.** The only non-timeout movement anywhere is one test that started passing.

Row counts are floored and equal on both sides (48,619/48,619, 0 present-in-one-only) so an empty or truncated side could not read as "no flips".

## 2. Compile-time delta — acorn standalone, `-O3`

`tests/dogfood/acorn-standalone-compile.mjs`, which reports its own `compileMs`. **One process per rep** (a long-lived process accumulates JIT state and the second arm would inherit the first's warm-up), **interleaved with the order alternating each round**, OFF arm **spelling `"0"`** (unset is now the ON arm), and a **loadavg gate before every rep** because other agents share this box.

Two sets, 16 compiles total:

| set | reps/arm | ON median | OFF median | delta |
| --- | --- | ---: | ---: | ---: |
| loaded (gate < 6.0) | 5 | 133,439 ms | 127,035 ms | +6,404 (+5.0 %) |
| quiet (gate < 3.0) | 3 | 118,234 ms | 121,824 ms | −3,590 (−2.9 %) |

Ranges overlap in both sets. Per-round **paired** ON−OFF differences, all 8 pairs:

`−19,953 · −10,771 · −9,689 · −6,119 · −4,023 · −401 · +6,404 · +8,688` ms — median **−2,212 ms**, mean **−4,483 ms**.

**Read: no measurable compile-time cost.** The two summaries disagree in sign between sets (+5.0 % loaded, −2.9 % quiet), which is itself the proof that this box does not resolve an effect of this size; the combined paired estimate is slightly *negative*, which is not a real speedup — it is zero within noise. What can be claimed is a **bound**: a cost larger than roughly ±8 % of a ~2-minute compile would have separated, and none did.

This is consistent with what the code does after the exclusion: with the slot lever off the satellite graph has no consumer and does not execute, so the ON arm's extra analysis is the `new`-site parameter narrowing plus the seeds/reads/bindings consumers — not the #4246 fixpoint.

## 3. Binary size — the one clean, separated measurement

acorn standalone `-O3`: **1,041,731 B → 1,046,104 B = +4,373 B (+0.42 %)**, *identical in all 8 reps of each arm* (zero variance). Canaries `2,3,4,5` and `imports 0` in both arms throughout — the artifact stays behaviorally intact.

Part of this is a known, previously-measured cost of splitting the ctor-param halves: `param-return-inference.ts` records that narrowing the parameter while the slot stays boxed re-boxes on every store (+27 B on a one-field fixture). This PR ships exactly that configuration deliberately, and this is its acorn-scale price.

## 4. Local suites

180/182 in the affected set (#743 ×9, #4155 P2, #2660 S3b, #3683 S4a, #4194, #2660 ×3, #3683 twin, #4155 provenance). The 2 failures are **pre-existing on unmodified `upstream/main` @ `15c3c9375`**, verified by checking out main's `src` and re-running: `#3683 S4a > excludes presence-tracked (conditionally assigned) slots` and `#3683 S4a > a promoted slot still marshals through the reflection arms`. A third, `#4155 Phase 0 > field added by a method, never seeded in the ctor → NaN`, is an `it.fails` that no longer fails on main either — also pre-existing, also not touched here.

`tests/issue-743-derivation-defaults.test.ts` is new: unset ⇒ ON for all four, per-pass predicate ≡ leaf predicate on every spelling, only `0`/`off`/empty disable, each flag independently disablable, and the slot lever OFF unless an explicit `1`. It runs no compiles, so it fails first and unambiguously if a default is re-inverted.

---

https://claude.ai/code/session_012Mjntn1rYJ7sH8kgfFDoqM


ttraenkler:
## Follow-up: `quality` (#3008 changed-root gate) — fixed, plus a sweep

The gate failed on `tests/issue-3683-numeric-fields.test.ts`, 2 of 15 red. My original "pre-existing on main" note was reached the **wrong way** — I restored main's `src` but kept my own test file, which compares main's compiler against my expectations and cannot distinguish "pre-existing" from "my flip inverted an unset-env assertion".

Re-established properly, two ways:

- **pristine `upstream/main` @ `d0019f86e`** (src *and* test file, nothing of mine): same two failures, **same assertion values**;
- a probe across **all four** derivation-flag configurations (family OFF / new defaults × S4a on / off): identical results in every cell.

So: stale assertions against main's own drift, not the inversion hazard. Both fixed, file now **15/15**, `pnpm run test:changed-root` exits 0 across all 11 changed root files.

| assertion | was | now | why it drifted |
| --- | --- | --- | --- |
| `fieldType(wat,"A","$has_maybe")` | `"i32"` | `presenceWordTypes() == ["i32"]` | presence became a **packed word** (`$$presence_0`) instead of one boolean companion per field, so the old lookup had been reading `undefined` — it could no longer fail for the right reason |
| reflection bitmask | `3` | `11` | `Object.keys` over a standalone fnctor instance started working (arm `8`); `getOwnPropertyDescriptor` (4) and `JSON.stringify` (16) are still gaps |

### Sweep for the same hazard elsewhere

**The specific delete-the-var pattern is fully swept:** 12 files reference the five family flags, and this PR touches **all 12**. Zero untouched files carry it.

Widening to the plausible blast radius — the **19** root test files that assert on `__fnctor_` struct shapes — of the 16 I do not touch, **14 green, 2 red**, both red on pristine main with identical values:

- `tests/issue-3685-presence-tracked-proven-reads.test.ts` — `expected false to be true`
- `tests/issue-3617.test.ts` — `expected +0 to be 1`

Left alone deliberately and filed as **#4253**: each needs its own diagnosis (stale pin vs real regression — `#3617` reads like a genuine enumerability change), and fixing them here would pull them into this PR's gate and bury that question.

**#4253 also records the structural gap this exposed:** PR CI never runs the root `tests/*.test.ts` population (**2,646** files after the gate's own exclusions), so a file that goes red stays invisible until an unrelated PR happens to edit it — which is exactly how all four of these accumulated. A change-scoped gate answers "nothing wrong" for every file it did not look at.

No compiler behaviour changed in this follow-up: the diff is one test file plus one issue file.

PR #4256, opened 2026-08-08
fix(#4249): ES5 standalone wave 4 — eval-spliced accessor crash + catch-body finally depth (+9 standalone, +4 gc)

ttraenkler:
Wave 4 of the ES5-standalone-90 % program (`plan/goals/es5-standalone-90.md`).
Issue: `plan/issues/4249-standalone-eval-accessor-crash-finally-continue.md`.

> **Stacks on #4249** (the wave-3 PR, branch `claude/test262-es5-pass-rate-vdseyg`),
> which is still open. This branch was cut from it and re-merged at its current
> head, so the diff-vs-main includes wave 3 until that lands. **Enqueue only
> after #4249 merges.** The wave-4 content is the last three commits.

### 1. An eval-spliced object-literal ACCESSOR crashed codegen

`Internal error compiling expression: Cannot read properties of undefined (reading 'declarations' | 'escapedName')` — 7 compile errors in `language/expressions/object`, and the crash class is generic (it kills the whole file).

The eval-inline lifter parses the eval string with a bare `ts.createSourceFile`, so those nodes were never **bound** and carry no `symbol`; every checker query resolving through one *throws* instead of returning `undefined`. `allNodesInlineSupported` exists to bail on exactly the node kinds whose codegen reaches the checker — it lists function/arrow expressions and classes and **misses object-literal accessors**, which `emitObjectLiteralAccessorFn` feeds to `compileArrowAsClosure` through an `as unknown as ts.FunctionExpression` cast.

**Widening the eval bail list would have been the wrong fix, measurably:** on the **gc** lane the same splice compiles fine (it routes to `compileArrowAsCallback`, which asks the checker nothing) and **5 of these 12 files PASS through it**. A blanket bail would have traded a standalone crash for a gc regression.

`computeClosureWrapperSig` now answers a never-bound declaration **syntactically** — all-externref params, externref return iff the body has a value-carrying `return` (nested functions not descended into). That is also the right answer for untyped JS.

### 2. `catch { break; } finally { continue; }` fell out of the loop

```js
var n = 0;
while (n < 5) { n += 1; try { throw "x"; } catch (e) { break; } finally { continue; } n += 100; }
// n must be 5; was 101
```

The finally body is pre-compiled once at `+1` (the try block) and cloned into each control-flow path; `finallyInlineDelta` retargets a clone by `current − baseline`. The catch-body `finallyStack` entry snapshotted its baselines **after** the extra `+1` for the inner try that wraps the catch body — so the baseline read `+2`, matched the break site exactly, and every break/continue inside a catch body inlined the finally at **delta 0, one label too shallow**. The cloned `continue` branched to the enclosing **try** instead of the loop, leaving the try and running the statements after it. The sibling catch_all sites already pass an explicit `cloneFinallyAtDepth(1)`; the derived-delta site is the one that got it wrong.

Fix: snapshot the baselines **before** that `+1`.

## Measured (sequential, per bucket, `runTest262File`, Node 25)

| bucket | files | base | after | Δ |
| `language/expressions/object` | 12 | 0 pass (7 CE) | **5 pass, 0 CE** | **+5** |
| `language/statements/try` | 8 | 0 | **4** | **+4** |
| `built-ins/RegExp/prototype/exec` | 15 | 6 | 6 | 0 |
| `built-ins/RegExp` residue | 35 | 9 | 9 | 0 |
| `language/types/object` | 13 | 1 | 1 | 0 |

**+9 standalone, zero regressions.** gc-lane A/B against the unmodified sources: `language/expressions/object` 6 → 6 (unchanged, as designed), `language/statements/try` 3 → **7** (+4 collateral — the finally fix helps both lanes).

Node 25 is mandatory in these directories: Node 22's host RegExp rejects ES2025 modifier syntax during early-error validation and manufactures phantom failures (#4233).

## Verification

- New: `tests/issue-4249-eval-accessor-and-finally.test.ts` — 22 tests (both root causes; both lanes for the accessor shapes; while/for/do-while × module/function scope; the non-abrupt-finally and `finally{break}` counter-cases; a labeled-break guard for the `bumpOuterBranchDepths` interaction).
- `tests/es5-standalone-*.test.ts` — all 10 wave suites, **150 passed**, including the 23 #4233 pins (brand guard).
- `scripts/equivalence-gate.mjs` — **no new equivalence regressions** (and 12 baseline failures now pass; baseline deliberately not ratcheted here).
- try/finally suites (`finally-block`, `finally-duplicate`, `issue-1858`, `issue-2061`, `issue-2903`, `issue-2906-3c`, `issue-2906-gap3`, `issue-993`, `try-catch`, `issue-2623`): unchanged. Their pre-existing failures were A/B'd against the unmodified files and fail identically on the base.

## The largest lever found, deliberately NOT taken

Standalone throws error **strings**, not `Error` objects (`typeErrorThrowInstrs`, `src/codegen/property-access.ts`), so every `e instanceof TypeError` answers false. **42 ES5 standalone failures** carry that signature (23 raw `TypeError: Cannot access property on null or undefined` + 19 `e instanceof TypeError`) — larger than anything left in these buckets. It is a generic path with suite-wide blast radius and belongs to the `error-model` goal; the issue file sizes it and names the second, independent half (native-proto method closures are not dispatch candidates, so a bare or element-access call never reaches the brand-recovery prologue — confirmed from the emitted WAT, where the argument is even passed in the `this` slot).

Other leftovers with their mechanisms named in the issue file: the four `exec`/`test` brand receiver shapes, `obj.prop++` on a non-numeric slot (a representation decision, not a coercion tweak), the for-in overlay-enumeration family, and the non-`Error` catch-binding reads.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PmwHrtyzm28H9YuH2nYLYy)_


ttraenkler:
**Deliberate `hold` (program coordinator) — predecessor-stacked on the wave-3 branch.**

This PR's diff-vs-main currently includes all of wave 3, which is in the merge queue as #4249. Holding so auto-enqueue doesn't race two entries carrying the same base diff. Once #4249 merges, the hold comes off, this PR auto-refreshes to its true ~+9-flip delta, and the queue takes it normally.

---

PR #4258, opened 2026-08-08
feat: ES5 standalone conformance — wave 4 (#4246 this/new, #4247 array keys, #4248 wrapper protos, #4251 harness canary, #4252 computed-key calls)

ttraenkler:
## Goal

Continuation of the ES5-standalone-90% program (waves 1–3 merged as #4234 and #4249). This consolidates five finished wave-4 packages (~+47 measured standalone flips plus gc collateral). The #4249-issue package ships separately as PR #4256. Spec: `plan/goals/es5-standalone-90.md`.

## Work packages

| Issue | Flips | What |
| --- | --- | --- |
| #4246 | +15 | `new` on a non-constructor throws TypeError (not a crash); sloppy-mode `this` is ToObject(thisArg) per §10.4.3; inlined function-expression callees bind their receiver |
| #4247 | +7 standalone, +7 gc | non-array-index element keys (`a[4294967295]`, both spellings) route to the array's named store instead of saturating/silently dropping — §10.4.2.2 had two halves, the string spelling invisible because write and read saturated to the same wrong index |
| #4248 | +25 | builtin wrapper prototypes as wrapper objects: own members visible to `hasOwnProperty` (CSV token scan), `[[PrimitiveValue]]` for the three prototypes, inherited-method value reads answer the prototype's own function object |
| #4251 | canary | standalone harness self-test ratchet suite: pins pass/fail of every test262 `harness/` self-test with root-cause notes; regressions AND silent fixes both fail the suite until the table is updated |
| #4252 | +2 self-tests | `obj[runtimeKey]()` on a plain-object receiver silently dropped the call (pushed `ref.null.extern`) — a vacuous-pass generator; both tail-dispatch fallback arms now route through `tryEmitInlineDynamicCall` behind a measured plain-object oracle gate |

Two canary notes: `verifyProperty-restore.js` flipped to `pass` in this PR (fixed by wave-3's #4230 between recording and integration — the ratchet caught it), and #4252's fix raised the harness floor 44 → 46 of 116.

## Validation

- Per package: sequential per-file A/B via the `runTest262File` seam on Node 25, zero regressions; #4247 verified both lanes (+7/+7); #4248 gc lane byte-identical (same sha256); #4252 demand gate proven by byte-hash A/B on five non-target shapes.
- Post-integration on the consolidated branch: `tsc` clean; suites `es5-standalone-this-and-construct` (22), `es5-standalone-harness-selftests` (19), `es5-standalone-array-nonindex-keys` (22), `es5-standalone-wrapper-prototype` (6), `issue-4252-computed-key-call` (7) all green; all quality gates pass locally with grants documented in the owning issue files.
- The merge queue's `merge_group` re-validation remains authoritative.


ttraenkler:
Park diagnosed and fixed; removing `hold`.

**The merge-group failure was real and is #4257** (`plan/issues/4257-late-spliced-arms-emit-undeclared-ref-func.md`): the declarative element segment is collected by a single mid-finalize scan that runs *before* the late-spliced arm fills, so a `ref.func` whose only occurrence is in a late arm (here: #4248's method-identity arm referencing the `Object.prototype.toString` closure) is never declared, failing WasmGC validation with `undeclared reference to function #580` — but only on the full runtime-eval provider module, where the usual accidental cover (another scanned site declaring the same closure) was removed by #4256's `computeClosureWrapperSig` change. Neither PR is wrong alone; only the merged state failed, which is exactly what the merge-group re-validation exists to catch. Third instance of this class (#2963, #2175 hand-patched it per-arm).

Fix (`558a6cb2`): `collectDeclaredFuncRefs` re-runs **additively** at the end of each driver's fill sequence, before dead-elimination — shift-safe by construction, byte-identical on modules that were already covered. Verified on the merged state: provider builds + canary-verifies (was failing at the parent commit), `tsc` clean, 115/115 across the pinned wave-4 suites plus the new `tests/issue-4257-late-funcref-declaration.test.ts`, budget gates green.

---

PR #4313, opened 2026-08-09
feat(npm-compat): advance real package execution frontiers

ttraenkler:
## Summary

This publishes the accumulated npm-compat work and suspends further package optimization with reproducible handoffs.

- advances generic compiler support uncovered by Hono, Redux, React, Lodash, UUID, and TypeScript package graphs
- runs real upstream test bodies where available: React is 64/64 scored, the selected ESLint unit lane is 44/44, UUID is honestly 6/75, and Redux's runtime workload is 1/1
- reports jsdom as 0/318 executed while its implementation still cannot emit a binary, and keeps ReactDOM's entry barrel distinct from its failing implementation
- adds the consumer-driven TypeScript source experiment (82 -> 31 files) without claiming compilation; the dynamic probe still emits no binary
- gives only Stylelint an explicit Node filesystem capability and labels that capability on the npm-compat card
- records exact suspended frontiers in the catalog issue plus dedicated Hono private-brand, package async-CFG, and Stylelint TDZ markdown issues

The public npm-compat report is not populated with cached or hand-entered results. Existing standalone harnesses that are not yet integrated into the generator remain documented as integration work.

## Remaining frontiers

- Hono: valid Wasm, then a runtime private `#routes` brand failure
- ReactDOM: malformed `updateForwardRef` local frame; zero scored tests
- jsdom: no binary within the 180-second budget; zero of 318 tests execute
- TypeScript: CPU-active through the 300-second cap after body compilation, but no binary
- Prettier/Axios/Stylelint: safe refusal for unsupported `await`-inside-`try` shapes
- styled-components: compiles but emits an invalid captured-reference local
- Stylelint additionally reaches a module-TDZ ordering failure

The exact commands, timings, branch state, and resume steps live in:

- `plan/issues/3995-pin-and-adapt-original-upstream-test-suites-for-catalog.md`
- `plan/issues/4301-hono-private-router-brand.md`
- `plan/issues/4302-package-async-await-inside-try-shapes.md`
- `plan/issues/4303-stylelint-module-tdz-global-order.md`

## Validation

- 36 changed Vitest files: 36 passed; 128 tests passed, 11 intentional skips
- `pnpm run typecheck`
- `pnpm run check:ir-fallbacks`
- `pnpm run check:issues`
- `pnpm run check:issue-ids:against-main`
- `pnpm run check:issue-ids:against-open-prs`
- full pre-commit and pre-push hooks
- branch-wide `git diff --check origin/main...HEAD`

Performance regressions are reported but are not a merge gate.


ttraenkler:
Shepherd note — this PR is now **`mergeable_state: dirty`**, and the conflict needs a decision from whoever owns the host-call ABI.

It read `clean` earlier only because that was computed against a stale base. Merging current `main` in produces one conflicted file, `src/codegen/expressions/calls.ts`, with two hunks that are the same collision:

```
<<<<<<< HEAD  (main)
    ensureHostCallFallbackImports(ctx, hostCallPlan);
=======       (this branch)
    ensureLateImport(ctx, "__js_array_new", …);
    ensureLateImport(ctx, "__js_array_push", …);
    ensureLateImport(ctx, "__call_function", …);
>>>>>>>
```

```
<<<<<<< HEAD  (main)
    dispatch = buildHostCallFallbackArm(ctx, fctx, hostCallPlan, anyLocal, argLocals) ?? dispatch;
=======       (this branch)
    …open-coded arm building a host args array via __js_array_new / __js_array_push,
      then calling __call_function…
>>>>>>>
```

`main` has refactored the host-call fallback into `ensureHostCallFallbackImports` / `buildHostCallFallbackArm`. This branch still open-codes the **three-crossing `__js_array_new` / `__js_array_push` / `__call_function` ABI** — the one #4295 replaced today with fixed-arity `__call_function_N`. This branch predates that landing.

So the resolution is a semantic call, not a mechanical one: should this dynamic-dispatch site adopt `buildHostCallFallbackArm`, and is `hostCallPlan` in scope where it is needed? Taking `HEAD` blindly may drop imports this branch relies on; taking this branch's side reinstates the ABI #4295 deliberately removed. I have not attempted a resolution — per CLAUDE.md, `src/**/*.ts` conflicts route to a senior dev, and a wrong resolution here would also silently corrupt any regression measurement taken afterwards.

**Why this blocks more than the merge:** `auto-enqueue` skips `DIRTY` PRs entirely, so while this stands the PR cannot enter the queue, cannot run a `merge_group`, and therefore cannot produce a regression report. Removing the `hold` label would not help.

Once the conflict is resolved, the park is actually useful: **#4359 landed today**, so a `merge_group` run now emits `merged-reports/test262-regressions-detail.txt` listing every regressed file. Previously the gate ran `diff-test262.ts --quiet` and recorded only counts — which is why the earlier "222 regressions" could never be triaged. The remaining verdict blockers are unchanged and tracked in **#4352**: the oob trap-ratchet +1 on `test/built-ins/Temporal/PlainDateTime/from/limits.js` (baseline `fail`, so the gate's own policy points at a named `trap-growth-allow`, #3596), and the catastrophic guard's raw 221 > 200 on a net-positive (+41 / +52) diff.

The `hold` label stays on deliberately.

---

ttraenkler:
Removing the `auto-park-bot:merge-group-failure` hold, on the owner's instruction to admit this as an exception. Diagnosis first, per the park-handling rules — the cited failure was investigated, not waved through.

**What was measured.** This branch has been merged onto current `main` (the conflict against main's `host-call-fallback.ts` extraction was resolved by keeping main's helper structure and porting this branch's two semantics into it: the `__extern_is_undefined` import and a real `undefined` for the bare-call `thisArg`, threaded in as an optional parameter rather than hardcoded). `lint` and `typecheck` are green; a `format:check` failure introduced by that resolution is fixed.

Reproducing the cited `oob` growth locally against current `main`, for `test/built-ins/Temporal/PlainDateTime/from/limits.js`:

| target | `main` | this branch |
| --- | --- | --- |
| gc / host | `RuntimeError: dereferencing a null pointer` (null_deref trap) | `Test262Error: Expected a RangeError but got a ReferenceError` — clean failure, **no trap** |
| standalone | `compile_error` (#2046) | `compile_error`, identical — excluded by #3595 |

So on the current merged state the file moves *out* of the trap categories, not into them, which is consistent with the gate's own `null_deref 1629 → 145`. The three prior gate runs also report byte-identical totals (221 regressions, +52, 32533 → 32585) despite being a day apart, which indicates the measured state had not moved — this branch only reached current `main` tonight.

**No `trap-growth-allow` is being declared.** A declaration carrying `tests:` is machine-verified and verification can only refuse, never admit. Declaring a reclassification that no longer occurs risks `REFUSING baseline push — the change-scoped trap-growth-allow did NOT verify`, which is what wedged promotion permanently on 2026-07-25 (#3644). Writing an allowance for a trap that has stopped happening would also permanently mask a genuine future regression on that file.

The queue therefore re-measures on the current merged state rather than being handed a valve. If the `oob` growth is genuinely gone, this passes. If it re-parks, the run will name what still fails against *current* `main` — which is the fresh data point #4352 has been missing.

Note this does not address #4352's second blocker: the catastrophic guard tripping at 221 raw regressions against a threshold of 200, on a net-positive diff (+41 fine-gate, 262 improvements). That guard counts raw regressions and ignores improvements, and whether it should consider net — or whether this PR should be split — remains an open decision. If it fires again, that is the reason, and it needs a policy call rather than a re-admission.

---

ttraenkler:
## Fixed — one blocker, not two, and the valve is now declared and machine-verified

Pushed `d2c0c11` to this branch. Diagnosis first, per the park-handling rules.

### The catastrophic guard was never an independent blocker

This is the part that kept the PR looking twice as stuck as it was. `Catastrophic regression guard (#1668)` is **downstream** of the fine gate, not parallel to it — `test262-sharded.yml` branches on `diff-test262`'s exit code and says so in its own comment:

> `exit 0` → the script's gate PASSED → … (authoritative, #3303)
> `exit 1` → the script's gate FAILED → apply the coarse catastrophic threshold

and on the pass branch, at 221 > 200, it prints *"Raw count exceeds 200 but the script's own gate approved it (deliberate re-baseline)"* and **exits 0**.

So 221 > 200 fails **only because `diff-test262` already exited 1**. On run [`31456242513`](https://github.com/loopdive/js2wasm/actions/runs/31456242513) the sole `GATE FAIL:` line is the oob trap ratchet — the #3457 ratio warning is already WAIVED on the net-positive diff. Clear the ratchet and the guard passes at the same 221. **No net-awareness policy call is needed to land this, and the PR does not need splitting on this account.**

### The ratchet failure is a reclassification, and it verifies

I checked the authoritative baseline directly (`loopdive/js2wasm-baselines` `test262-current.jsonl`, 48,735 entries) rather than reading the gate's one-line summary:

```json
{ "file": "test/built-ins/Temporal/PlainDateTime/from/limits.js",
  "status": "fail",
  "error_category": "null_deref",
  "error": "dereferencing a null pointer [in __module_init()]" }
```

The test was **already trapping**, as `null_deref`. It did not start trapping — it moved *between* trap categories, which is the same movement that takes `null_deref` from **1629 → 145** in that very run. The #3189 ratchet counts per category, so it sees the `oob` +1 without seeing the −1484.

The declaration is a `trap-growth-allow` (#3596) in `plan/issues/4352-…md`'s frontmatter, which is exactly where the gate's own policy text routes a baseline-`fail` file.

**Verified before pushing, not after.** #3596 can only refuse a claim, never admit one, and an unverifiable declaration wedges baseline promotion (#3644) — so I drove `evaluateTrapReclassification` with the real baseline row and the run's real growth numbers. Zero failures:

```
=== trap-growth-allow (#3596): reclassification VERIFIED for 1 declared test(s) —
    each was non-passing on the baseline, and no undeclared trap growth was observed. ===
```

All three conditions hold: **named** · **non-passing on the baseline** · **complete** (the run grew exactly one category by exactly one file, and it is the file named).

### On the 03:38 refusal

The allowance was declined earlier this morning on the grounds that the trap "no longer occurs", from a local run showing a clean `Test262Error` on gc/host. The merge_group 25 minutes later reported `oob 35 → 36` on the same file — against a baseline the gate itself certified content-current (*"0 test262-relevant commits separate the baseline from main HEAD"*), which rules out drift. Whatever that local run measured, it was not the merged state the queue gates on. This declaration is made on the queue's measurement.

### One trap worth propagating

`parseFrontmatterCountReason` is a **line-based** reader — it ends the block at the first line that is not `key: value`. My first draft wrapped `reason:` across several lines, which silently dropped the nested `tests:` list, leaving a declaration that parses as valid but **unnamed** — precisely what #3596 refuses. It would have re-parked with a confusing "must NAME the reclassified tests" error. Caught by running the parser rather than by eye. Keep `reason:` on one line.

Removing `hold` now. If it re-parks, the run will name which condition failed — that is a fact about the declaration, not grounds to re-litigate the diagnosis.

---

PR #4338, opened 2026-08-09
fix(runtime-eval): carrier-wrap closure VALUES for the eval seam (#4307)

ttraenkler:
Closes #4307. **Stacks on #4335** (`issue-4245-membrane-slice1`), which itself stacks on #4321 — those land first; this branch carries their commits.

## The defect

#4245 slice 1 made a compiled top-level function **DECLARATION** callable from evaluated code, because `__module_init` seeds its module global with the #2928 AOT-callable carrier. A closure **VALUE** is not a declaration and never met that adapter, so it crossed the seam as a raw module-local closure struct. The membrane classifies callability with the *adapter's own* closure classifier, and a caller's closure wrapper hierarchy is module-local by construction — so `typeof` answered `"object"` and calling it was a loud TypeError inside evaluated code.

## Root cause

Three caller-side sites hand a value to the provider; only one wrapped.

1. `emitRuntimeEvalGlobalBindingPushBody` applies the AOT-callable adapter **only** in the `ctx.topLevelFunctionNames` arm — a script-scope `var` holding a function expression takes the `else` arm and is pushed raw.
2. `emitStandaloneDirectEvalRuntime`'s `bindingPushInstrs` pushes direct-eval binding cells untouched; `reifyCurrentDirectEvalBindings` boxes them with a plain `coerceType` to `externref`, i.e. `extern.convert_any` of the raw closure.
3. A closure read off a **plain compiled object** — deliberately out of scope, with evidence below.

## The fix

| file | why |
| --- | --- |
| `src/codegen/runtime-eval-callable.ts` | `__runtime_eval_wrap_callable(externref) → externref` (mint the canonical carrier iff the value is a raw closure), its one-entry identity memo, and `emitRuntimeEvalCarrierUnwrapAny` — the AOT-side inverse |
| `src/codegen/expressions/runtime-eval-provider.ts` | call it at the two crossing sites: direct-eval cells (in place) and the script-scope-`var` globals push (write-back) |
| `src/codegen/expressions/calls-closures.ts` | carrier front-guard at the three closure-typed cast sites of `compileClosureCall` |
| `src/codegen/context/types.ts` | one optional field on the carrier record |
| `tests/issue-4307-closure-carrier-wrap.test.ts` | new self-gating lane, 9 cases |
| `tests/quickjs-eval-membrane.test.ts` | retires #4245's pinned residual (`12` → `1042`) |

**Why a helper function, not an inline ladder** — the closure base-wrapper set is not final while an expression is being compiled, so an inline `ref.test` ladder would test a partial hierarchy and silently miss the closures declared after it. The body is rebuilt at finalize, the same discipline the carrier trampolines use.

**Why the cell is mutated in place** — the cell IS the interpreter's write-back target; wrapping a copy would drop assignments made inside the evaluated source. In-place also makes the wrap idempotent, which is what keeps `f === f` true across two evaluations.

**Why the AOT side needed an unwrap** — this is what turned it from "wrap at two sites" into real codegen. `typeof` (#2928) and dynamic method dispatch (#4197) already handled the carrier, but `compileClosureCall` guard-casts the cell/global value to the lifted self-carrier struct; a carrier fails that cast, the cast yields null, and the call **traps**. Unwrapping to `target` restores the fast path with no dispatch and no `__current_this` bookkeeping.

**Why the memo** — `var g = f` gives two cells holding one closure; each would mint its own carrier and `f === g` would read *false* where it reads true today. That is a regression, not a leftover gap.

## Measured (quickjs engine, artifact `d8a5a91d6f183b87`)

| case | before | after |
| --- | --- | --- |
| local `var f = function(){}` — `typeof` in direct eval | `"object"` | **`"function"`** |
| local `var f = function(){}` — `f(20)` in direct eval | TypeError | **41** (compiled body) |
| a function CREATED by eval calls the caller's local closure | TypeError | **21** |
| script-scope `var fn = …` via indirect eval | TypeError | **42** |
| script-scope `var fn = …` via direct eval | TypeError | **42** |
| AOT `f(1)` after the binding crossed | 42 | **42** (traps without the unwrap) |
| AOT `f(1)`, TYPED closure binding | 42 | **42** |
| `f === g` across two evaluations | 11 | **11** |
| non-closure bindings (object/number/string) | 9 | 9 |

**Poisoning check** — reducing *only* `callableWrapHelperBody` to identity, leaving every call site, the memo and the unwrap intact, reverts every fixed case to its exact pre-fix reading. A green run with the change doing nothing is impossible.

**Default path byte-identical** — 126 binaries (playground examples + `tests/fixtures` + `examples/`, each in default/standalone/wasi) hash-identical with and without the change.

**Scoped test262, `language/eval-code/` under the flag: 560 / 816 — UNCHANGED, and unchanged per sub-corpus** (260/286, 48/61, 155/309, 97/160, all identical to #4245 slice 1). No regression, and no gain: the residual in that corpus is EvalDeclarationInstantiation and annexB value/ordering, not closure callability. Stated plainly rather than extrapolated.

**Suites** — the #4307 + membrane + provider lanes: 55 passed. Default-path proof set with no env: 71 passed / 3 skipped. Equivalence gate: *no new regressions*. `tsc --noEmit`, lint, `format:check`, `check:oracle-ratchet` (+0 raw-checker), `check:func-budget` all clean.

## Deliberately out of scope, with evidence

A closure read off a **plain compiled object** (`bag.m`) is **not** caller-side reachable. `__membrane_get` does `target[key]` in the *adapter's* compiled code and answers plain-object reads by walking the caller's structurally-canonical object vector directly. Two probes pin it: an accessor on a compiled object read from evaluated code returns 0 with its side-effect counter still 0 (the caller's getter never ran), and a prototype method reads `undefined`. Fixing it needs the membrane to route object property reads back to the owning module — #4245 slice-2 territory.

## Budget grants

`loc-budget-allow: src/codegen/expressions/calls-closures.ts` (+7) — the guard must sit at the three closure-typed cast sites in `compileClosureCall`; all 30 lines of logic live in the subsystem module, only the import, three one-line calls and the explaining comment land there. The func-budget growth was avoided by extracting `directEvalCellWrapInstrs` rather than taking a grant.

---


ttraenkler:
## `quality` fails on `check:loc-budget` — one line fixes it

[Job 93623167516](https://github.com/loopdive/js2wasm/actions/runs/31440219181/job/93623167516) — the only failing gate. The `calls-closures.ts` grant already in the issue file is honoured; a **second** god-file grew and is not declared:

```
LOC budget gate: intentional growth allowed by this change-set's issue file:
    src/codegen/expressions/calls-closures.ts: 1643 → 1650 (+7) granted by plan/issues/4307-closure-value-carrier-wrap.md

LOC budget gate FAILED (#3102):
  God-files grown past their allowed size:
    src/codegen/context/types.ts: 3643 > 3641 (+2)
```

### Verified: the growth is this PR's, and it is the +2 the description already claims

| | lines |
| --- | ---: |
| `origin/main` | 3641 |
| `scripts/loc-budget-baseline.json` | 3641 |
| this branch | **3643** |

The whole diff on that file:

```diff
@@ -2037,6 +2037,8 @@ export interface CodegenContext {
     interpretedTrampolineFuncIdx?: number;
+    /** (#4307) `(externref) -> externref` closure→carrier wrap helper. */
+    wrapHelperFuncIdx?: number;
   };
```

One doc comment and one optional field on an existing sub-object of `CodegenContext` — exactly *"`src/codegen/context/types.ts` | one optional field on the carrier record"* from the PR body. It cannot be relocated to a subsystem module: the field is on the context type. So this is a declare-it case, not a move-it case.

### The fix

In `plan/issues/4307-closure-value-carrier-wrap.md`, extend the existing `loc-budget-allow:` list:

```yaml
loc-budget-allow:
  # …existing calls-closures.ts comment…
  - src/codegen/expressions/calls-closures.ts
  # One optional field + its doc comment on the runtime-eval carrier record in
  # `CodegenContext` (`wrapHelperFuncIdx`, +2). The wrap helper's index has to be
  # reachable from the context; the 30 lines of logic live in
  # `runtime-eval-callable.ts`.
  - src/codegen/context/types.ts
```

Do **not** touch `scripts/loc-budget-baseline.json` — the gate output says so explicitly (#3131: per-PR bumps make every open PR conflict on that file).

### Scope

**#4338 only.** #4340/#4343/#4346/#4348 are siblings on #4335, not descendants of this branch — all four still read `types.ts` at 3641 and none carries the #4307 issue file, so none inherits this failure.

Every other gate in the job is green: `check:ir-only` READY, `dead-exports` OK, `oracle-ratchet` +0, `pushraw` +0.

I could not push this myself — the branch is on `ttraenkler/js2`, which is outside this session's authorized repository set (`git push` → *"access denied by the git proxy"*; the maintainer `update-branch` API also refuses).

---

PR #4343, opened 2026-08-10
feat(#4308): slice B — EvalDeclarationInstantiation for global-caller and indirect eval, eval-code 624→710/816

ttraenkler:
Slice B of #4308. **Stacks on #4340** (slice A) → **#4335** (membrane slice 1) → #4321 → #4319. Those land first; do not enqueue ahead of them.

## Measured

| run | pass / 816 |
|---|---|
| quickjs, pre-slice-B (base `2c8b8f3fd`) | **624** (re-measured on this tree, reproduces slice A exactly) |
| **quickjs, post-slice-B** | **710** (+86) |
| interpreter (`TEST262_FULL_RUNTIME_EVAL=1`) | **779**, unchanged |

**0 regressions**, diffed test-for-test across all 816 files. Done-signal was ≥700.

## Per-cluster gains

`global:existing-global-update` 16 (direct+indirect) · `global:existing-block-fn-update` 16 · `global:block-scoping` 14 · `lang:var-env-*` / `switch-*-decl-nostrict` / `non-definable-global-*` 16 · **plus 24 `func:*` files** (`existing-var-update`, `existing-fn-no-init`, `existing-block-fn-update`) that were budgeted to slice C and fell out of the widened write-back.

## The residual is one named dependency, not diffuse decay

71 quickjs-only files remain, and **48 of them are EDI-complete and blocked on a single different defect**:

```
Test262Error: Invalid descriptor field: __qjs_handle__
```

That is `propertyHelper.js` rejecting the *descriptor argument*. `verifyProperty(global, "f", {enumerable:true,…})` builds that object literal **inside the QuickJS realm**, and a plain QuickJS object crossing **outward** has no live view, so it arrives compiled-side as `{__qjs_handle__}`. `global:init` moved from *"f should be an own property"* to this message — meaning the binding, its value **and** its descriptor are all correct by the time the harness chokes on how it received the descriptor.

That is **#4245 slice 2** (outward live views), and this measurement puts it on #4242's critical path **ahead of #4308 slices C and D**.

Full residual: 48 → #4245 slice 2 (`global:init` 16, `existing-global-init` 16, `existing-non-enumerable-global-init` 16) · 16 → slice C · 3 → slice D · 4 pre-existing (`lex-env-distinct-cls`, `global-env-rec-eval`; byte-identical errors before and after).

**On the second-order discount** the projection was hedged against: it is real — the target list went ~110 → 62 of its own — but it is **not diffuse**. It is one dependency, countable file by file, with no unattributed remainder.

**The 16 non-enumerable files did not land, but not for P3's reason.** The push widening works: a `defineProperty`'d non-enumerable global is now visible inside the eval, proved in the lane. They sit in the 48-file descriptor block above.

## Trap worth propagating

The call-time globals sync (the 14 `block-scoping` files) first gated on `qjsIntrinsicRealm`, which is only set when a module reads `eval`/`Function` **first-class**. It compiled, linked, ran clean — and gained **exactly zero** files. A full scoped run diffed 0 gained / 0 lost at an identical total of 696.

A totals-only comparison would have read that as both *"does nothing"* and *"harmless"* simultaneously; only the test-for-test diff showed the path was dead. Fixed by recording the realm at the head of `qjsPushGlobals`. This is the third distinct silent-no-op in this workstream, after slice A's `typeof` guard and the #4245 stub fallback.

## Scope and safety

- **Artifact unmoved**: key `d8a5a91d6f183b87`, sha256 `b0662069c241d0430d91c53a3b0e2d1281fd9eb78dd1c93490b0a9dfa70eec5b`; `scripts/quickjs-artifact/` untouched.
- Adapter-only — `scripts/quickjs-eval-provider.mjs`, `tests/quickjs-eval-membrane.test.ts`, and the issue file. **No `src/`**, no `src/interp/`, no shim change. The interpreter, acorn and the IR substrate are untouched, and the interpreter tier is re-measured at 779 as the standing proof.
- The `Object.keys` → non-enumerable widening is deliberately narrow, per P3's recorded hazard: a blanket swap to `getOwnPropertyNames` would leak the lexical-cells carrier plus 8 compiled error constructors into the realm and fight slice A's fix.
- Lane: 19 new readings, **11 of which fail on the pre-slice-B adapter** — anti-vacuity measured in both directions. Full lane 33/33, provider lane 28/28.
- The 15 default-path failures in `issue-2923`/`issue-3632`/`issue-3633` are **pre-existing** — identical counts on the base tree with these files reverted.

#4309 is noted in the implementation record as a signature to recognise (works braced / fails unbraced ⇒ #4309, not slice B); #4341 is **not** merged into this branch.

This branch also carries the probe results (P2/P3/P4/Q5) and the §5/§6 amendments, which were not on its base.

---


ttraenkler:
## Conflict resolved (locally) — one file, three hunks, none of them compiler source

Merging `main` (`4919f502`) into `54776db1` conflicts in exactly one file:

```
CONFLICT (content): plan/issues/4308-quickjs-eval-declaration-instantiation.md
```

`scripts/quickjs-eval-provider.mjs` and `tests/quickjs-eval-membrane.test.ts` auto-merge clean.

### How each hunk resolves

| # | region | take | why |
| --- | --- | --- | --- |
| 1 | frontmatter `status` | **main** | branch predates the update and still reads `status: ready`; main has `status: in-progress` + `assignee: ttraenkler/senior-dev`. #4308 is demonstrably in progress with slices A–D open, and the assignee exists only on main. |
| 2 | trap #5 (#4305) | **main** | the branch carries the **original, wrong** shape ("a succeeding direct eval followed by a throwing one with an `instanceof` catch"). Main carries the correction landed during slice A — the trigger is *static* (`catch` → direct eval → `catch` whose body reads its parameter), root-caused to `fctx.boxedCaptures` being keyed by name. Keeping the branch's text would revert a correction. |
| 3 | `## Slice B — implementation record`, +226 lines at EOF | **branch** | main's copy of the file ends exactly where this block begins, so nothing from main is dropped. Pure append, and it is the substance of this PR. |

Both the Slice A record (line 1094) and the Slice B record (line 1233) are present afterwards.

### One trap for whoever applies this

The tempting shortcut — *"take main's file wholesale, then append the Slice B block"* — is **wrong**. I checked it against the real 3-way merge and it silently drops two branch-only regions that git auto-merges correctly outside the conflict markers:

- `updated: 2026-08-09` → `2026-08-10`
- the 6-line **`AMENDED 2026-08-09 by probe Q5 — DO NOT IMPLEMENT THE SCANNER BELOW`** block above §1.1′

So resolve the three hunks in place; do not rebuild the file from main's copy.

### Gates re-run on the resolved tree

| gate | result |
| --- | --- |
| `update-issues --check` | exit 0 |
| `check-issue-spec-coverage` (#2093) | exit 0 — *all gated done-flips carry a probe/test reference* |
| `check-issue-ids --against-main` | exit 0 |

I have the resolved merge commit but **cannot push it** — the branch lives on `ttraenkler/js2`, which is outside this session's authorized repository set (`git push` → *"access denied by the git proxy"*; the maintainer `update-branch` API 422s; the GitHub API refuses reads on that repo). Same blocker as the loc-budget fix I posted on #4338.

---

PR #4346, opened 2026-08-10
feat(#4245): slice 2 — outward live views for QuickJS objects, eval-code 710→758/816

ttraenkler:
Slice 2 of #4245 — the **outward** half of the eval membrane. **Stacks on #4343** (#4308 slice B) → #4340 (slice A) → #4335 (membrane slice 1) → #4321 → #4319.

## Measured

| run | pass / 816 |
|---|---|
| quickjs, pre-slice-2 (base `2aa44ab21`) | **710** (re-anchored on this tree) |
| **quickjs, post-slice-2** | **758** (+48) |
| interpreter (`TEST262_FULL_RUNTIME_EVAL=1`) | **779**, unchanged |

**All 48 target files landed. 0 regressions, and 0 error-text changes** among the still-failing files — diffed test-for-test across all 816, not by totals. Clusters: `global:init` 16 · `existing-global-init` 16 · `existing-non-enumerable-global-init` 16.

**Artifact NOT moved**: key `d8a5a91d6f183b87`, sha256 `b0662069c241…`. Own-key enumeration and `delete` are realm-side helpers installed once per context, so none of the planned shim exports were needed. Adapter-only, no `src/` change.

## Both designs the plan named were built, measured, and rejected

- **Proxy box** (the decision table's choice) — **rejected**. The premise validation holds as far as it went: an adapter-minted Proxy *is* canonical, and `__extern_get`/`__extern_set`/`__extern_has`/`__getOwnPropertyNames` do dispatch its traps cross-module (measured: read returned 41, `getOwnPropertyNames` returned the trap's list). What it never covered is **`__hasOwnProperty`/`__object_hasOwn`, which have no `$Proxy` arm** — `Object.hasOwn(proxyBox, "a")` measured **false**. `verifyProperty` gates every descriptor check behind that call, so a Proxy box would have turned all 48 files **green having verified nothing**.
- **Per-key accessors** (the plan's own documented fallback) — **also rejected**. `__call_accessor_get` → `__call_fn_method_0` guards only the AOT-callable carrier (#4197); the interpreted-callback marker is recognised by `__apply_closure` alone. Measured: the getter installs (gOPD reports an accessor) and returns null.

Landing the third design instead of the specified one is the substantive deviation in this PR, and it is deviation on evidence: the specified design's failure mode was **silent false-green**, which is the single worst outcome for a conformance slice.

## Two `src/` follow-ups, reported not attempted (per the issue's constraint)

1. **A `$Proxy` arm on `__hasOwnProperty`/`__object_hasOwn`.** This is a real correctness gap for **any user-written Proxy**, entirely independent of eval — `Object.hasOwn` on a Proxy silently answers `false` instead of dispatching the `has`/`getOwnPropertyDescriptor` trap.
2. The interpreted-callback guard on `__call_fn_method_N`.

Either would upgrade this slice from seam-granular to per-read liveness.

## The half that measurement caught rather than reasoning

The descriptor work **alone** read 726 (+16). The two `existing-*` clusters merely moved from `Invalid descriptor field` to `Cannot convert undefined or null to object` — those tests do `var global = fnGlobalObject()` *inside* the eval and then call `verifyProperty(global, …)` from top-level compiled code, and object-valued globals were excluded from the pull, so compiled `global` stayed `undefined`. Widening the pull from FUNCTION to any non-callable OBJECT is what took it to 758. Reasoning about the descriptor path alone would have shipped +16 and called it done.

`qjs_wrapper_gc_handle` now has its caller (outward identity collapse) — the named loose end from slice 1 is closed.

## Residual: the parity gap is 23, not 71

8 `func:existing-fn-update` + 8 `func:no-skip-param` (#4308 slice C) · 3 `var-env-var-strict-*` (slice D) · 2 `lex-env-distinct-cls` + 2 `global-env-rec-eval` (pre-existing).

**Correction to slice B's residual table**: its 16 `func:existing-var-update` files are **both-engines** failures, so they were never part of the parity gap.

Liveness is at **seam granularity**, stated honestly: a QuickJS getter's side effects run at sync time, not at compiled-read time. The full residual list (own-properties-only, arrays, descriptor flags, sync cost, cycles) is in the issue record.

## Anti-vacuity

Build-time canary `outwardProbe` = 6543, whose top digit is exactly the own-keys property that a silently-passing box fails — **verified to fail the build when poisoned** (`returned 3, expected 6543`). 10 new lane cases (43 total, 71 with the provider suite, all green), **4 of which go red when the box read path is poisoned**. Default-path suites green with no env set: 73 passed, 1 skipped.

`src/interp/`, acorn and the IR substrate are untouched; the interpreter tier re-measures at 779.

---


ttraenkler:
## Same conflict as #4343 — identical resolution

`DIRTY` against main (`842b472c3`), 9 commits behind. One conflicted file, inherited from the stack parent:

```
plan/issues/4308-quickjs-eval-declaration-instantiation.md
```

Three hunks, byte-for-byte the same regions as #4343, and they resolve the same way: **main** for the frontmatter `status` (`in-progress` + `assignee`), **main** for trap #5 (#4305 — the branch carries the superseded shape), **branch** for the slice records appended at EOF.

Full reasoning and the gate results: **[#4343's comment](https://github.com/loopdive/js2wasm/pull/4343#issuecomment-5247169454)**.

Cheapest path: resolve once on #4343, then merge `issue-4308-slice-b-edi` into this branch and the resolution comes with it — no second hand-merge.

Note #4348 is *not* the same on hunk 1: it deliberately carries `status: in-review` as the stack's handoff marker, so taking main there would discard it. Flagged on that PR.

I can't push this — the branch is on `ttraenkler/js2`, outside this session's authorized repository set.

---

ttraenkler:
Closing as superseded — the content is already on `main`. Verified: this PR's head is an ancestor of `main`, landed via #4366 (slices C+D, which stack on this slice and carried its commits).

This PR could not be merged on its own: as a fork-head PR it reported `mergeable_state: dirty` against a base its head already contained as a parent, `git merge-tree` said clean, and `PUT /update-branch` answered `422 merge conflict between base and head`. Moving the same commits to a head branch in this repository made them `CLEAN` and auto-enqueued within minutes. Nothing was lost.

---

PR #4348, opened 2026-08-10
feat(#4308): slices C+D — function-caller EDI and strict-caller eval, eval-code 758→783/816 (parity reached)

ttraenkler:
Slices C and D of #4308, shipped together — 16 and 3 files, same adapter file, splitting them would have cost a serialization round-trip for no benefit. **Stacks on #4346** (#4245 slice 2) → #4343 (slice B) → #4340 (slice A) → #4335 (membrane slice 1) → #4321 → #4319.

## Parity reached, and exceeded

| run | pass / 816 |
|---|---|
| quickjs, base `1c721c8fa` (re-measured on this tree) | **758** |
| **quickjs, post-C+D** | **783** |
| interpreter (`TEST262_FULL_RUNTIME_EVAL=1`), re-checked after the `src/` change | **779** — unchanged, **zero per-file movement** |

**+25, zero regressions, zero error-text changes**, diffed test-for-test across all 816. The 758 baseline reproduced slice 2's record with zero per-file disagreement, so it is gate-grade rather than design-grade.

**The quickjs-only gap is 0 files.** All 34 remaining quickjs failures are files the interpreter fails too. Going the other way, quickjs passes **4 files the interpreter does not** (`non-definable-global-{function,generator}`, `var-env-{var,func}-init-local-new-delete`).

**#4242 can claim parity on `language/eval-code/` outright — no accepted-residuals block is needed for this set.** That reverses the projection this issue was filed under, which put the realistic landing zone at 740–765 with a permanent 15–40 file remainder.

## Per-cluster breakdown

| n | cluster | slice |
|---|---|---|
| 8 | `annexB func:existing-fn-update` | C |
| 8 | `annexB func:no-skip-param` | C |
| 3 | `var-env-var-strict-{caller,caller-3,source}` | D |
| 2 | `lex-env-distinct-cls` (direct+indirect) | D — one of the "pre-existing, not yours" pairs |
| 2 | `global-env-rec-eval` (direct+indirect) | D — the other pair |
| 2 | `var-env-{var,func}-init-local-new-delete` | C, unplanned |

Both pairs that were scoped out as pre-existing fell out of the design anyway.

## The silent-no-op trap fired again — and totals hid half of it

The first full run read **+20** and looked clean. The test-for-test diff said **+25 / −5**:

- four `global-env-rec*` files trapped with `memory access out of bounds` — the in-realm `%eval%` capture was lazy, so on the indirect path it ran *after* the first globals push and could capture a membrane wrapper of the compiled `eval` marker, giving infinite re-entry;
- `var-env-func-init-local-update` flipped pass→fail.

Both fixed; the capture now happens in `qjsEnsureContext`. This is the **fifth** occurrence of this class in the workstream, and the second where a totals-only comparison would have shipped a net-positive number over a real regression. Per-file diffing is the only reason it was caught.

## The authorized `src/` line

One activation-seed sentinel per function-scoped call site (`directEvalCallerIsFunctionScoped`, purely syntactic — oracle-ratchet clean at +0). P4 measured that a **declaration-free arrow** caller is indistinguishable from global code at the seam, so its `var`s would misroute onto the global object.

It buys **zero** test262 files, exactly as predicted — the corpus uses `function` wrappers throughout. `arrowGlobalProbe` is its only evidence: **0 without it, 1 with it**. It was authorized as a correctness fix, not a conformance one, on the grounds that a silent wrong-scope write is the class that resurfaces later as an unattributable bug.

Gates owed for touching `src/` and run: equivalence shards 1–4/4 (no new regressions), oracle-ratchet OK, loc +51 OK, func-budget OK, `tsc --noEmit` clean, prettier clean.

## Artifact

Unmoved: key `d8a5a91d6f183b87`, sha256 `b0662069c241d0430d91c53a3b0e2d1281fd9eb78dd1c93490b0a9dfa70eec5b`. No shim change, no rebuild — every slice from A through D held the engine binary constant.

## Two findings that are not this PR's to fix

1. **`check:godfiles` fails on this tree** — 5 regressions in `calls.ts` / `index.ts` / `object-runtime.ts`, none of which this branch touches. Pre-existing drift on the stacked base.
2. **New caller-side codegen defect**: `for (var i = …)` in the same function as a direct eval emits an **invalid module** (`CompileError: local.tee[0] expected type (ref null N), found local.get of type i32`). `let` in the loop head is fine; ordering is irrelevant. Reproduced with this branch's `src/` change reverted, so it is engine- and slice-independent. It forced a lane rewrite here and needs its own issue.

## One test expectation deliberately changed

`tests/quickjs-eval-provider.test.ts` case 11 pinned the #4238 slice-3 residual *"a strict caller's assignment THROWS"*. Slice D retires that residual, so the case now asserts the assignment **updates** the binding — still distinguishing success from both the old TypeError and a silently-lost write, rather than being loosened.

`src/interp/`, acorn and the IR substrate are untouched, and the interpreter tier re-measures at 779 with zero per-file movement — the standing proof that the default-flip has not degraded the engine it replaces.

---


ttraenkler:
## Same conflict as #4343 and #4346 — one file, but hunk 1 resolves *differently* here

`DIRTY` against main (`842b472c3`), 9 commits behind. The conflict is entirely in the shared record:

```
plan/issues/4308-quickjs-eval-declaration-instantiation.md
```

Structurally identical across all three PRs in the stack — three hunks, same regions. Full per-hunk reasoning is on **[#4343](https://github.com/loopdive/js2wasm/pull/4343#issuecomment-5247169454)**; the short form:

| # | region | #4343 / #4346 | **#4348** |
| --- | --- | --- | --- |
| 1 | frontmatter `status` | take **main** (`in-progress` + `assignee`) | ⚠️ **take the branch** — see below |
| 2 | trap #5 (#4305) | take **main** — the branch carries the superseded shape | same |
| 3 | slice records at EOF | take **branch** — main's file ends where the block begins | same (larger: +526 lines here vs +226 on #4343) |

### Why hunk 1 differs on this PR

This branch deliberately sets, with its own explanatory comment:

```yaml
# Slices A–D are all implemented and measured (see the four implementation
# records below); the branch is handed off for PR + merge, so this is the
# HANDOFF case of the status lifecycle — whoever observes the merge flips it to
# `done`.
status: in-review
```

Main says `status: in-progress` + `assignee: ttraenkler/senior-dev`. Blindly taking main here — correct for #4343/#4346 — would **discard the deliberate `in-review` handoff marker** on the PR that completes the stack. The right resolution is a genuine merge of the two sides: keep the branch's comment and `status: in-review`, **and** keep main's `assignee:` line, which the branch does not have.

### Cheapest order

Resolve once on #4343, then let the stack carry it: #4346 merges `issue-4308-slice-b-edi`, #4348 merges `issue-4245-membrane-slice2`. Each PR then only has its own hunk-1 decision left instead of all three.

### One trap

Do **not** rebuild the file as "main's copy + the slice records". I checked that shortcut against the real 3-way merge and it silently drops branch-only regions git merges correctly outside the markers — `updated:` and the `AMENDED … DO NOT IMPLEMENT THE SCANNER BELOW` Q5 block. Resolve the hunks in place.

I can't push this — the branch is on `ttraenkler/js2`, outside this session's authorized repository set. (Maintainer edits *are* enabled, so the "Update branch" API works on the clean PRs in this stack — it just can't resolve a conflict.)

---

ttraenkler:
Closing as superseded by #4366, which carried these exact commits and has merged. Verified: this PR's head is an ancestor of `main`.

#4366 was opened only because this PR's fork-based head was stuck reporting a merge conflict that did not exist — `git merge-tree` clean against the same base, and `PUT /update-branch` returning `422`. The same commits from a head branch in this repository went `CLEAN` and were enqueued immediately.

One follow-up this PR's frontmatter assumes: it set `#4308` to `status: in-review` as the deliberate HANDOFF case, expecting whoever observes the merge to flip it to `done`. That flip has not happened yet.

---

PR #4429, opened 2026-08-12
docs(#4157): __extern_get decomposed 73/27; helper-level acorn work closed out, one lever specced

ttraenkler:
## Description

Docs-only follow-up to #4424. Seven commits, two `.md` files, no `src/` change — every experiment below was built, measured, and reverted.

**Superseded description:** this PR originally covered three commits. It now carries the whole session, including a **retraction of one of its own original claims** (§4).

### 1. Five measured attempts on `__extern_get`, and what they add up to

| # | attempt | result |
| --- | --- | --- |
| 1 | skip the declared-field ladder for plain-`$Object` receivers (in #4424) | **null**, ±0.3 pp |
| 2 | abstract-`eq` casts instead of RTT casts on the hit path | **null**, ±0.5 pp |
| 3 | outline the cold path so `wasm-opt` can inline the rest | **net null**, −0.13 pp |

Each was order-reversed ON/OFF/OFF/ON on `standalone-dynamic`, checksum 422 held throughout, and each is written up in enough detail to rebuild.

### 2. The result that pays for the other two

Attempt 3 is a bad optimisation and a good instrument. Splitting `__extern_get` at the cache arm decomposes it for the first time:

| half | self % | share |
| --- | ---: | ---: |
| cache-hit prologue | **6.38 %** | **73 %** |
| entire cold path — 15,032-instruction ladder, own-property walk, proto walk | 2.20 % | 27 % |

`__extern_get` alone fell 8.71 % → 6.38 % with both ON blocks cleanly below both OFF blocks — the only real separation of the session. But the 2.2 pp reappeared in `__extern_get_cold`: relocation, not saving, and only the **bucket total** exposed that. `wasm-opt` declined to inline the prologue because I had sized it wrong — "four instructions" was four *top-level* ones, one an `if` carrying ~45.

This retroactively explains attempts 1 and 2: **attempt 1 optimised the 27 %, attempt 2 shaved two instructions off a path that is ~45.** Neither could have worked.

Also first-measured here: the per-key cache (#3673) runs at **87.24 % of 506,752 calls per parse**, with 3.77 % thrash — which prices out "make the cache smarter" as well.

### 3. Levers priced out without building

| lever | verdict |
| --- | --- |
| smarter per-key cache | thrash is only 3.77 % |
| rewrite the `applyRefNullFixups` walk | 1.36 % of calls, nearly all in dead code |
| typed-receiver `__dc_*` ABI | ≈ 0.6 %, below floor |

The typed-receiver entry also records a **soundness fact**: guarded `__dc_*_g` trampolines *cannot* take a typed receiver. Their `ref.test` exists because the receiver's shape is a whole-program inference; a typed parameter would force the call site to cast, reintroducing exactly the trap the guard prevents.

### 4. Retraction of this PR's own §2 (as originally filed)

The original §2 said the `__dc_*` trampolines are stuck on an `externref` ABI because `applyRefNullFixups` walks backwards one instruction per parameter, and prescribed rewriting that walk. **The rewrite already exists.** #4077 added `locateCallArgProducers`, an exact *forward* stack model; the backward walk is only its fallback, and the ABI note in `typed-this.ts` predates it. Measured: **48,670 calls modelled exactly, 672 (1.36 %) falling through**, and the abandons are on terminators (`return` 5,812, `throw` 4,316, `br` 1,445) after which the code is unreachable.

Left as a written retraction rather than silently amended. The **~4 % self-time figure stands** — that came from the profile; only the diagnosis was wrong.

### 5. Where this leaves acorn

Helper-level work is finished. The 2026-08-08 cross-runtime table already gave the floor: Node spends 92.9 % of its parse inside acorn's own code, this build 37.1 %; removing **every** helper, GC and regexp millisecond still leaves ≈ **3.0×**. That residue is register allocation, inlining and IC-class work on emitted code — which nothing in #4157 targets.

One lever survives above the measurement floor: **call-site inline caching**, now aimed at the 6.38 pp prologue. §10 of the issue is its executable spec — verified insertion point (`fillMemberGetDispatch`, per-name rather than per-site), the full instruction sequence with field indices checked against source, the required widening of dispatcher reservation, and the four traps this session paid for once each.

### Verification

- prettier clean; `git diff --stat origin/main...HEAD -- src/ tests/ scripts/ .github/ benchmarks/` is empty.
- Every experiment checksum-verified at 422 before being reverted; none is in the tree.

## CLA

### DIFF
--- a/plan/agent-context/session-2026-08-07-acorn-perf-handoff.md
+++ b/plan/agent-context/session-2026-08-07-acorn-perf-handoff.md
@@ -238,3 +238,92 @@ What finally converged it, named independently by both lanes:
 verified pre-existing by swapping base blobs. See #3552 for why CI does not
 surface these, and its still-open follow-up: detection exists (`issue-tests.yml`
 post-merge), nothing consumes it.
+
+---
+
+# Update — 2026-08-12
+
+**The allocation program described above is finished, and the picture has
+changed.** Full detail is in #4157 under the four `2026-08-12` headings; this is
+the short version so the next lane knows what is still true.
+
+**Standing: still ~7× on `standalone-dynamic`.** The wins landed, but they moved
+GC, not the total.
+
+## What changed
+
+| bucket | 08-07 | 08-12 |
+| --- | ---: | ---: |
+| gc-engine | 23.1 % | **2.97 %** |
+| dynamic-lookup | 13.5 % | **21.15 %** |
+| call-dispatch | 8.1 % | **11.39 %** |
+
+GC went from largest bucket to ninth. `__extern_get` (9.22 %) is now the hottest
+frame in the profile — ahead of every compiled acorn function. The two buckets
+this handoff flagged as "untouched by anything" are now 32.5 % together, and are
+the whole remaining addressable story. Of the 100.3 ms gap in the 08-08
+cross-runtime table, ~9 ms is closed; helpers with no Node counterpart carry
+about 55 % of what remains.
+
+## Do NOT re-attempt (two more nulls, same helper)
+
+The exhaustion list above grows by two, and this pair is stronger than the
+others because they attack from opposite ends and both land on zero:
+
+| lever | result |
+| --- | --- |
+| skip the declared-field ladder for plain-`$Object` receivers | null, ±0.3 pp |
+| abstract-`eq` casts instead of RTT casts on the cache hit path | null, ±0.5 pp |
+
+The first removed **14,770 instructions** of provably-dead work from the miss
+path; the second removed two RTT checks from the hit path and shrank the binary.
+Neither moved the bucket. **`__extern_get`'s 9 % is not made of instructions
+removable from inside it** — at 21.7 ns / ~45 cycles per call, what is left is
+call overhead, the props-array pointer chase, and branch behaviour. Do not spend
+another A/B cycle on the body.
+
+Both experiments are written up in #4157 in enough detail to rebuild; neither is
+in the tree.
+
+## The one thing that is priced above the measurement floor
+
+**Site- or name-local inline caching — remove the CALL, not the work inside it.**
+
+The numbers that make it the only live candidate:
+
+- **506,752 `__extern_get` calls per parse**, **87.24 % served from the existing
+  per-key cache** (#3673). First time that cache has been measured.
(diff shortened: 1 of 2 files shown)


ttraenkler:
Pushed a correction to §2 of this PR's own description — the diagnosis there was wrong, and it's better caught before review than after.

§2 says the `__dc_*` trampolines are stuck on an `externref` ABI because `applyRefNullFixups` walks backwards mapping one instruction per parameter, and that the fix is to rewrite that walk to use `instrPopsPushes`. **The rewrite already exists.** #4077 added `locateCallArgProducers`, an exact *forward* stack model that threads `instrPopsPushes` through the instruction list and records, per call, which instruction produced each argument. The hand-rolled backward walk is only its fallback. The ABI note in `typed-this.ts` was written for #3683 and predates it.

Measured with env-gated counters over the standalone acorn build (reverted after; checksum 422 held):

| | count |
| --- | ---: |
| calls modelled exactly by the forward model | **48,670** |
| calls falling through to the legacy backward walk | **672** (1.36 %) |

And the lists that abandon do so on **terminators** — `return` 5,812, `throw` 4,316, `br` 1,445, `return_call` 664 — after which the rest of the flat list is unreachable. Abandoning there is correct, not a defect, and the "lost" calls are overwhelmingly in dead code, where Wasm validation is polymorphic and a mis-rewrite is inert. Only six abandons (`local.tee`/`local.set`) are a genuine modelling gap.

Two consequences:

1. **Do not rewrite the walk.** It serves 1.36 % of calls, nearly all unreachable. §2 should not be read as scoping that work.
2. **The typed-receiver experiment is far cheaper than §2 implied.** The ABI note's own worked failure — `this.parseExprOp(this.parseMaybeUnary(null, false, false, forInit), …)`, where each `false` is two instructions — is exactly what the forward model handles. So the test is just: reserve a `__dc_*` trampoline with `(ref $__fnctor_F)` as parameter 0, drop the `extern.convert_any` at the call site and the `any.convert_extern; ref.cast` in the body, and see whether it validates.

The **~4 % self-time** figure for the `__dc_*` family is unaffected — that came from the profile. Only the diagnosis of what blocks it was wrong.

Full correction is in the new commit as entry (6) of #4157, written up rather than silently amended.

---

PR #4432, opened 2026-08-13
fix(#4394): Test262 harness self-tests — GC 54→98/116, standalone 66→80/116

ttraenkler:
## Description

Driving the Test262 **harness self-test** cohort (`test262/test/harness/*.js`, 116 files, admitted to the census by #4304) toward 100%, in **both lanes**.

| lane | before | after |
| --- | ---: | ---: |
| host / GC | 54 / 116 | **98 / 116** |
| standalone | 66 / 116 | **80 / 116** |

Full inventory, per-bucket measurements and the remaining work live in `plan/issues/4394-test262-harness-selftest-conformance.md`.

| # | fix | lane | pass |
| --- | --- | --- | ---: |
| — | baseline | GC | 54 |
| 1 | `Test262Error` constructor identity | GC | 80 |
| 2 | second-level function-static write dropped | GC | 84 |
| 3 | partial object-literal argument arrives as `null` | GC | 88 |
| 4 | numeric-first array literal drops string elements | GC | 90 |
| 5 | host-lane dynamic-receiver `valueOf` | GC | 92 |
| 6 | `$DONE` exposed on the sandbox unconditionally | GC | 93 |
| 7 | bare `EvalError`/`URIError`/`AggregateError` read as `null` | GC | 94 |
| 8 | host callback bridge had no `[[Construct]]` | GC | 94 |
| 9 | `defineProperty(globalThis, …)` took the struct path | GC | 96 |
| 10 | `instanceof` a module-declared `Test262Error` | GC | 97 |
| 11 | `new TypeError()` ignored a user shadow | GC | **98** |
| 12 | §9.1.1.4.18 — script functions bind on the global object | standalone | 75 |
| 13 | `Array.prototype.{map,filter,forEach,…}` callable as a value | standalone | **80** |

All but #6 are general compiler bugs that happen to be *observable* through the harness; #6 is a test-infrastructure correction. Several are worth more outside the cohort than inside it — see the sample deltas.

### 1 — `Test262Error` constructor identity (26)

`new Test262Error(msg)` is intercepted **by name, with no shadow guard**, and in the JS-host lane lowered to `env::__new_Test262Error` — a real `Error` subclass built by the runtime. The literal upstream harness always **declares** `Test262Error` itself (`sta.js`), so the module's identifier reads a compiled WasmGC closure while the thrown object's `.constructor` is the *host* class. Compiled `===` on two externrefs uses `ref.eq` when the right operand is statically a GC struct, so it never reaches the `__host_eq` shim written for this case.

Keep the host `Error` and pass the module's own ctor value alongside the message via a new `__new_Test262Error_ctor(msg, ctor)` builtin import that stamps it as a non-enumerable own `constructor`.

### 2 — the second-level function-static write is dropped (4)

`collectDeclarations` drops a top-level assignment whose root identifier is not a module global. #2671 keeps `F.<prop> = …` with a bare-identifier receiver; #3666 added the nested-receiver keep **for standalone only**. So `assert.deepEqual._compare = (function () { … })();` compiled to **nothing** in the host lane.

### 3 — a partial object-literal argument arrives as `null` (4)

Under `allowJs` the checker honours JSDoc, so `verifyProperty(obj, prop, desc, { restore: true })` lowered to `struct.new <{restore}>` plus a **guarded** downcast to `<{label,restore}>` — and a guarded cast that misses yields `ref.null`. `{ restore: true }` silently restored nothing.

### 4 — a numeric-first array literal drops its string elements (2)

The widening scan carved out `StringLiteral` elements for the native-strings lanes; the host lane has no equivalent scan, so `[0, s]` widened and `[0, "a"]` did not — the latter read back `NaN`. `compareArray` therefore called two arrays differing only in their string elements equal.

### 5 — host-lane dynamic-receiver `valueOf` (2)

The blanket "`valueOf()` returns the receiver" is `Object.prototype.valueOf`, right only when nothing overrides it. #4201 fixed this for standalone; the host lane kept the shortcut, so `Object("a").valueOf()` returned the wrapper.

Landed as a dedicated in-module decision, then **superseded during the merge with main**: `3f614151` (Moment dogfood) extracted the whole fallback into `expressions/valueof-fallback.ts`. Verified that main's route also preserves `o.valueOf() === o` — the contract the dedicated module existed to protect — so the module and its two runtime builtins were **removed** rather than kept as a second implementation. The test stays: it is what proves the surviving route holds the contract.

### 6 — `$DONE` exposed unconditionally (1)

`_buildFreshSandbox` set `sandbox.$DONE` for every test (#3428), so `asyncHelpers.js`'s `hasOwnProperty(globalThis, "$DONE")` guard could never observe its own absence. Withholding it unconditionally is equally wrong — one test declares its own `$DONE` with no `async` flag (measured: gating on the flag alone traded one pass for one fail). Now gated on "the script declares one". See #12: the real binding was missing in both lanes.

### 7 — bare `EvalError` / `URIError` / `AggregateError` read as `null` (1)

Two registries in `extern-declarations.ts` have to agree and did not. `LIB_GLOBALS` gates whether `collectDeclaredGlobals` runs **at all**; `AMBIENT_BUILTIN_CTORS` is the loop inside it that registers `env.global_<Name>`. `Error`/`TypeError`/`RangeError`/`SyntaxError`/`ReferenceError` were in both; their siblings — plus `BigInt`, `Proxy`, `SharedArrayBuffer`, `Atomics` — were in neither, so a bare value read lowered to `ref.null.extern`.

`new EvalError()` was never affected. Only the *identifier* was null, which is what kept this quiet: an identity comparison against null does not throw, it answers `false`.

### 8 — the host callback bridge is an arrow, so it has no `[[Construct]]` (0 here, +29 in sample)

Every compiled callable handed to a host API was rejected by `Reflect.construct` / `new`, whatever it was written as. The harness's `isConstructor` is `Reflect.construct(function () {}, [], f)` — its own **target** is an inline function expression, so the probe threw before ever inspecting `f`, the `catch` swallowed it, and the helper answered `false` for everything, in all **644** test262 files that include `isConstructor.js`.

Only an ordinary function definition has `[[Construct]]` (§15.2.4), so the compiler picks a sibling import `__make_callback_ctor` whose bridge is a plain `function`; arrows, generators, async functions and accessor callbacks keep the arrow bridge. The regression test pins both halves — widening this into "every compiled callable is a constructor" would be equally wrong.

### 9 — `defineProperty(globalThis, …)` took the compiled-struct path (2)

`Object.defineProperty(this, 'Object', {configurable: true, value: Object})` threw `TypeError: Object method called on null or undefined`, naming a receiver that was never null. The harness prefix opens with `var $262 = { global: globalThis, … }`, in front of **every** test, and that one VALUE use makes the checker mint a struct for `typeof globalThis`. The define then took the struct arm, whose guarded cast can never match a host externref.

The trigger is the earlier value use, not the define — which is why this only showed up under the assembled harness. Standalone/WASI are excluded: there `globalThis` really is a native `$Object` struct.

### 10 — `instanceof` against a module-declared `Test262Error` (1)

Fix 1 handled `.constructor`, but `instanceof` walks the **prototype chain**, and the value is deliberately a host `Error` subclass, so its chain can never reach the module's compiled closure. `__new_Test262Error_ctor` already receives the carrier; it now records it, and `_instanceofResult` answers true when the value is a host Test262Error **and** the RHS is a carrier that has minted one. Both halves required, so an unrelated compiled constructor never starts matching.

### 11 — `new TypeError()` ignored a user shadow (1)

The NativeError names were claimed **by name with no scope check**, so a shadowing declaration produced the intrinsic while the identifier of the same name read the user's binding. Measured: `e.constructor === TypeError` false, `e.constructor === intrinsicTypeError` true — exactly the collision `assert-throws-custom-typeerror.js` exists to detect.

`shadowed-error-ctor.ts` walks the enclosing scope chain and declines when a user binding is found. `Test262Error` is excluded on purpose (fix 1 exists to reconcile it). A **top-level** shadow is claimed by a different `new` path before this arm is consulted and is still wrong — pinned by an explicit failing-behaviour test rather than left silent.

### 12 — script functions never bound on the global object (standalone, 9)

§9.1.1.4.18 CreateGlobalFunctionBinding: a SCRIPT's top-level function declarations are own properties of the global object. We never implemented it. Measured for `function $DONE() {}`:

| lane | `hasOwnProperty(globalThis, "$DONE")` | `typeof globalThis.$DONE` |
| --- | --- | --- |
| host/GC | false | `"function"` |
| standalone | false | — |

**Neither lane implemented it** — the GC lane only passes the affected tests because the runner *fakes* the own-property on its sandbox (fix 6). Seeding happens at the top of `__module_init`, ahead of every user statement, which is what declaration hoisting requires. Scripts only; a module's top-level declarations live in the module environment record.

Gated to standalone for now, and it fixes `language/global-code/decl-func.js` — the spec test for this exact binding.

### 13 — `Array.prototype.map` as a VALUE (standalone, 5)

Nine standalone tests reported a constructor mismatch that had **nothing to do with constructors**. The cause is one line of the harness — the code that FORMATS a failure:

```js
compareArray.format = function (arrayLike) {
  return "[" + Array.prototype.map.call(arrayLike, String).join(", ") + "]";
};
```

`emitArrayProtoMemberBody` implemented only `slice`; every other member degraded to a catchable `not yet callable as a value`. So the moment a comparison failed, the TypeError from the *formatter* replaced the Test262Error the test was asserting about.

Routed to `__hof_<name>`, the native standalone loop that already existed for the dynamic-receiver call form. **The receiver guard is load-bearing**: the old refusal threw for every receiver, so `Array.prototype.map.call(undefined)` passed its "must throw TypeError" test *by accident* — routing without a §23.1.3 step-1 check silently returns an empty result, measured as 3 regressions before the guard was added.

### Diagnosed, deliberately NOT landed

Repairs implemented, measured, and **reverted** rather than shipped, each recorded in the issue file with its evidence:

- **Callee-side async Promise envelope** (7 tests) — correct and working, but `tests/equivalence/promise-chains.test.ts` *codifies* the legacy sync pass-through (`double(21) as any as number === 42`). Landing it means amending that expectation — a project-level behaviour decision — and it gained **zero** harness tests on its own. Local sampling did not catch the conflict: 620 async test262 files showed 0 status changes. `equivalence-gate` is the check that sees this class of change.
- **Naive `valueOf` fall-through** — resolves the wrapper but breaks `o.valueOf() === o`, shape-sensitively, so four category samples showed 0 diffs *while the contract was broken*.
- **Duplicate top-level `var` widening** — 0 diffs on a 250-file sample and the `$262` guard fixed, but incomplete and 0 harness tests gained.
- **Bare-`symbol` parameter carrier** — a raw i32 handle; `type-mapper.ts` documents why the local repair was rejected (#2792/#2785). Filed as evidence for the deferred #2610 value-rep pass.
- **JSDoc-typed parameter makes `typeof` lie** — under `allowJs`, `@param {PropertyDescriptor|undefined} desc` makes the carrier a struct, so a primitive argument answers `"object"`. The repair is a change to core parameter typing in JS modules, for one harness test; it wants its own issue.

### Regression evidence

Every landed commit measured both ways on independent samples; **no sample lost a passing test**.

| sample | result |
| --- | --- |
| `equivalence-gate`, all 8 shards, re-run per commit and after the merge | no new regressions |
| 400-file sample of `isConstructor.js` consumers | +29 pass, 0 lost |
| 600-file sample over `built-ins/{Error,NativeErrors,AggregateError,Proxy,BigInt}` | +9 pass, 0 lost |
| 220-file standalone sample over `language/statements/function`, `language/global-code`, `built-ins/Function` | 0 regressions, +1 |
| 200-file standalone sample over `Array/prototype/{map,filter,forEach}` | 0 regressions, +1 |
| 200-file sample over `built-ins/Object/define*`, `built-ins/global{,This}`, `language/global-code` | 0 regressions |
| `built-ins/Object/defineProperty`, first 400 | 325 → 326 pass, 0 lost |
| 205-file sample of the 1,841 tests calling `assert.throws(Test262Error, …)` | 0 status changes |
| all 76 suite tests including `deepEqual.js` | 4 fail → pass, 0 lost |
| `built-ins/Reflect` (153), `built-ins/JSON` (165) | 0 status changes |
| `built-ins/Array`, first 400 | 120 → 121 pass, 0 lost |
| `built-ins/Number` (340), `built-ins/Boolean` (51), `built-ins/Date` (400) | 0 status changes |
| 400-file sample of `flags: [async]` tests | 0 status changes |

Plus eleven regression test files (49 cases) and a clean `tsc --noEmit`.

### Remaining — GC 18, standalone 36

GC: **7** async function expressions (blocked on the codified equivalence expectation, not the lowering) · **3** bare-`symbol` parameter carrier (#2610) · **2** `detachArrayBuffer` · **6** singletons.

Standalone: **15** async family (fix 12 cleared the `$DONE` gate; these now fail on real async causes) · **6** null deref in `toString` (the `deepEqual-*` family) · **4** the rest of bucket 13, blocked on `String` callable as a value — `compareArray.format` now returns `[null, null, null]` rather than throwing, because the loop runs and the callback is null · **11** misc.

> **Measuring standalone locally requires the QuickJS runtime-eval provider.** Without it 25 tests fail on `provider is not built` and the lane reads 41/116 instead of 66/116. Build with `npx tsx scripts/build-quickjs-eval-provider.mjs` — note `npx tsx`, not bare `node`: the plain-node run builds `libquickjs.wasm`, then fails on the *adapter* and leaves a half-populated cache that still reports "not built".

## CLA


ttraenkler:
## Park diagnosed — removing `hold`

Cited run: [31725434849](https://github.com/loopdive/js2wasm/actions/runs/31725434849), `check for test262 regressions` → **Fail on regressions**. Net on the merged state was **+147 pass**.

The run's own detail step could not name the regressed files — it crashed with `ENOENT: no such file or directory, open ''` (an empty `$BASELINE`), so `test262-regressions-detail.txt` was never written and only aggregate counts reached the log. The list below was reconstructed by local A/B against `origin/main`.

### 1 — genuine regression, fixed

`built-ins/Reflect/construct/newtarget-is-not-constructor-throws.js`

§26.1.2 treats an **omitted** and an explicit-`null` newTarget oppositely — step 2 defaults to the target, step 3 throws a TypeError for a *present* non-constructor. The wasm import boundary is fixed-arity and pads an omitted third argument with a null externref, so the runtime shim could not tell them apart and collapsed both to the 2-argument form. `Reflect.construct(fn, [], null)` quietly constructed instead of throwing.

Presence is a compile-time fact, so it now rides on the import **name** (`__reflect_construct_newtarget` when the call site passes three arguments).

Worth noting *why this branch surfaced it*: the test only passed before because its **target** was itself un-constructible (the arrow callback bridge), so the call threw at step 1 and never reached step 3. Fix 8 in this PR made an ordinary compiled function constructible and exposed the real defect underneath.

### 2 — not a pass regression, named allowance

`built-ins/Proxy/construct/call-result.js` was **already failing on the merge-base and still fails**; only the mechanism moved.

| | status | error |
| --- | --- | --- |
| main | `fail` | `function () { [native code] } is not a constructor` (catchable) |
| branch | `fail` | `illegal cast` in `__call_fn_method_3` (uncatchable trap) |

Construction now correctly proceeds into the proxy construct trap and hits a pre-existing bad cast in the method-dispatch trampoline. That is the `fail`-baseline reclassification the gate routes to a named `trap-growth-allow` (#3596); declared in the issue frontmatter with the test named.

### Verification

863-file A/B against `origin/main` over `built-ins/Function`, `built-ins/NativeErrors`, `built-ins/Error`, `language/expressions/new` and `language/statements/try` — the areas the failing gate pointed at:

```
REGRESSIONS 0
IMPROVEMENTS 15
```

(93 `language/statements/try/dstr/*` files are excluded: they throw `meta.features is not iterable` inside the runner and take the process down. That reproduces identically on unmodified `origin/main`, so it is a pre-existing runner issue, not a property of this branch — filing separately.)

Targeted `built-ins/{Proxy,Reflect}` A/B (464 files): 0 regressions, +2.

### Also fixed since the park

- **`DIRTY`** — merged `origin/main`; two import-line conflicts, resolved as unions.
- **`quality`** — the merge brought in #4395's `check:host-import-policy` ratchet, which holds `src/runtime.ts`, `resolveImport` and the eight owned adapters at exactly their current size, with no per-issue allowance path. Rather than just raising the ceiling, the host-side `Test262Error` support moved out to `src/runtime/test262-harness-host.ts` (88 of 106 lines recovered); the residual +18/+11 is dispatch plus the comments recording why each arm is load-bearing, and the two numeric ceilings move by exactly that. The new module is deliberately **not** added to `ownedAdapterPaths` — whether harness-compat host code belongs in that counted set is #4395's call.

---

PR #4456, opened 2026-08-13
fix(ir): Function.prototype member reads armed the runtime-eval boundary; js2→js2wasm reference sweep; #4394 evidence record

ttraenkler:
## Description

Follow-up to #4432 (merged). Three independent deliverables:

1. **A `main` IR-selection regression fixed** — the one that has kept the `quickjs eval-engine lane` red on `main` and every PR since the #4395-era runtime-eval boundary inventory landed.
2. **The repository-rename reference sweep** (`loopdive/js2` → `loopdive/js2wasm`, 136 files).
3. **The #4394 evidence record** — two implemented-measured-REVERTED attempts, so the next attempt starts from what is true.

## 1 — `Function.prototype.call` demoted every standalone IR unit

`buildIrRuntimeEvalBoundaryPlan` (src/ir/runtime-eval-boundary-plan.ts) records any bare `eval`/`Function` identifier that is neither a direct callee nor a member *name* as an "intrinsic-value" escape site. `Function` in **receiver** position — `Function.prototype.call`, `Function.name` — is neither, so one ubiquitous idiom set `callableBoundaryRequired` for the whole module; in standalone/WASI that flag demotes **every IR unit** to the legacy frontend (the #2928 arm in `src/codegen/index.ts`).

`tests/issue-2960.test.ts` ("does not treat Function.prototype member reads as an escaped Function constructor") pins the opposite contract:

```
expected [] to include 'add'   ← irCompiledFuncs was empty
```

Confirmed pre-existing on `main` (fails at `7797d690`, before #4432 merged) — see the [triage comment](https://github.com/loopdive/js2wasm/pull/4456#issuecomment-5286207970).

**Fix**: receiver-position reads of `Function` are excluded from the inventory — the value that flows onward is the read property, not the eval-capable constructor. `eval` deliberately **stays** a site in receiver position: `eval.call(null, src)` performs a real indirect eval, so member reads through `eval` still route the provider.

**Verified** with the exact suites the CI lane runs: interpreter config **66/66**, quickjs config **139/139**, `check:ir-fallbacks` clean, `tsc --noEmit` clean.

## 2 — repository-rename sweep (`loopdive/js2` → `loopdive/js2wasm`)

Every reference that names the **repository** — CI workflows, tooling defaults (`claim-issue.mjs` PR-scan repo, `release.mjs` targets, `REPO` constants), agent instructions, contributor docs, published links. 278 references across 136 files.

Deliberately **not** renamed:
- **`@loopdive/js2`** — the published npm/JSR package name (39 occurrences preserved; the substitution anchored on a negative lookbehind for `@`). Renaming a repo does not rename a package; rewriting it would break every installer and the `publish-npm.yml` proxy-dependency check.
- **`loopdive/js2wasm-baselines`** — already correct.
- **Historical records** (`plan/log/`, `plan/issues/`, `plan/agent-context/`, `CHANGELOG.md`) — they describe what was true when written; the GitHub redirect keeps their links live.

One fixture needed a real fix, caught by the suite: `pre-push-labs-remote.test.ts` asserts case-**insensitive** normalization, so its deliberately upper-case input moved with the expectation (`LoopDive/JS2WASM.git`).

## 3 — #4394 evidence record (docs)

Two repairs implemented, measured, and **reverted**, recorded in `plan/issues/4394-test262-harness-selftest-conformance.md`:

- **`String` as a value in standalone**: a conversion-closure registration fixed 1 harness test and broke 3 `built-ins/String` identity tests (`new String(NaN).constructor === String` → `[object Object]`) — the bare `String` read was already resolving correctly to the constructor carrier. The real fix is making the **carrier** callable; recorded so the cheap version is not re-attempted.
- **The function-valued `toString` diagnosis was wrong**: the recorded "make the arm decline" fix was implemented and measured — the generic dispatch it declines into produces a **null string ref**, strictly worse. The probes established the prototype-assignment idiom (`P.prototype.greet = fn; p.greet()` → `null`) is broken for **any** method name, that the property is stored correctly (only direct-call dispatch misses it), and that two structurally identical object literals in one module can share a method body — all recorded, each wanting its own issue.

Harness cohort state: host/GC **98/116**; standalone currently measures **72/116** under the now-default QuickJS eval engine (#4242) — the earlier 80/116 figure was not measured under identical engine conditions, so the delta is not all regression. Parallel bucket work on the remaining failures is in flight on top of this branch.

## CLA


ttraenkler:
## `quickjs eval-engine lane` is red on `main`, not on this PR

[Run 31742116412](https://github.com/loopdive/js2wasm/actions/runs/31742116412/job/94587984084) — `tests/issue-2960.test.ts:146`:

```
AssertionError: expected [] to include 'add'

  it("does not treat Function.prototype member reads as an escaped Function constructor")
    expect(r.irCompiledFuncs ?? []).toContain("add");
```

Not caused by this PR. Reproduced with `JS2WASM_EVAL_ENGINE=interpreter npx vitest run tests/issue-2960.test.ts` in a clean worktree at each revision:

| revision | | result |
| --- | --- | --- |
| `be2ca8a6` | this PR | 1 failed / 12 passed |
| `713a3f08` | `main` (tip) | 1 failed / 12 passed |
| `7797d690` | `main` **before** #4432 merged | 1 failed / 12 passed |

The third row is the one that settles it: the failure predates #4432, so it is neither the harness work nor the repository-rename commit. This PR does not touch `tests/issue-2960.test.ts` — its last change was #4335 — nor anything in the IR selection path.

**Why it still matters:** the lane is non-required, but a red non-required check drives `mergeStateStatus` to `UNSTABLE`, and `auto-enqueue` takes only `{CLEAN, HAS_HOOKS}`. So while this is red, nothing on this branch can be auto-enqueued even with every required check green.

The substantive finding is about `main`, not this PR: a standalone-target IR selection regression (`Function.prototype.call` read as a value causes the function to fall off the IR path) is live on `main` and only visible in a non-required lane, so nothing gates on it. Worth its own issue.

---

PR #4484, opened 2026-08-14
fix(#4394): harness standalone wave 2 — verifyproperty, Test262Error identity, typedarray-intrinsics (72→99/116)

ttraenkler:
## Description

Wave 2 of the parallel attack on the standalone Test262 harness self-test cohort (#4394). **Stacked on #4467** (in the merge queue) — retarget to `main` once #4467 lands.

## Measured on the complete tree

| lane | session start | after wave 1 (#4467) | after wave 2 (this PR) |
| --- | ---: | ---: | ---: |
| standalone | 72 / 116 | 90 / 116 | **99 / 116** |
| host / GC | 98 / 116 | 98 / 116 | **98 / 116** (unmoved) |

Full-cohort verification: 99/116 is the exact union of every bucket's predicted fixes with **zero interaction losses** — one cross-bucket interaction was found by re-measurement and reconciled (checkpoint 4 below).

## Checkpoint 1 — `verifyproperty` (`82746952`)

Fixes both `configurable-object` tests and `verifywritable-array-length`. Dominant lever: propertyHelper.js's `Function.prototype.call.bind(...)` uncurry lines made **two independent scanners** (deadcode-elide, runtime-eval boundary plan) classify `Function`-as-receiver as an eval hazard — linking the ~24 MB interpreter provider into every propertyHelper-including module and flipping them into the shared-realm carrier representation where descriptor round-trips lose fields. The boundary-plan arm supersedes the #2960 carve-out with a narrower, sounder predicate (`Function.prototype.constructor` stays an escape; `eval` stays a site in receiver position); issue-2960 suite 13/13 after reconciliation.

## Checkpoint 2 — `test262error-identity` (`1aff3b06`)

Fixes `sta.js`, `verifyProperty-same-value`, `verifyProperty-value-error`, `asyncTest-func-throws-sync`. A stack of three defects whose error messages pointed at the wrong lines; the load-bearing one attributed by `git log -L` + a minimal repro to `dbccc693` (#4395) widening the bound-function arm to standalone **ahead of** the #3571 uncurried dispatch. Measured at checkpoint 2: standalone 95/116.

## Checkpoint 3 — `typedarray-intrinsics` (`92561b55`)

Fixes `testTypedArray.js`, both `conversions` tests, `detachArrayBuffer.js`; `wellKnownIntrinsicObjects.js` moves from **CompileError — the compiler emitted invalid wasm** — to running (remaining failure is a separate cross-module intrinsic identity semantic).

The invalid-wasm root cause, proven by funcMap instrumentation: `tryEmitInlineDynamicCall` baked helper func indices into detached dispatch-arm arrays, then compiled the callee — and a callee like `new Function(src)()` adds a late import mid-lowering, shifting every index while the baked copies stay stale. After dead-import elimination the stale arm called the *adjacent* function (`__str_to_number`), yielding `call[0] expected externref, found call_ref of type f64`. This is the recurring #329/#1677/#1839/#1886/#2043 index-desync family; other detached-array bake sites remain to audit (recorded in the issue file).

Also converts 4 **false passes** into real ones: `view[Symbol.toStringTag]` was `undefined` and `TA.name` was `undefined`, so `assert.sameValue(undefined, undefined)` passed vacuously in `built-ins/TypedArray/prototype/Symbol.toStringTag/*`; both sides are now implemented (§23.2.3.34) and the tests pass on substance.

## Checkpoint 4 — cross-bucket reconciliation (`532befed`)

The full-cohort re-measurement caught one interaction the isolated agents could not see (the typedarray worktree predates the async bucket): the new nullish-receiver TypeError guard in the closed-method dispatchers composed with the async then-chain machinery into an illegal cast on a user-thenable receiver. `.then` is now exempt from the generic guard — its dispatcher has a dedicated owner (async-scheduler + then-thenable-miss) whose miss arms must observe the receiver themselves. A/B on the single file confirmed the guard alone was the flip.

## Remaining 17, all with recorded causes

- **#2610 symbol carrier** (×5): `compare-array-symbol`, both `verifyenumerable` symbol tests, `verifyProperty-restore-accessor-symbol`, `deepEqual-primitives` — blocked on the deferred value-rep work.
- **Realms** (×2): `assert-throws-same-realm`, `throwsAsync-same-realm` — `$262.createRealm` doesn't exist standalone; candidate permanent exclusions.
- **Async singletons** (×5): incl. one whose fix was built and *withdrawn* after it flipped 7 corpus tests (the coercion-boundary ref.null defect is recorded as the durable fix site).
- **Cross-module intrinsic identity** (×1): `wellKnownIntrinsicObjects` — QuickJS-provider-obtained `%Array%` is not `Object.is`-identical to the compiled carrier.
- Singletons (×4): `deepEqual-mapset`, `deepEqual-primitives-bigint`, `verifyProperty-desc-is-not-object` (JSDoc typeof lie), `verifyProperty-restore-accessor`, `detachArrayBuffer-host`.

## Verification

Every bucket: isolated-worktree implementation, targeted standalone A/B (0 pass→fail each — 227-file async, 300-file Object/Error, String guard trio + 450 files, 250-file defineProperty/gOPD, 200-file TypedArray), GC lane unmoved, per-test before/after in `plan/issues/4394-*.md`-referenced reports. Integrated tree: full 116-file cohort both lanes, every flip accounted; `tsc --noEmit` clean; all budget-gate growth granted per-file/per-function with measured amounts and reasons.

## CLA


ttraenkler:
## Park diagnosed — `hold` removed after fixing what the run named

Cited run: [31777774874](https://github.com/loopdive/js2wasm/actions/runs/31777774874). Two failures, diagnosed separately:

### 1. `merge shard reports` — "213 unclassified failures" → fixed (`8321023d`)

Not a classification-map gap — a **rendering regression this PR introduced**. The identity fix makes `new Test262Error(msg)` construct an ordinary user-fnctor instance (that's what fixed `sta.js`), but the #2962 exception renderer's `__any_to_string` has no arm for user structs: every thrown harness error rendered `[object Object]` instead of `Test262Error: …`, and the report's root-cause map lost all those signatures. Bisected to `new-builtin-globals.ts` by single-file main-version swap; local standalone sampling showed ~100+ `[object Object]` rows matching the CI count's shape.

Fix: the renderer now bakes a per-fnctor arm for each declined harness error fnctor (`"Name: " + message`, message through the same `__any_to_string` chain as in-module `String(x)`), recorded at the decline site, resolved fully before body-build per the emitter's index-shift-safety contract. The bisect repro renders `Test262Error: Expected SameValue…` again; `sta.js` and the other identity fixes hold.

### 2. `check for test262 regressions` — host net −1: **the run cannot say which test, and that blind spot is now also fixed**

The gate's summary runs `--quiet` (no file list) and the "Enumerate regressed files for triage" step has crashed with `ENOENT: open ''` on **every run since it was added**: it reads `$BASELINE`/`$META_ARGS` from the *gate step's* shell, and workflow steps don't share environments. Two parks in two days each cost a full local A/B reconstruction because of this. Fixed in this PR (`8321023d`): the step now defines `BASELINE` itself.

What local reconstruction could establish about the −1: the boundary-plan narrowing is exonerated (all 18 corpus files containing the re-added `Function.*` site patterns match baseline), the `$262.evalScript` shim-drop theory is exonerated (the only baseline-passing `evalScript` test still passes), and a 1,529-file targeted host sweep produced only concurrency flake (all spot-checked "regressions" pass serially on both `main` and the branch). The companion `compile_error→fail` transition in the same run is consistent with the stale-capture fix *repairing* previously-invalid wasm so a test now runs.

If the −1 recurs in the next merge_group, the fixed detail step will name the file and it gets fixed with full information; if it was unmanifested host flake (the canary quarantine manifest holds 932 known-flaky paths from only 2 canary runs), the re-run passes.

---

PR #4507, opened 2026-08-14
fix(marked): bound upstream compilation and preserve class object shapes

ttraenkler:
## What changed

- preserve class static/instance method identities and method ABI keys;
- keep callable method receivers valid across object-shape trampolines;
- materialize open object-spread sources before storing them in closed fields;
- run each upstream compatibility compile and Wasm execution in a killable worker with a hard deadline;
- keep the exact upstream Marked tests and report compile, validation, and runtime outcomes separately.

## Current result

For Marked v18.0.2 `test/unit/Hooks.test.js`, the module now compiles and validates (`4,550,040` bytes, about 11.2 seconds). The 15 admitted synchronous upstream tests currently pass `0/15` in Wasm with `br is not a function`; the remaining dynamic host-method bridge is explicitly recorded in [#4412](https://github.com/loopdive/js2wasm/blob/main/plan/issues/4412-marked-upstream-host-runtime-gap.md).

## Verification

- `pnpm run typecheck`
- `DOGFOOD_MARKED_TIMEOUT_MS=60000 pnpm run dogfood:marked-upstream-suite`
- pre-push typecheck, lint, formatting, oracle/coercion ratchets, numeric-local parity, and issue-integrity gates

This is intentionally a draft: it ships the compile watchdog and validated Wasm path while the runtime method-dispatch gap remains for follow-up.


ttraenkler:
Heads-up: this merge (`6756ed8c`) regressed private-method + super HomeObject dispatch, and it's currently failing every merge-queue entry via the uncatchable-trap ratchet (`null_deref` 140→142, #3189 gate — first casualty was #4578's auto-park).

**Bisected** (first-parent probes over today's main): `c3ff8a1f` (14:21) passes → `6756ed8c` (this PR's merge, 14:58) traps. All earlier candidates pass.

**Repro** (both `experimentalIR: true` and `false`, standalone target): `test262/test/language/statements/class/elements/super-access-inside-a-private-method.js` — a subclass private method `#m() { return super.method(); }` invoked as `this.#m.call(o)` with a foreign receiver now traps with `dereferencing a null pointer`. Baseline status was **pass**, so this is a genuine regression, not drift. `private-method-get-and-call.js` (baseline fail) also moved into the null_deref category.

Likely in the class-method-identity / receiver-trampoline changes this PR made. A surgical fix is being prepared as issue #4469 on branch `claude/ir-4469-homeobject-regression`, constrained to keep `tests/issue-4507-class-method-dispatch.test.ts` green — if you're already on it in your lane, comment there so we don't double-fix.

---

PR #4580, opened 2026-08-15
fix(codegen): fix-forward the three #4507 object-shape regressions (#4468)

ttraenkler:
## Description

Issue #4468 (plan by the Fable lane, implementation by an Opus subagent). PR #4507 was merged past its failing merge-group gate (project lead chose fix-forward over revert); this restores the 7 regressed test262 tests and the `null_deref` trap ratchet without giving back #4507's Marked win. Three separate defects, each isolated by per-file revert of #4507's 17-file src subset and A/B'd at `c3ff8a1f` (its parent):

1. **Trampoline null-deref** (`method-trampolines.ts`): #4507's `coerceTrampolineThisSlot` appended `ref.as_non_null` to a receiver slot that is deliberately nullable (the #2025 "receiver present but not this struct" arm) — the recorded `this` type read non-nullable while the emitted param is `(ref null …)` on both sides. Fix: early-return for any `ref`/`ref_null` target; the receiver path bridges only genuinely different carriers and can no longer emit a trapping instruction.
2. **Destructured-default params** (`closed-method-dispatch.ts`): #4507's under-application synthesized missing optional args with `defaultValueInstrs`, which for `externref` is `ref.null.extern` = JS `null` — a *real* argument on the host lane, so callee defaults never fired. Fix: shared `missingArgInstrs` — constant defaults inline, `externref` gets host `undefined`, anything else declines the closed-dispatch entry back to the pre-#4507 host fallback.
3. **Spread evaluation order** (`literals.ts`): #4507 removed a `native-first` gate, so the host lane snapshotted statically-declared fields instead of observing getters/deletes/expandos at spread time. Fix: gate restored, marked load-bearing.

## Validation

- The 7 regressed test262 files: pass at parent, fail on main, **pass with fix** (chunk runner, original-harness lane). The 3 already-failing neighbors stay failing (pre-existing, not chased).
- Collateral slice over `object/dstr`, method-definition, spread, private-method families: **1,034 tests, 0 failures**. Equivalence subset 367/369 with both failures pre-existing on unpatched main.
- **Marked dogfood byte-identical**: compile 1/1, validate 1/1, per-test status and per-test wasm error identical to main (a broader first version of fix 2 *did* regress marked — measured, rejected, narrowed before landing).
- `tests/issue-4468-shape-regressions.test.ts`: 3 per-family cases — fail at `6756ed8c`, pass at parent, pass with fix.
- Typecheck/lint/format and all ratchet gates green; budget growth granted in the issue frontmatter.
- The merge queue's regression diff remains the authority for the 7 returning to pass and `null_deref` ≤ 140.

Harness note recorded in the issue for future runs: the test262 pool needs `scripts/compiler-bundle.mjs` and `scripts/runtime-bundle.mjs` built, else every test times out behind `[pool] worker failed before ready`.

Issue #4468 → done in this PR (self-merge path).

## CLA

https://claude.ai/code/session_01GhCthcoeCGK29GwmdyiqGm

---
_Generated by [Claude Code](https://claude.ai/code/session_01GhCthcoeCGK29GwmdyiqGm)_

### DIFF
--- a/plan/issues/4468-marked-shape-regressions.md
+++ b/plan/issues/4468-marked-shape-regressions.md
@@ -1,10 +1,11 @@
 ---
 id: 4468
 title: "PR #4507 regressed 7 test262 tests + 2 uncatchable null_deref traps on main — object-shape trampolines / spread-source materialization"
-status: in-progress
+status: done
 sprint: current
 created: 2026-08-15
 updated: 2026-08-15
+completed: 2026-08-15
 priority: high
 horizon: m
 feasibility: medium
@@ -12,6 +13,12 @@ reasoning_effort: max
 task_type: bug
 area: codegen
 goal: correctness
+loc-budget-allow:
+  - src/codegen/closed-method-dispatch.ts
+  - src/codegen/literals.ts
+func-budget-allow:
+  - src/codegen/closed-method-dispatch.ts::fillClosedMethodDispatch
+  - src/codegen/literals.ts::compileObjectLiteralWithAccessors
 ---
 
 # #4468 — fix-forward the #4507 merge-group regressions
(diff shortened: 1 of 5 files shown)


ttraenkler:
Closing as superseded: a parallel lane fixed the same #4507 regressions first — #4579 (issue #4466, the 7 test262 restorations, `null_deref` ratchet 142→140) and #4581 (issue #4469, the nullable trampoline receiver = this PR's family 1). Both are on main and the sharded baseline has been refreshed past them, so this branch now conflicts in exactly the two files both lanes fixed. The independent root-cause analysis here agrees with theirs (same nullable-receiver mechanism); no unique fix content remains worth rebasing. Issue #4468's analysis is preserved in this PR's branch for the record.

---

PR #4617, opened 2026-08-16
feat(#4513): IR-adopt statically-foldable computed object keys

ttraenkler:
## Description

Implements #4513 (issue file in this PR): object literals with a **statically foldable** computed key — string/numeric/template-with-no-substitution/parenthesized-literal — now claim and emit in the IR instead of rejecting the whole function with `objectlit-computed-key`.

(Reserved as #4511 first; `--allocate` ran `pr_scan="degraded"` because `gh` is unavailable in-container, and main landed the usage-limit monitor as #4511 concurrently. The `check:issue-ids:against-main` pre-push gate caught it — renumbered to #4513, the #4511 reservation released. Working exactly as designed.)

**Claims now:** `{ ["a"]: v }`, `` { [`a`]: v } ``, `{ [0]: v }`, `{ [0x10]: v }`, `{ [("a")]: v }` → `emitted(body=true)`. **Still demotes** (unchanged): runtime/variable keys, `{ [`${p}b`]: v }`, `{ [1+1]: v }`, `Symbol.iterator` keys, computed **method/accessor** names (`{ ["m"]() {} }`, `{ get ["a"]() {} }`), and `class A { [k]() {} }` (different gate, different node kind). `{ a:1, ["a"]:v }` now rejects with the more precise `objectlit-duplicate-key`. No new reject code minted.

Two things measurement corrected: (1) a numeric-canonicality guard (`String(Number(text)) === text`) was drafted then **deleted as dead code** — TS's scanner already stores the canonical decimal form (`0x10`→`16`, verified across 16 spellings), so the guard documented a rejection that never fired; replaced by a runtime assertion that `{[0x10]}`/`{[16]}` are one key. (2) `{ ["m"]() {} }` is a **legacy backend bug** (emits a non-callable `m`) — recorded so nobody "fixes" the IR to match it.

**Evaluation order** pinned with the #4459 counter (`t = t*10 + k`), asserted on both paths and against V8: `{ ["b"]: p(1), a: p(2) }` → 12 (not 21, though `a` sorts first) — initializers run in source order, name-sort applies only after.

**Gates** (on the merged tip): `tests/issue-4513.test.ts` 28/28; neighbours #4471/#4459 62/62; `check:ir-fallbacks` OK; `check:ir-only` (+`--policy=hybrid`) READY; `check:ir-adoption` clean (ObjectLiteralExpression row regenerated); typecheck 0 (base 0); lint/format clean; equivalence shards 1–8 all exit 0, no new regressions; loc-budget `select.ts` +17 / `from-ast.ts` +9 granted. The single fold lives in a new leaf module `src/ir/property-key-fold.ts` so selector and lowerer consult one function and cannot drift into a post-claim invariant; each god-file takes a one-call-site swap at the data-property site only.

Two measured caveats, both **not this branch's**: `check:ir-only` standalone reads 22/15 vs 19/18 pre-merge — that delta is main's (A/B against revert copies reads 19/18 with and without the change); and `check:godfiles` fails identically without this change on files the branch doesn't touch (and no workflow runs it).

## CLA

https://claude.ai/code/session_01G8EzRKZovuUremQM9dfJAC

---
_Generated by [Claude Code](https://claude.ai/code/session_01G8EzRKZovuUremQM9dfJAC)_


ttraenkler:
## D1a handoff — 2026-09-01

The repair session is stopped. No unfinished compiler or test draft was committed or pushed.

A stacked plan amendment adds the missing frontend-neutral physical-evidence leaf and records the continuation contract. The unpublished draft closes the singleton, owner-transition, private-layout, prewrite-reauthorization, staged-install, native-witness, and occurrence-identity seams, but it is not acceptance-ready: the authority and materialization leaves still exceed their hard limits, the evidence leaf and settlement reprojection remain unfinished, and the test migration remains incomplete and unvalidated.

Continuation order is exact: land the plan amendment; extract and wire the bottom evidence leaf; bring authority, evidence, and materialization to their ratchets without allowances; finish the lifecycle tests; run the strict load-gated validation with normal hooks; then obtain a fresh exact-byte Sol review before making an implementation PR ready.

PR #4624, opened 2026-08-16
fix(codegen): Function residuals (#4483) + operator smalls (#4484) + builtin-surface smalls (#4485) + module-global undefined seed (#4489)

ttraenkler:
## Description

Closing wave (16) of the #4426 ES5-standalone campaign — four independently verified lanes, ~+45 measured flips, every sweep zero-regression:

**#4483 (Function.prototype call/apply/bind residuals) — +15/−0 on the full 509-file `built-ins/Function` sweep.** apply/call §15.3.4.3-4 TypeError guards with real instances, reflective `Function`-ctor calls, `__get_builtin` CE→decline. The finishing agent caught a real defect in its predecessor's snapshot: the apply-TypeError arm fired on plain objects owning their own `apply` — fixed with an `isCallableReceiver` guard and pinned as a negative control. Pins verified under both eval tiers (no tier arm needed — verified, not assumed).

**#4484 (operator/coercion smalls) — +9/−0 over 604 files, accepted as partial (decision recorded).** instanceof step-order fix (`S11.8.6_A6_T2`) with a subtle catch: reordering widened an over-throw on non-callable `@@hasInstance` (§13.10.2 calls it at step 4, IsCallable is step 5) — invisible to pass/fail sweeps, found by diffing error text on unflipped rows, gated by `moduleInstallsCallableHasInstance`. Plus §9.10 ToObject TypeErrors for null/undefined member access and spec-non-writable builtin strict-write throws. Shortfall vs the ≥15 bar is structural: seven rows blocked on the missing `{}` → `Object.prototype` [[Prototype]] edge, routed to #4506 (fnctor-representation XL).

**#4485 (builtin-surface smalls) — +18/−0 vs a ≥14 bar** (annexB Date 14→23; encodeURI +5 — the ONE §19.2 global function missing from the value-read name table; Error name/toString +4 — the own-`name` half was a WRITE landing where the struct.get read never looks; new `error-instance-field-write.ts`). `toGMTString` aliases `toUTCString`'s singleton per §B.2.6 (same function object). Family D (Array RangeError + `[object Array]`) found already-passing and pinned as a regression guard, not claimed. Reimplemented after a lost worktree from recovered written findings — all re-verified with fresh runs.

**#4489 (module-global undefined seed) — the substrate change, verified by a 6,310-paired-row stratified corpus A/B (25.1% of the 25,148-file scope, 1.7× the agreed floor): 3 fixes, 0 regressions.** Module-scope `var x;` reads now yield the undefined singleton instead of `ref.null.extern` (indistinguishable from the closure ABI's absent-arg pad). The corpus instrument earned its keep: phase 1 caught a real `pass → uncatchable-trap` regression (`S13_A17_T1`, call-before-init) at row ~3,900 — a #789 guarded-cast consumer whose null-test meant "unset" read the non-null singleton as "don't throw"; the predecessor's catalogue had that row marked "No change" from source reading. Fixed via `emitIsNullishAnyAt` (`is_null ∨ tag-1-singleton`, `undefinedSingletonActive`-gated, host lane byte-identical), which also repairs a years-latent function-scope twin of the same trap. Truthiness/numeric-coercion matrix pinned (`Number(x)` 0→NaN, `x === undefined` false→true, etc.).

Also: 3 inherited harness-selftest ratchet entries flipped to pass (healed by the #4480/#4482-era waves, per the ratchet's own instruction); residuals filed as #4518 (reflective String arm renders the undefined singleton as `[object Object]` — blocks 3 of #4489's R1 rows) and #4519 (member-get backup guard: `undefined.foo` wrongs instead of throwing; corpus-grade verification mandated per the #4489 precedent).

Verified post-merge here: 53/53 (issue-4489/4485/4465), 75/75 (issue-4485/4484/4483), 57/57 (issue-4484/4483 + adjacents), harness-selftests 19/19. One merge conflict (assignment.ts) resolved by keeping both new arms — #4484's spec-non-writable and #4485's Error-field write key on disjoint receivers.

## CLA

https://claude.ai/code/session_01XdMAVayXGLs5Syf9GaQzT7

---
_Generated by [Claude Code](https://claude.ai/code/session_01XdMAVayXGLs5Syf9GaQzT7)_


ttraenkler:
Park diagnosis before re-enqueue (per the auto-park protocol):

**The batch itself was anomalous.** The group's other member, PR #4489 (auto-park runner-kill classifier), is a single-file `.github/` script change with zero compiler code — and it **merged 2026-08-14**, two days before this group ran. A merge group batched with an already-merged PR is a stale queue entry; the group's merged state was not `main + two live PRs`.

**The 44-row delta is one family and does not intersect this PR's diff.** ~22 `dstr/*dflt-ary-ptrn-rest-id-exhausted` shapes (`Array.isArray(x)` false on a rest-pattern array) + 3 `Object/entries/getter-*` rows (entry not an array) + tail. This PR's four lanes touch Function call/apply guards, instanceof/`in`/assignment arms, Date/Error/encodeURI surface, and the module-global undefined seed — none touch destructuring/iterator lowering or Object.entries. (Object.entries was recently reworked by PR #4598/issue #4493, a different lane.)

**Local cross-check:** both representative rows fail with byte-identical errors on a pure `origin/main` worktree and on this branch — the delta the gate observed does not reproduce as branch-caused in the local lane.

Re-enqueueing solo (one-shot, per the bot's own batched-group instruction) as the attribution step: a solo merge-group run against a fresh baseline arbitrates. If it parks again solo, the regression is this PR's and gets fixed on the branch before any further re-admission.

---

PR #4631, opened 2026-08-16
fix(#2668): standalone ordinary indexed set must create an all-true data property

ttraenkler:
## What

In `--target standalone`, an ordinary indexed assignment that **creates** an
array element produced a non-writable, non-enumerable, non-configurable own
property. One flag constant; the rest of the diff is a comment and a test.

```js
var a = []; a[0] = 101;
Object.getOwnPropertyDescriptor(a, "0")
// was  {value: 101, writable: false, enumerable: false, configurable: false}
// now  {value: 101, writable: true,  enumerable: true,  configurable: true}
```

The knock-on was worse than the descriptor read: a later **legal** same-value
redefine threw `configurable attribute of a non-configurable property` and
aborted the whole module — that is test262 `15.2.3.6-4-212` and
`15.2.3.7-6-a-208`.

## Why (root cause)

An indexed write on an `any`/externref-typed array receiver is routed by the
`__extern_set` overlay prologue to the vec define native `__vec_dp_value`. That
call passed only the `hasValue` flag bit, which specifies **none** of
writable/enumerable/configurable — so when the index did not already exist,
CompletePropertyDescriptor (§6.2.6.6) filled the three omitted fields with
`false`.

Control reaches that call site only when the key has **no** companion descriptor
entry: every non-null-entry arm above it returns (deleted → re-seed, accessor →
setter, writable → merge-write, non-writable → silent reject). So the index is
either an implicit dense element — whose effective attributes are all true, and
which `__vec_dp_value`'s own seed already materialises — or it is brand new,
where CreateDataProperty also requires all true. All-true is correct in both
cases, which is why this is a constant and not a new branch.

## Why it survived this long

The four most obvious probes all say "this works", because each one makes the
index **backed** before the write, and a backed index gets seeded with all-true
attributes inside `__vec_dp_value`:

| shape | before this PR |
| --- | --- |
| `var a = []; a.length = 1; a[0] = 101` | correct |
| `var a = [101]` | correct |
| `var a = []; a.push(101)` | correct |
| `var a: number[] = []; a[0] = 101` | correct |
| **`var a = []; a[0] = 101`** | **wrong** |

Those four are kept in the test as controls rather than deleted — they are what
made the defect look impossible.

## Measurement

Base vs head, same source, same harness, 11 rows per lane
(`.tmp/probe4.mts`, both arms run back-to-back on this box):

- **standalone: 5 rows fixed, 0 regressed.** `t_grow_from_empty_literal`,
  `t_grow_from_new_Array`, `t_presized_then_write` (all `0` → `111`),
  `t_same_value_redefine_survives` (threw → did not throw),
  `t_empty_redefine_keeps_defaults` (`0` → `111`).
- **host: byte-identical on all 11 rows.** The change is inside a
  `ctx.standalone` branch and cannot reach host codegen. Five of those rows are
  wrong in **both** arms — separate pre-existing host gaps, listed by name in
  `HOST_SKIP` rather than pinned.

A same-population test262 A/B over the 86-file defineProperty/defineProperties/
create family (`--target standalone`, via `scripts/harness-flip-probe.ts`) is
running; the flip partition will be posted as a comment before this is
considered ready to land.

Not claimed: no test262 gain is asserted here beyond what that partition
measures.

## Scope

Deliberately narrow. This is the first landed slice of
[#2668](https://js2wasm.loopdive.com/dashboard/issue.html?slug=2668-es5-object-defineproperty-descriptor-fidelity-residual)'s
vec-index MOP work and does **not** close it. Untouched, and separately
measured as still broken in standalone:

- `Object.keys` / `getOwnPropertyNames` enumerate the unbacked tail left by
  `a.length = N` (`[0,1,2]` with `length = 6` reports 6 keys, not 3);
- an array `length` above `2^31-1` (`defineProperty(a, "length", {value:
  4294967294})` leaves `length` at 0);
- the host lane's `_vecDefineOwnProperty` "in-bounds index with no sidecar is a
  first definition" workaround.

Also untouched: the plain-object descriptor-attribute cluster owned by another
lane.

## Budget grants

`src/codegen/vec-overlay.ts` is a god-file at its LOC cap and
`fillVecOverlayHelpers` at its function cap. The +6 lines are the comment
explaining why the constant is the fix; granted via `loc-budget-allow:` /
`func-budget-allow:` in this PR's own issue file.


ttraenkler:
## Test262 partition — same population, matched settings

Both arms run back-to-back on the same box through the **real** runner
(`scripts/harness-flip-probe.ts`, which drives `runTest262File` /
`assembleOriginalHarness` with upstream `propertyHelper.js` inlined), over the
exact 86-file `Object/{defineProperty,defineProperties,create,…}` family from the
2026-08-16 ES5-standalone census. The two arms differ only in
`src/codegen/vec-overlay.ts`.

- base: `d38224d53` (this branch's point off `upstream/main`)
- head: the same tree with the one flag constant changed
- invocation, identical for both arms except `--out`:

```
npx tsx scripts/harness-flip-probe.ts \
  --files .tmp/dp-family.txt \
  --out   .tmp/2668-{base180,head}-standalone.jsonl \
  --target standalone --timeout 180000
```

```
before : {"fail":86}            after : {"fail":82,"pass":4}
union 86 · partition verified 86 == 86

fail -> pass (GAINED) : 4
pass -> fail (LOST)   : 0
other status change   : 0
unchanged             : 82
only in before / after: 0 / 0
NET                   : +4
```

The four gained:

```
built-ins/Object/defineProperty/15.2.3.6-4-210.js
built-ins/Object/defineProperty/15.2.3.6-4-212.js
built-ins/Object/defineProperties/15.2.3.7-6-a-206.js
built-ins/Object/defineProperties/15.2.3.7-6-a-208.js
```

That is the dense-default-element sub-cluster exactly — the rows whose contract
is "an element created by ordinary assignment begins writable/enumerable/
configurable, and an empty or same-value redefine preserves that". Nothing
outside the mechanism moved. **No claim is made beyond this partition**; the
other 82 rows fail for the mechanisms listed under "Scope" and are untouched.

### Two instrument caveats, stated because they bound the number above

1. **A first base arm was discarded, not reported.** It ran with the probe's
   default 60 s timeout and, under box load, turned 20 of 86 rows into
   `compile_error: compilation timeout (91 s)`. Those rows were *unmeasured*,
   not failing, and diffing them against a 180 s head arm would have
   manufactured ~20 phantom flips. Both arms above use `--timeout 180000`.
2. **The QuickJS eval provider cannot be built in this container** (no
   `clang-18`, no `cmake`, no cached artifact), so eval-shaped rows report a
   manufactured "provider is not built" failure. One row in this family
   (`15.2.3.6-4-594`) is affected; it is `unchanged` in the partition, and it is
   unmeasured in *both* arms rather than a real failure.

A broader standalone sweep over `built-ins/Array/|built-ins/Object/` is running
as a regression check on the hot path this touches (every ordinary indexed
create in standalone routes through the changed call site); result to follow.

ttraenkler:
## Park diagnosis — the 42/44 regressions are inherited from `main`, not from this PR

The auto-park bot flagged `check for test262 regressions` + `merge shard reports
→ Standalone regression guard (#1897)` on
[run 31933649385](https://github.com/loopdive/js2wasm/actions/runs/31933649385):
standalone `improvements=8, regressions=42, net −34`; host `44 regressions`.

Per the park-handling rules I diagnosed the cited run before touching the label,
and I have **not** removed the `hold`. Three independent checks say this cluster
is not this change:

**1. An unrelated PR reports the identical fingerprint.**
PR #4627 (`refactor(#4514): directional reverse-callers edge`, no overlap with
this diff) reports the **same standalone bucket signature
`37d9311a9cb806e4`, the same 42 non-CT files**:

| PR | improvements | regressions | net |
| --- | ---: | ---: | ---: |
| #4631 (this) | 8 | 42 | −34 |
| #4627 (unrelated) | 4 | 42 | −38 |

The guard prints its own verdict on that: *"Same signature on another PR ⇒
identical cluster ⇒ likely baseline drift."*

**2. This change cannot reach the host lane, yet host regressed 44.**
The edit lives inside `fillVecOverlayHelpers`, whose first statement is
`if (!ctx.vecOverlayReserved || !ctx.standalone) return;`. Host codegen is
unreachable from it. A host regression therefore cannot be caused here — and the
standalone 42 are the same cluster as the host 44.

**3. The regressed files are all `Array.isArray` assertions.**
~40 × `dstr/dflt-ary-ptrn-rest-id-exhausted` (arrow / function / generator /
async-generator / class / object destructuring — "Expected true but got false")
and 3 × `built-ins/Object/entries/getter-*` ("first entry is an array Expected
SameValue(«false», «true»)"). That is the subject of
`ef1a41d0d fix(#1888): Array.isArray static fast path claimed every ref is an
array` (PR #4629), merged to main **after** the baseline snapshot the guard
compares against — the run's own warning says
`baseline JSONL is 1 test262-relevant commit behind main HEAD (83m clock;
baseline=2926a9a)`.

### The useful part: this measures the PR in isolation

The two parked PRs share the drift, so the **difference** between them is signal.
Improvements differ by exactly **4** (8 vs 4) — an independent, full-corpus
standalone confirmation of the +4 measured locally on the 86-file family, on the
same four files. Net effect of this PR alone: **+4 improvements, 0 regressions.**

Escalated to the tech lead: this blocks every PR entering the queue, not just
this one, and the resolution (revert `ef1a41d0d`, fix forward, or deliberately
promote the baseline to the new floor) is a main-branch decision, not mine.

`main` has since been merged in and the new test is re-verified green on the
merged tree.

PR #4644, opened 2026-08-17
refactor(ir): #3954 phase 2 slice 1 — split the JS dialect out of nodes.ts

ttraenkler:
## Description

Draws the **producer axis**: instruction kinds whose meaning comes from a *language specification* now live in `src/ir/dialect/js.ts`, separate from the language-neutral core in `src/ir/nodes.ts`. This is #3954 phase 2, first slice, plus the sub-issue ([#4551](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4551-ir-instruction-neutrality-verdict)) that owns what is left.

**Sequenced ahead of the `ir-full-coverage` push by project-lead decision**, on this measurement — `IrInstr` union arms across `origin/main`:

| date | arms |
|---|---|
| 2026-05-01 | 51 |
| 2026-06-01 | 57 |
| 2026-07-01 | 58 |
| 2026-07-15 | 71 |
| 2026-08-01 | 78 |
| 2026-08-18 | 78 |

Phase 2 is O(instruction kinds), so its cost tracks that curve; `ir-full-coverage` is expected to add ~40 more. Phase 1's surface (58 `JsTag` references across 24 files) is *not* growing the same way, which is why the tag seam is the cheaper half to defer.

### What moved

23 **uncontested** ECMAScript kinds: `dyn.*` (5), `iter.*` + `forof.iter` (6), `gen.*` (4), `await`/`async.*` (3), `extern.*` incl. RegExp (5). `dyn.truthy` is ECMA-262 §7.1.2; `iter.next` is the JS iterator protocol — none of them means anything to a non-JS source language.

Declaration moves and re-exports only:
- `nodes.ts` 3,441 → 3,032 lines; **all 54 importers of `nodes.js` unchanged**.
- `import type` throughout, so the core↔dialect cycle has no runtime edge.
- `nodes.ts` still assembles the `IrInstr` union — the single sanctioned core→dialect edge.

### The boundary is enforced, not conventional

`scripts/check-ir-dialect.mjs`, wired into `quality`:
- **R1** only `nodes.ts` may import a dialect;
- **R2** `nodes.ts` re-exports every name a dialect declares.

Both were **negative-tested against deliberate violations** before being wired in — a gate nobody has watched fail is not known to work.

### What deliberately did NOT move

`vec.*`, `class.*`, `object.*`, `string.*`, `box`/`unbox`/`tag.test`, `forof.vec`/`forof.string`, `coerce.to_externref`.

Whether those are neutral is genuinely unsettled, and spot-checks reversed the intuitive reading more often than they confirmed it: `vec.*` array holes live in `src/codegen/array-holes.ts`, **above** the IR, and `string.*` is already parameterized by `IrStringEncoding` rather than hardcoding UTF-16. [#4551](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4551-ir-instruction-neutrality-verdict) owns the per-kind verdict. **An unresolved kind stays in core** — placing one in the dialect on a hunch would give a guess the authority of a lint rule.

### Verification

- `pnpm typecheck` clean (`tsconfig.ts7`).
- `biome` clean on every touched file.
- `tests/ir-*.test.ts` run **serially**: the same 2 failures as the pre-split base (`ir-bytecode-proof` OP.CALL, `ir-scaffold` selector shape), both pre-existing on `main` and untouched here. The 2 extra failures on a parallel run were timeout flakes in the ~28s WasmGC-VM cases; they pass in isolation with the split applied.
- The base copy of `nodes.ts` was captured before the first edit, so both arms of that comparison are runs actually executed rather than an inherited figure.

`#3954` was claimed on `origin/issue-assignments` before starting.

### Review is POST-MERGE, deliberately

`backend-agnostic-ir` is **Lane B's** goal; this was written by the Opus lane at the project lead's direction. The PR was held as a draft to gate on a Fable-lane review; the lead's call was to mark it ready once it was a working checkpoint, which it is — self-contained, behaviour-neutral by construction, dependent on neither #4551 nor phase 1. Holding a green finished checkpoint out of `main` buys merge-conflict risk and blocks nothing else.

The review still happens, as [#4552](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4552-fable-review-js-dialect-split) (`sprint: current`, `model: fable`, `priority: high`). Its questions are unchanged — the boundary rule, the 23 placements (`extern.*` is the likeliest disagreement), the phase inversion, and two corrections to confirm. Only the outcomes changed: accept → close; accept-with-changes → follow-up PR, cheap while the dialect is one file; reject → revert PR, since `main` is append-only and this is 23 declaration moves plus a gate.

#4552 also records **two schema questions with a clock on them**: `instrKind` in `docs/ir/ir-module.schema.json` is a closed enum of 60 entries, 19 of them JS ops, so whether a non-JS producer is possible at all is decided there rather than by the in-tree union. Recommendation on the record: open the namespace now (one edit plus a version bump while consumer count is plausibly zero), defer the in-tree open arm.

## CLA


ttraenkler:
**Review tracked as [#4552](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4552-fable-review-js-dialect-split)** (`sprint: current`, `model: fable`, `priority: high`), filed in the latest commit.

`backend-agnostic-ir` is **Lane B's** goal (fable / porffor / symphony, per `plan/method/lane-partition.md`). This PR was written by the Opus lane at the project lead's direction — a deliberate cross-lane action, now recorded rather than left implicit. The reviewer this needs is the one who did not write it.

#4552 asks for a verdict on four specific things rather than a rubber stamp:

1. **The boundary rule** — one dialect file or per-family; union-in-core (here) vs union-with-dialect (MLIR's shape); whether the gate should also assert the converse of R1/R2.
2. **The 23 placements** — `extern.*` is flagged as the likeliest disagreement, since it is arguably a *host*-boundary concern, which is the same argument that kept `coerce.to_externref` in core.
3. **The phase inversion** — phase 2 landed before phase 1 on a cost-of-delay argument. Specific risk to weigh: phase 1 may want `box`/`unbox`/`tag.test` placed differently than a dialect-first world assumes, and those three are currently in core.
4. **Two corrections** where this lane's first reading was wrong — `IrInstr` has 78 arms, not 82; `BackendEmitter` has 3 JS-shaped methods, not 12. Please confirm the corrections, not the originals.

Outcome is accept / accept-with-changes / reject-the-sequencing, and it also decides whether [#4551](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4551-ir-instruction-neutrality-verdict) unblocks or folds into #3954.

**This PR stays a draft until that review lands**, so it cannot be auto-enqueued ahead of it. All 41 checks are green and `mergeStateStatus` is `CLEAN`, so it is ready to go the moment the review says so.

---

PR #4658, opened 2026-08-19
perf(es5): standalone ES5 conformance 94.2% -> 95.4% (109 of 523 residue rows)

ttraenkler:
## ES5 standalone conformance: 94.2% → 95.4%

Seven-lane parallel push against the **standalone** (pure-Wasm, no JS host) ES5
residue. Every figure below was measured at `target=standalone` on a **frozen
tree**, not summed from lane reports.

| | |
| --- | ---: |
| standalone ES5 before | **8,506 / 9,029 — 94.2 %** |
| standalone ES5 after | **8,615 / 9,029 — 95.4 %** |
| residue converted | **109 of 523 — 20.8 %** |

The 523-row corpus is the full standalone ES5 failure set, classified with the
authoritative `scripts/generate-editions.ts` edition classifier over a fresh
baseline, so the denominator matches the published
`test262-standalone-editions.json` exactly.

## Bugs that reach ordinary programs, not just conformance rows

- **`for (var i = 0; i < a.length; i++) a[i]` read `a[0]` every iteration.** A
  constant-key resolver folded a `var` binding to its initializer regardless of
  mutability. Caught by a GC-lane unit suite while the standalone lane list, the
  551-row guard, `tsc` and both budget gates were **all green** (#4556).
- **`break` inside `for-in` was a no-op — and inside an enclosing loop it broke
  the WRONG loop.** The standalone static-unroll path emitted bodies as a bare
  straight-line sequence with no break/continue target. `continue` was broken
  too (#4561).
- **`u() == 1` never called `u`.** The equality operators emitted their operands
  and then **rolled the instructions back**, substituting a constant (#4515).
- **`Array.isArray` answered `true` for any GC reference** — in standalone that
  is every non-primitive (#4556).
- **`a["1"]`, `a[new Number(2)]` silently compiled to element 0**, on read and
  write (#4556).
- **`--this.x` on a `var`-declared script global wrote NaN** — #4500's read/write
  pair was missing a third member (#4515).
- **`++x` on an unresolvable reference threw nothing**, while `x + 1`, `x()` and
  `x += 1` all threw correctly (#4515).
- Two modules emitted **invalid Wasm** the engine refused to instantiate (#4560).

## Verification

Beyond each lane's own row list:

- **551-row standalone guard** — every row verified passing locally on a clean
  tree, so the target is a clean 551/551 rather than a partial score.
- **121-module prototype-write corpus** — built for this push after the guard
  proved blind to a class of regression. Run **one process per test**; it caught
  a real 5-row regression the guard could not see.
- **Unit suites relative to the merge base**, including GC-lane suites. One lane
  reported byte-identical sorted failing-test *name* sets, not merely equal
  counts.
- `tsc --noEmit`, LOC and function budget gates green throughout. Every budget
  allowance is justified in the owning issue's frontmatter; several lanes earned
  the gate instead by extracting into subsystem modules, leaving the god-files
  smaller.

## Deliberately not done

- **Work measured as host-only was parked, not merged** (#4559): 740 standalone
  runs across two corpora moved zero rows with byte-identical failure sets. It is
  real js-host work (0/101 → 68/101 there) and stays reachable.
- **Fixes that moved zero rows were reverted**, twice, rather than shipped
  unmeasurable into hot shared code.
- **Bucket E retired rather than deferred** (#4556): ≥146 vec-length read sites,
  only 17 converting to f64 — the other 129 consume the raw i32 as loop bounds
  and `array.new_default` sizes, so ToUint32's bit pattern would hand them −1.
  It is a representation change, not a signedness flip.
- **A `bind` `length` implementation was measured at +2/−2 and reverted**,
  exposing two pre-existing blockers instead (#4562, #4563).

## Guard accounting

The guard reads **547/551**, not 551/551, and the four are **documented
unmaskings, not regressions**: `verifyProperty` was *skipping* its
writable/configurable branches while the underlying reads returned `undefined`.
Each was verified individually, one process per test, and both minimal repros
behave identically on `main`. A static scan found **25 of the 551 rows** are the
same at-risk shape.

Also fixes `scripts/check-test-vacuity-shapes.ts`, which used
`new URL(import.meta.url).pathname` and so scanned **zero files** in any checkout
whose path contains a space.

## Issues

Filed: #4555, #4556, #4559, #4560, #4561, #4562, #4563, #4564.
Updated: #4163 (its 90 % goal was already met at 94.2 %), #4206, #4491, #4492,
#4515.

#4163 also carries the measurement caveats this push produced: vacuous passes,
parallel pollution of prototype-writing tests, contention-shaped timeouts, that
`jobs=1` is *not* one-process-per-test, that a long run needs a frozen tree, and
that "which code actually runs?" is answered by the emitted module.

## Not reachable, and why

100 % was not attainable here and the push measured the reasons rather than
assuming them. QuickJS-blocked rows counted per lane (31 / 11 / 6 / 3 / 2) —
that tier needs clang-18, which needs Xcode Command Line Tools; `libquickjs.wasm`
otherwise builds fine on macOS after two GNU-isms in
`scripts/quickjs-artifact/build.sh` were made portable (included). **The blocked
pool grows as fixes land**, because tests get past an earlier failing assertion
into eval-touching code. Where lanes measured it, the locally-decidable pool is
roughly half the nominal count.

https://claude.ai/code/session_015vAL9KZvTPwwBcFovPJ8aH


ttraenkler:
Post-merge performance handoff (measured, not extrapolated):

- The npm-compat history pins the first catastrophic standalone-dynamic artifact to `0d87f216`, immediately after this PR. The preceding `6049c004` artifact measured Acorn at `0.11443x` Node and clsx at `0.17670x`; the first bad artifact measured `0.0009507x` and `0.0001240x`, respectively. The interval contains this PR and the artifact refresh commit.
- The clsx cause was commit `536a3c0`: strict `arguments` eagerly installed the `%ThrowTypeError%` `callee` accessor through the generic object runtime on every activation. PR #4669 has fixed that in code. Its exact O4 A/B is `0.143173 us/op` fixed versus `729.364 us/op` eager, checksum `14/14`.
- Acorn is not fixed by that change. In the refresh after #4669, Acorn's O4 build still fails in Binaryen with `unexpected expr type` / `UNREACHABLE ... Flatten.cpp:231`. The npm-compat generator then deliberately ships an **unoptimized** binary and warns that its performance and size are not comparable. That explains why Acorn remains catastrophically slow on the published page.

Please have the owning ES5 session isolate the #4658 code shape that makes Acorn invalid for Binaryen Flatten, fix it, and add an exact regression covering optimized (`-O4`) standalone-dynamic Acorn. The site is currently stale at source revision `7b2a1f94`; the replacement refresh is running at https://github.com/loopdive/js2/actions/runs/32432607766.

ttraenkler:
Correction to my preceding handoff after instruction-level isolation: the Acorn O4 crash is **pre-existing and independent of this PR**. The raw module contains eight `try_table` instructions, including two in `__named_this_call_finishNodeAt_104` and six in parser closures. Binaryen 125's O4-only `Flatten` pass classifies `try_table` as control flow but implements only block/if/loop/legacy-try, so it reaches `Flatten.cpp:231`; Binaryen's current main branch still explicitly rejects `TryTable` in this pass.

This PR remains the source of the measured hot-path cliff through eager strict-arguments poison accessors; PR #4669 correctly fixes that clsx-owned regression. Please do **not** spend the ES5 session on the Binaryen failure. I am tracking and implementing the separate optimizer fallback in a repository markdown issue: retry O4 with only `flatten` skipped when this exact unsupported-pass signature occurs. The current Acorn proof already succeeds with checksum `422/422`, zero imports, and `58,826 us/op` versus Node `4,060 us/op` (`0.0690x`), compared with the stale published `0.000838x` ratio.

PR #4663, opened 2026-08-20
feat(ir): retire standalone Builtins through explicit DOM capability

ttraenkler:
## Summary

This retires the four standalone Builtins legacy bodies through the same prepared IR program already used by the JavaScript-host lane.

- prepare `el`, `crd`, `rw`, and `main` as one sealed IR component with no duplicate direct bodies;
- register an exact `dom@1` embedder capability with eight signature-checked imports and an explicitly supplied subtree root;
- authenticate the native-string boundary per runtime lifecycle, rejecting missing, foreign, donor, stale, malformed, or tampered authority;
- preserve the existing JavaScript-host DOM adapter path and keep unregistered DOM surfaces typed Unsupported;
- reuse the native string search and number-format engines for one-argument `indexOf`/`includes`, `toFixed`, and proven-ASCII case helpers;
- preserve literal CSS, batched concat, immutable search, constant bitwise, native number/string carrier, and external-boundary inlining optimizations;
- record the completed checkpoint, host-policy surface, IR census, and optimization decisions in the plan and retirement ledger.

## Migration result

- standalone IR bodies: **27 -> 31 of 37**;
- standalone legacy/typed Unsupported bodies: **10 -> 6**;
- post-claim Invariants: **0**;
- host-surface blockers: **4 -> 2**;
- call-graph blockers: **3 -> 1**;
- remaining bodies: exactly the six Calendar owners, to be retired atomically next;
- JavaScript-host lane: unchanged at **37/37 IR**.

The standalone artifact imports exactly `global_document`, `Document_createElement`, `Document_get_body`, `Element_set_innerHTML`, `Element_set_textContent`, `CSSStyleDeclaration_set_cssText`, `HTMLElement_get_style`, and `Node_appendChild`. No ambient document fallback, generic extern dispatch, or broad host-string bridge is admitted.

## Optimization and runtime envelope

The optimized IR artifact is smaller than the valid direct control on every size/work metric: **23,955 vs 24,500 raw bytes**, **16,867 vs 17,065 gzip-9 bytes**, **455,697 vs 465,391 compiler-WAT characters**, **49,949 vs 61,257 terminal-body characters**, **105 vs 136 locals**, and **105 vs 116 calls**. Both contain **124 functions** and the same eight DOM imports.

The final fresh-process direct/IR/direct medians were **88.375 / 76.750 / 72.833 us**, but the direct endpoints drifted **17.59%**, so that bracket is deliberately inconclusive. The paired control measured **96.334 vs 97.334 us** with median IR/direct ratio **0.991725x** and p10-p90 **0.7051-1.4552**. This supports runtime parity, not a speed claim. Every sample reproduced the complete 81-element/24-value/216-boundary oracle checksum.

## Validation

- focused standalone Builtins ownership, runtime, authority, shadow, optimization, and artifact suite: **14/14**;
- broader DOM/IR/inliner/async/timer/target-profile regression matrix: **115/115**;
- commit-hook changed-root selection: **32/32**;
- strict census: host **37/37**; standalone **31 IR / 6 typed Unsupported+legacy / 0 Invariants**;
- host-policy census: **33 probes / 393 imports / 0 legacy-semantic / 0 unknown**;
- optimization ledger: **50 decisions / 36 IR-owned / 3 retirement-ready / 2 source-anchored**;
- fallback, adoption, oracle, coercion, allocation, dead-export, rollback, kind-neutrality, LOC/function, stack, codegen-fallback, and harness-budget gates pass;
- typecheck, lint, formatting, issue/ID integrity, and diff integrity pass.

## Next slice

Calendar is now the only standalone residue. Its module initializer, nullable module storage, readers, callback, and main must move together with an exact clock capability for **31 -> 37 IR** and **6 -> 0 legacy/Unsupported**.


ttraenkler:
Post-merge npm-compat regression found in the live artifact (generated 2026-08-20T16:48:59Z, source 084cf798b / merged #4663), compared with artifact 245fa7d34:

- Acorn standalone runtime-dynamic: 96,006 → 18,410,114 µs/op (~192× slower); relative factor 0.1137× → 0.000576×. The binary remains import-free (2,051,667 → 2,054,300 bytes), but the benchmark wrapper is legacy and only the helper subset is IR-emitted.
- Acorn standalone static: 0.0837 → 0.1420 µs/op (~1.7× slower).
- clsx standalone runtime-dynamic: 0.319 → 240.062 µs/op (~752× slower); factor 0.1729× → 0.000269×. Binary 41,969 → 48,410 bytes, import-free, with no IR functions in the dynamic driver.

This points at the standalone builtin/string path introduced by the DOM-capability migration, not a JS↔Wasm host boundary. Please take the follow-up in the IR migration lane.

The package test counts do not reproduce a pass-rate regression in committed artifacts: Acorn is 3490/3518 → 3494/3518 and clsx remains 20/32. If the page shows 3507 or another denominator, that is a separate stale/mismatched artifact issue.

ttraenkler:
Follow-up on the regression handoff: the authoritative standalone Test262 snapshots isolate the slight pass-rate drop to one real pass loss in this merge.

- Before (`0d87f216`, baseline commit [`2f0a1329`](https://github.com/loopdive/js2wasm-baselines/commit/2f0a13293ac909c0848c863f4be7b530df2eadea)): **31,348** standalone passes.
- After (`084cf798`, this merge, baseline commit [`ce38e56c`](https://github.com/loopdive/js2wasm-baselines/commit/ce38e56c73ec148ebcf23f3ce1b285a2f6ed3db9)): **31,347** standalone passes.
- JS-host stayed **32,639**, so this is standalone-specific.

Exact lost pass:

`test/language/expressions/object/scope-meth-param-rest-elem-var-close.js`

It changed from `pass` to:

`null_deref: dereferencing a null pointer [in __module_init()]`

The neighboring `scope-meth-param-rest-elem-var-open.js` also changed from its prior assertion failure to the same `__module_init` null dereference. Both are sloppy direct-eval-in-parameter-scope tests and neither uses DOM. This makes the generic standalone runtime/IR changes in this PR—not the exact DOM capability surface—the likely shared cause.

The current baseline later recovered past this count, but that does not make the semantic regression in `084cf798` safe. Please add these two exact tests to the [#4576 IR slice](https://github.com/loopdive/js2wasm/blob/main/plan/issues/4576-standalone-dom-builtins.md) regression matrix while fixing the Acorn/clsx standalone slowdown reported above.

PR #4693, opened 2026-08-21
test(npm-compat): unblock Lodash CommonJS upstream units

ttraenkler:
## Summary

- expose a reusable Node-oriented upstream test shim without the late-initialized browser `global` alias
- compile Lodash and lodash-es's modular CommonJS graphs with the Node platform
- keep the original selected QUnit callbacks and native oracle unchanged
- add a runner regression test for the Node shim contract

## Why

Imported Lodash CommonJS helpers execute before the generated entry module initializes the shared `var global = globalThis` shim alias. Its `_root` probe therefore observed an uninitialized alias and fell back to the unavailable `Function` constructor in Wasm. The Node platform already lowers bare `global` correctly, so the Node shim omits only that alias.

## Results

Before this change, both adapters reached native 11/11 but failed all 11 Wasm callbacks at module initialization. After this change:

- lodash: 11/11 admitted original callbacks pass in Wasm; native 11/11
- lodash-es: 11/11 admitted original callbacks pass in Wasm; native 11/11
- each report keeps 1,742 unselected upstream registrations explicitly classified as deferred/unavailable infrastructure

## Validation

- `pnpm exec vitest run tests/dogfood/upstream-suite-runner.test.ts`
- `DOGFOOD_LODASH_UPSTREAM_SUITE=1 DOGFOOD_LODASH_ES_UPSTREAM_SUITE=1 pnpm exec vitest run tests/dogfood/lodash-upstream-suite.test.ts`
- `pnpm run typecheck`
- Prettier, Biome, LOC/function/oracle budgets, and pre-push checks

Related tracking: [#4585](https://github.com/loopdive/js2wasm/blob/main/plan/issues/4585-npm-compat-refresh-resilience.md)

### DIFF
--- a/tests/dogfood/lodash-upstream-suite.mjs
+++ b/tests/dogfood/lodash-upstream-suite.mjs
@@ -8,7 +8,7 @@ import { setupNpmCompatCatalogPackage } from "./npm-compat-catalog.mjs";
 import { setupLodashUpstreamSuite } from "./setup-lodash-upstream-suite.mjs";
 import {
   UPSTREAM_TEST_EXPORTS,
-  UPSTREAM_TEST_SHIM,
+  UPSTREAM_TEST_SHIM_NODE,
   cliUpstreamHarness,
   compileAndRunUpstreamModule,
   summarizeUpstreamRuns,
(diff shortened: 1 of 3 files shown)


ttraenkler:
Additional checkpoint: jsdom selected slice is 6/6 in Wasm, styled-components 6/6, webpack 13/16; Stylelint 8/9 and Redux 13/82 with remaining failures scored as runtime/compiler semantics, not unavailable infrastructure. The Lodash Node-shim fix itself remains 11/11 for both lodash and lodash-es, with 1,742 registrations explicitly deferred.

PR #4747, opened 2026-08-22
refactor(abi): #3520 C35 — retire the positional fallback's last four families

ttraenkler:
Continues [#3520](https://js2wasm.loopdive.com/dashboard/issue.html?slug=3520-ir-r1-source-qualified-identity-program-abi) (IR-only R1) after C34 (PR #4729) and PR #4733.

## What this does

C30–C34 each moved one compiler-authored callable family off the generic `retained-module-function` role, whose derived ordinal **is the function's final index** — a value that moves whenever any unrelated import or function is added or eliminated. That is a positional label, not an identity, and removing exactly that class of assumption is what R1 exists to do.

Four families were left, 15 rows on the five `SINGLE_HOST_ENTRIES`. This retires all four, taking the generic bucket **15 → 0** on that corpus.

| family | role (ordinal) | anchor | derived ordinal | rows |
| `__\0js2_call_fn_method_argc_0..5` | `closure-argc-dispatcher` (15) | entry source | dispatcher arity | 6 |
| `__async_resume_f*` + its two step adapters | `async-frame-machinery` (16) | the async function's **own unit** | resume 0 / fulfill 1 / reject 2 | 6 |
| `__vec_from_extern_N` | `vec-from-extern-materializer` (17) | entry source | position in the sorted **pre-elision** vec struct-name list | 1 |
| `__math_reduce_trig`, `Math_atan` | `stdlib-math-helper` (18) | entry source | index in a closed constant table | 2 |

Two of these are deliberately stronger than a sorted-survivor order:

- **Async is unit-anchored**, so a second async function elsewhere in the module cannot renumber the first.
- **The Math table is a compile-time constant** derived from the stdlib builtin definitions, so nothing a program contains can move an entry at all.

The vec family follows C34's rule exactly (PR #4733): the canonical order comes from the **pre-elision** record. It is keyed by the vec **struct name**, not by `vecTypeIdx` — that index is a position in the type space, so a dead-type compaction moves it.

Recording is split from planning as in C34. Planning runs from the finalization seam after DCE and — unlike C34 — **last** among the registries, which is what lets the Math family be recorded wholesale and still leave `Math_sin` with its `intrinsic-provider` owner.

**An ambiguous slot declines the role.** C35's first cut assumed a second allocator object claiming one slot was always an ownership contradiction and let the session's plan-contract invariant reject it. That assumption is false and it broke the #2980 guard suite: when the async host fallback fires the SAME declaration is lowered twice and **both** lowerings' functions stay live (measured: two `__async_resume_fg`, two `_fulfill`, two `_reject` from one `async function* g()`). A unit with two rival machineries has no single owner to name, so the slot now declines the role in a pre-pass — order-independent — and every claimant falls through to the generic sweep, restoring the pre-C35 outcome exactly. Nothing is hidden: the duplicate stays visible as two generic rows.

## Evidence

Measured in fresh processes on both sides, base `origin/main` `81edcbcaa` vs this branch, over `SINGLE_HOST_ENTRIES`:

| measure | base | C35 |
| --- | --- | --- |
| defined functions | 139 | 139 |
| generic `retained-module-function` | 15 | **0** |
| the four new roles | 0 | **6 / 6 / 1 / 2** |
| `struct-field-accessor` / `vec-host-bridge` / `closure-host-bridge` | 25 / 24 / 14 | 25 / 24 / 14 |
| `date-civil-support` / `data-struct-host-bridge` | 1 / 3 | 1 / 3 |
| `imported-function` / `body` / `intrinsic-provider` | 80 / 46 / 27 | 80 / 46 / 27 |
| **total callable rows** | **237** | **237** |
| terminal / emitted / unsupported / invariant | 37 / 32 / 5 / 0 | 37 / 32 / 5 / 0 |

Rows **moved, not created**: the callable-row total and every routing figure are unchanged, and 6 + 6 + 1 + 2 = 15 exactly accounts for the emptied bucket.

- **Byte identity** — all 30 `gc`/`standalone`/`wasi` × tracked/untracked sha256 digests match the base copies, re-verified after every merge of main.
- **`check:ir-only` and `check:ir-fallbacks --verbose`** — output **byte-identical** to base (`diff` exit 0 on both). Readiness READY, 38 terminal / 38 emitted / 0 unsupported / 0 invariants in both lanes.
- **#2980 guard suite** 5/5. **`test:changed-root`** (the gate that failed) green on all 3 changed files, 32 tests. **C35 suite** 10/10.
- **Equivalence gate, 8 shards** — no new regressions.
- Typecheck, Prettier, Biome, LOC/function budgets, `check:issues`, `check:test-vacuity-shapes` green.

**Anti-vacuity:** with `planCompilerSupportCallableAbi` disabled, 5 of the C35 suite's tests fail; the ones that stay green are the pure table/helper assertions, which correctly do not depend on the wiring. The new ambiguous-slot regression test carries its own floor (the duplicate row count of 2).

## Two census suites converted; two attempted and withdrawn

The C30–C33 suites pinned `definedFunctions: 166`, `genericRows: 45`, `closureRows: 26`, `dataRows: 5` and the routing tuple `37/30/7/0/37/30`. Every one has moved and all four have been red for weeks. **The corpus is not the reason** — `git log` reports no commit touching any of the five example files since before 2026-07-25, so those suites were reporting compiler evolution rather than the ownership property they defend. The largest measured move is legacy bodies **37 → 5**, the R-series moving terminal bodies onto the IR path.

The **data-struct and date-civil** assertions are converted to self-derived invariants (one structural owner per emitted helper of that family, none on the positional fallback, routing total and invariant-free, anti-vacuity floor), following PR #4733's argument. Both are green.

The **closure and vec** assertions were converted too, then **reverted to their unmodified `main` state**. Converting them promoted their long-red `$v0` tests into required `quality` in this PR, because `test:changed-root` runs a root test file only when the branch modifies it. The first attempt at unblocking that pinned the observed *wrong runtime values* — and that was correctly rejected on review: #4743's precedent pins a selector/bucket outcome, not a wrong value, and encoding a miscompile as expected output is a different and worse thing. Those two files are now byte-identical to main, their census assertions stay stale-red exactly as on main, and the defect keeps its full characterization in the issue file instead.

**Suite delta** across all 61 `tests/issue-3520-*` files: 22 failing tests in 17 files on `81edcbcaa` → **20 in 15**. The failing-test-name diff is a pure subtraction of the two converted census tests, with no new failure.

## Findings recorded, not fixed here

Documented in the issue file with per-file verdicts:

- **The `$v0` cluster (C30 vec export displacement).** A user export named `__vec_len` / `$v0` is displaced by the compiler's vec bridge under **every** name in the family — logical label, physical base, and appended `$` slots — so the user's function is unreachable. The closure family in the same module is intact, which localizes the defect to the C30 publication. Mechanism hypothesis (a `funcMap` name collision) recorded and labelled unverified.
- **#2980 duplicate machinery.** The module carries a complete duplicate set of async frame machinery (three extra live functions per affected async function) when the host fallback fires. Reachability of the losing set was not examined.
- Two other suites fail on an IR-first internal throw and a support-resolver slot failure — compiler defects surfacing as IR fallbacks.

## What this does not do

The `retained-module-function` code path stays as the totality guarantee: the claim is that it carries **no rows on this corpus**, not that no program can produce one. Resume-checkpoint items for `LegacyAbiAdapter` routing, exports as ABI aliases, and session notification from allocator replacement / dead-type elimination / compaction are untouched, and R1 acceptance still requires the red suite to clear. The issue stays `in-progress`.


ttraenkler:
Orchestrator verdict: HOLD on current head 794233e50c08.

The latest test commit makes changed-root green by asserting the three user-visible collision defects as expected output. That is not a C35 ownership invariant: the affected user functions still return compiler helper values (`0` instead of `801`, `901`, and `201`).

Please take the bounded stack-local route: revert only the census-conversion hunks in `tests/issue-3520-closure-host-bridge-abi.test.ts` and `tests/issue-3520-vec-support-callable-abi.test.ts`, while retaining the C35 suite and the green data/date conversions. Alternatively, fix the collision in a separately owned slice and restore the positive user-value assertions. Do not enqueue this head.

ttraenkler:
Hold addressed on head `cd6cfe11f` via your bounded route: the census-conversion hunks in both files are reverted (byte-identical to `main`; `git diff` empty), so no user-visible wrong value is asserted as expected anywhere — the `$v0` characterization lives in the issue file as prose, with a note not to re-attempt the value-pin route. The C35 suite and the genuinely-green data/date conversions are kept.

The same push fixes the guard-suite regression (`issue-2980-carrier-fallback`): under the #2980 host fallback one async declaration lowers twice with both machineries live, so a single-owner claim is genuinely ambiguous — such slots now decline the `async-frame-machinery` role in an order-independent pre-pass and fall to the generic sweep, restoring the pre-C35 outcome with the duplicate visible as two generic rows. Guard suite 5/5, changed-root green, telemetry and all 30 digests byte-identical to base. Auto-merge re-armed.

---

ttraenkler:
Released corrected head cd6cfe11f after independent review and terminal required checks. It is already queued; leave the branch and PR state unchanged.

PR #4769, opened 2026-08-22
fix(react-dom): preserve dynamic host values in project graphs

ttraenkler:
## Summary

- keep unknown `createElement` receivers on the dynamic dispatcher instead of binding them to the ambient `Document` extern method
- widen directly reassigned anonymous object/reference parameters and scalar-to-object parameters to `externref`
- keep top-level fnctor prototype writes in the host lane so React-style constructors retain their methods
- preserve callback-aware timer and stored `queueMicrotask` calls across multi-file host graphs without changing standalone/WASI policy
- collect referenced ambient callable globals while excluding same-named module bindings
- keep implicit-any property receivers dynamic for ordinary and computed property access
- record the remaining ReactDOM scheduler-flush handoff in [the tracking issue](https://github.com/loopdive/js2wasm/blob/main/plan/issues/3982-react-dom-own-unit-tests-against-compiled-wasm.md)

## Evidence

- focused Vitest: **30/30 passed** across legacy and experimental IR modes
- `pnpm run typecheck`: passed
- Prettier check: passed
- `pnpm run check:ir-fallbacks`: passed

The full ReactDOM upstream corpus is intentionally not claimed green here. The remaining blocker is compiled scheduler work/`act` Promise-microtask flush synchronization; no generated diagnostic or test-harness workaround was shipped.


ttraenkler:
Reproduced the previous merge-group high-water breach against the exact main baseline: the broad implicit-any property rule caused a 64-pass standalone drop, with 60 Promise receiver failures. The root cause was widening read-only Test262 `$DONE(error)` callback parameters and changing their Wasm ABI.

This update narrows widening to direct property assignments. ReactDOM's mutable `root.cancelPendingCommit = null` path remains dynamic, while read-only callback/error carriers retain their inferred ABI. The affected async harness and focused ReactDOM tests now match the baseline compiler output byte-for-byte.

PR #4890, opened 2026-08-25
fix(regexp): make prototype flag deletes observable

ttraenkler:
## Summary

- add a standalone, demand-gated RegExp native prototype delete arm
- limit it to RegExp-branded instances and the global, ignoreCase, and multiline accessor keys
- remove matching padded member-CSV tokens without seeding accessor getters, preserving existing plain reads
- refresh current-upstream residual pins and document the bounded A/B result

## Verification

- current upstream baseline 56afab8c32ffdf11cdd8c72e36a3299144b9622b: target rows 0/3 pass
- this branch: target rows 3/3 pass
- issue 2885: 5/5 pass
- neighboring stable controls: 6/6 pass-to-pass
- full normal commit and pre-push hooks passed; no bypass flags used


ttraenkler:

PR #4894, opened 2026-08-25
fix(#4556): honor Array overrides for inline String arrays

ttraenkler:
## Summary\n\n- Allow the standalone builtin-prototype override arm to handle the zero-argument `new Array` receiver synthesized by `String(new Array)`.\n- Keep side-effecting inline expressions on the existing single-evaluation path.\n- Add focused #4556 controls plus the assembled Test262 pin for `built-ins/String/S15.5.1.1_A1_T8.js`.\n\n## Evidence\n\n- Upstream baseline `9bed3802f`: the focused standalone Test262 row was `fail` (controls: must-pass `pass`, must-fail `fail`).\n- Rebasing onto current upstream `c4c831aa1` produced `abef934d6`; the row is `pass` and the #4556 focused suite is 6/6.\n- TypeScript 7 and TypeScript 5 typechecks pass.\n- Normal pre-commit and pre-push hooks pass, including lint, format, budget, oracle/coercion ratchets, numeric-local parity, and issue integrity.\n\nAll validation ran without `--no-verify`.

### DIFF
--- a/src/codegen/builtin-proto-member-override.ts
+++ b/src/codegen/builtin-proto-member-override.ts
@@ -49,10 +49,14 @@
  * byte-identically — `protoNamedDirty` is a pre-scan flag, so the arm is not
  * merely dead, it is never built.
  *
- * IDENTIFIER RECEIVERS ONLY. The else arm re-dispatches the whole call so the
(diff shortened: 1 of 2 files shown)


ttraenkler:

PR #4895, opened 2026-08-25
fix(array): preserve accessor-backed values through filter

ttraenkler:
## Summary

Preserve ES5 Array.prototype.filter behavior when indexed properties are accessor-backed and the source length changes during iteration. Keep descriptor-backed receivers and filter results on the externref vec carrier so getter values retain identity while length snapshots remain spec-correct.

## Test262 impact

Fresh-upstream baseline fail -> branch pass for two exact ES5 rows:

- built-ins/Array/prototype/filter/15.4.4.20-9-b-14.js
- built-ins/Array/prototype/filter/15.4.4.20-9-b-16.js

## Verification

- Exact rows: 2/2 pass
- Permanent filter suite: 11/11
- TypeScript 5 and 7 typechecks pass
- Normal pre-commit and full pre-push hooks pass, including lint/format, budgets, oracle/coercion ratchets, #3765 parity 18/18, and issue integrity
- No verification bypasses

Commits: 37c3fe6e4, a08738ab7.

### DIFF
--- a/src/codegen/array-filter-length-set.ts
+++ b/src/codegen/array-filter-length-set.ts
@@ -1,0 +1,83 @@
+// Copyright (c) 2026 Loopdive GmbH. Licensed under Apache-2.0 WITH LLVM-exception.
+/**
+ * Standalone ArraySetLength routing for typed `Array.prototype.filter` inputs.
+ *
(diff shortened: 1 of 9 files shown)


ttraenkler:

PR #4915, opened 2026-08-25
fix(es5): preserve direct-eval assignment targets

ttraenkler:
## Summary

- resolve assignment targets before a direct-eval RHS can introduce a same-named binding
- preserve potentially shadowed standalone closure results in the externref carrier
- pin the two ES5 Test262 assignment-scope regressions

## Test262 impact

- `language/expressions/assignment/S11.13.1_A6_T1.js`: fail → pass
- `language/expressions/assignment/S11.13.1_A6_T2.js`: fail → pass
- focused rows: 2/2
- composed aggregate tests: 18/18 across assignment, Array constructor, and Object.keys

## Verification

- rebased onto current `loopdive/js2:main`
- full normal pre-push hooks passed (typecheck/lint, Prettier, oracle ratchet, coercion ratchet, numeric-local parity 18/18, issue integrity)
- no verification bypasses used

### DIFF
--- a/src/codegen/closures.ts
+++ b/src/codegen/closures.ts
@@ -199,6 +199,7 @@ import {
   enclosingFunctionOwnScopeMayReachDirectEval,
   functionMayReachDirectEval,
   RUNTIME_EVAL_STATE_POOL_CAPTURE_NAME,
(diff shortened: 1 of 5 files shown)


ttraenkler:

PR #4918, opened 2026-08-25
fix(standalone): reify ES5 global own-property-name carriers

ttraenkler:
## Summary
- seed standalone global carriers for eval and the ES5 constructors with builtin descriptor flags
- preserve concrete string-vector keys through inline for-in heads
- make the global Object.getOwnPropertyNames ES5 row observe the required names

## Test262 flip
- test/built-ins/Object/getOwnPropertyNames/15.2.3.4-4-1.js

## Verification
- exact upstream source via synthetic standalone runner: pass
- focused global carrier controls: 3/3
- aggregate Array constructor pair: 2/2
- aggregate defineProperties wave: 18/18
- full normal pre-push hooks: typecheck, lint, format, oracle/coercion ratchets, numeric-local IR parity 18/18, issue integrity

Base is loopdive/js2:main; ttraenkler/js2 is used only for the head branch.


ttraenkler:

PR #4919, opened 2026-08-25
fix(es5): preserve forEach array-like coercion

ttraenkler:
## Summary
- keep constructor instances dynamic when stored in object-valued length fields
- apply ToPrimitive/ToNumber to externref array-like lengths
- terminate unresolved callback identifiers after their ReferenceError path instead of emitting an impossible standalone callback import

## Test262 flips
- test/built-ins/Array/prototype/forEach/15.4.4.18-3-23.js
- test/built-ins/Array/prototype/forEach/15.4.4.18-4-2.js

## Verification
- exact official rows: 2/2
- focused regressions and borrowed-array controls: 19/19
- clean-upstream comparison confirmed the issue-3200/2773 neighbor failures are pre-existing
- typecheck, format, LOC/function, oracle/coercion gates pass
- full normal pre-push hooks pass, including numeric-local IR parity 18/18 and issue integrity

Base is loopdive/js2:main; ttraenkler/js2 is used only for the head branch.

### DIFF
--- a/plan/issues/3169-standalone-array-hof-arraylike-receivers.md
+++ b/plan/issues/3169-standalone-array-hof-arraylike-receivers.md
@@ -5,7 +5,7 @@ status: done
 completed: 2026-07-12
 assignee: ttraenkler/dev-array-hof
 created: 2026-07-12
-updated: 2026-07-13
+updated: 2026-08-25
 priority: high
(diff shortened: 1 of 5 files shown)


ttraenkler:

PR #4924, opened 2026-08-25
fix(arguments): preserve mapped descriptor flags and accessors

ttraenkler:
Restore the measured mapped-arguments descriptor candidate from #4695. Keep numeric descriptor attributes in the vec sidecar and host proxy, route mapped accessor reads and writes through ordinary property hooks, and retain default flags for existing mapped elements so parameter-map synchronization remains spec-correct.

The bounded split passes all 20 owned Test262 rows and three controls while leaving the inferred-f64 string reverse-sync row for its ABI follow-up.


ttraenkler:

PR #4928, opened 2026-08-25
fix(for-of): throw TypeError for nullish RHS

ttraenkler:
Use the canonical JavaScript TypeError lowering at synchronous for-of null guards so ToObject(null/undefined) produces a catchable error instance. Preserve guarded-cast fallback behavior and add the exact Test262 regression with iterator controls.

# Codex session


ttraenkler:

PR #4932, opened 2026-08-25
fix(for-of): preserve destructuring-head TDZ during receiver eval

ttraenkler:
Install a bounded temporary TDZ environment for synchronous array/vec destructuring heads while compiling the receiver. Restore surrounding descriptors before existing iteration lowering so post-loop and excluded paths remain separate.


ttraenkler:

PR #4937, opened 2026-08-25
fix(for-of): close native generators on abrupt exits (#4716)

ttraenkler:
Resume native generator state machines with an implicit return(undefined) when for-of exits through break, return, or an outer continue. Recover native state refs held in generator bindings on the host lane, and cover the exact Test262 rows plus IteratorClose controls in both lanes.

# Codex session


ttraenkler:

PR #4942, opened 2026-08-25
codex/4718 es2015 forof generator close throw

ttraenkler:
Depends on #4937.

Implements #4718: close native generators when a for-of body exits via throw while preserving the break/return controls.

Validation: exact target plus six controls pass 14/14 across host and standalone; focused Vitest 14/14; TS5/TS7, lint, format, pre-push ratchets, parity, and issue integrity pass.


ttraenkler:

PR #4945, opened 2026-08-25
fix(reflect): validate setPrototypeOf Symbol and undefined arguments (#4722)

ttraenkler:
Route statically typed Symbol values through the shared non-object TypeError guard and allow only null as a primitive Reflect.setPrototypeOf prototype. Add host and standalone regressions plus live baseline and validation evidence.


ttraenkler:

PR #4944, opened 2026-08-25
codex/4720 es2015 forof unresolvable destructuring writes

ttraenkler:
Depends on #4939.

Implements #4720 by preserving unresolved identifier assignment semantics across for-of array element/rest destructuring writes.

Validation: exact/control matrix 11/11 in host and 11/11 standalone; focused Vitest 2/2; TS5/TS7, targeted lint, format, and full pre-push gates pass.

PR #4946, opened 2026-08-25
fix(#4723): preserve array rest values in accessor setters

ttraenkler:
Depends on #4944.

Implements #4723 by preserving object-literal setter callback values as externref in the host lane so for-of array-rest property writes deliver the array rather than a scalar coercion.

Validation: exact/control matrix 12/12 across host and standalone; focused Vitest 12/12; TS5/TS7, lint, format, and full pre-push pass.


ttraenkler:

PR #4947, opened 2026-08-25
fix(#4721): forward absent Proxy get traps with receiver

ttraenkler:
## Summary
- keep top-level `new Proxy` and `Proxy.revocable` bindings on the externref carrier so the Proxy MOP receives the native proxy
- preserve the receiver through `Object.create` when a standalone Proxy prototype has an absent, `undefined`, or `null` `get` trap
- enforce the ES2015 non-callable `get` trap distinction in standalone
- add host/standalone focused regression coverage and bounded issue evidence

## Validation
- exact Test262 target plus minimal controls: host 5/5, standalone 5/5
- focused `tests/issue-4721.test.ts`: 6/6
- TypeScript 5 and 7, lint, format, oracle/coercion ratchets: passed
- pre-push hook: typecheck/lint, format, ratchets, numeric-local parity 18/18, and issue integrity passed
- source delta: 131 changed lines across four source files (under the 180-line cap)

The broader nested target-is-Proxy controls remain documented as pre-existing prototype-carrier gaps and are intentionally outside this bounded fix. Latest `upstream/main` was merged without rebasing before this PR. Do not merge from this agent.


ttraenkler:

PR #4962, opened 2026-08-25
fix(function): expose hasInstance descriptor standalone

ttraenkler:
## Summary
- expose `Function.prototype[Symbol.hasInstance]` as an own standalone native method
- return its exact non-writable, non-enumerable, non-configurable descriptor
- use syntax-aware prototype mutation detection and add tracked #4739 evidence

## Validation
- exact host and standalone Test262 `prop-desc.js`: pass
- sibling standalone controls: 4/4 pass
- focused #4739 + existing #4676 tests: 5/5 pass after latest upstream merge
- TS5, TS7, lint, format, budgets, commit and pre-push hooks pass

Issue: `plan/issues/4739-es2015-function-hasinstance-descriptor.md`

PR #4967, opened 2026-08-25
fix(array): expose mutable Symbol.iterator property

ttraenkler:
## Summary
- expose `Array.prototype[Symbol.iterator]` as the exact same intrinsic closure as `values`
- seed symbol-keyed prototype companion entries without user-visible key coercion
- preserve the intrinsic in modules that do not demand the prototype companion
- track the implementation and verification in `plan/issues/4742-es2015-array-prototype-symbol-iterator.md`

## Validation
- exact ES2015 `Array/prototype/Symbol.iterator.js` standalone row
- exact `not-a-constructor.js` standalone row
- host identity and descriptor control
- #4739 dependency regression
- TS5 and TS7 typechecks; LOC/function/oracle/coercion/issue gates; commit and pre-push hooks

Depends on #4962.


ttraenkler:

PR #4970, opened 2026-08-25
fix(for-of): preserve escaping proxy globals

ttraenkler:
Follow-up to #4931 after its merge-group retry exposed and then landed four Proxy regressions.\n\nModule-global Proxy bindings now apply the existing call/new escape gate: values passed to native generic operations retain the target-shaped representation, while the non-escaping iterator Proxy remains externref-backed. Focused host Test262 verification passes the exact iterator row plus all four regressed Proxy rows, and dedicated WAT pins cover both slot contracts.


ttraenkler:

PR #5125, opened 2026-08-28
fix(#5095): Array.prototype.at() with no argument answers element 0

ttraenkler:
Fixes [#5095](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5095-array-prototype-at-no-argument) — `[10,20,30].at()` answered `0` (the *index*) instead of `10` (the element).

## Root cause

`compileArrayAt` (`src/codegen/array-methods.ts`) opened with a hard reject:

```ts
if (callExpr.arguments.length < 1) {
  reportError(ctx, callExpr, "at() requires 1 argument");
  return null;
}
```

`index` is an ordinary parameter, so `arr.at()` is legal JS and is exactly `arr.at(0)` — `ToIntegerOrInfinity(undefined)` is `+0` (§7.1.5 step 2). The reject returned `null`, whose diagnostic the caller **swallows** (`compile()` reports `success: true` with an **empty** `errors` array), and the call collapsed into the caller's degraded fallback. The two spellings in the report — `0` in string position, `undefined` in value position — are that one collapse seen through two coercion paths.

Fix: replace the reject with an `i32.const 0` index arm, the same shape `Array.prototype.includes` already uses for its absent `searchElement` (`emitIncludesSearchValue`, [#2872](https://js2wasm.loopdive.com/dashboard/issue.html?slug=2872-standalone-typedarray-prototype-cluster)).

## Measured before/after

A/B on `src/codegen/array-methods.ts` alone (file-copy revert, same probe, same worktree), sha256 of the emitted module:

| shape | gc before → after |
| --- | --- |
| `a.at()` | `0f1e06b6d00d` → **`f3eea868812c`** |
| the other 17 probed shapes | unchanged, byte-identical |

Byte-identical across the fix: `at(0)`, `at(undefined)`, `at(NaN)`, `at(-1)`, `at("1")`, `at(i)`, `ta.at(0)`, `ta.at(-1)`, `"abc".at()`, `"abc".at(1)`, `indexOf(20)`, `indexOf()`, `includes(20)`, `includes()`, `lastIndexOf()`, `slice(1)`, externref-vec `at(0)`.

Two things fall out of the hashes:

- After the fix, gc-mode `a.at()` is byte-identical to `a.at(0)` **and** `a.at(undefined)` — pinned as a test so the three spellings of index 0 cannot drift apart.
- Before the fix, `a.at()` had the **same** hash as `a.indexOf()` and `a.lastIndexOf()`. That is the shared degraded-fallback collapse, identified by hash rather than inferred — and it is what surfaced the sibling defect below.

Values (gc): `[10,20,30].at()` `0` → **`10`** in value position, `"0"` → **`"10"`** in string position.

## Two findings beyond the report

- **The TypedArray receiver was also broken.** The issue's table lists it as already-correct; it is not. It shares this lowering, so `new Int32Array(3).at()` answered `0` rather than element 0. Fixed by the same change, pinned in both lanes.
- **`indexOf()` / `lastIndexOf()` with no argument are broken the same way** — `[10,20,30].indexOf()` answers `0` where §23.1.3.13 requires `-1`. **Not fixed here**: the correct answer is a different search value with different comparison semantics (strict equality, holes skipped via HasProperty), not a defaulted index, so it carries its own test262 blast radius. Written up in the issue file; the follow-up issue is being allocated by the orchestrator.

`includes()` with no argument was already correct and is guarded so. Empty-array `at()` is pinned as an **equality against `at(0)`**, not a literal: the native vec renders an out-of-bounds read of a number vec as `NaN` rather than `undefined`, which is pre-existing and general (`[10].at(5)` does the same) — pinning it as a literal would turn a separate bug into a fixture.

## Tests

- `tests/issue-5095-array-at-no-argument.test.ts` — 13 tests: no-argument case in value and string position (gc + standalone), empty-array equality against `at(0)`, TypedArray receiver in both lanes, the gc byte-identity assertion, and the already-correct forms as regression guards.
- `tests/equivalence/array-at-no-arg.test.ts` — 5 rows, modelled on the sibling `array-includes-no-arg.test.ts`.
- `tests/issue-3481-step2-symbol-arg-revalidation.test.ts` — its `[].at()` case is tightened from "does not throw" to `toBe(10)`, as that file anticipated.

## Gates

`typecheck` · `lint` · `format:check` · `check:func-budget` · `check:coercion-sites` · `check:oracle-ratchet` · `check:dead-exports` · `check:issue-ids` · `check:done-status-integrity` · `check:issue-spec-coverage` · `check:test-vacuity-shapes` — all 0. `check:ir-fallbacks` unchanged. `check:loc-budget` 0 with a `loc-budget-allow` grant in the issue frontmatter (+4 lines: the fix is a 3-line arm replacing a 4-line reject; the growth is the comment recording why the zero-argument form must not be rejected).

Equivalence gate run as **8 shards**, all green: 24 failing / 1,685 passing against 24 known-failures in the baseline — zero new regressions, failure count exactly the baseline.

test262 `built-ins/Array/prototype/at/` and `built-ins/TypedArray/prototype/at/` are not run locally; CI's `merge_group` re-validation is the gate.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Namds9phYeHvpLKu735io3)_

### DIFF
--- a/plan/issues/5095-array-prototype-at-no-argument.md
+++ b/plan/issues/5095-array-prototype-at-no-argument.md
@@ -1,10 +1,17 @@
 ---
 id: 5095
 title: "Array.prototype.at() with NO argument returns the index, not the element (native vec lowering only)"
-status: ready
+status: done
+completed: 2026-08-28
 sprint: current
 created: 2026-08-27
-updated: 2026-08-27
+updated: 2026-08-28
+# +4 lines in src/codegen/array-methods.ts (2026-08-28). The fix itself is a
+# 3-line arm that REPLACES a 4-line hard reject; the growth is the comment that
+# records why the zero-argument form must not be rejected, so the next reader
+# does not restore the reject. Net source delta excluding comments is negative.
+loc-budget-allow:
+  - src/codegen/array-methods.ts
 priority: medium
 horizon: s
 feasibility: easy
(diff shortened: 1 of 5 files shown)


ttraenkler:
Hold removed after diagnosis — the parked regression is not this PR's diff.

Evidence (A/B via file-copy revert of `src/codegen/array-methods.ts` on merge base `b1cc63d1b1`, authoritative `runTest262File` harness path):

- The failing test (`built-ins/TypedArray/prototype/map/return-new-typedarray-conversion-operation-consistent-nan.js`) contains zero `.at(` occurrences, so this PR's changed lowering is never reached; its compiled module is **byte-identical** with and without the diff on both lanes (gc `7c5aa1fd2069`, standalone `2662fc0c647b`).
- The module hash the gate saw *did* move — on **main**: plain `origin/main` `09a016fd5f` and this branch merged with it both produce `5f661843bce3`. The mover is a main-side commit between the merge base and the queue run (likely `7ce653f416`, Array.from argument evaluation — the failing iteration's `makeArray` is `Array.from(...)` in `testTypedArray.js:70`; the "Testing with X and Y." suffix names the arg-factory, not the receiver).
- The test passes deterministically on base, branch, merged state, and plain main, both targets, 4 repeats — not reproducible locally.

Re-admitting once per the park protocol. If it parks again with the same signature (`ca79065c32d815d4`), the cause is on main and will reproduce on any PR queued after `09a016fd5f` — that becomes a main-side investigation, not this PR's.

Separate systemic finding filed for follow-up: the #1222 "wasm-unchanged ⇒ runner variance" exclusion in `scripts/diff-test262.ts` can never fire — the baseline jsonl carries no `wasm_sha` field (0 of 48,735 entries), so every flaky flip is reported as a hard "wasm-hash-changed" regression.

---

Issue #5091, opened 2026-08-27
ES2015 standalone Set.prototype.forEach rejects Symbol callbacks


ttraenkler:
---
id: 5091
title: "ES2015 standalone Set.prototype.forEach rejects Symbol callbacks"
status: done
sprint: current
created: 2026-08-27
updated: 2026-08-27
priority: high
horizon: s
feasibility: easy
reasoning_effort: medium
task_type: conformance
area: codegen
language_feature: set-prototype-foreach
es_edition: es2015
goal: standalone-mode
related: [3573, 2162]
assignee: "ttraenkler/codex-es2015-residual-20260827"
files:
  - src/codegen/map-runtime.ts
  - tests/issue-5091-set-foreach-symbol-callback.test.ts
  - plan/issues/5091-es2015-set-foreach-symbol-callback.md
loc-budget-allow:
  - src/codegen/map-runtime.ts
---

# #5091 — ES2015 standalone `Set.prototype.forEach` rejects Symbol callbacks

## Problem

The maintained ES2015 Test262 row
`built-ins/Set/prototype/forEach/callback-not-callable-symbol.js` passes in the
host target but fails in standalone. `Set.prototype.forEach(Symbol())` must
throw a `TypeError`; the standalone native collection path does not classify a
Symbol callback as statically non-callable and falls through to the host
`Set_forEach` import instead.

This is the one Symbol-shaped tail left by #3573, whose five literal callback
cases (`null`, `undefined`, number, boolean, and string) are already green.
The change is limited to the native `Set`/`Map` forEach callback guard. Dynamic
callbacks and Wasm closures must retain their existing routing.

## Exact cohort and baseline (2026-08-27)

The focused cohort is exactly one ES2015 row:

- `test/built-ins/Set/prototype/forEach/callback-not-callable-symbol.js`

Fresh authoritative snapshots were fetched with
`node scripts/fetch-baseline-jsonl.mjs --force` and
`node scripts/fetch-baseline-jsonl.mjs --standalone --force` from the pinned
Test262 revision `b363f29d3c43c626dc852744ad64a0b48a003693`. The baseline row
is **host 1/1 pass** and **standalone 0/1 pass, 1/1 compile error**. The
standalone error is a `host_import_leak` for `Set_forEach`; there are no
timeouts or skips in this cohort. The neighboring five #3573 rows remain
standalone pass controls.

## Implementation plan

1. Extend the existing static non-callable check in
[...]

PR #5198, opened 2026-08-29
fix(codegen): Symbol sub-families — symbol[] vec brand fixed, Map rows re-diagnosed (#3481) ✓

ttraenkler:
Works the two Symbol sub-families [#3481](https://js2wasm.loopdive.com/dashboard/issue.html?slug=3481-bigint-symbol-coercion-value-rep) left open after PR #5174: the `symbol[]`-declaration gate (fixed) and the Map rows (re-diagnosed, deliberately not fixed — see below).

Every row was re-derived on `origin/main` before any code was written; all six still reproduce with the exact cluster signature, so the enumeration is accurate as a membership list — but the Map sub-family's stated **cause** is not.

## `symbol[]` — fixed, 2 rows flip fail → pass

A symbol VALUE lowers to an i32 id, so `symbol[]` was registered as the SAME `__vec_i32` type as `boolean[]` and the i32 typed-array views. `buildElemCoerce` therefore materialised `Object.getOwnPropertySymbols(o)` — a real host array of real host Symbols — by `__unbox_number`-ing every element, i.e. `Number(Symbol())`, which throws §7.1.4. It threw at the **declaration**: a program that only ever reads `syms.length` died too.

The fix is a brand, not a new mechanism. `{kind:"i32", symbol:true}` and `symbolBoundaryCoercionInstrs` already existed ([#2847](``https://js2wasm.loopdive.com/dashboard/issue.html?slug=2847-acorn-cosmetic-marshalling-quirks)/[#3676](https://js2wasm.loopdive.com/dashboard/issue.html?slug=3676-symbol-host-i32-rep))`` and already cover both lanes; `symbol[]` simply never carried the brand.

- `src/codegen/index.ts` — `resolveWasmType`'s Array arm applies `symbolBrand` and registers the branded vec under its own key (`__vec_i32_symbol`). **The separate key is what makes the blast radius provable**: a module with no `symbol[]` never registers the type, so its type section is untouched.
- `src/codegen/type-coercion.ts` — `buildElemCoerce` routes a branded element through the existing seam, **host lane only**. Standalone/WASI keep the [#2866](https://js2wasm.loopdive.com/dashboard/issue.html?slug=2866-standalone-symbol-carrier) guarded `$Symbol`-carrier arm: the helper's standalone branch is an unguarded `ref.cast` that would trap on a non-symbol element, and with no carrier registered it reaches for a host import a host-free module must never acquire.
- `src/codegen/vec-access-exports.ts` — `__vec_get`'s box chain gains an `i32_symbol` arm using `__box_symbol`.

**No new host import** — both `__unbox_symbol` and `__box_symbol` are already in the manifest. `check:host-import-policy` passes with no baseline bump.

### Reachable cohort — 638 rows, run COMPLETELY on both sides

`Object.getOwnPropertySymbols` is the only signature in the TypeScript lib returning `symbol[]` (verified against `lib.*.d.ts`), and no test262 harness file mentions it — so the 43 rows whose own source names it are the complete set reachable that way. The other 595 are a deterministic every-90th stride control.

| | pass | fail | compile_error | skip |
| base | 381 | 154 | 3 | 100 |
| fix | **383** | **152** | 3 | 100 |

**+2 fixed, 0 regressed.** Per row, **624 of 638 identical on `wasm_sha` AND `status` AND the full error string.** 14 moved bytes. Of those, **11 keep an identical status AND an identical error string**; a 12th keeps its status and changes only its error text (the Proxy row below); the remaining 2 flip fail → pass:

- `built-ins/Object/getOwnPropertySymbols/object-contains-symbol-property-with-description.js`
- `built-ins/Object/getOwnPropertySymbols/object-contains-symbol-property-without-description.js`

The control sample earned its keep: one of its 595 rows moved bytes (pass → pass) and exposed a **second** way a `symbol[]` arises that the grep missed — an array literal of symbols (`[Symbol.toPrimitive]` in `testIntl.js`). That form never reaches `buildElemCoerce`; the only behaviour it picks up is the `__vec_get` box arm, which makes a dynamic read answer the Symbol instead of the raw id.

### Byte identity — 39 shapes × host / standalone / wasi

117 compiles per side: **87 byte-identical**; the 30 that move are exactly the 10 `symbol[]` shapes × 3 targets. Standalone and WASI are unchanged for everything else **by measurement, not by argument**.

| shape (host) | base | fix |
| --- | --- | --- |
| `const syms = Object.getOwnPropertySymbols(obj)` | THROWS | `1｜symbol｜true` |
| annotated / `readonly` / no-description / well-known / registered | THROWS | correct, identity holds |
| declaration only, array never read | THROWS | `1` |
| dynamic `(syms as any)[0]` | THROWS | `symbol｜true` |
| `(symbol｜number)[]` — not branded | `2｜1` | `2｜1` (byte-identical) |

### Correction to the PR #5174 record — the `Uint8Array` pin

PR #5174 hands forward that `new Uint8Array([Symbol()])` must keep throwing TypeError. **Measured on base it does not throw — it answers `101`, the raw symbol id**, as do `Int8Array`, `Uint16Array`, `Uint32Array`, `Int32Array`, `Float64Array`, `number[]` and `boolean[]`. An array LITERAL never reaches `buildVecFromExternref`. All eight are byte-identical base vs fix.

The shape that DOES exercise the shared path is a **host array** materialised into a numeric/boolean vec, and there the spec throw is real and preserved — byte-identically:

| shape | base | fix |
| --- | --- | --- |
| `const src: any = [Symbol()]; new Uint8Array(src)` | TypeError | TypeError (byte-identical) |
| same for `Uint32Array` / `Int32Array` | TypeError | TypeError (byte-identical) |
| `const a: boolean[] = src` / `const a: number[] = src` | TypeError | TypeError (byte-identical) |

The pin holds; it was pinned on the wrong shape. Both are now regression guards.

### The third `symbol[]` row does not flip — and the blocker is not symbols

`Proxy/ownKeys/call-parameters-object-getownpropertysymbols.js` loses its Symbol coercion (error changes to `SameValue(«undefined», «Symbol(a)»)`) and then fails on a pre-existing defect: **the `ownKeys` trap's return value is ignored entirely.** Proved on base with a trap returning a plain literal — `ownKeys: t => ['x']` also yields length `0`. No Symbol involved. Worth its own issue.

## Map rows — re-diagnosed, NOT fixed

The cluster table calls these "statically-symbol key type". Measured, that is not what blocks them.

`new Map([[4, 4], ['foo3', 3], [s, 2]])` is inferred by **TypeScript itself** as `Map<number, number>`. The `forEach` callback's parameters therefore lower to `f64`, and the WebAssembly JS API's own argument coercion applies ToNumber to whatever the host passes: a Symbol key throws, a **string** key silently becomes `NaN`.

| probe | on base |
| --- | --- |
| `Map<symbol, number>` + `forEach((v, k) => …)` | THROWS |
| `Map<string｜number｜symbol, …>` + same | works (a union lowers to `externref`) |
| the real construction with the symbol REMOVED, asserting a string key | fails: `SameValue(«NaN», «"foo3"»)` |
| the real row with the callback bound to a `var` first | **passes, all three rows** |

A fix was built, measured and **reverted**: widening the callback's parameter *representation* to `externref` for `Map`/`Set` `forEach` fixed the shapes where the parameter is never stored, but flipped zero test262 rows — the row bodies do `results.push({ value, key })`, and the object literal's field types and the array's element type re-narrow to `f64` from the same TS types (confirmed in the WAT: `struct.new` with `__unbox_number` on both fields). Shipping blast radius across every `Map`/`Set` `forEach` callback for no measured conformance is not worth its budget.

What it actually needs: the parameters must be `any` in the **type system**, not merely `externref` in the representation — a node-level dynamic-parameter override consulted by every site deriving a representation from `getTypeAtLocation`. That is not a coercion-site slice, and it is neither the `__module_init` blocker nor [#2949](https://js2wasm.loopdive.com/dashboard/issue.html?slug=2949-ir-dynamic-value-representation).

Adjacent defect recorded, not fixed: `Map.prototype.forEach` on a map built from a heterogeneous entries literal **silently corrupts** non-numeric keys — `'foo3'` reads back as `NaN`, no throw, no diagnostic. The Symbol case is the loud instance of an otherwise silent bug.

## Tests

`tests/issue-3481-symbolarr-vec-brand.test.ts` — 26 cases: eight `symbol[]` materialisation shapes, the dynamic `any` read, the empty array, a pair of lane assertions (the JS-host module DOES acquire `__unbox_symbol` — the positive half, without which the negative one is near-vacuous — and the standalone module acquires no import at all), and regression guards for the shared i32-vec path (host array with a Symbol still throwing a real `instanceof TypeError` for `Uint8Array` / `Uint32Array` / `Int32Array` / `boolean[]` / `number[]`, `(symbol | number)[]` unbranded, host-array `boolean[]` / `number[]` / `string[]`, numeric typed arrays, symbol property keys, `Symbol.keyFor`, `typeof`). One pre-existing gap is pinned at its **observed** value with the reason inline rather than left untested.

**Non-vacuity: 10 of 26 fail against the base compiler** — the nine materialisation cases plus the assertion that the JS-host lane now reaches the symbol seam at all. The 16 that pass on base are exactly the guards, the empty-array case and the standalone zero-import floor.

## Gates

`typecheck` · `lint` · `prettier --check` · `check:loc-budget` · `check:func-budget` · `check:coercion-sites` · `check:oracle-ratchet` · `check:dead-exports` · `check:ir-fallbacks` · `check:host-import-policy` (no baseline bump) · **8/8 equivalence shards, "No new equivalence regressions" each**. LOC and func budgets were also run with `LOC_GATE_BASE=$(git rev-parse origin/main)` to simulate CI's merge-preview base; both green, with the growth granted by this PR's own issue file (the grants are added here, so they cannot be stranded).

The [#3481](https://js2wasm.loopdive.com/dashboard/issue.html?slug=3481-bigint-symbol-coercion-value-rep) suites (174 tests across six files) and the adjacent Symbol/vec suites ([#1732](https://js2wasm.loopdive.com/dashboard/issue.html?slug=1732-builtin-method-value-construct-brand-and-descriptors) / [#2378](https://js2wasm.loopdive.com/dashboard/issue.html?slug=2378-fn-symbol-bigint-weak-proto-value-read) / [#2610](https://js2wasm.loopdive.com/dashboard/issue.html?slug=2610-standalone-symbol-wellknown-value-read-refusal) / [#2866](https://js2wasm.loopdive.com/dashboard/issue.html?slug=2866-standalone-symbol-carrier) / [#3676](https://js2wasm.loopdive.com/dashboard/issue.html?slug=3676-symbol-host-i32-rep) / [#3961](https://js2wasm.loopdive.com/dashboard/issue.html?slug=3961-symbol-valued-struct-field-loses-identity) / [#4616](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4616-cross-module-shorter-arity-callback-dispatch) / [#4776](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4776-es2015-symbol-prototype-valueof-standalone) / [#5091](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5091-es2015-set-foreach-symbol-callback) / [#5107](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5107-es2015-symbol-toprimitive-descriptor), 117 tests) are green.

`origin/main` was merged in after the measurements above and every gate, the six scoped rows and all 226 tests in the [#3481](https://js2wasm.loopdive.com/dashboard/issue.html?slug=3481-bigint-symbol-coercion-value-rep) + symbol suites were re-run on the merged state — both flips still hold, all gates still green. The equivalence 8/8 above was run on the pre-merge branch, which is the cleaner attribution for this change in isolation; a second 8/8 run on the merged tree is reported in the PR thread.

[#3481](https://js2wasm.loopdive.com/dashboard/issue.html?slug=3481-bigint-symbol-coercion-value-rep) stays **open** — it is a multi-slice issue and the Map sub-family, cause 4 and the remaining blocked rows are unresolved.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Namds9phYeHvpLKu735io3)_


ttraenkler:
### Merged-tree equivalence — 8/8 green

The PR body promised this run in the thread. `origin/main` (`23bc3ddece`) is merged into the branch, and all 8 equivalence shards were re-run on that merged tree in separate processes (`EQUIVALENCE_FORK_HEAP_MB=4096`, `VITEST_FORK_MAX_OLD_SPACE_SIZE=4096`). **`failed shards: none`** — every shard reports "No new equivalence regressions".

| shard | failing | passing | verdict |
| --- | --- | --- | --- |
| 1 | 5 | 181 | ✓ no new regressions |
| 2 | 4 | 212 | ✓ |
| 3 | 3 | 294 | ✓ |
| 4 | 0 | 166 | ✓ |
| 5 | 2 | 196 | ✓ |
| 6 | 7 | 213 | ✓ |
| 7 | 2 | 271 | ✓ |
| 8 | 1 | 185 | ✓ |

The failing counts are **identical, shard for shard, to the pre-merge run** (5 / 4 / 3 / 0 / 2 / 7 / 2 / 1), all against the same 24 known-failures baseline. The two shards whose *passing* counts rose — 7 (268 → 271) and 8 (175 → 185) — gained tests that arrived with `main`, not from this branch.

So both runs agree: the pre-merge one isolates this change, and the merged-tree one confirms nothing in the interaction with `main` moves the set.

Also re-run on the merged state and green: `typecheck`, the full ratchet chain (`check:loc-budget`, `check:func-budget`, `check:coercion-sites`, `check:oracle-ratchet`, `check:dead-exports`, `check:host-import-policy`, `check:issues`, `check:done-status-integrity`), both budget gates under `LOC_GATE_BASE=$(git rev-parse origin/main)`, the six scoped test262 rows (both flips still hold), and 226 tests across the #3481 and adjacent Symbol/vec suites.

Left as a **draft** deliberately — for the orchestrator to validate and release. Not un-drafted, no auto-merge, not enqueued.

---

PR #5210, opened 2026-08-29
feat(#3520): establish structural Program ABI ownership

ttraenkler:
## Summary

- Establish source-qualified structural Program ABI ownership for inherited class-method aliases and graph-global module initialization, for [#3520](https://js2wasm.loopdive.com/dashboard/issue.html?slug=3520-ir-r1-source-qualified-identity-program-abi).
- Preserve exact allocator/function and binding identity through planning, module bindings, and final publication.
- Add fail-closed identity, source, cardinality, collision, and mutation coverage.

## Hold repairs (`2904ff70`)

The two review HOLDs below are cleared, and so is a third item CI raised on `ee5db545` after this PR was opened. The three had one thing in common: an invariant that was written down but never actually reached by the code or the test that claimed to exercise it.

### HOLD 1 — graph-global module-init planning was unbounded — CLEARED

The old loop planned one support callable per surviving observation, at that observation's own ordinal. It therefore accepted a graph with **no** live pass, with **several** passes, and with a pass published at **ordinal 1+** — none of which the single public entry point can describe.

`planRetained` now requires the graph-global population to be exactly: one raw observation, one live observation, the *same* observation, ordinal zero, on its exact allocator object, and retained on it (`hasPlan` + `hasLocator`). Absence is only a defect when the module actually publishes the surface, so the requirement is anchored on the public `__module_init` export rather than firing on every module with no top-level state.

The public half cannot live in that method: export aliases are raised by `ProgramAbiExportRegistry.planRetained`, which runs *after* this registry. `assertGraphGlobalPublicAlias` is therefore called from the finalization driver once exports exist, and proves exactly one public `__module_init` whose `aliasOf` **and** `intent.targetId` are pass zero.

Four fail-closed mutations, applied to **production planning state** during a real multi-source compile — not to a copy of the published entry list:

| Mutation | What it does | Rejected by |
| --- | --- | --- |
| zero observations | clears the registry's observation list | `found 0 raw and 0 live` |
| two observations | appends a duplicate pass | `found 2 raw and 2 live` |
| ordinal one | republishes the single pass at ordinal 1 | `found 1 raw and 1 live at ordinals [1]` |
| retargeted export | repoints public `__module_init` at `score`'s already-owned target | `public __module_init is not the exact alias of graph-global pass zero` |

The retarget deliberately reuses an already-owned callable: a descriptor aimed at an unowned function is caught by export planning's own "no Program ABI owner" guard, which would prove nothing about this invariant.

### HOLD 2 — the missing-canonical-locator negative did not reach the locator code — CLEARED

Confirmed exactly as reported. The mutation deleted the allocator, so `planRetained` resolved `definedFuncAt(observation.funcIdx)` to `undefined` and stopped at the **"lost its exact allocator"** guard, several checks before the locator validation. Both guards raise `missing-required-locator`, so the old `code`-only assertion could not tell the two apart and the intended lines were never executed.

Reworked into two mutations that keep the canonical `WasmFunction` **live** and corrupt only the required Program ABI draft locator:

- **missing** — the unit observations that would plan the required canonical draft are removed, so `getDraft(aliasOf)` is `undefined`;
- **wrong ownership** — the canonical draft is planned with the identical reference and signature but its allocator object is a *second, equally live* function, so every predicate in the block matches except `locatorObjectForBinding(aliasOf) !== liveFunc`.

**Non-vacuity is proven in both directions, inside the test:**

| Direction | Assertion |
| --- | --- |
| the OLD mutation stopped early | deleting the allocator is asserted to throw `/lost its exact allocator/`, i.e. the earlier guard, never the locator block |
| the unmutated fixture is not already failing | the same fixture with no mutation is asserted `not.toThrow()` |
| the NEW mutations reach the intended lines | both are asserted to throw `/has no exact canonical Program ABI locator/` — the message raised only by the locator-validation block |
| the allocator really stayed live | `mod.functions` is asserted to still contain the canonical (and, for case B, the decoy) after the guard fires |

### CI equivalence regressions — FIXED IN THE CODEGEN

`super-property-access :: super.prop accesses parent class property via getter` and `super-element-access :: super['getter'] accesses parent getter via element access` both failed on `ee5db545` with

```
inherited class callable … disagrees with its exact canonical signature
```

One root cause, as suspected — inherited-alias handling:

| Draft | Raised at | Receiver param |
| --- | --- | --- |
| alias | class-body compilation | `ref typeIdx 14` |
| canonical | `planRetained`, after `eliminateDeadImports` | `ref typeIdx 1` |

Same function, same contract, two type-index *epochs*. `applyTypeLayoutRemap` rebases the session's `callableTypeContracts` but never rewrites a draft's `intent.signature`, which is frozen when the draft is raised — so the exact-signature invariant was comparing dead-type elimination's renumbering, not the ABI, and rejected a correct module.

Moving the alias plan later is not available: class-body compilation is the only point at which its prepared component scope is still open, and a draft raised afterwards is refused outright (`would mutate sealed prepared scope` — verified, that was the first attempt). The comparison instead now reads the **rebased contracts** through a new read-only accessor, `ProgramAbiSession.currentCallableSignature`, and additionally requires both to equal the live function's own current signature — a strictly stronger check than the one it replaces. Nothing was added to any baseline.

### WASI / start-section smoke failure — FIXED (`0ccfd486`)

The graph-global invariant above shipped asking the wrong question, and the `smoke` check caught it on `be9963f1`:

```
Codegen error: graph-global module-init expects exactly one public __module_init export, found 0
```

Publishing that name is only **one of three invocation policies**. `declarations.ts` derives `exportModuleInit = deferTopLevelInit && !wasi`; the Wasm `start` section and WASI's `_start` adapter reach the same body without publishing any name, so **zero public exports is the correct shape there**. Keying on the export *surface* rather than the *policy* rejected correct modules. The invariant now keys on the policy, and `assertGraphGlobalPublicAlias` returns early where no public entry point exists.

The defect was **wider than the WASI lane reported**: plain multi-source `start`-section builds, no WASI involved, were rejected too.

Regression coverage the invariant shipped without — both non-export policies are compiled and asserted to publish zero `__module_init` exports, with the deferred-export policy asserted to publish exactly one so the distinction is real rather than a blanket exemption. The test has to be **multi-source**: a single source has one module-init unit which the exact-unit path claims, so the graph-global branch is never entered and a single-source WASI compile cannot reproduce this at all. The first version of the test was single-source and passed against the buggy code; catching that vacuity is what surfaced the start-section half.

| Check | Result |
| --- | --- |
| `examples/native-messaging/smoke-test.sh` | **PASS** end-to-end under real wasmtime 44.0.0 — compile, byte-exact 17-byte stdout frame, clean stderr |
| new test vs. the buggy invariant | fails with the exact CI error at `program-abi-module-init-planning.ts:374` |

## Validation

All commands run bare, never piped, and re-run in full on `0ccfd486`.

### Tests

| File | Passed / total |
| --- | --- |
| `tests/issue-3520-module-init-callable-abi.test.ts` | 5 / 5 |
| `tests/issue-3520-class-method-alias-abi.test.ts` | 3 / 3 |
| `tests/issue-3520-integration-population-identity.test.ts` | 8 / 8 |
| `tests/issue-3520-module-binding-class-identity.test.ts` | 4 / 4 |
| `tests/equivalence/super-property-access.test.ts` | 7 / 7 |
| `tests/equivalence/super-element-access.test.ts` | 4 / 4 |
| **Total** | **31 / 31** |

Both equivalence files are fully green, not just the two named tests.

### Gates

| Gate | Command | Exit | Result |
| --- | --- | --- | --- |
| LOC budget | `node scripts/check-loc-budget.mjs` | 0 | OK — 5 changed src files, net +573 LOC; `program-abi-session.ts` 3561 → 3574 granted by the issue file |
| Function budget | `node scripts/check-func-budget.mjs` | 0 | OK — no unallowed growth in 5 changed src files |
| Coercion sites | `node scripts/check-coercion-sites.mjs` | 0 | OK — no net vocabulary growth across 4 changed codegen files |
| Oracle ratchet | `npm run -s check:oracle-ratchet` | 0 | OK — `getTypeAtLocation +0`, `ctx.checker +0` |
| Dead exports | `npm run -s check:dead-exports` | 0 | OK — 23 known entries, 0 new |
| LOC budget (CI base) | `LOC_GATE_BASE=$(git rev-parse origin/main) node scripts/check-loc-budget.mjs` | 0 | OK — same grant resolves against the merge preview |
| Function budget (CI base) | `LOC_GATE_BASE=$(git rev-parse origin/main) node scripts/check-func-budget.mjs` | 0 | OK |
| IR fallbacks | `pnpm run check:ir-fallbacks` | 0 | OK — 0 unintended, 0 post-claim, 0 module-level increases |
| Typecheck | `npm run -s typecheck` | 0 | OK — 0 diagnostics |

**9 / 9 gates green.** No gate was skipped and `--no-verify` was not used; the pre-commit and pre-push hooks ran on every commit.

### Wider equivalence sweep — 4 pre-existing failures, not from this PR

A full `tests/equivalence` run reports 4 failures (and then OOMs the vitest worker, as `CLAUDE.md` notes the full suite can). Re-running the two affected files with **this PR's production files reverted to `origin/main`** reproduces the same 4 failures identically, so they are pre-existing on `main` and untouched here:

| Pre-existing failure | File |
| --- | --- |
| `for-loop with function declaration in body` | `arguments-nested-and-loops` |
| `isNaN(void x) is true` | `logical-conditional-identity` |
| `void x in numeric context produces NaN` | `logical-conditional-identity` |
| `void numeric literal in numeric context produces NaN` | `logical-conditional-identity` |

Those two files are 65 / 69 both with and without this change.

## Notes

- The `program-abi-session.ts` allowance is placed away from the R1-A slice's edit to the same frontmatter list, so the two PRs do not conflict on it.
- Append-only: no rebase, amend, or force-push on `ee5db545`.
- Draft state untouched — promotion is the orchestrator's call.

---


ttraenkler:
CI note for the repair lane on head `2904ff70`: the `smoke` check fails in the native-messaging example (`--target wasi`):

```
nm_js2wasm_node_fs.ts:170:1 - error: Codegen error: graph-global module-init expects exactly one public __module_init export, found 0
```

That's the new exactly-one `__module_init` invariant firing on the WASI lane, where the init is wrapped into `_start`/the WASI adapter rather than published as a public `__module_init` export — so "found 0" is the expected shape there, not a defect in the example. The invariant needs to key on the invocation policy (public export vs start-function/WASI wrapping), not on the public export surface alone. No push from this side — the active repair session owns the branch; folding the WASI lane into the invariant belongs with repair item (1).

---

PR #5211, opened 2026-08-29
feat(ir): adopt the comma operator and bounded dynamic-lane `in` (#5164)

ttraenkler:
Implements [#5164](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5164-ir-comma-and-in-operators) — three IR adoption slices, each bounded by what the IR can already lower rather than by a new producer.

- **S1 — value-position pure comma.** `const c = (a, b)` is §13.16.1 evaluate-left / discard / evaluate-right, which is exactly `lowerDiscardedExpression(left)` then `lowerExpr(right)`. The selector recurses the left operand through #4459's discard predicate, so its `discard-mutating-operand` line is inherited verbatim and `(a = 1, b)` stays legacy-owned. Nested commas need no extra arm.
- **S2 — for-incrementor comma.** `for (…; …; i++, j--)`: each side re-enters `isPhase1ForUpdateExpr` / `lowerForUpdateExpr`, so both keep the update-clause mutation bookkeeping — which is what makes the mutating `i++` idiom admissible here while the pure-discard gate still refuses it in value position.
- **S3 — bounded dynamic-lane `in`.** Only the non-fast dynamic externref receiver, probed through `__extern_has` — the same dual-mode helper legacy's own `in` calls (host import / standalone object-runtime native), so inherited properties answer identically by construction. Key lowered before receiver per §13.10.1 steps 1-4.
- **S4 stays deferred**, with its reason now written into the `InKeyword` capability row.

Also corrects the adoption table's `**` note, which was stale for a second time — #4787 landed exponentiation.

## Probe flips

Probes are `.tmp/ci-*.ts`; selector lines via `JS2WASM_IR_SHAPE_DIAG=1`, emission via a real `compile({ trackIrOutcomes: true })`. Claim alone is not evidence — a flip is "FALLBACK line gone" **and** `kind: emitted` + `irBodyEmitted: true`.

| probe | before | after |
| --- | --- | --- |
| `const c = (a, b)` | REJECT `expr-binary-op-,` | emitted |
| `(a, b, c)` + nested `(a, b)` | REJECT `expr-binary-op-,` | emitted |
| `for (…; …; i++, j++)` | REJECT `expr-binary-op-,` | emitted |
| `(a = 1, b)` | REJECT `expr-binary-op-,` | REJECT `expr-binary-op-,` — unchanged, purity line held |
| `"x" in o` / `k in o` / `i in o`, all with an `any` receiver | REJECT `param-type-not-resolvable` | emitted |
| `"a" in {a:1}`, and `in` on a `number` receiver | REJECT `expr-binary-op-in` | REJECT `expr-binary-op-in` — unchanged |
| `"zzz" in new Proxy(…)` | REJECT | REJECT `constructor-resolution-unsupported` |
| `in` under `target: standalone` / `wasi` | REJECT | REJECT `body-shape-rejected` |

Zero post-claim demotions in every row.

## Three things measurement forced past the plan

1. **Flipping a capability row off `defer` also opens `isPhase1BinaryOp`** (`binaryOpCapability(op) !== "defer"`), so the generic binary tail began accepting *every* `in`, including typed-object and primitive receivers, which then post-claim-demoted. Both new arms are therefore terminal: they answer accept-or-`expr-binary-op-<op>` themselves and never fall through.
2. **S3's lane was unreachable one layer above the operator.** The pre-claim dynamic-value flow scan had no `in` arm, so every `any`-receiver function rejected at `param-type-not-resolvable` before the `in` arm ran. The added three-line arm changes no representation — the receiver reaches `__extern_has` as the dynamic carrier it already is, exactly as the #2952 for-in arm does.
3. **`__extern_has` was registered only as a side effect of legacy compiling the same function.** IR-first skips that body, making it an `unknown-function-ref` **hard error**, not a demote — the #3143 `__extern_is_undefined` failure mode. `preregisterInOperatorSupport` mirrors that fix; it cannot share with `preregisterForInSupport`, whose host-mode liveness helper is the different `__for_in_has` import.

S3 also needs the host-string carrier certificate: with native strings a key is `(ref $AnyString)`, which no externref host-arg position accepts, so without it the whole standalone lane claimed and then demoted (2 post-claim demotions → 0).

## Validation

- `tests/issue-5164-comma-and-in.test.ts` — **17/17**. Every runtime case asserts legacy ≡ IR ≡ the JavaScript reference for the value **and** for the ordered side-effect log. `in` covers 10 receiver/key shapes: string-literal hit/miss, present-but-undefined, inherited (prototype chain), variable key, `"length"` on an array, and a numeric key on an array receiver.
- `pnpm run check:ir-fallbacks` — OK, no unintended / post-claim / module-level increases.
- Ratchet chain run bare: `check-loc-budget`, `check-func-budget`, `check-coercion-sites`, `check:oracle-ratchet`, `check:dead-exports` — all exit 0, also under `LOC_GATE_BASE=origin/main`.
- `node scripts/gen-ir-adoption.mjs --check` — up to date. `pnpm run typecheck` and `pnpm run lint` — exit 0.
- LOC delta +258 net across 4 `src/` files; growth and the four function-budget entries are granted in this PR's own issue frontmatter with a dated rationale.
- **Pre-existing, not caused by this change** (identical failures on the `origin/main` base, verified by revert-and-measure): 3 failing tests plus 2 collection errors across `issue-2949-s5-2-eq`, `issue-2949-s5-3-relational`, `issue-3053-u2-claim-flip`, `issue-3529-dataflow-outcomes`.

## Follow-ups left open

- The mutating value-position comma (`(a = 1, b)`) is still legacy — the test262 `(NUMBER = Number, "MAX_VALUE") in NUMBER` idiom needs both that *and* S3's comma-key fold, so neither half alone unlocks it.
- S3 consumes two certificates named for for-in (`isDynamicForInReceiver`, `forInHeadValueIsHostString`) because both ask a *carrier* question, not a statement-shape one. Renaming them is a pure-rename cleanup, deliberately not taken here to keep this PR off #2949's surface.


ttraenkler:
This park is collateral, not this PR's failure. Both failing steps are the two known main-side issues diagnosed on the earlier #5199 park (byte-level A/B evidence in [#5181](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5181-standalone-stack-balance-error-shift)):

1. **"Build merged standalone test262 report"** fails on 24 unclassified `stack-balance (#1058)` compile errors introduced to main by the selfhost commit `8f161cbf15` (PR #5204) — nothing in this PR's comma/`in` diff touches stack-balance or those files. The classifier fix is **PR #5220**, currently in the merge queue; this group ran on a merged state that predates it.
2. **"Fail on regressions"** reports the same ~937-regression drift cluster (bucket signature `f524a794246996c6`) the #5199 group reported — the baselines JSONL is behind main, and the six named "baseline-pass" files fail identically on main *without* any queued PR's diff.

Plan: leave the `hold` in place until #5220 lands (its classifier travels with each subsequent merged state), then re-enqueue this PR once. The gate-bypass that let earlier red groups merge anyway is tracked in [#5182](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5182-merge-group-gate-bypass).

---

PR #5214, opened 2026-08-29
feat(ir): adopt mutating expression statements with property/element LHS (#5163)

ttraenkler:
Closes [#5163](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5163-ir-mutating-expression-statements).

[#4459](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4459-ir-discarded-expression-statements) adopted the value-**discarding** expression statements. The **mutating** shapes had no dedicated arm, so a single `o.x += 1;` rejected the whole containing function. This lands all four measured sub-slices.

## What changed

| Slice | Shape | Was rejecting as |
| --- | --- | --- |
| S0 | top-level local `a = 5;` and `x++;` | the top-level walker had **no arm at all** |
| S1 | property compound `o.x += e;` and `o.x++;` | `nontail-compound-or-binary-stmt` / `nontail-incdec-stmt` |
| S2 | element compound `a[i] += e;` and `a[i]++;` | same, plus `body-exprstmt-other` in a loop body |
| S3 | chained assign `a = b = e;`, `o.x = o.y = e;` | `nontail-assign-nonprop-lhs` / `expr-binary-op-=` |

Each ships its selector arm and its lowering arm **atomically** in both statement walkers and both dispatchers.

These are **not** a widening of the plain-`=` arms. A compound assignment is a read-modify-write, while `=` lowers receiver → RHS with *no read at all*. So `lowerPropertyCompound` / `lowerElementCompound` factor the **write tail** out of the `=` lowerers and put the read where the spec puts it:

```
receiver  →  index  →  READ  →  RHS  →  binop  →  WRITE
```

with the receiver and index each lowered **exactly once** and reused for both the read and the write. That is what makes `o.x += f(o)` (where `f` mutates `o.x`) write `oldRead + rhs`, and `a[g()] += h()` call `g` once. A chained assignment lowers its RHS once and stores *that value* into every target, never re-reading the inner one. Bounds behaviour is inherited unchanged from the paths S2 composes, so an out-of-bounds `a[i] += 1` still grows and stores NaN exactly as legacy does.

## Probe flip

Measured with `JS2WASM_IR_SHAPE_DIAG=1` on the issue's probe set (`.tmp/stmtmut-*.ts`).

**Top level — `.tmp/stmtmut-probe.ts` (7 functions)**

| function | shape | before | after |
| --- | --- | --- | --- |
| `pE1` | local `a = 5;` | FALLBACK `nontail-assign-nonprop-lhs` | **CLAIM + emitted** |
| `pA8` | local `a++;` | FALLBACK `nontail-incdec-stmt` | **CLAIM + emitted** |
| `pS1a` | `o.x += 1;` | FALLBACK `nontail-compound-or-binary-stmt` | **CLAIM + emitted** |
| `pS1b` | `o.x++;` | FALLBACK `nontail-incdec-stmt` | **CLAIM + emitted** |
| `pS2a` | `a[0] += 1;` | FALLBACK `nontail-compound-or-binary-stmt` | **CLAIM + emitted** |
| `pS2b` | `a[0]++;` | FALLBACK `nontail-incdec-stmt` | **CLAIM + emitted** |
| `pS3` | `a = b = 7;` | FALLBACK `nontail-assign-nonprop-lhs` | **CLAIM + emitted** |

claimed 0/7 → **7/7**, fallbacks 7 → **0**.

**Body buffers — `.tmp/stmtmut-body.ts` (8 functions, 3 of them controls)**

| function | shape | before | after |
| --- | --- | --- | --- |
| `bS1a` | `o.x += 1;` in a loop | FALLBACK `body-exprstmt-other` | **CLAIM + emitted** |
| `bS1b` | `o.x++;` in a loop | FALLBACK `for-body` | **CLAIM + emitted** |
| `bS2a` | `a[i] += 1;` in a loop | FALLBACK `body-exprstmt-other` | **CLAIM + emitted** |
| `bS2b` | `a[i]++;` in a loop | FALLBACK `for-body` | **CLAIM + emitted** |
| `bS3` | `a = b = i;` in a loop | FALLBACK `expr-binary-op-=` | **CLAIM + emitted** |
| `ctlLocalCompound` | control: local `a += n;` | CLAIM | CLAIM (unchanged) |
| `ctlPropAssign` | control: `o.x = 5;` | CLAIM | CLAIM (unchanged) |
| `ctlElemAssign` | control: `a[0] = 5;` | CLAIM | CLAIM (unchanged) |

claimed 3/8 → **8/8**, fallbacks 5 → **0**. All three controls keep claiming.

Every flipped function is emission-backed, not merely claimed: `compile(src, {trackIrOutcomes: true})` reports `kind: emitted`, `irBodyEmitted: true` for **12/12**.

## Two findings worth flagging

**The plan's one UNVERIFIED soundness assumption is now verified.** It asked whether any IR pass could reorder the old-value read across the RHS's side effects. It cannot: `effectsConflict` (`src/ir/effects.ts`) returns true for `a.readsHeap && b.writesHeap`, and `object.get`/`class.get` are `readsHeap` while `object.set`/`class.set` and every call are `writesHeap`. The `o.x += f(o)` equivalence test is the empirical backstop.

**A build-stage demote is not always a fallback — it depends on the compile route.** Two separate bugs came from this, both caught by re-running the probes rather than by reasoning:

1. `planI32Slots` promoted `b` in `a = b = 7` (its write is canon-i32-lowerable in isolation), so the chain met a mixed f64/i32 target set and demoted. Because the selector had already claimed, that demote surfaced as a **`kind: invariant` hard compile error** (`unpatched-slot`), not a fallback. Fixed at the source: `writeShapesAreLowerable` no longer promotes a slot whose write is a link in a `=`-chain, per #3741 invariant W.
2. After merging `origin/main`, the *string* chain `a = b = "z"` hit the same wall — a **string-route** function has its legacy body skipped (`multi-prepared-body-skips.ts`), leaving a demote with nothing to fall back to. Fixed by lowering the shape instead of refusing it: the chain lowering required f64 destinations only to sidestep per-target coercion, but every write emit it uses is type-agnostic, so the restriction bought nothing.

The general rule is pinned as a test over nine sources, each **compiled alone** — a multi-function module often keeps the legacy body and hides exactly this failure, which is why the first hazard sweep missed it.

## Validation

| Check | Result |
| --- | --- |
| `tests/issue-5163-mutating-statements.test.ts` (new) | **44/44** pass |
| `tests/issue-4459.test.ts` | **37/37** pass |
| Equivalence gate, 8/8 shards | **1,718 passing**, 24 known-failures in baseline, **0 new regressions** |
| `pnpm run check:ir-fallbacks` | OK — unintended (none), post-claim (none), module-level (none) |
| `node scripts/check-loc-budget.mjs` | OK — +471 `from-ast.ts`, +242 `select.ts`, both granted by the issue file |
| `LOC_GATE_BASE=origin/main` LOC + func budget | OK — no unallowed growth |
| `node scripts/check-func-budget.mjs` | OK — **no allowance grant needed** |
| `node scripts/check-coercion-sites.mjs` | OK — no net vocabulary growth |
| `npm run check:oracle-ratchet` | OK — `getTypeAtLocation` +0, `ctx.checker` +0 |
| `npm run check:dead-exports` | OK — 23 known entries, 0 new |
| `pnpm run typecheck` | 0 errors |
| `biome lint src tests scripts --diagnostic-level=error` | 0 errors |

The new test file's 44 cases are ordering-first, since that is where a read-modify-write desugaring goes wrong silently: `o.x += f(o)` must write `oldRead + rhs`; `a[g()] += h()` must call each of `g`/`h` once, in that order; `o.x++;` and `++o.x;` must be indistinguishable in statement position; `a = b = e()` must evaluate `e` once and give both targets its value; out-of-bounds `a[i] += 1` must match legacy. Side-effect counters are carried on **local objects passed as arguments**, never module-level `let`s — a module-binding counter is itself outside IR scope, which would demote the function under test and make the equivalence assertion vacuous (comparing legacy against legacy).

Ordering cases are checked against **Node** as well as legacy, so a wrong order cannot pass by matching a wrong reference.

## Deliberately out of scope

These demote **typed** (asserted as `unsupported`, never `invariant`): non-f64 fields and elements, accessor-backed properties, extern/DOM receivers, TypedArray views, #3734 narrowed-i32 vecs, the inferred linear-vector lane, nested-receiver property compounds, element targets inside a chain, and module-binding chain targets. Value-position twins (`const v = (o.x += 1)`) remain a separate follow-up, as the plan states.

## One edit to an existing test

`tests/issue-4459.test.ts` pinned five shapes as "still rejects". Four of them — `o.x += 1;`, `a[i] += 1;`, `a = b = 1;`, `o.x++;` — are exactly the residual this issue closes. They are **inverted** there (now asserting claim + emission) rather than deleted, so the boundary that file guards stays checked, just on the other side. `new.target;` still pins as rejecting. The file is fully green (37/37).

## Structure

Both walkers share one gate (`phase1MutatingStatementVerdict` in `select.ts`) and both dispatchers share one lowering entry (`lowerAdoptedMutatingStatement` in `from-ast.ts`), so the top-level and body-buffer positions cannot drift. That dedup is also what kept both budget gates green with no allowance grant: `isPhase1StatementListInScope` went from +90 over its ceiling to 0, and `lowerStatementList` back under 300 LOC.

---


ttraenkler:
Park attribution, from the cited run's own logs ([job 99084983857](https://github.com/loopdive/js2/actions/runs/33245968708/job/99084983843)): this is **collateral from main's #5204 regression, not this PR's diff**.

- The regression list is the [#5204](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5204-selfhost-ts5-parser-graph) damage signature measured against a **pre-#5204 baseline** — the log's own drift warning says the baselines JSONL is 13 test262-relevant commits (411 min) behind main. The named rows are RegExp `match-indices` (filed as [#5187](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5187-regexp-exec-index-broken)), destructuring `dstr/array-elem-put-*`, and a 62-test `String/prototype/split` bucket — none of them touched by this PR's mutating-statement adoption, which carried a 0-regression equivalence gate (1,718 passing, 8/8 shards) at PR level.
- The `Standalone pass-count high-water floor (#2097)` step fails on main's own −2,273 breach ([#5181](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5181-standalone-stack-balance-error-shift)), which every merge group inherits.

Not removing `hold` yet: re-enqueueing before main recovers would fail the identical way and spend the one re-admission. Unblock order: [#5186](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5186-stack-balance-hard-error-family)'s fix-forward PR #5228 (stack-balance family, ~2,580 compile_errors) plus the corpus-B fixes (#5237 for the RegExp half) need to land; once the floor and baseline recover on main, the `hold` comes off and auto-enqueue re-takes this PR.

---

PR #5232, opened 2026-08-29
revert(selfhost): back out the TS5 parser graph compile — −2,273 standalone passes (#5181) ✓

ttraenkler:
## This is the PREPARED REVERT OPTION — for the owner to choose against fix-forward

This PR is prepared, validated and ready, but it is **one of two options**. It is
not a claim that reverting is the right call. The alternative is to fix
`8f161cbf15` forward on its own branch. Whoever owns the call should pick; this
branch exists so that "revert" is a decision rather than a project.

## What and why

Reverts `8f161cbf15` — *"feat(selfhost): compile TypeScript 5 parser graph to
Wasm"* — which landed via PR #5204.

The commit regressed standalone test262 conformance at scale. Measured in
[#5181](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5181-standalone-stack-balance-error-shift)
against the commit's own parent (`70e8e3c1ca`), two independent samples:

| corpus | sample | result |
| --- | --- | --- |
| A — rows carrying the `stack-balance (#1058)` shared-body refusal | 25 | **21 went `pass` → `compile_error`** (21/22 of the locally measurable) |
| B — pass-losses carrying **no** stack-balance signature | 25 | **17 went `pass` → `fail`** — silent wrong answers |

Corpus B is the worse half: destructuring binds the wrong value and
`RegExp.exec` reports the wrong index. Those are silent miscompiles, a
correctness class strictly worse than the loud stack-balance refusal.

Re-weighted by real population size, roughly **2,030–2,115 of the 2,575 gross
standalone pass losses** trace to this one commit, against a floor delta of
**−2,273**:

```
[standalone-highwater] current pass=31371, mark=33644 (floor=33594, tolerance=50, delta=-2273)
```

The metric is `host_free_pass`, so the [#2879](https://js2wasm.loopdive.com/dashboard/issue.html?slug=2879-standalone-metric-measure-host-free-ness) §4 carrier-migration exemption does
not apply. The breach does not predate the commit (the high-water mark was set
by a passing run before it landed, and the promoted baseline carries **zero**
rows with the signature).

## Restoration — measured on this branch

**Both columns below were measured on this branch**, not inherited from an
earlier artifact: each test was run through the standalone single-file runner
(`runTest262File(..., "standalone")`) twice — once with `main`'s `src/`+`tests/`
checked into the worktree (the "on `main`" column) and once on the reverted tree.
Positive controls run first in both arms, so a uniform result is a property of
the tree and not of the harness.

| # | test262 file | on `main` (`ddab1b0743`) | this branch |
| --- | --- | --- | --- |
| control | `language/statements/if/if-stmt-else-async-gen.js` | pass | **pass** |
| control | `language/expressions/class/dstr/async-gen-meth-dflt-obj-ptrn-prop-eval-err.js` | fail | **fail** |
| A10 | `built-ins/TypedArrayConstructors/Float32Array/proto.js` | compile_error | **pass** |
| A5 | `built-ins/Promise/create-resolving-functions-resolve.js` | compile_error | **pass** |
| A24 | `harness/deepEqual-array.js` | compile_error | **pass** |
| A4 | `built-ins/Function/prototype/bind/15.3.4.5.2-4-5.js` | compile_error | **pass** |
| A7 | `built-ins/TypedArray/from/mapfn-is-not-callable.js` | compile_error | **pass** |
| A2 | `built-ins/Atomics/sub/non-shared-int-views-throws.js` | compile_error | **pass** |
| B5 | `built-ins/RegExp/prototype/exec/S15.10.6.2_A4_T10.js` | fail | **pass** |
| B7 | `language/expressions/arrow-function/dstr/obj-ptrn-id-trailing-comma.js` | fail | **pass** |

**8 of 8 targeted regressions return to PASS**, both controls behave, and the
two error modes (loud `compile_error` and silent `fail`) both clear.

All six corpus-A failures on `main` carry the identical diagnostic, e.g.
`stack-balance (#1058): function "__closure_69" reaches an instruction array
from incompatible control-flow or function-local contexts` (`cacheComparison`
for `deepEqual-array`). The two corpus-B failures are wrong answers, not
refusals: `__executed.index` `Expected SameValue(«2», «null»)` and
`Expected SameValue(«null», «23»)`.

## What the revert removes

The TypeScript 5.9.3 parser-graph selfhost work: 55 added files — **7 codegen
modules**, **44 `issue-1058-*` test files**, the dogfood TypeScript parser
harness (2 files), `tests/issue-3946-compile-profile.test.ts` and
`tests/issue-4757.test.ts` — plus the codegen changes across 78 existing files.
133 files total, ~14.9k lines removed against ~1.8k restored (net **−6,206 LOC**
by the budget gate).

`tests/issue-1058-consumer-driven-barrels.test.ts` is deliberately **kept** — it
predates the reverted commit (which modified it, not added it) and still passes.

## What the revert preserves

The reverted files were also touched by later, independent commits. Each of
these was kept, hunk by hunk, rather than rolled back with the feature:

- **[#5096](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5096-user-class-shadowing-global-not-constructable)** — `builtinSymName` / builtin-shadow guard in `resolveWasmType`
- **[#4406](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4406-return-type-unboxing-abi) Phase 3** — `noteParamUnboxStats` in both codegen lanes
- **[#5093](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5093-formalful-method-spread-argument-misalignment)** — the `arguments`-based spread path (`if (!viaArguments)`)
- **[#5154](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5154-es2015-standalone-lang-semantics-wave1)** — `maybeRecordArrayProtoIteratorTombstone` in `compileDeleteExpression`
- **[#4646](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4646-structmap-class-name-keying-shared-body)** — per-declaration class identity (`compiledClassBodies`)
- **redux** — the heterogeneous-callable capture guard in `arrow-phases.ts`
- **`68cd60e8bb`** — `finalizeMultiPreparedModuleInitStartup` ownership

## Re-land condition

Re-land the selfhost feature once the **stack-balance shared-body interaction**
is fixed. In plain terms: the selfhost graph makes two callers that need
*different* stack repairs point at the **same** `Instr[]` array, so the repair
pass refuses to mutate it rather than corrupt one of them —
`src/codegen/stack-balance.ts:2804`, *"emit distinct instruction arrays at the
producer."* That producer must be found and made to give each owner its own
array.

The corpus-B silent miscompiles (destructuring, `RegExp.exec`) must be
root-caused as well — they carry no stack-balance signature, so fixing the
shared-body refusal alone would not clear them.

## Dependency check

No later commit builds on the reverted change. 39 later non-merge commits touch
the same source files, but **none** references any of the 7 modules or 44
exported symbols the commit introduced — verified by scanning every later
commit's added lines for those identifiers. The conflicts were textual overlap
only.

## Validation

- Standalone restoration: 8/8 above, with both controls
- `tests/issue-5092-*`, `tests/issue-5164-*`: **38/38 pass** (no `issue-5163-*`
  test exists on `main` yet — that work is still in flight)
- `tests/issue-1058-consumer-driven-barrels.test.ts`: 8/8 pass
- `typecheck:ts5`: clean · `lint` (biome): clean · prettier: clean
- Ratchet chain bare: LOC ✅ · func ✅ (with a documented allowance, below) ·
  coercion-sites ✅ · oracle-ratchet ✅ · dead-exports ✅
- CI-base simulation `LOC_GATE_BASE=ddab1b0743`: LOC ✅ · func ✅

### One note on the function-budget allowance

Three functions trip the func-budget gate — `updateTypeStack`,
`emitClosureCallExportN`, `emitClosureMethodCallExportN`, each **+29 lines**.
This is the revert restoring their **pre-selfhost bodies**: the reverted commit
had refactored each ~29 lines *shorter*, and the committed baseline was reseeded
on `main` after it landed, so restoration reads as growth against a ceiling that
only exists because of the commit being reverted. `src/codegen/stack-balance.ts`
is line-for-line identical to its `70e8e3c1ca` content. The allowance is granted
in `plan/issues/1058-compile-the-typescript-compiler-itself.md` with that dated
rationale, and retires with the re-land. No baseline JSON was touched.

---


ttraenkler:
Parked as **draft** — the fix-forward option now exists and is the better one: PR #5228 root-causes the stack-balance shared-body refusal in the pass itself (stack-polymorphic bodies skip the context-key comparison), keeps the selfhost work, and preserves both fail-closed guards. This branch also went DIRTY when #5216/#5220 landed, so it would need a fresh 133-file conflict resolution to be usable.

Keeping this PR open as the fallback until #5228's merge group confirms the standalone floor recovers. Note for whoever closes the loop: #5228 does **not** claim the corpus-B silent miscompiles (destructuring, `RegExp.exec().index` — filed as #5187); if the floor stays breached beyond tolerance after #5228 lands, the residue is corpus-B and needs its own root-cause, not this revert.

---

ttraenkler:
Closing — the fix-forward path won: PR #5228 (stack-balance shared-body fix) is on main at `2ba09662a9`, restoring the ~2,580-row refusal family while keeping the selfhost work. Reverting `8f161cbf15` now would rip out both and need a third conflict resolution.

For the record, #5228's own merge group still failed its regression gate (run 33247654446) — the residue is the corpus-B silent damage this revert also measured: `RegExp.exec().index` ([#5187](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5187-regexp-exec-index-broken-by-5204), fix open as PR #5237) and the destructuring family ([#5185](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5185-heterogeneous-destructured-callback-param-null)). Until those land, code-PR merge groups may keep tripping the standalone floor against the pre-#5204 baseline. The restoration table in this PR's description remains the measured reference for what full recovery looks like (8/8 including both corpus-B samples).

---

PR #5260, opened 2026-08-29
ci(test262): add skip_promote dispatch input — measure without promoting

ttraenkler:
## Description

Adds a `skip_promote` boolean to `test262-sharded.yml`'s `workflow_dispatch` inputs, and one clause to `promote-baseline`'s `if:`. Six added lines, no source touched.

### The gap

`promote-baseline`'s `if:` admits `push || workflow_dispatch`, and its "Push baseline artifacts to js2wasm-baselines repo" step carries no guard of its own. So **any** manual dispatch, on any ref, overwrites `loopdive/js2wasm-baselines` — the regression-gate baseline every open PR is diffed against — with that ref's numbers, re-stamps the #2097 high-water mark, and can push a `[skip ci]` commit to `main` (rebuilding every queued merge group, the cost #1951 exists to avoid) plus redeploy Pages.

The failure mode is inverted, which is what makes it a trap: promote only runs when `merge-report` succeeds, so the *clean* measurement — the one that finds nothing wrong — is exactly the one that triggers the overwrite. It cannot be relied on to self-block.

The three existing promote suppressors each alter the measurement:

| input | suppresses promote | measurement comparable to the published baseline |
|---|---|---|
| `ir_first` | yes | no — IR-first compile path |
| `layout_emit` | yes | no — per-type layout emission |
| `eval_engine: interpreter` | yes | no — different standalone eval lane |
| `skip_promote` (this PR) | yes | **yes** — identical compile, run and engine |

So before this change there was no way to dispatch a measurement at a non-tip ref without side effects — which is precisely what a bisect needs.

### Scope check

Both write paths — the baselines-repo SSH push (~L2735) and the deploykey push to `main` (~L3063–3332) — live wholly inside `promote-baseline` (L2453–3454), so gating that one job is sufficient. `write-run-cache-bot`, the only other job with side effects, is `push` + `github-merge-queue[bot]` only and never fires on dispatch.

### Motivation

`merge shard reports` has failed on every push-to-main run since 2026-08-29T03:05Z on the #2097 standalone host-free floor (−2,505 vs the mark). That skips `promote merged report to main baseline`, so the baselines repo, the in-repo mirrors and the deployed landing-page pass rate have all been frozen at the 02:28 snapshot. Bisecting the drop needs measurements at two historical refs, which this input makes safe to take.

Not a fix for the regression itself — this only unblocks measuring it.

## Verification

- YAML parses; `skip_promote` present in the dispatch inputs and as a clause in the `promote-baseline` `if:`
- All five source-ratchet gates green — `check-loc-budget`, `check-func-budget`, `check-coercion-sites`, `check:oracle-ratchet`, `check:dead-exports` (each run bare, not piped, statuses captured individually)
- Diff touches no `src/`, so the gates are unchanged by construction

## CLA

---
_Generated by [Claude Code](https://claude.ai/code/session_01RJew5ksYvThAbG7qFJLXT4)_


ttraenkler:
`quickjs eval-engine lane (non-required)` is red on `40f2f4fd` ([run 33266753991](https://github.com/loopdive/js2/actions/runs/33266753991/job/99138121493)). It is **not** this PR's, but it **does** block enqueue, and I lack the permission to re-run it — so flagging rather than fixing.

**What failed.** A vitest worker OOMed, then the pool collapsed:

```
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
[2715] 30145 ms: Mark-Compact 547.6 (559.2) -> 513.8 (531.2) MB ... allocation failure
⎯⎯⎯⎯ Unhandled Rejection ⎯⎯⎯⎯⎯   Error: Channel closed
Serialized Error: { code: 'ERR_IPC_CHANNEL_CLOSED' }
```

`tests/issue-4242-eval-engine-parity.test.ts` had just passed 40/40. Nothing asserted false — the process ran out of heap on the last of the five suites in that step.

**Why it isn't this PR's.** The diff is one `workflow_dispatch` input plus one `if:` clause in `test262-sharded.yml`, and one prose section in `docs/ci-policy.md`. No source, no test, no dependency, no lockfile — nothing that can move heap usage in this lane. This is also the only failure in the last 25 runs of `quickjs-wasi-artifact.yml`, so it is not red on the base branch either.

**Proposed patch** (deliberately not applied here — it is unrelated to this PR's change, and the rules say propose rather than widen). The lane's two vitest steps set no heap cap, unlike both sibling lanes — `eval-interpreter-lane.yml` uses `NODE_OPTIONS: --max-old-space-size=3072` and `test262-sharded.yml` uses `VITEST_FORK_MAX_OLD_SPACE_SIZE`. It died at ~500 MB, i.e. the runner default. In `.github/workflows/quickjs-wasi-artifact.yml`, on the "Run all QuickJS semantic suites and parity-gate tests" step:

```yaml
         env:
           JS2WASM_EVAL_ENGINE: quickjs
           JS2WASM_QUICKJS_ARTIFACT_DIR: .tmp/quickjs-artifact
+          NODE_OPTIONS: --max-old-space-size=3072
```

Worth its own issue: five suites in one vitest process with no cap will keep drifting into this as they grow, and the failure presents as a channel-closed error rather than as memory pressure.

**One correction for the record.** `.claude/skills/steward/SKILL.md` §2 states that a red non-required check does not block enqueue. `scripts/enqueue-green-prs.mjs` says otherwise — `classifyChecks` carries an explicit *"Zero-FAILURE rule: applies to EVERY check, required or not (#3878/#3904)"*, and any non-passing, non-pending state lands in `failures` before the required-name filter is reached. So this red lane does gate `auto-enqueue`, and the skill's guidance is stale on that point.

A re-run is the cheapest resolution if someone with Actions write access can trigger one; `POST /actions/runs/33266753991/rerun-failed-jobs` returns 403 for me.

---

PR #5263, opened 2026-08-29
ci(quickjs): cap the eval-engine lane's heap — it OOMs deterministically

ttraenkler:
## Description

One line (plus a comment) on `quickjs-wasi-artifact.yml`'s "Run all QuickJS semantic suites and parity-gate tests" step:

```yaml
+          NODE_OPTIONS: --max-old-space-size=3072
```

That step runs five suites in one vitest process and set **no heap cap at all**, so it inherited the runner default (~500 MB) and now dies on it:

```
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
... Mark-Compact 544.9 (560.5) -> 513.9 (530.5) MB ... allocation failure
Unhandled Rejection: Error: Channel closed
Serialized Error: { code: 'ERR_IPC_CHANNEL_CLOSED' }
```

**Not a flake.** It reproduced identically on two different SHAs of #5260 (`40f2f4fd`, `ec7490ff`), both times *after* `tests/issue-4242-eval-engine-parity.test.ts` passed 40/40. Nothing asserts false — the process runs out of heap on the last of the five suites, and the pool collapse surfaces as a channel-closed error. That presentation is what makes it read as infrastructure noise rather than memory pressure.

`3072` matches what `eval-interpreter-lane.yml` already sets; `test262-sharded.yml` caps the same way via `VITEST_FORK_MAX_OLD_SPACE_SIZE`. This lane was the outlier in having no cap.

## Why this is worth its own PR

The lane is non-required, but `classifyChecks()` in `scripts/enqueue-green-prs.mjs` applies its zero-FAILURE rule to **every** check *before* the required-name filter (#3878/#3904) — so a red non-required lane blocks `auto-enqueue` outright, and nothing retries such a lane on its own.

Observed on #5260: all six required checks green or skipped, `mergeable: true`, no `hold` label, and every `auto-enqueue` sweep logging

```
- #5260 skip (failing-checks: quickjs eval-engine lane (non-required): fail)
```

So this strands any PR touching paths that trigger this workflow, and #5260 is blocked on it today. Split out rather than folded into #5260 because it is unrelated to that PR's change (a `skip_promote` dispatch input).

## Verification

- YAML re-parsed after the edit and the step's `env` asserted to carry `NODE_OPTIONS` — not eyeballed
- All five source ratchets green (`check-loc-budget`, `check-func-budget`, `check-coercion-sites`, `check:oracle-ratchet`, `check:dead-exports`), each run bare with its own status captured
- Diff touches no `src/`, so the ratchets are unchanged by construction
- The real proof is this PR's own `quickjs eval-engine lane` run: it exercises the patched step, so a green result here *is* the reproduction-then-fix

## Follow-up worth considering (not in this PR)

`3072` raises the ceiling; it does not address five suites sharing one process as the corpus grows. Splitting that step, or setting the cap workflow-wide, would be the durable fix.

## CLA

---
_Generated by [Claude Code](https://claude.ai/code/session_01RJew5ksYvThAbG7qFJLXT4)_


ttraenkler:
**The cap works. It uncovered a real membrane regression that the OOM had been hiding.**

On `1bd14d42` the lane no longer OOMs — no `Reached heap limit`, no `ERR_IPC_CHANNEL_CLOSED`, and the step ran to completion for the first time: **1 failed | 138 passed (139)**, 5 test files instead of 1.

That is the fix doing exactly its job, and it is also why the lane is still red.

### What the OOM was masking

Every OOM run completed only `tests/issue-4242-eval-engine-parity.test.ts` (40 tests) before dying. `tests/quickjs-eval-membrane.test.ts` — 57 tests — is the memory hog that was being killed, so **none of its assertions had run in any recent CI run**. With the worker cap raised it runs, and one assertion fails:

```
FAIL tests/quickjs-eval-membrane.test.ts > #4245 slice 1 — quickjs eval membrane (inward)
     > descriptor VALUES cross faithfully, including a false flag
AssertionError: expected +0 to be 1 // Object.is equality
  ❯ tests/quickjs-eval-membrane.test.ts:1125:37
```

Line 1125 is `expect(outward.enumValProbe!()).toBe(1)` — a descriptor's `enumerable: true` reads back as `0`/false after crossing the membrane outward. The test's own comment calls this the anti-vacuity check: it exists to catch precisely a synthesized descriptor rather than a faithfully-crossed one, so a `0` here is the failure mode it was written to detect, not an incidental mismatch.

This is unrelated to a heap cap, so per the repo's rules I am not fixing it in this PR.

### Why it stayed invisible

`quickjs-wasi-artifact.yml` triggers on **`pull_request` only**, and only for a narrow path list — it never runs on `main` and has no cron. So a regression in the membrane is undetectable until some PR happens to touch one of those paths. The last green run was 2026-08-28 13:41; every run since has been an OOM, which masked the assertion entirely. The window between those two states is where this landed.

That trigger design is arguably the more important finding than either bug: a lane that only runs on unrelated PRs, gates enqueue via the zero-FAILURE rule, and never runs on the branch it is protecting will keep accumulating invisible breakage and then block whoever next touches those paths.

### Status

- The heap cap is correct and verified — worth landing on its own merit.
- The lane stays red on the membrane assertion, so the zero-FAILURE rule keeps blocking both this PR and #5260.
- I cannot reproduce locally: `scripts/build-quickjs-eval-provider.mjs --require-cache` fails with no artifact cache present, and CI restores that cache. Diagnosing the membrane bug from here would be CI-only iteration.

Flagging for a decision rather than guessing at a membrane fix.

---

PR #5264, opened 2026-08-29
fix(#5204): bridge f64, rest and getter members of builtin-derived classes

ttraenkler:
Closes [#5204](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5204-host-class-bridge-f64-params).

**Stacked PR — land order: #5252 → #5258 → #5262 → this.** Branched from `issue-5203-init-static-dispatch` (PR #5262). Do not merge before all three predecessors.

## The bug

On `class D extends Array`, instance members beyond a zero-argument method never reached the host dispatch path — **at init AND after init**, so unlike #5193/#5202/#5203 this is a capability gap, not the module-init timing window (measured with an after-init-only probe: no top-level call at all).

```
add(x: number, y: number)  →  "add is not a function"
sum(...xs: number[])       →  "sum is not a function"
get g()                    →  NaN
tag(s: string)             →  works
```

That last row is the diagnosis in one line: a `string` formal is already externref, so **f64 was the entire gap** between "a zero-arg method bridges" and "a method with arguments does".

## Three defects behind one bridge

`class D extends Array` is **externref-backed**: its instance is a real host object, so the generic `__member_kind_<key>` `ref.test` cascade over WasmGC struct types can never match it, and only the class-qualified bridge can serve it. That is also why methods worked and property reads did not — a method call reaches `__extern_method_call`'s not-a-function tail, which consults the class resolver itself; a plain read goes through `_safeGet`, which did not.

| symptom | cause |
| --- | --- |
| `add(x,y)` | `supportsHostClassBridgeParam` rejected `f64`, suppressing the ENTIRE per-key surface — not even `__member_kind_inc` was emitted |
| `sum(...xs)` | only the STRUCT-path `__class_call_sum_vararg` existed; no class-qualified vararg bridge |
| `get g()` | (a) accessor bodies compiled with a WasmGC-struct receiver while METHODS of the same class compiled with externref; (b) `_safeGet` consulted `_resolveClassMember` only inside its `_isWasmStruct(obj)` block |

**The getter needed a codegen fix, not just a bridge.** Measured directly: `D_get_g` had `params: [{kind:"ref",typeIdx:28}]` while `D_z` had `params: [{kind:"externref"}]` — same class, same declaration. So no getter bridge was even *emittable*. `compileClassBodiesInner` hardcoded the struct receiver for accessor bodies while methods used the #4534 externref rule; fixed for getters and setters from the same rule.

## What changed

- `supportsHostClassBridgeParam` admits `f64`, and `emitExternrefClassMethodDispatch` unboxes each argument through `callArgCoercionInstrs` — the same coercion the struct-path bridge had all along and simply never reached. The RECEIVER check is now spelled out separately so widening the argument allowlist cannot silently widen the receiver one.
- `classBridgeNeedsNumberBox` scans parameters too — `addUnionImports` must run before the bridge loop captures any `funcIdx`.
- New `__class_call_<Class>_<key>_vararg`, locating the rest slot by TYPE from the end of the signature (#4644's lesson: `restInfo.restIndex` is a SOURCE index with no receiver slot). Resolver learned the `_vararg` suffix.
- New `__call_get_<Class>_<key>`, checked first in the resolver's class-qualified branch.
- `_safeGet` consults `_resolveClassMember` on its non-struct tail, **after** the native read and the sidecar so nothing that already resolved changes precedence, reading the #5193/#5202/#5203 start-export channel so it also works during init.
- Emission gate widened to `classExternrefBackedSet.size > 0`: a module whose only host-side access is a getter READ registers no dynamic method name, so the emitter used to bail entirely. The extra keys feed ONLY the class-qualified emitters.

## Allowlist rejections that REMAIN (recorded, not widened)

`i32`, `i64`, `f32` parameters (the native-annotation lane) and `ref`/`ref_null` struct/vec parameters. Each needs its own representation contract at the host boundary; none is on the Temporal path.

## Validation

- `tests/issue-5204-host-class-bridge-f64-params.test.ts` — 9 cases, each measured at init AND after init. **6 fail on base**; the 3 that pass on base are controls and still pass.
- `tests/issue-5191 / 5193 / 5202 / 5203` — 53/53 unchanged.
- `classes`, `class-methods`, `class-method-calls`, `class-expression`, `class-expressions`, `getters-setters`, `accessor-side-effects`, `abstract-classes` — 68/68.
- `issue-2992-accessor-merge` / `-widening` — 6 failures, **identical on base and here** (A/B revert). Pre-existing.
- **Equivalence gate, all 8 shards, exit 0** — 0 new regressions.
- LOC / function / coercion-site / oracle-ratchet / dead-export gates clean. Growth allowances with rationale in the issue file; the coercion-site +1 is a single `funcMap.get("__unbox_number")` passed into the coercion engine, not a hand-rolled ToNumber.

## Temporal harness (#4628 critical path)

All three probe rows executed 2026-08-29 (#5256 merged into a local probe tree for measurement only):

| probe tree | `moduleInitError` |
| --- | --- |
| #5252 + #5256 + #5258 | `TypeError: __clz30 is not a function` |
| + #5262 (#5203) | `object is not iterable (cannot read property Symbol(Symbol.iterator))` |
| + #5262 + this | same as above |

**`moduleInitRuns` is still false.** This PR does not move the harness by itself — the next blocker fires first — but it is on the path: jsbi's instance methods with arguments are exactly the shape it unblocks.

Next blocker, located (stack `Object.fromEntries → src/runtime.ts:14436 → __module_init`): `__object_fromEntries` passes the compiled value straight to the host `Object.fromEntries`, which needs `Symbol.iterator`; an opaque WasmGC vec has none. Its neighbour `__object_assign` marshals via `_wrapForHost`. Being filed by the coordinator.

---


ttraenkler:
Closing: all of this PR's content landed on main via PR #5324's merge (its branch stacked on the full Temporal chain), leaving this PR 0-diff. The issue file (status: done) merged with it.

---

PR #5266, opened 2026-08-29
fix(runtime): marshal compiled iterables and tuple pairs for Object.fromEntries (#5205)

ttraenkler:
Fixes [#5205](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5205-object-fromentries-compiled-value-marshal) — the seventh @js-temporal/polyfill module-init blocker ([#4628](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4628-temporal-runtime-object-spike)).

**STACKED** on `issue-5204-bridge-f64-params` (sanctioned predecessor-stacking). Land order: #5252 → #5258 → #5262 → #5264 → this. `origin/main` is merged in as of 2026-08-29.

## The issue named one defect; the call site had two

`__object_fromEntries` was a one-liner handing the compiled value to the host `Object.fromEntries`, which needs `@@iterator` (§7.4.4 GetIterator) — an opaque WasmGC vec has none, so **every** compiled source threw `object is not iterable`.

Fixing that half alone is **not** enough. §7.1.19 AddEntriesFromIterable then reads `Get(entry, "0"/"1")` off each *entry*, and a heterogeneous `[key, value]` pair is a **tuple struct** (fields `_0`, `_1`) readable only through `__struct_field_names` + `__sget_*` — exports, hence unreachable during the wasm `start` section. Measured on the polyfill with only the source marshalled: both `Object.fromEntries` calls **returned**, with `{ undefined: undefined }` instead of the ten-key unit table. Silently wrong data is worse than the original throw, so the tuple half is not optional.

## What changed

| file | change |
| --- | --- |
| `src/runtime.ts` | materialise the source (`_materializeIterable`), decode each entry with the new `_decodeCompiledPair`, hand the normalised entries to the native implementation so ToPropertyKey / CreateDataPropertyOrThrow stay the engine's |
| `src/codegen/expressions/call-builtin-static.ts` | the `Object.fromEntries` lowering asks for #5193's funcref prologue (`ctx.needsInitMarshalHelpers`) — without it the runtime fix is inert at module top level |
| `src/codegen/init-class-dispatch-helpers.ts` | #5202's name-based CSV registration now also carries the struct-read family (`__sget_*`, `__struct_field_names`, `__is_data_struct`); per-field names cannot live in #5193's fixed positional ABI |

`_decodeCompiledPair` is deliberately **shallow** — only the pair container is rebuilt; the two slots keep the compiled references they were, so `fromEntries([[k, a]]).k === a` and live mutation both survive. The deep `_convertIterableForHost` was the tempting reuse and would sever both.

## Temporal harness (acceptance criterion 2)

Both rows executed on the same tree, ESM lane:

| tree | `moduleInitError` |
| --- | --- |
| base (this branch without the fix) | `TypeError: object is not iterable (cannot read property Symbol(Symbol.iterator))` |
| + this fix | `TypeError: Cannot access property on null or undefined at 4:10198` |

`moduleInitRuns` is **still false** — this clears blocker 7, not the last one. Position 4:10198 is `ct = Intl.DateTimeFormat`: `Intl` is simply not provided, and a scoped probe fails identically *after* init too, so the eighth blocker is a missing-global capability gap, not an init-window one. Being filed by the coordinator.

Both `fromEntries` calls now return the correct ten-key tables (verified by tracing host imports during instantiation).

## Validation

- New `tests/issue-5205-fromentries-init-marshal.test.ts` — 8 cases, each pairing an at-init call with an after-init control. **On base: 6 failed / 2 passed** (the two passing are the deliberate controls). With the fix: 8/8.
- No regressions: `issue-5191`, `issue-5193`, `issue-5201`, `issue-5202`, `issue-5203`, `issue-5204` (69 passed) + scoped Object runs `issue-965`, `issue-2042`, `issue-2042-fromentries-objvec`, `issue-2042-r2-topropkey-object`, `issue-2190b-anytuple-nested`, `issue-3161`.
- `issue-2042-s3` + `issue-3160` have 4 failures in the **standalone** `getOwnPropertyDescriptors` lane — verified identical on base (pre-existing).
- `equivalence-gate.mjs` shards 1–8, lint, typecheck, and the loc/func/coercion/oracle/dead-export ratchets green. Budget allowances declared in the issue file's frontmatter.

---


ttraenkler:
Closing: superseded by PR #5324's merge (full Temporal chain landed), 0-diff.

---

PR #5271, opened 2026-08-29
fix(codegen): materialize the Intl host global (#5206)

ttraenkler:
Fixes [#5206](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5206-intl-global-missing) — the eighth `@js-temporal/polyfill` module-init blocker for [#4628](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4628-temporal-runtime-object-spike).

**STACKED.** Land order: #5252 → #5258 → #5262 → #5264 → #5266 → **this**. Branched from `issue-5205-fromentries-marshal` (PR #5266) per sanctioned predecessor-stacking, with `origin/main` merged in. Do not enqueue before the predecessors land; if any of them changes, this branch needs a re-merge.

## What was wrong

There was no `Intl` global at all. `Intl` is `declare namespace Intl` in `lib.es5.d.ts` — a namespace, not a `declare var` — so none of the value-shaped arms in `compileIdentifierCore` claimed it and every read lowered to the graceful `ref.null.extern` default. The emitted WAT for `Intl.DateTimeFormat` was literally `ref.null extern` followed by the null-check throw. Unlike #5193/#5202/#5203/#5205 this was **not** an init-window timing bug: a scoped probe failed identically after init.

## Measured call surface

30 `Intl` occurrences in the linked bundle. **At module init**: `Intl.DateTimeFormat` as a value (twice), `.supportedLocalesOf` copied off the constructor, `Intl.DurationFormat?.prototype` destructured with `??`, and a guarded write onto `Intl.DurationFormat.prototype`. **Lazily**: `new Intl.DateTimeFormat(locale, options)`, `.resolvedOptions()`, `.formatToParts()`, `.formatRange()` / `.formatRangeToParts()`, `Intl.supportedValuesOf?.("timeZone")`.

## The fix

One condition on the #3087 host-global-materialization arm in `src/codegen/expressions/identifiers.ts`, which lowers a bare ambient name to `__extern_get(__get_globalThis(), name)` — the route `Buffer`, `process` and `crypto` already take. Both helpers are **imports**, not module exports, so no #5193 start-export channel was needed; it resolves during `start` unchanged.

- **Standalone/WASI untouched** (the arm is `!standalone && !wasi`); a test asserts neither lane gains a `__get_globalThis` import.
- **#2838 / `CLOSURE_UNSAFE_HOST_AMBIENTS` re-verified**: `Intl` was already in that deny set, and the check is by identifier text *before* `oracle.valueDeclarationOf`, so making `Intl` resolvable does not change which HOF callbacks take the closure lane. No edit needed.

## Temporal harness

`moduleInitRuns` is **still false** — this clears the eighth blocker, not the last one. Init now advances past all five top-level `Intl` statements and throws later with `RangeError: Invalid era data: eras are required`.

**Ninth blocker, diagnosed and reduced — a general codegen scope leak, not an Intl or init-window bug, and it fails after init too:**

```js
function C(e, t) {
  return (function (e) { let t, n = e; return n === null ? "NULL" : "len" + n.length; })(t);
}
C("x", [1, 2]);   // native: "len2"   ·   js2wasm: "NULL"
```

An **immediately-invoked** function expression / arrow whose body declares any binding named `X` makes a bare `X` in its own argument list read as `null` — the argument is evaluated in the callee's scope rather than the caller's. Storing the same function in a variable first, or calling a hoisted named function, is correct. It is a **silent wrong-value** defect, not a throw, and minifiers reuse short names constantly. Pre-existing on base; being filed by the coordinator.

Also noted, deliberately **not** fixed: a compiled `Date` is a plain object with a `timestamp` field, not a host `Date`, so `formatToParts(new Date(e))` throws while `formatToParts(0)` works — compiled-`Date` ↔ host-`Date` bridging, separate issue.

## Validation

- New `tests/issue-5206-intl-global.test.ts`, 8 cases, each at-init read paired with an after-init control. **On base 6 fail / 2 pass** (controls); with the fix 8/8.
- `issue-5191/5193/5201/5202/5203/5204/5205` + this file: **85 passed**.
- Scoped host-global / namespace runs green: `issue-1058-ambient-performance-global`, `issue-2603-process-ambient-typing`, `issue-2929-cd-global-materialization`, `issue-4150-declared-global-cache`, `issue-3956-global-object-binding-aliasing`, `issue-3176-json-namespace-reflection`, `issue-3081-number-namespace-const-receiver`, `issue-1690b-var-shadows-module-global`. `issue-3061-host-buffer-bytelength` has 1 failure verified **identical on base**.
- `equivalence-gate.mjs` shards 1–8, `typecheck`, `biome lint`, and the loc / func / coercion / oracle-ratchet / dead-export gates green. LOC + func growth (+13 lines, 12 comment) granted in the issue frontmatter.

---


ttraenkler:
Closing: superseded by PR #5324's merge (full Temporal chain landed). The remaining diff is only generated-report formatting churn from auto-refresh, no source content.

---

PR #5279, opened 2026-08-30
fix(codegen): evaluate inline-IIFE arguments in the caller's scope (#5207)

ttraenkler:
Fixes [#5207](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5207-iife-argument-callee-scope-leak) — an IIFE's argument list was evaluated in the **callee's** scope, so a bare caller identifier whose name the callee happened to reuse silently read as `null`.

```js
function C(e, t) {
  return (function (e) { let t, n = e; return n === null ? "NULL" : "len" + n.length; })(t);
}
C("x", [1, 2]);   // native: "len2"   ·   js2wasm before: "NULL"
```

No throw, no diagnostic — a wrong value. Minifiers reuse short names constantly, so every minified bundle containing an IIFE was exposed.

**Land order:** after #5252 → #5258 → #5262 → #5264 → #5266 → #5271. This branch already carries `issue-5206-intl-global` (the last of those) so the fix could be measured on the full Temporal stack.

## Root cause

One arm, in `src/codegen/expressions/call-tail-dispatch.ts`. The inline-IIFE fast path entered `enterInlineIifeBindingScope(fctx, iifeBindingNames)` and **then** compiled the argument expressions.

That scope exists for a good reason (#3128): the inline path splices the callee's body into the **caller's** `FunctionContext`, so it must temporarily hide every name the callee declares or the body would resolve to the caller's same-named locals. `collectDirectEvalBindingNames` returns exactly that set — parameters, `arguments`, and every *recursively* nested `var`/`let`/`const`, class name, function-declaration name and catch parameter — plus the function expression's own name.

Hiding those is right for the **body** and wrong for the **argument list**, which is caller-scope by §12.3.6.1 and is evaluated before the callee's environment exists. An argument identifier in the hidden set therefore missed `fctx.localMap` entirely and lowered to a null/absent read.

It also explains the two matrix rows that were already correct: "function stored in a variable" and "hoisted named function" never take the inline arm.

## Fix

Argument evaluation moves **out** of that scope, into a new `src/codegen/expressions/inline-iife-arguments.ts` (`compileInlineIifeArguments`). `compileDiscardedArgument` moved with it — the two arms that moved were its only callers. Net effect on the god-file: `call-tail-dispatch.ts` shrinks by 60 lines and `compileTailDispatch` by the same, so no LOC/function-budget allowance is needed.

Two details are load-bearing:

- The parameter slots are allocated there but deliberately **not** registered in `fctx.localMap`. Registering a parameter's name eagerly would shadow the caller's binding for the *remaining* arguments — `(function (a, b) {…})(b, a)` must read the caller's `a`. The slots keep their real parameter names so `deduplicateLocals` (which merges only `__`-prefixed temps) cannot collapse a parameter into an unrelated slot.
- `call-tail-dispatch` binds the names to those slots immediately after entering the callee scope — the first moment the body can observe them. Binding-pattern parameters are still materialized by `destructureParamObject`/`destructureParamArray` further down, unchanged.

Extra arguments past the declared parameter list were caller-scope too and had the same defect; they moved with the rest.

## Matrix

All nine reported rows, host **and** standalone (measured on this branch, 2026-08-29):

| shape | before | after | native |
| --- | --- | --- | --- |
| IIFE param `x`, inner `let t`, called `(t)` | `NULL` | `len2` | `len2` |
| IIFE param `e`, inner `let q`, called `(t)` | `len2` | `len2` | `len2` |
| inner `var t` / `const t` | `NULL` | `len2` | `len2` |
| arrow instead of function expression | `NULL` | `len2` | `len2` |
| caller binding is a `const`, not a param | `NULL` | `len3` | `len3` |
| inner PARAMETER shadows (no `let`) | `NULL` | `len2` | `len2` |
| argument is `t.length` rather than `t` | `undefined` | `2` | `2` |
| function stored in a variable, then called | `len2` | `len2` | `len2` |
| hoisted named function called normally | `len2` | `len2` | `len2` |

Two shapes **beyond** the reported matrix were also wrong on base and are covered by the new tests: later arguments seeing the caller's binding for a name an earlier parameter shadows, and extra arguments past the parameter list.

## The empty-array caveat: ONE root cause, not two

The issue flagged that the reduced repro delivered `null` at the argument while the polyfill delivered an *empty host array*, and asked whether that was a second cause. It is not — both are the same missing read at the same site. The faithful constructor repro (`GregorianBaseHelper`'s shape: IIFE inside a constructor over a parameter the callee shadows, destructured result) moves from `Cannot read properties of null (reading 'filter')` on base to a different, *later* error with the fix, and the harness's era-data throw is gone. Nothing needs separate filing on that axis.

## Temporal harness — advances past "Invalid era data"

`node --import tsx tests/dogfood/temporal-polyfill-harness.mjs`, full stack, before and after:

| | before | after |
| --- | --- | --- |
| compile / validate | success, 0 errors, binary validates | unchanged |
| `moduleInitRuns` | `false` | **`false`** |
| ESM module-init error | `WebAssembly.Exception thrown from module init (payload unreadable: no instance, so no __exn_tag)` | `TypeError: filter is not a function` |

**`moduleInitRuns` is still false** — this does not un-gate the [#4628](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4628-temporal-global-runtime) integration. But the era blocker is cleared: the opaque wasm exception was the polyfill's compiled `throw new RangeError("Invalid era data: eras are required")`, and init now runs past it to the next statement in the same constructor.

Exact stop position, with the fix:

```
TypeError: filter is not a function
    at src/runtime.ts:14131          (the extern-method dispatcher's "no arm matched" throw)
    at invokeReusable                 src/runtime/fixed-extern-method-call.ts:27
    at GregorianBaseHelper_init       wasm-function[357]:0x782fd
    at OrthodoxBaseHelper_init        wasm-function[372]:0x79994
    at EthiopicHelper_init            wasm-function[378]:0x79e63
    at __module_init                  wasm-function[8452]:0xe1af1
```

Source: the `n.filter((e => null != e.reverseOf)).length > 1` guard immediately after the era-count checks (linked bundle offset ~108,217). The UMD lane is unchanged and still fails `WebAssembly.compile()` on `__closure_35` — a pre-existing, unrelated blocker.

## New defects found (for separate filing — both PRE-EXISTING, unmasked, not caused here)

Both reproduce identically on pristine `origin/main`; the first is what the Temporal harness now stops on.

1. **`.filter` on an array-literal constructor argument reaches the host extern-method dispatcher as a compiled vec struct → `TypeError: filter is not a function`.** Reproduces with **no IIFE at all**:

   ```js
   class HelperBase { constructor() {} }
   class G extends HelperBase {
     constructor(e, t) { super(); this.eras = t.filter((x) => x.code); }
   }
   class Sub extends G { constructor(e, t) { super(e, t); } }
   const inst = new Sub("c", [{ code: "a" }, { code: "b" }]);
   // ERR: filter is not a function
   ```

2. **Wasm validation error in a derived-class constructor whose field is assigned from an inlined IIFE result**: `Compiling function "G_init" failed: type error in fallthru[0] (expected externref, got (ref null N))`. Occurs with the IIFE returning the parameter unchanged (no `.filter` involved).

## Validation

- **New tests** — `tests/issue-5207-iife-argument-caller-scope.test.ts`, 36 cases (18 shapes × host/standalone) including the polyfill's constructor shape with a populated **and** an empty era table. **28 of the 36 are red on base**; the 8 green ones are the deliberate controls (no-collision row, variable-held function, hoisted function, `arguments` carrier).
- **Equivalence suite** — all 218 files in `tests/equivalence/`, run in batches of 20 (the full suite OOMs in one vitest process). Result below.
- **Closure / IIFE / scope / arguments / shadow / hoist scoped run** — all 90 `tests/*.test.ts` matching those terms. Result below.
- **Focused re-run after the module extraction**: `issue-5207`, `issue-4031-iife-capture-scan`, `issue-4380-empty-object-widening-iife-body`, `issue-4555-typeof-arguments`, `generator-iife`, `arguments-object` — 86 passed.
- **Gates** — `check-loc-budget` · `check-func-budget` · `check-coercion-sites` · `check:oracle-ratchet` · `check:dead-exports` all green, run chained after the final edit. `npm run typecheck` clean. Biome clean on the changed files.

### Pre-existing failures — every one reproduced on pristine `origin/main`

Both suites were run in full on this branch, and **every** failing test was then re-run in a scratch worktree checked out at pristine `origin/main`. The failing sets are **identical**, test for test. Nothing here is a regression from this PR.

Equivalence (218 files) — 24 failing tests across 11 files:
`arguments-nested-and-loops` (1) · `array-inline-return` (1) · `delete-sentinel` (1) · `logical-conditional-identity` (3, the `void` → NaN family) · `misc-small-patterns` (1) · `new-non-constructor` (2) · `null-dereference-guards` (5) · `optional-direct-closure-call` (2) · `reflect-api` (1) · `tdz-reference-error` (6) · `yield-as-expression` (1).

`tests/equivalence/multi-file-compilation.test.ts` OOMs the vitest worker even run alone — **also on pristine `origin/main`**, so it is this container's 16 GB ceiling, not a code change.

Scoped run — 11 failing tests across 8 files:
`issue-1053-arguments-global-staleness` (1) · `issue-1528-closure-construct` (1) · `issue-1712-capture-closure-dispatch` (1) · `issue-2637-b2-ctor-closure-registration` (1) · `issue-3036-late-microtask-closure` (2) · `issue-3522-ir-closure-destructuring-ownership` (3) · `issue-4206-with-closure-ir` (2) · `issue-4525-moment-closure-validation` (1).

## Notes for the reviewer

- This branch also carries the two issue files from `docs-5207-5208-filing` (merged in, not re-authored) so the implementation PR can set `status: done` on #5207 per the issue-status lifecycle without an add/add conflict at merge time. **If that docs PR (#5273) is still open, it becomes a no-op once this lands.**
- `plan/issues/5207-iife-argument-callee-scope-leak.md` gains an `## Implementation notes` section recording the root cause and the two design constraints above.

---


ttraenkler:
Closing: superseded by PR #5324's merge (full Temporal chain landed), 0-diff.

---

PR #5283, opened 2026-08-30
fix(runtime): give compiled vecs and callbacks a host identity during module init (#5209)

ttraenkler:
Tenth `@js-temporal/polyfill` module-init blocker
([#4628](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4628-temporal-runtime-object-spike)). Fixes
[#5209](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5209-ctor-arg-array-vec-host-filter).

## Land order

Stacked on the Temporal fix stack — must land AFTER it. In **pull-request** numbering: PR 5252 (merged) → 5258 → 5262 → 5264 → 5266 → 5271 → 5279 → **this PR**. Those PRs implement, in order, issues
[#5193](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5193-host-typedarray-ctor-opaque-value-marshal),
[#5202](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5202-init-window-class-prototype-population),
[#5203](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5203-init-window-static-method-closure-wrap),
[#5204](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5204-host-class-bridge-f64-params),
[#5205](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5205-object-fromentries-compiled-value-marshal),
[#5206](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5206-intl-global-missing) and
[#5207](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5207-iife-argument-callee-scope-leak).
The branch already merges `origin/issue-5207-iife-arg-scope`, which carries that whole stack, so the diff against `main` includes it until the predecessors land.

## What was wrong

`t.filter((x) => x.code)` inside a constructor running at module top level failed **three separate ways, each masking the next** — all in the JS-host module-init window (top-level code runs in the wasm `start` section, so `callbackState.getExports()` is `undefined` for all of init; the #5193 → #5202 → #5203 → #5205 series has been closing this window facet by facet):

1. **Dispatch — the reported throw.** `_wrapForHost` gated its vec→host Array facade on `getExports()`. During init the vec became the *generic object proxy* instead, which has no `filter`, so the extern-method dispatcher fell through every arm and threw `filter is not a function`.
2. **The callback never ran.** With (1) fixed, `filter` resolved and returned `[]`. `__make_callback` builds a bridge whose body is `exports.__cb_<id>`; `__cb_*` was not on the start-export channel, so the bridge took its "park until exports exist" arm — **returning `undefined` on the spot**. Correct for an async reaction, silently wrong for `filter`.
3. **The field read answered `undefined`.** `x.code` off an *untyped* callback parameter is a dynamic `__extern_get` needing `__sget_<field>`. #5205 already registers that family at init, but `_safeGet`'s probe still asked `getExports()`.

Defects 2 and 3 are **silent**: fixing only the reported throw would have turned a TypeError into an empty array — the polyfill's `n.filter((e) => null != e.reverseOf).length > 1` era guard would simply stop guarding. The new tests assert values, and two rows pin that init and post-init give the **same** answer.

## Why the host path, not the compiled path

The issue suggested preferring the compiled path. The evidence says the routing is not the defect: `t` is an untyped constructor parameter, so `t.filter(cb)` is a dynamic member call codegen cannot resolve statically — and the identical statement *after* init already went through the host Array facade correctly. Only the timing differed. Keeping the value compiled would have needed a runtime `__is_vec` discriminator at every dynamic member call site, and would not have fixed defects 2 or 3.

## Changes

| File | Change |
|---|---|
| `src/runtime.ts` | `_exportSlotFor` — one live, shared export slot per struct; `_wrapVecForHost` reads through it instead of capturing a snapshot; the extern-method dispatcher's marshalling path and its `_VEC_PRIMITIVE_READ_METHODS` arm use `marshalExports()`; `_safeGet`'s `__sget_` probe uses `marshalExports()` |
| `src/runtime/native-function-source.ts` | `callbackDispatcher` resolves `__cb_<id>` from exports **or** the start-export channel; the bridge parks only when the compiled body is genuinely unreachable |
| `src/codegen/init-class-dispatch-helpers.ts` | `__cb_*` joins the #5202 CSV registration channel |
| `src/codegen/init-marshal-helpers.ts` | registration also fires when the module imports any `__extern_method_call*` — decided from the import list, one reviewable place |
| `tests/dogfood/temporal-polyfill-harness.mjs` | opt-in `JS2WASM_DOGFOOD_STACK=1` prints the module-init stack for a host-side TypeError |
| `tests/issue-5209-init-window-vec-dispatch.test.ts` | new, 18 assertions |

The `_wrapVecForHost` live slot is load-bearing: views are cached for the vec's lifetime, so one born during init with the partial helper set would otherwise have kept that set forever.

## Validation

- **New tests**: 18/18 pass. On pristine `origin/main`, **8 fail** — all host lane (`filter is not a function` ×4, `indexOf is not a function`, and two silent wrong-value rows). Every standalone row already passed; the standalone rows guard that the fix stayed in the host timing window.
- **`equivalence-gate`**: 24 failing / 1718 passing / 24 known-failures in baseline → **no new equivalence regressions**.
- **Predecessor stack** (#5193/#5202/#5203/#5204/#5205/#5206/#5207 test files): 76/76 pass.
- **Array-method scoped run** (10 files): 100 pass, 5 fail — **identical 5 failures on base** (`string_constants` harness wiring, unrelated).
- **Gates**: loc-budget, func-budget, coercion-sites, oracle-ratchet, dead-exports, typecheck, lint — all green, chained. No `--no-verify`. No baseline JSON touched.

### Temporal harness, before and after

Both runs executed in this branch (`node --import tsx tests/dogfood/temporal-polyfill-harness.mjs`), the "before" with the four changed source files reverted to their `HEAD` copies:

| | before | after |
|---|---|---|
| ESM lane init | `TypeError: filter is not a function` | `TypeError: The comparison function must be either a function or undefined: [object Object]` |
| `moduleInitRuns` | `false` | `false` |

`moduleInitRuns` is **still false** — this advances the polyfill one statement, it does not un-gate #4628.

## Next blocker (being filed by the coordinator)

Same function, one statement further — the polyfill sorts the era table right after the filter guard:

```
TypeError: The comparison function must be either a function or undefined: [object Object]
    at Array.sort (<anonymous>)
    at invokeMethod                 src/runtime.ts:10952
    at GregorianBaseHelper_init ← OrthodoxBaseHelper_init ← EthiopicHelper_init ← __module_init
```

`invokeMethod` wraps struct arguments with `_wrapForHost(args[i], exports)` where `exports = callbackState?.getExports()` (undefined at init) and never applies `_maybeWrapCallableUnknownArity` to arguments at all, so a compiled comparator closure crosses as an opaque struct. Deliberately **not** done here — `invokeMethod` is the DOM lane's hot path and callable-wrapping of arguments is a behaviour change deserving its own issue and regression run.

## Other pre-existing defects found (also unfiled)

Both reproduce on pristine `origin/main`; neither is caused or worsened by this PR:

1. **Dynamic-receiver call at module init with object elements traps** with a bare `WebAssembly.Exception` from `__module_init` — possibly the same family as [#5210](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5210-derived-ctor-inline-iife-fallthru-type-error).
2. **In-place vec mutation by a host method at init** (`t.sort(cmp)` in a ctor): the vec WRITE-side exports (`__vec_set_elem`, `__vec_set_len`, `__vec_mut_supported`) are not on the init channel; a prototype adding them as positional ids 6-8 was dropped because no demonstrably improving case could be built and that wire ABI is permanent/append-only.

---


ttraenkler:
Closing: superseded by PR #5324's merge (full Temporal chain landed). The remaining diff is only generated-report formatting churn from auto-refresh, no source content.

---

PR #5308, opened 2026-08-30
feat(ir): prepare fast scalar functions before direct emission

ttraenkler:
## Summary

- replace the blanket fast-mode preparation rejection with a fail-closed explicit scalar ABI proof
- require each `number` / `boolean` / `void` syntax position to match the final IR override and the exact allocated Program ABI slot
- preserve every existing body, ownership, caller, function-value, and component-closure guard; unsupported fast signatures remain direct
- align three stale #3521 routing assertions with behavior already present on current main while strengthening their compile-once and exact-size evidence
- record the bounded #3521 implementation plan and issue-local LOC allowance

## Validation

- focused fast-scalar routing controls: 4/4 passed
- former changed-root failures: 3/3 passed
- positive numeric and mixed f64/i32 components bypass poisoned direct bodies and report direct=0 / IR=1
- IR-off poison control and fast default-parameter negative prove the direct route remains live
- module-init control poisons the scalar callee direct body while proving module init itself remains direct
- function-value identity pins exact GC/standalone binary deltas: 0 / 119 bytes
- 8,000,000,000 numeric result retained beyond the i32 range
- TS7 no-emit, scoped Biome lint, Prettier, and all mandatory pre-push gates passed

Refs #3521

### DIFF
--- a/plan/issues/3521-ir-r2-prepared-program-free-function-compile-once.md
+++ b/plan/issues/3521-ir-r2-prepared-program-free-function-compile-once.md
@@ -2904,3 +2904,11 @@ bodies, `irFirstSkipped`, and a prepared component identity while preserving
 the `8_000_000_000` result above the i32 range. A route-off poison control proves
 the direct path is still live, and a fast default-parameter negative remains
 direct and fails under the same poison.
+
+The changed-root CI gate also exposed three assertions already stale on
+current `main`. Their coverage is retained rather than suppressed: function
+value identity now pins the exact GC/standalone binary deltas (`0` / `119`
+bytes), the independent `directOnly` scalar owner proves compile-once beside a
+blocked function-value component, and the module-init boundary poisons the
+scalar callee's direct body while proving that the module initializer itself
+remains direct.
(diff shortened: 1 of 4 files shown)


ttraenkler:
Queue triage update:

- The solo merge-group failure was infrastructure-only: shards 2, 8, 20, and 42 received hosted-runner shutdown signals before producing complete Test262 evidence.
- Targeted reruns made shards 2 and 42 green. Shards 8 and 20 were again runner-cancelled; a serial shard-20 retry received the same shutdown signal for a third time.
- No compiler/Test262 verdict failed in this solo group. The earlier high-water breach came from a different cumulative group with #5295 and cannot be attributed to #5308 alone.

I am removing hold so repository automation can create a fresh solo merge group instead of continuing to rerun the parked group.

ttraenkler:
Temporary recovery hold: PR #5317 reseeds the standalone high-water mark to the already-published forced baseline. This PR is still non-draft and otherwise merge-ready; remove `hold` immediately after #5317 lands so repository automation re-enqueues it.

PR #472, opened 2026-05-21
ci: auto-merge origin/main into PR branches to stop wasting runner hours on stale-branch reruns

ttraenkler:
## Problem

When a wave of PRs merges in sequence (today: 17 PRs in one wave), each merge advances `origin/main` and invalidates other PRs' just-completed test262 runs that ran against an older base. The existing `Verify origin/main is merged into branch` quality gate then fails on the stale PRs and forces a manual `git merge origin/main && git push` cycle — which kicks off another **5-hour test262 run** on each branch.

The result of today's wave: a lot of expensive runner time burned re-running test262 on branches whose only difference from a passing run was a few [skip ci] baseline metadata commits, ci-status JSON files, or unrelated other-PR merges to main.

## Fix

Have CI do the merge **itself** before running tests, instead of just verifying:

- Local to the runner — the PR branch on the remote is **unchanged**, no extra push, no extra CI trigger
- No-op when `origin/main` is already an ancestor of `HEAD`
- Hard-errors on conflicts (still requires human resolution)
- Applied to all PR-touched jobs:
  - `ci.yml` quality job
  - `test262-sharded.yml` `test262-shard`, `merge-report`, `regression-gate`
- `fetch-depth: 0` on the checkouts so the merge has the history it needs

The `promote-baseline` job is push-only and is skipped via the `if: github.event_name == 'pull_request'` guard on the new step.

## Why merge instead of rebase

Matches the project's existing rule (`CLAUDE.md`): "Never rebase. Merge preserves history and is safely reversible."

## Test plan

- [x] `npx js-yaml .github/workflows/ci.yml` parses OK
- [x] `npx js-yaml .github/workflows/test262-sharded.yml` parses OK
- [ ] The real proof is the next stale-branch CI run: a PR opened against a slightly-behind base should auto-pass the quality gate without manual merge-push churn.

Refs task #103.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5314, opened 2026-08-30
fix(runtime): callable-wrap struct arguments in invokeMethod (#5211) — moduleInitRuns flips TRUE

ttraenkler:
## Summary

`invokeMethod` — the generic `extern_class` method shim in `src/runtime.ts` — marshalled every WasmGC-struct argument with `_wrapForHost` **alone**. `_wrapForHost` builds a *data* facade and never builds a callable, so a compiled closure handed to a host method crossed with no `[[Call]]`.

The reported case is the @js-temporal/polyfill era table:

```js
n = n.filter((e) => e.code);   // -> a REAL host Array (post-#5209)
n.sort(((e, t) => { … }));     // comparator crosses as a raw struct
```

`n` is `any`, so the first-match ambient binding resolves `sort` to `Uint8ClampedArray_sort`; at runtime `self` is the host Array and `self["sort"]` is `Array.prototype.sort`, which rejected the argument with *"The comparison function must be either a function or undefined: [object Object]"* from `GregorianBaseHelper_init ← OrthodoxBaseHelper_init ← EthiopicHelper_init ← __module_init`.

Fixes [#5211](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5211-init-sort-comparator-opaque-struct).

## 🎉 `moduleInitRuns` is now TRUE — this un-gates the #4628 integration

The @js-temporal/polyfill ESM lane now compiles, validates **and finishes `WebAssembly.instantiate()`**. This was the eleventh and last module-init blocker in the [#5193](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5193-host-typedarray-ctor-opaque-value-marshal) → [#5209](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5209-ctor-arg-array-vec-host-filter) series.

| Harness field | Before (this branch's base) | After |
| --- | --- | --- |
| `headline` | compiled + validates, but module init THROWS | **compiled + validates + module init ran** |
| `moduleInitRuns` | `false` | **`true`** |
| `moduleInitError` | `TypeError: The comparison function must be either a function or undefined: [object Object]` | `null` |
| `compileErrorCount` | 0 | 0 |
| `binaryValidates` | true | true |

Command: `JS2WASM_DOGFOOD_STACK=1 node --import tsx tests/dogfood/temporal-polyfill-harness.mjs`, run on the full stack (…#5279 → #5283 → this) before and after.

**Exact stop position after the fix: there is none for the ESM lane — init completes.** The UMD lane is unchanged and still fails one gate earlier, at `WebAssembly.compile()`: `Compiling function #472:"__closure_35" failed: type error in return[0] (expected externref, got i32) @+336185`. That is a pre-existing codegen defect on a different lane, identical before and after.

## Which facet was real

| Facet | Real? | Evidence |
| --- | --- | --- |
| Capability — never callable-wraps arguments | **YES, this is the reported failure** | The reduced repro throws the exact TypeError **at init AND after init**. Timing is not the axis. |
| Timing — `getExports()` → `marshalExports()` | **YES, but a different symptom** | A non-closure **vec** argument crosses as the generic `[object Object]` proxy during init instead of its array facade. Separately reproducible, own failing-on-base row. |

## The change

One block in `invokeMethod`'s argument loop:

- **callable bridge first, host facade otherwise** — the same two-step every sibling path already uses (`wrapHostValue` in `__extern_method_call`, the keyed-collection `Map`/`Set` arm 130 lines above in the same `switch`, `wrapLinkedProviderValue`). This shim did only the second step.
- **`marshalExports(callbackState, exports)`** for the facade. The strict `exports` local is untouched, because the arms below read it as "init has finished"; only the marshalling question gets the wider #5209 view.

Net: `src/runtime.ts` +38 / −1.

## How the DOM hot path stays fast

`invokeMethod` is the DOM lane's hot path (#4150: 5,000 of `dom/set-attributes`'s 7,000 crossings land here):

1. **The added work is entirely inside the existing `hasStructArg` branch.** No new predicate, no new loop, no new allocation on the common path — the `callArgs = args` no-copy path is byte-identical.
2. **The DOM lane never enters that branch.** Its arguments are strings and host element handles, never WasmGC structs.
3. **The fixed-arity wrappers below still short-circuit** before `invokeMethod` runs at all when no argument is a struct — unchanged.
4. **A non-closure struct argument keeps exactly its previous result.** `_maybeWrapCallableUnknownArity` asks the module's own `__is_closure` discriminator and returns the value unchanged for anything else. The one added cost is a single `__is_closure` call per *struct* argument.

Point 2 is pinned by a **DOM-lane control test**: it drives the `dom/create-elements` + `dom/set-attributes` shapes against a recording mock and asserts the recorded attributes and child order.

## Tests

`tests/issue-5211-invoke-method-callable-arg.test.ts` — 15 assertions. **6 fail on base**, all on the HOST lane (measured by reverting only `src/runtime.ts`): 4 throw the comparison-function TypeError, 2 answer `0` instead of `1`.

Two load-bearing repro properties (documented in the file): `const _re = /x/;` registers the TypedArray extern classes so the first-match loop finds a `sort` candidate (otherwise the call falls to `__extern_method_call`, a different dispatcher with no bug); the comparator is held in a **variable** (an inline arrow crosses as a real JS function via `__make_callback` and never enters `invokeMethod` — kept as controls).

## Regressions run

| Suite | Result |
| --- | --- |
| issue-5209 / issue-5207 / issue-5205 / issue-5206 test files + `dom-style-assignment` | 71 passed |
| `dom-capability-adapter`, `extern-boundary-routing`, `fixed-host-method-calls`, `issue-2502-sort-externref`, `issue-1623-extern-doublewrap`, `issue-1888-any-extern-roundtrip`, `issue-2875-transferred-proto-method-call` | passed |
| `dom-containment` (4), `externref` (1), `issue-3058-dyn-view-proto-methods` (3) | 8 failures — **pre-existing on pristine `origin/main`**, byte-identical failure set before and after |
| `node scripts/equivalence-gate.mjs` (full suite) | **24 failing / 1718 passing / 24 known-failures in baseline — no new equivalence regressions** |

Gates chained before commit, all green: loc-budget · func-budget · coercion-sites · oracle-ratchet · dead-exports. Growth allowances for `src/runtime.ts` / resolveImport are declared in this PR's own issue file frontmatter.

## Defects found and deliberately NOT fixed here (for separate filing)

1. **Standalone: `.sort(cmp)` on a dynamic receiver is a silent no-op.** `[3,1,2]` in, `[3,1,2]` out, no throw, init and post-init alike; statically-typed receivers sort correctly. Pre-existing on `origin/main`.
2. **An omitted optional argument is lowered to `null`, not `undefined`.** `n.sort()` on an `any` receiver reaches native `Array.prototype.sort` as `sort(null)` and throws. Pre-existing; pinned by a `known-unfixed` row.
3. **The #5193 init-marshal helpers are registered only when the module needs them elsewhere.** A module with no vec-producing call registers none, so a vec argument still crosses as `[object Object]` during its init — an emission-condition gap in `src/codegen/init-marshal-helpers.ts`, not a runtime gap.
4. **UMD lane** — still fails `WebAssembly.compile()` on `__closure_35` (pre-existing, unrelated lane).

## Land order

Stacked on the Temporal module-init series. Land **after**: #5262 → #5264 → #5266 → #5271 → #5279 → #5283, then this. The branch carries `origin/issue-5209-ctor-arg-vec-filter` merged in (sanctioned predecessor-stacking) plus `origin/docs-5211-filing`, and `origin/main` re-merged after #5258 landed. Because the predecessors are still open, the diff against `main` shows their hunks too; **this PR's own change is one block** — the `hasStructArg` argument loop inside `invokeMethod`.

---


ttraenkler:
Closing: superseded by PR #5324's merge (full Temporal chain landed — including this PR's `moduleInitRuns: true` fix). The remaining diff is only generated-report formatting churn from auto-refresh.

---

PR #5318, opened 2026-08-30
feat(temporal): Temporal is a real runtime global via a compile-once linked provider (#4628)

ttraenkler:
## `Temporal` is a real runtime object now

[#4628](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4628-temporal-runtime-object-spike) step 3.

`#661` shipped a compile-time *lowering* that pattern-matches `Temporal.*` syntax — it can never define a global, which is why **1,589 test262 rows read `Temporal is not defined`**. The #4628 spike chose Option A (compile `@js-temporal/polyfill`); the 2026-08-29 attempt found a third gate the spike never reached — module init **threw** — and stopped there.

**That gate is open on this base.** Re-run 2026-08-30: `tests/dogfood/temporal-polyfill-harness.mjs --no-umd` → compile success, 0 errors, `WebAssembly.compile()` OK, **`moduleInitRuns: true`**. So this PR does the wiring.

### Land order

Branched from `origin/main` and merged `origin/issue-5211-init-sort-comparator` (sanctioned predecessor-stacking — that branch carries the whole eleven-fix Temporal stack). **Land after** PRs #5262 → #5264 → #5266 → #5271 → #5279 → #5283 → #5314.

### What a user program observes (measured, not asserted)

| Probe | Result |
| --- | --- |
| `typeof Temporal` | **`"object"`** (base: `undefined`) |
| `Object.getOwnPropertyNames(Temporal)` | `Duration,Instant,Now,PlainDate,PlainDateTime,PlainMonthDay,PlainTime,PlainYearMonth,ZonedDateTime` |
| `Temporal.PlainDate` through a function boundary | `"object"` — survives |
| `Object.getOwnPropertyNames(Temporal.PlainDate)` | `compare,from,length,name,prototype` |
| `new Temporal.PlainDate(2020,3,4)` → `y/m/d` | `2020/3/4` |

That is the issue's **acceptance criterion 1**, met: the three things the syntactic lowering structurally cannot do.

### How — compile ONCE, then link

`src/temporal-provider.ts` presents the linked `@js-temporal/polyfill@0.5.1` + `jsbi@4.3.0` bundle to the **existing** npm package linker (#2527) as a one-file package, so it inherits that machinery: its own Wasm binary, a content-addressed cache, an embedded provider manifest, and the frozen cross-module type group. `compileWithTemporalGlobal` then compiles user source with a **one-line** prelude binding bare `Temporal` to the provider's getter, and publishes the artifact in `result.linkedModules` so `instantiateLinkedProject` wires it with no caller-side provider handling.

Source *acquisition* stays in `tests/dogfood/` — `src/` carries no fixture path.

| | cost |
| --- | --- |
| Provider build, cold | **38.0 s** (once) |
| Provider build, warm | 0.75 s disk / 0 ms in-process |
| **Second consumer, compile + instantiate + run** | **430 ms** |

The rejected source-level prepend pays the full compile *every* time — ~41 h across the Temporal bucket. That number is why the shape is what it is.

### What still does not work — recorded, not hidden

| Shape | Result | What was measured |
| --- | --- | --- |
| `Temporal.PlainDate.from("…")` | `RuntimeError: dereferencing a null pointer` | **Fails identically when the polyfill is compiled as ONE module with no provider and no linking.** A pre-existing gap in the polyfill's own intrinsic / `Object.create(proto)` machinery — the provider seam neither causes it nor can fix it. |
| `Temporal.Now.instant` | `undefined` | `"function"` single-module, `undefined` through the provider. This one **is** linking-specific: `Now`'s methods do not survive the cross-module value crossing. Best next follow-up. |
| `new Temporal.PlainDate(…).toString()` | `"[object Object]"` | prototype-method / `Symbol.toStringTag` dispatch not wired. |

The test262 runner is **deliberately not wired** to the provider here: these three would move rows from `not defined` to `dereferencing a null pointer` without a net conformance gain. **Acceptance criteria 2 and 3 are not claimed.** The wiring is now small when they clear — `referencesTemporal(source)` → `compileWithTemporalGlobal(...)` → `instantiateLinkedProject(...)`.

### Three changes carried alongside

**1. `temporal-native.ts` stands down for a real `Temporal` binding.** It matched on *spelling alone*. With the provider wired and no gate, the lowering silently won and capped everything at its 3 classes — `total is not a function`, `instant is not a function`, `ZonedDateTime`/`PlainDateTime` threw. The gate is a resolution question, not a flag. **Ambient declarations do not count** — `declare` emits no value, it asserts a host global, and `tests/issue-661.test.ts` opens every fixture with `declare const Temporal: any;` in an ordinary `.ts` file. `temporal-native.ts` is **kept** — the 594 currently-passing rows are its.

**2. `src/runtime.ts` — class object `.prototype` through the dynamic lane.** This was the actual blocker, and it is a general defect: `class C { m() {} } function probe(e){ return typeof e.prototype; }` → `"undefined"`. A class object is a `$ClassName` WasmGC struct, not a *closure* struct, so the host reader (`_getOrVivifyFnPrototype`, gated on `__is_closure`) declined it — while the **static** lane answered the real prototype singleton all along. The polyfill's `ae()` helper hits it on all nine classes; that is what threw at module init. Same family as #5191. Fix returns the **raw** struct, so the lanes agree: `probe(C) === C.prototype` went from **false to true**.

**3. `CLOSURE_UNSAFE_HOST_AMBIENTS` drops `Temporal`, keeps `TemporalHelpers`.** The PR #2838 hazard (212 tests flipped pass→fail) measured **both ways** on the #3126 ref-element HOF shape: only the case the provider prelude creates moves (declared `const Temporal = {…}`: 0 → 2 passing). The deny list is consulted *before* the generic `decl === undefined || isDeclarationFile` test, so an undeclared `Temporal` was already denied there — the entry was redundant for the hazard and load-bearing only against a real binding. The test262 runner does not inject a `declare const Temporal` (checked).

### Standalone scope — OUT, stated loudly

`--target gc` / JS host only. **No new host import is introduced** — the provider's import set is whatever the polyfill's own compile needs, and the linker refuses any namespace outside `env` / the string namespaces / declared `link:` targets. But a **standalone `Temporal` does not exist**: the linker requires a deferred provider-init export, which `src/package-linker.ts` documents as unavailable for WASI (`_start` startup contract). That lifecycle comes first; deferred to a follow-up.

### Validation

- 256-row deterministic `built-ins/Temporal` slice, before **and** after (file-copy A/B on the three changed `src/` files): pass **23/23**, fail **233/233**, `Temporal is not defined` **102/102**, per-row verdict lists **byte-identical**. Expected — the runner is not wired — and a non-zero diff would have been a regression.
- `tests/issue-4628-temporal-global.test.ts` (11, incl. the heavy child-process provider lane), `tests/issue-4628-class-value-prototype.test.ts` (7), `tests/issue-661.test.ts` (5/5), `issue-3126`, `issue-4787-temporal-merge-group-regressions`, `issue-4627-captured-global-coercion` — all green.
- `#5191…#5211` family green, run one file per process. `#5194`/`#5197` OOM the vitest worker on this box — **confirmed to OOM identically on the base sources**, environmental.
- Gates: `typecheck`, `lint`, loc/func/coercion/oracle/dead-exports all exit 0. `check:host-import-policy` maxima bumped to the CI-measured 18,995 / 7,754 (most of that growth is the #5211 stack this branch carries).

### New defects found, for follow-up issues (being filed by the coordinator)

1. `Temporal.PlainDate.from(…)` — null deref, single-module too.
2. `Temporal.Now.*` methods lost across the provider boundary.
3. Compiled-class instance `toString` / `Symbol.toStringTag` dispatch.

---


ttraenkler:
Closing: this PR's entire content — the #4628 provider and the `Temporal` global — landed on main via PR #5324's merge (its branch stacked on this one), leaving this PR 0-diff. The milestone shipped; follow-up conformance work continues in PRs #5334/#5339/#5343/#5347/#5350.

---

PR #5412, opened 2026-09-01
feat(ir): #3526 F1-S3 — generator setReturn boxing under manifest authority

ttraenkler:
Implements the 2026-09-01 F1-S3 plan for [#3526](https://js2wasm.loopdive.com/dashboard/issue.html?slug=3526-ir-r6-semantic-runtime-contract), mirroring the landed F1-S1/F1-S2 machinery rather than re-deriving it.

## What this migrates

The `gen.setReturn` seam pinned `__box_number` by spelling a runtime symbol at the attachment site and choosing it by value-type presence — outside the frozen manifest's authority. That is the follow-up row both F1 checkpoints named.

Now: a resolved `generatorNumberBox` policy selects a provider, the frozen manifest carries the row (even for a generator-only module, which emits no intrinsic uses), and `attachIrGeneratorSupport` consumes that provider's physical symbol instead of one written at the seam.

It is a **sibling** of `numberBoundary`, never a widening of it — this seam boxes *natively* on the GC native-strings lane, which the number boundary's box arm deliberately cannot. A test pins that the number box arm did not acquire a native member.

## Route decisions the plan left open (both taken from measurement)

**V1 — attachment totality: the `??` fallback is unreachable, so it is retired.** One producer (`builder.emitGenSetReturn`), one production lowering site (`integration.ts` → `lowerIrEntryFunction`), attachment strictly precedes Phase 3, and every later `healthyForLower` assignment is a filter — nothing joins the lowered set afterwards. The one splice risk (`canInline` admitting a trivial single-block generator into a non-generator owner) is already rejected earlier in the same lowering arm by the `generatorBufferSlot` guard, before the boxing reference is read. `lower.ts` now throws instead of re-deciding the symbol locally: a missing attachment demotes one owner rather than silently reintroducing a second authority.

**V2 — the import-bound route is not available; the sanctioned runtime-bound alternative was taken.** The plan asked whether an import-bound `boxProvider` preserves import membership and order. The measurement did not get that far. Attaching `irImportFuncRef("env", "__box_number", "__box_number")` — one line, everything else unchanged — makes the compile **fail** on both reachable lanes:

| fixture | gc-host | gc-native-strings |
| --- | --- | --- |
| `VALUE_RETURN_GEN` | compile FAILS — `invariant/unexpected-internal-throw` | compile FAILS — same |
| `I32_RETURN_GEN` | compile FAILS — same | compile FAILS — same |
| `REF_RETURN_GEN` (no boxing) | unaffected | unaffected |

The error is `callable-provider resolution requires a runtime or intrinsic reference` — a designed precondition of `resolveAndObserveCallableProvider`, which the generator observation path feeds. So the manifest decides **which** physical symbol answers the seam (via the central `number.box` capability record on the host arm, via the union-native runtime symbol on the native arm) and the seam binds that symbol the one way its observation path admits. Sealing itself was not the constraint: `recordExternalCallable` keys on `irCallableBindingKey` for every binding kind, and the agreement check reads only `needsBoxing === (boxProvider !== undefined)`.

**V3 — freeze scan and attachment classify one population by construction:** both go through `forEachIrGeneratorSetReturn` (same `funcKind` gate, same `valueTypesOf` map, same deep walk, same predicate), with a per-case test battery asserting the two verdicts are equal.

**V4 — the plan's i32 fixture had to be replaced.** Every `type i32 = number` generator variant demotes at IR selection with `type-resolution-unsupported`, so its `f64.convert_i32_s` is legacy output and the cell would have been vacuous. The replacements are `return n | 0` and `return n > 2`, both verified IR-claimed (`legacyBodyEmitted: false`) and WAT-verified to emit `f64.convert_i32_s` → `call $__box_number` → `call $__gen_set_return`.

## Measured neutrality

**Byte parity — 35/35 cells identical, WAT included.** Seven fixtures × five lanes, compiled before and after on the same tree, then re-measured against the merged base. Every cell matches on byte length, sha256, import set **and order**; a file-by-file diff of all 35 emitted WAT texts is empty. **No WAT diff of any kind** — the F1-S1 purity-diff allowance does not apply here and none was needed.

| fixture | gc-host | gc-native-strings | standalone | WASI | linear |
| `VALUE_RETURN_GEN` | 376 ✓ | 22240 ✓ | 49963 ✓ | 49990 ✓ | demote ✓ |
| `I32_RETURN_GEN` (`n \| 0`) | 466 ✓ | 22329 ✓ | 50026 ✓ | 50053 ✓ | demote ✓ |
| `BOOL_RETURN_GEN` (`n > 2`) | 367 ✓ | 22231 ✓ | 49929 ✓ | 49956 ✓ | demote ✓ |
| `FOROF_GEN` | 416 ✓ | 22274 ✓ | 50256 ✓ | 50283 ✓ | demote ✓ |
| `REF_RETURN_GEN` (no boxing) | 344 ✓ | 22225 ✓ | 49952 ✓ | 49979 ✓ | demote ✓ |
| `VOID_GEN` (no setReturn) | 390 ✓ | 22253 ✓ | 49915 ✓ | 49942 ✓ | demote ✓ |
| `CLEAN` (no generator) | 113 ✓ | 21973 ✓ | 22588 ✓ | 22615 ✓ | 4874 ✓ |

**Imports and order — identical in every cell.** Both host lanes carry `__box_number, __gen_create_buffer, __gen_push_f64, __gen_set_return, __create_generator` in that order (gc-host prefixed by its `string_constants` globals). `REF_RETURN_GEN` carries the same list minus `__box_number` — the control showing the boxing import appears only when the seam demands it.

**Census** — `check:ir-fallbacks` output-identical (diffed, not eyeballed); unintended, module-level and post-claim buckets still empty. **Standalone/WASI/linear** — generators still demote at build; `tests/issue-2951.test.ts` needed no edit.

**Non-vacuity.** Reverting only the attachment threading (hardcoded runtime ref restored, whole schema kept) fails exactly 2 tests — the `boxProvider`-shape assertion (which also closes the measured gap that *no* test pinned `boxProvider` before this slice) and the fail-closed attachment error — while all 20 schema, policy, freeze-discipline and derivation tests stay green. The byte cells deliberately do not carry this argument; they are identical by construction, which is the point of the slice.

## Out-of-scope observation (recorded, not fixed)

A generator returning a boolean yields `{done: true, value: 1}`, not `true` — the branded i32 is boxed through `__box_number`. Measured identical on the IR path and the legacy path, so it is a pre-existing whole-compiler conformance gap, not an IR-path defect, and not something a byte-neutral slice may change. Worth its own issue.

## Validation

TypeScript 7 and 5; `check:ir-fallbacks` (bare, output-identical); the ratchet chain bare **and** under `LOC_GATE_BASE=$(git rev-parse origin/main)`; `check:ir-dialect`, `check:ir-layering`, `check:ir-only`, `check:linear-ir`, `check:host-import-policy`, `check:ir-kind-neutrality`, `check:standalone-ir-cutover-corpus` (`derived=19/19`, `units=47/47`, `terminal=38/38`). Focused suites: 134/135 across 14 files.

Two expected divergences, both precedented: `tests/issue-4104-…` gains one field in its whole-shape policy pin (as F1-S1 and F1-S2 each did), and `check:ir-kind-neutrality` needed its sanctioned evidence-line refresh — **two citation lines only**, patched surgically after establishing the semantic delta by normalising and diffing both JSON documents (committing the regenerator's output would have been a 269/85-line diff for a 2-line change). No verdict, kind, placement, ratchet count or rationale changed.

**Pre-existing failure, not caused by this change-set:** `tests/issue-2951.test.ts` › "standalone generators stay compile-twice" fails identically with the change-set reverted to `origin/main` `2dfb8396`.

## Scope discipline

`src/ir/integration.ts`'s diff is limited to the freeze scan, the policy projection, the owner-local partition and the attach call site — #3525 (codex lane, claim verified before touching the file) overlaps that file. Untouched: the from-ast generator build gate, `gen.push`/`gen.epilogue`/`gen.yieldStar` and the `__gen_*` family, legacy generator codegen, `compiler-timer-shim-preparation.ts`, and both `numberBoundary` and `booleanBoundary`.

---
_Generated by [Claude Code](https://claude.ai/code/session_0145MNKtnLSHp3Fgoo69hU2o)_


ttraenkler:
## Park diagnosed: the parked failure is not this PR's, and the "wasm-hash change" line is an artifact

Diagnosed the cited run ([33497198917](https://github.com/loopdive/js2/actions/runs/33497198917), [job log](https://github.com/loopdive/js2/actions/runs/33497198917/job/99827325421)) before touching the label, per the auto-park rules.

**The failing check:** `check for test262 regressions`, one regression —
`test/language/statements/class/subclass/class-definition-null-proto-super.js`, pass → fail, `Maximum call stack size exceeded` (`range_error`). Net −1, so the net gate hard-failed.

### The binary is byte-identical — measured four ways

The strongest reading of the log is its `Regressions with wasm-hash change: 1` line: a non-generator class test whose emitted binary appears to move under a generator-boxing slice. It does not move. Compiled on clean `origin/main` and on this branch, on the same tree:

| shape | base sha256 | branch sha256 | |
| --- | --- | --- | --- |
| raw test body | `dccb0352fdfbdb06` | `dccb0352fdfbdb06` | identical |
| body + `assert.js`/`sta.js`, sloppy | `e49d895080761e36` | `e49d895080761e36` | identical |
| body + harness, strict | `8ffaf3ba014271d4` | `8ffaf3ba014271d4` | identical |
| runner's exact `compileOptions` | `b78983fd8a3789b3` / `1ef8ae90fbacf5b1` | same | identical |
| **runner's own `assembleOriginalHarness` + `computeWasmSha`** | `aa0313d0d7f6` (primary), `c0a8b0c96fcb` (strictRerun) | same | **identical, WAT diff empty** |

The last row is the honest lane's real shape: the runner's own harness assembly, its own compile options, its own hash function, both the primary and the strict-rerun variant. Output is also deterministic across processes (3 runs, identical shas), so this is not sampling luck.

### Why the gate said "wasm-hash change" anyway

`scripts/diff-test262.ts:1747`:

```js
const wasmUnchanged = typeof baseSha === "string" && typeof curSha === "string" && baseSha === curSha;
```

`wasmUnchanged` requires **both** sides to carry a `wasm_sha`. In the current baseline JSONL (`loopdive/js2wasm-baselines`, fetched fresh):

- **0 of 48,735 entries carry `wasm_sha`** — including 0 of the 35,659 `pass` entries.

So `wasmUnchanged` can never be true against this baseline. Every pass→fail transition is counted under "Regressions with wasm-hash change", and the companion `Wasm-identical noise (pass → other, same wasm_sha): 0` line is structurally guaranteed rather than measured. The #1222 byte-identity noise filter — the exact mechanism meant to absorb a failure like this one — **is currently inert**. That reads as a baseline-schema gap worth its own issue; it is not something this PR can fix.

### What the failure actually is

With bytes identical, behavior is identical, so the `Maximum call stack size exceeded` is runner-side variance on a stack-depth-sensitive test. The same run's other signals agree: `compile_error → compile_timeout` +1 (population 15→16) and aggregate compile time +1.8% — a loaded runner. This PR is not the cause of that either; on the honest-lane module (241 KB, the test's own compile) the branch is marginally *faster*: median 792 ms vs 824 ms over 5 runs (min 725 vs 746).

The gate's `LIKELY-REAL` banner keys on baseline content-currency (0 test262-relevant commits behind main), which is true and orthogonal — it says the baseline is fresh, not that the binary moved.

### Action

Removing the `hold` for **one** re-admission, on a confirmed not-this-PR determination backed by the measurements above — not a fix, because there is nothing in this diff to fix. No code change is being pushed, and this is the single sanctioned re-run; if the same test regresses again on the re-validation I will treat that as new evidence rather than re-admitting a second time.


---

PR #5425, opened 2026-09-01
feat(#3523): gap-3 — WASI prepared module-init adapter (last direct-body producer retired) ✓

ttraenkler:
Replaces #5424 (identical head branch and content) — that PR was fully green but authored by `claude[bot]` (`authorAssociation: NONE`), which `auto-enqueue.yml`'s author-trust gate skips indefinitely; reopened here under a trusted identity per the diagnosis on #5424. No code changes; checks are keyed to head `4df1b8fd5` and are already green.

## Description

Closes gap 3 of [#3523](https://js2wasm.loopdive.com/dashboard/issue.html?slug=3523-ir-r4-module-init-compile-once) R4: WASI was the **last** direct-body/legacy-body producer in the 12-cell module-init census. Every cell now reads `direct 0 · legacy 0 · IR 1`.

| shape | host-start | host-deferred | standalone | wasi (before → after) |
| (a) `const memo = new Map()` | 0·0·1 | 0·0·1 | 0·0·1 | **2·1·1 → 0·0·1** |
| (b) `let v = 7` | 0·0·1 | 0·0·1 | 0·0·1 | **1·1·1 → 0·0·1** |
| (c) `let total = 0; total = total + 1` | 0·0·1 | 0·0·1 | 0·0·1 | **1·1·1 → 0·0·1** |

`module-init-pass1` and `pass2` both drop to 0 on every WASI cell.

### The problem

`applyModuleInitGuard` made `__module_init` idempotent by **splicing** a prologue into an already-emitted body: mint `__init_done`, prepend `global.get $done / if(return) / $done = 1`, then prepend `call __module_init` to every other export. A Prepared body cannot take that splice — its identity is fixed at the preparation snapshot, and the early `return` is exactly the return-class op that `bodyContainsReturnClassOp` withdraws the IR patch over (#3142/#3168), because every later epilogue would become unreachable.

### The change

The guard becomes **invocation-policy-driven** rather than post-hoc:

- `__init_done` is **reserved at preallocation**, before any body is emitted.
- The prepared body is **constructed around a wrapping `if`** — no return-class op, so the withdrawal scan passes and every appended epilogue (`finalizeInModuleInitFlag`'s `__in_module_init = 0` above all) still runs on the already-initialized path, which the early-`return` form would skip. Planting is gated on an option set only by the prepared preparation call, so the post-direct overlay can never plant a second guard.
- `applyModuleInitGuard` becomes prepared-aware and **fails closed**: it authenticates the planted instructions by object identity and never reassigns the body. An unguarded prepared init is an `IrInvariantError`, never a silently unguarded binary.
- **Invariant 7 grows its third arm** (`wasi-start-export`) on both the single-module and M2 routes — the `!ctx.wasi` gate meant a WASI admission would have shipped with *zero* adapter reconciliation.
- `assertGraphGlobalInvocationPolicy`'s `wasi-start-export` case now **runs against the Prepared unit**. It previously short-circuited, so the one `_start` adapter — built after declarations — was checked by nothing.
- Selector admission is last. Multi-source WASI stays rejected; no field is added to `plan.invocation`.

**One divergence from the plan:** dropping `ctx.wasi` from the refusal admits nothing on its own — `strictEnvImportGate` is `strictNoHostImports ?? target === "wasi"`, so `ctx.strictNoHostImports` is unconditionally true under WASI and refuses the lane by itself. Narrowed to `(ctx.strictNoHostImports && !ctx.wasi)`, which keeps refusing an **explicit** `--no-host-imports` gc/host build.

### Verification

- **Byte neutrality (mandatory):** file-copy A/B with `cmp` — **9/9 identical** binaries across shapes (a)(b)(c) × {host-start, host-deferred, standalone}. This slice touches nothing non-WASI and the binaries prove it.
- **Poison seam:** `JS2WASM_TEST_POISON_DIRECT_MODULE_INIT_BODY=1` is **RED on all three WASI cells at base** and **GREEN on all three after**.
- **WASI sizes** (raw / gzip -9): a 52,012→51,365 / 30,963→30,623 · b 49,236→49,162 / 29,545→29,512 · c 49,262→49,176 / 29,555→29,516.
- **Behavior pin landed first** (commit 1, green before the mechanism moved): there was no idempotency/direct-export pin for any admitted shape under WASI, so the relocation would have self-certified.
- **Fail-closed:** the four adapter mutations `issue-3520` pins on the multi-source route get single-source *prepared* equivalents, each with a positive control, plus a guard-strip seam.
- `issue-1789`, `issue-1411`, `issue-4376` legacy-lane pins untouched and green; `check:ir-fallbacks` and all five ratchet gates clean.

Probes P1–P4 and the full verification matrix are answered in the `## 2026-09-01 gap-3 checkpoint note` in the issue file, including one honest negative (an `issue-4376` test that timed out under parallel load and A/B's as pre-existing marginal fragility, not a regression).

### Growth allowance

`multi-prepared-program.ts` (+14) and `ir-prepared-free-functions.ts` (+9) added to `loc-budget-allow` in the issue frontmatter with a dated rationale. `scripts/*-baseline.json` untouched.

## CLA

---


ttraenkler:
## Park diagnosis: runner-variance flip, wasm byte-identical — re-admitting (the one sanctioned re-enqueue)

The parked run ([33538485830](https://github.com/loopdive/js2/actions/runs/33538485830)) reports exactly **one** regression: `test/language/statements/class/subclass/class-definition-null-proto-super.js` pass → fail with `Maximum call stack size exceeded` (range_error), net −1.

**Measured byte identity.** Compiled that test with the runner's own pipeline (`assembleOriginalHarness` → `compile` with the exact `runOriginalHarnessVariant` options → `computeWasmSha`) in two clean worktrees:

| variant | PR head `4df1b8fd5` | base `origin/main` |
| --- | --- | --- |
| primary | `aa0313d0d7f6` (241,802 B) | `aa0313d0d7f6` (241,802 B) |
| strictRerun | `c0a8b0c96fcb` (241,727 B) | `c0a8b0c96fcb` (241,727 B) |

The wasm this PR produces for the flipped test is **byte-identical** to main's, so the flip cannot be PR-caused. `Maximum call stack size exceeded` is the stack-depth-sensitive error class that flips under runner load — the same park mode as PR #5412, root-caused in [#5256](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5256-test262-baseline-wasm-sha-inert-noise-filter): the gate's own "Regressions with wasm-hash change: 1" line is structurally unable to excuse this (the promoted baseline carries `wasm_sha` on 0 of 48,735 entries, so the #1222 byte-identity filter is inert and **every** flip reads as a wasm-hash change).

Consistent with that: the PR's diff is WASI-module-init-scoped with a 9/9 file-copy `cmp` byte-neutrality proof across host/standalone lanes (PR description), and the flipped test is a gc-host class-hierarchy test untouched by the diff.

Removing `hold` per the auto-park protocol — one re-admission after posted diagnosis, no repeat if it parks again (that would go to a human).

---

PR #5224, opened 2026-08-29
fix(standalone): ES2015 ArrayBuffer/DataView conformance wave 1 (#5150) — undefined arg padding, ctor argument validation, module-global TA view slot, ArrayBuffer.isView as a value

ttraenkler:
## Description

ES2015 standalone buffers work package [#5150](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5150-es2015-standalone-buffers-wave1) (umbrella [#4444](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4444-es6-standalone-edition-closeout-umbrella)). This PR started as an unvalidated interrupted WIP; it now carries the validated re-implementation done on current `main` by an Opus implementer working the issue's plan, integrated by a review-and-fix workflow. The original WIP contributed nothing to the final tree (every WIP hunk resolved to `main`, then the four lane commits apply cleanly on top) — no history was rewritten.

- **Clusters A + F** — DataView setters return `undefined` and absent arguments pad with the `undefined` singleton (float setters genuinely write NaN); module-global `$__ta_view` slot for `var t = new Uint8Array(buf)` bindings; explicit-`undefined` `ArrayBuffer.prototype.slice` end (cluster C's end-default).
- **Cluster D** — buffer constructor argument validation: ToIndex/ToPrimitive order, allocation RangeError before `array.new`, DataView brand → detached → bounds order, `undefined` byteLength, plain-call TypeError (via `main`'s table-driven `CALL_WITHOUT_NEW_COLLECTION_CTORS`).
- **Cluster G** — `ArrayBuffer.isView` as a first-class value (`invoked-as-a-fn`, TA/DataView subclass instances).
- **Review fix** — the `$__ta_view` global pin is dropped when the global is rebound to a differently-typed view (`taViewGlobalIsRebound`), which otherwise trapped with a null deref on `t = new Uint8Array(2); t[0] = 9`; statically-numeric `slice` end no longer round-trips through externref boxing.

[...]

PR #5467, opened 2026-09-02
refactor(#3526): F2-S5 — string.concat under manifest policy

ttraenkler:
## What

Family 2's fourth slice of [#3526](https://js2wasm.loopdive.com/dashboard/issue.html?slug=3526-ir-r6-semantic-runtime-contract) moves the string **concatenation** seam — `a + b` on two statically-string operands, and the `+=` builder append — under the frozen runtime manifest, and retires the WasmGC emitter's no-provider fallback.

**Sub-A** adds a `stringConcat` policy with **two features** (`js.string.concat`, `js.string.concat.owned`) and **four provider rows**. The policy picks the *authority* (`nativeStrings ? native : host` — `standalone` and `wasi` both imply `nativeStrings`); the instruction's `concatMode` picks *which helper on it* (`__str_concat` vs `__str_concat_owned`). Both HOST rows name the one `string.concat` capability, because `wasm:js-string` has no owned-append builtin — the freeze projects capabilities through a set, so the collapse costs no second import. The resolve arm now reads `preparedStringConcatProvider(prepared, mode)` and fails closed; the host half is located by **import-section position** (`exactCallableImportIndex(ctx, arm.module, arm.field)`), never `ctx.funcMap` (`#1072` shadowing) and never a late import (which would shift every defined funcidx).

One new signature, `EXTERNREF_PAIR_TO_REF_EXTERN_INTRINSIC_SIGNATURE`: the `string.concat` host record returns a non-null `(ref extern)` and no existing signature carries that result. It is the only one that does — pinned.

**Sub-B** retires the emitter's `ctx.nativeStrings` fallback in `emitStringConcat`, which carried a second, independent copy of the lane decision *plus its own private mode-to-helper mapping*.

## Why

The seam's physical choice was decided twice, locally, off `ctx.nativeStrings` — once in the resolve table and once in the emitter. After this slice the frozen manifest is the single authority, and an owner whose policy cannot provide the seam is partitioned out before freeze rather than silently minting a body.

## Evidence

- **Byte-neutral: 65/65 matrix cells identical.** 13 fixtures × 5 lanes (gc-host, gc-native-strings, standalone, WASI, linear), comparing byte length, sha256, full WAT text, the ordered import list with func/global indices parsed from the binary import section, errors and demotions. `diff -r` over all 65 WAT texts is empty. **The batched many-arity cells (CAT3/CAT4/TPL/TPL3/TPLEQ — F2-S6's seam) are among them and did not move.**
- **Corpus: 104/104 cells identical** (26 `.ts` files under `website/playground/examples/**` + `examples/**` × 4 WasmGC lanes).
- **`check:ir-fallbacks` output byte-identical** to a base-tree run of the same command; `string-builder-candidate` stays at its baseline count of **2**. No `scripts/*-baseline.json` for LOC or fallbacks touched.
- **Reach counts, instrumented on both trees, identical counter for counter:** resolve arm 4 host / 19 `__str_concat` / 3 `__str_concat_owned`; `emitStringConcat` 40 `provider/immutable` + 4 `provider/owned-append`; batched arms 10; linear 9. The instrumented AFTER run is itself byte-identical to the clean BEFORE run, so the instrumentation is byte-inert.
- **Zero-reach, measured before the retirement:** a temporary `throw` replacing the fallback was reached **0 times** across the 65 byte cells (which stayed byte-identical *with the throw in*) and across 22 string suites / 352 passing tests, with a red set identical name-for-name to the base tree's.
- **Non-vacuity, three independent reverts:** the resolve arm → **3** failures (exactly the (d) arm pins); the retirement → **1** (the discriminator); the four manifest rows → **11** (the whole seam — see divergence 2 in the checkpoint note; the batched-family fence still passes, which is exactly right).
- **Runtime oracle:** `+`, a `+=` builder loop and a template checked against JavaScript on seven input pairs through an instantiated host-lane module — including two lone **surrogate halves that combine** into one astral code point and a 1000-iteration builder loop — plus native-strings and linear compile/validate.
- **Suites:** the new suite (27 tests) + every `issue-3526-*` suite + `issue-4103-ir-async-runtime-providers` / `issue-4104-ir-async-plan-runtime-consumer` + `issue-3520-callable-provider-abi` + `issue-3521-prepared-component-dependencies` + the 22-suite string set — **462 passing across 28 files**.
- **Gates green:** the five ratchets bare and under `LOC_GATE_BASE=$(git rev-parse origin/main)`; `typecheck`, `lint`, `prettier --check`; `check:ir-dialect`, `check:ir-layering`, `check:ir-only`, `check:linear-ir`, `check:host-import-policy` (**the host import set does not change** — no new `env` import, no new builtin), `check:test-vacuity-shapes`, `check:ir-kind-neutrality`.

### Baseline evidence lines touched

`scripts/ir-kind-neutrality-baseline.json`, **2 lines, patched by hand** (the regenerator emits a 269-line diff for the change): `forof.string` evidence `src/ir/integration.ts:6531 → 6629` and `string.len` evidence `src/ir/backend/linear-integration.ts:1630 → 1632`. Established by normalising both documents to sorted leaf paths: **462 leaves each, exactly 2 changed**, both evidence strings. No verdict, kind, placement, ratchet count or `settledBy` moved; `string.concat` stays `neutral`.

### Corrections to the plan (full detail in the checkpoint note)

1. **+353 net src LOC, not the plan's +170.** Structural: this is the first seam with two features under one policy, so every per-seam artefact is minted twice, and it is the first that could not reuse an existing ABI. Granted in the issue's frontmatter with a dated rationale.
2. **V-C's third revert fails 11 tests, not "(a)/(b) only".** After the migration the frozen row is the only source of the physical choice — the same interlock F2-S4 recorded.
3. **The plan's P3 suite list names `#3740`; no such test file exists.** The nearest neighbours are `issue-3741-i32-slot-promotion` (unrelated) and `issue-3744-ir-owned-append-string-builder` (already in the list). Ran the other 21 named suites plus `native-strings-roundtrip`.
4. **The (c) runtime oracle needed three owners, not one.** A fused `let s = a + b; … s += "x"` owner demotes with `string-evidence-unsupported` before the seam is reached — pre-existing string-builder-shape behaviour, the same sensitivity the census recorded for APPENDREAD/APPENDPLUS.
5. **The F2-S6 byte fence is sha-pinned on the `gc-host` cells only**; on standalone it is pinned structurally, because a sha pin on a ~23 KB module carrying the whole native-strings runtime would go red on any unrelated runtime edit.
6. **31 pre-existing reds on this container, not the 17 named by [#5274](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5274-standing-red-tests-string-and-3529-suites).** That set reproduces exactly; the extra 14 (`issue-1761` ×9, `issue-2598-2599` ×3, `issue-2163` ×1, `issue-3744` ×1) all run `optimize`, and 18 of the failures print `invalid type: distinct rec groups would be identical after binary writing (to resolve this, use --enable-gc)` — this box's Binaryen, not this slice. Identical name-for-name before and after.
7. **`tests/issue-3518-multi-prepared-string-leaf-planner.test.ts` OOMs the vitest worker in this container on either tree** (F2-S4 recorded the same). CI runs it.

## Not touched

The batched many-arity family (pass selection, `string.concat$arityN` arms, the `env.__concat_N` late import, `ensureNativeBatchedConcat`) — that is F2-S6. Also `src/ir/from-ast.ts`, `src/ir/string-builder-shape.ts`, `src/ir/passes/batch-string-concat.ts`, `string-support.ts`'s mode→symbol mapping, `tests/issue-3744-ir-owned-append-string-builder.test.ts`, and every existing policy.


ttraenkler:
Park diagnosis for [run 33603446656](https://github.com/loopdive/js2/actions/runs/33603446656): collateral from `main`, not this PR. Keeping `hold` on deliberately — re-admitting now would re-park on the identical rows.

**Byte identity.** Compiled with the runner's own pipeline at this PR's head `6d6425c8` and its base `3c99a83b`, every cited row is byte-identical on both trees:

| row | sha256 (head = base) |
| --- | --- |
| `built-ins/Atomics/notify/null-bufferdata-throws.js` | `52987c3dc0f1…` |
| `built-ins/Atomics/wait/cannot-suspend-throws.js` | `bb90c7dc46f9…` |
| `built-ins/TypedArray/…` (`illegal_cast` growth, 2 of 7 probed) | `088cbc5af180…`, `45f209b7a549…` |

This PR touches only the string-concat seam (65/65 matrix cells and 104/104 corpus cells byte-neutral, see the description); it cannot move Atomics or TypedArray rows.

**Source: PR #5224** (`5dd7a916`, merged 05:34 UTC). Its own merge group ([run 33593621223](https://github.com/loopdive/js2/actions/runs/33593621223)) reported the same two Atomics pass→fail rows and the same `illegal_cast` 28 → 35 (+7) growth; the queue merged it anyway (batched group), and its author confirmed all nine rows on `main` in [the post-merge note](https://github.com/loopdive/js2/pull/5224#issuecomment-5504974753), with a follow-up fix announced from `claude/es6-test262-standalone-g10c7u`.

**Why every full-test262 group parks until that fix lands.** `main`'s own post-merge promote refuses to move the baseline ([run 33597759219](https://github.com/loopdive/js2/actions/runs/33597759219/job/100149508335), `promote merged report to main baseline`): `illegal_cast grew 28 → 35 (+7)`, and no change-scoped `trap-growth-allow` can exist in an npm-compat refresh's change-set. The per-SHA run cache runs the same gate, so no post-#5224 `main` SHA has a baseline either. Every merge group since therefore diffs against the pre-#5224 floor and sees #5224's nine rows as its own.

**Plan.** Re-admit once (single sanctioned re-admission) after the #5224 follow-up merges and `main`'s push-event test262 run promotes successfully. If that fix stalls, the only alternative is the lead-sanctioned `BASELINE_TRAP_GROWTH_ALLOW` one-cycle valve on a `workflow_dispatch` of `test262-sharded.yml`, which bakes the nine rows into the floor — a human call, not one this lane will make.

---

ttraenkler:
Re-admitted (the one sanctioned re-admission): PR #5469 — the #5224 follow-up restoring the two Atomics TypeError paths and un-trapping the seven detached/resized TypedArray rows — merged as `2999d822`, and `main`'s push-event test262 run ([33610405931](https://github.com/loopdive/js2/actions/runs/33610405931)) promoted the baseline successfully at 08:47 UTC (`Push baseline artifacts to js2wasm-baselines repo` green, per-SHA cache written). Every row this park cited is fixed on the floor the next merge group will diff against. `hold` removed; `auto-enqueue` owns it from here. A second park goes to a human.

---

PR #5474, opened 2026-09-02
perf(#3523): gap-6a — pre-lift module-scope closures and skip module-init pass 1 when discovery is static

ttraenkler:
### What

`module-init-pass1` compiles the entire module initializer before any top-level function body, purely so those bodies can see what it discovers. Measured, it mutates 45+ `ctx` collections on a test262 harness population and exactly **one** is decision-changing for those bodies: the closure-binding family.

This adds `src/codegen/declarations/module-init-closure-prelift.ts`, which publishes that one family from the AST alone. When it can, the initializer compiles exactly **once**, after the bodies — pass 1 is skipped. `src/codegen/declarations.ts` gains wiring only; `closures.ts` and `closures/arrow-phases.ts` are untouched. Slice gap-6a of [#3523](https://js2wasm.loopdive.com/dashboard/issue.html?slug=3523-ir-r4-module-init-compile-once) (R4 module-init compile-once), implemented by an Opus lane from the planning lane's gap-6 design record; the lane was interrupted by a container restart mid-verification and resumed from its own worktree with nothing lost.

### Why

Removing pass 1 with nothing in its place breaks the async test262 runner: `doneprintHandle.js`'s `var __consolePrintHandle__ = function (msg) { print(msg); }` compiles from a typed `call_ref` down to the dynamic `__call_function_*` boundary, `$DONE` inlines it, and completion is never observed. The pre-lift is what makes the skip safe — and that is shown by mutation, not asserted: with the gate on and the inventory disabled, **5 files regress**.

Registering the shared **wrapper** is sufficient, and that is measured rather than argued. `compileClosureCall` derives its `ref.cast` target from `getClosureFuncSelfTypeIdx(info.funcTypeIdx)` — the lifted func type's self parameter — not from the closure's own struct. Both that func type and the wrapper root are keyed by **signature**. So the pre-lift mints the wrapper with an empty capture list, registers, and never compiles a body or mints a lifted function; the single later compile is free to mint whatever subtype it wants. A side effect is that the dead re-lifted `$__closure_N` twin disappears.

### Evidence

Compile samples, runner-faithful (`assembleOriginalHarness`, the runner's options), candidate vs forced two-pass:

| | host (deferred) | standalone |
|---|---|---|
| measured | 325 | 163 |
| admitted (discovery-static) | **223 (68.6%)** | **24 (14.7%)** |
| error-count divergence | **0** | **0** |
| success divergence | **0** | **0** |
| admitted modules smaller / equal / bigger | **223 / 0 / 0** | **22 / 2 / 0** |

- **Runtime, 90-file runner-faithful sample:** candidate `pass 71 / fail 18`, two-pass control `pass 70 / fail 19` — zero regressions, one gain. Same sample with the gate on and the inventory off: **5 regressions**, four of them the async `p2only` family. The inventory is load-bearing by mutation.
- **Corpus** (`website/playground/examples/**` + `examples/**`, 28 files × {gc, standalone, wasi}): **84/84 sha256-identical** to the base tree. The gate admits 0/56 there, so the slice is provably a no-op on the corpus.
- **Equivalence:** all **8 shards** exit 0, "No new equivalence regressions" (1,718 passing, 24 known-failures).
- **Suites, re-measured on the final tree:** 137 tests / 9 suites, 0 failures — `issue-3523-ir-module-init-compile-once` 20, `issue-3872` 28, `issue-2965` 11, `issue-4182-annexb-global-blockfn` 9, `issue-4195-eval-refusal-message-and-dedupe` 6, `issue-4376-module-init-chunking` 19, `issue-3765-numeric-locals` 18, `issue-3523-module-init-single-pass` 15, `issue-3523-module-init-discovery-static` 11 (new).
- **Post-merge re-validation** against `origin/main` `f64beb1a03`, which itself edits `declarations.ts`: typecheck clean, five ratchets green bare and under `LOC_GATE_BASE=origin/main`, target suites 44/44, module-init scar suites 39/39.
- **Gates:** `check-loc-budget`, `check-func-budget`, `check-coercion-sites`, `check:oracle-ratchet`, `check:dead-exports` all exit 0; `check:ir-fallbacks` exit 0 and byte-identical to the forced two-pass control. LOC/func growth in `declarations.ts` is granted in the issue frontmatter with a dated rationale; no baseline JSON touched. Full pre-commit hook on every commit; pre-push `#3765` check 18/18.
- Not run locally: `tests/issue-3518-multi-prepared-string-leaf-planner.test.ts` (OOMs the vitest worker on a 4-core box; CI covers it).

### Divergences from the plan

1. **No reuse plumbing.** The plan required the real compile to reuse pre-minted struct/func types via a `closureStructByNode` short-circuit. Measured: unnecessary — call sites cast to the signature-keyed wrapper root, and pass 2 already mints a different struct from pass 1 today. `compileArrowAsClosure` is unchanged. A simplification, recorded as a deviation.
2. **Refusals are not capture-based.** The plan built its refusal list around captures. Measured, captures are irrelevant to what the bodies read (a module-scope `let`/`const` is a module *global* here, not a module-init local), so the shipped refusals are the cases where the subtype/func-type relation does not hold, plus population-level ones. The second family was found by the pre-push `#3765` gate, not by a probe: write-once fnctor prototype methods (`Tok.prototype.run = function …`) mint the typed-`this` twin and direct-call carrier between the passes, so only the two `closureMap`-keyed shapes (`var/let/const <ident> = <closure>`, `<ident> = <closure>`) count as sites and a population with no site keeps pass 1. Three stricter rules were measured at 0/82 admission on the runner-faithful harness sample; the shipped rule admits 58/82.
3. **Refusals are pinned differentially, not by reading a reason string back.** Each refusal in the new suite ships with the admitted twin that removes exactly the refusing feature.
4. **A pass-1-only compile refusal moves to the JS runtime error channel.** `const [p, q] = <a number>` used to fail the compile; it now compiles and throws the §7.4.3 `TypeError` at run time — which is what JS does. Zero occurrences across 325 files; pinned as its own test.

### Residual risk

A **mixed** population — one keyed site *plus* a fnctor prototype method, which is exactly the test262 harness — is still admitted, so that method's `#3683` typed-`this` twin and `#3765` direct-call carrier still move from pass 1 to pass 2. It is at parity on everything this slice can measure (0/0 divergence over 325 host + 163 standalone compiles, 0 status divergence over 90 runtime files, 8/8 equivalence shards), but a twin-body difference is invisible to status and error counts, so it is **not pinned**. Refusing it outright was measured at **0/82** admission — it retires the slice. The `merge_group` standalone floor is the backstop.

Separately: `#4376`'s `preserves TDZ ordering for const writes and updates` runs **34,597 ms** against vitest's 35 s default on a 4-core box. It passes, and the isolated A/B has the pass-1 skip making it *faster* (28,452 ms vs a 31,304 ms two-pass control), so the flake exposure is ambient, not this PR's — but it is thin enough to warrant its own issue.

### Not touched

`src/ir/**` (the `#3526` F2 lanes own it), the IR selector, the prepared route, `"discover"`/`"skip"` multi-source modes, `closures.ts`, `closures/arrow-phases.ts`, the TDZ seam, the poison seam, the WASI `__init_done` reservation, `scripts/*-baseline.json`. gap-6b's integrity and capture inventories stay out of scope — but the standalone admission figure above says they are worth considerably more to that lane than the 68% host number alone suggests.

Full record, including the measurements that ruled out three stricter rules: the `## 2026-09-02 gap-6a checkpoint note — Opus lane` section of [#3523](https://js2wasm.loopdive.com/dashboard/issue.html?slug=3523-ir-r4-module-init-compile-once).

---


ttraenkler:
Park diagnosis for [run 33619016636](https://github.com/loopdive/js2/actions/runs/33619016636): **a real regression of this PR, not collateral.** The baseline is content-current (0 test262-relevant commits behind), the bucket signature `742f2e587506a8a1` is unique to this group, and every regressed row carries a wasm-hash change.

- **76 pass→other, 5 other→pass, net −71**, plus compile_error 500 → 542 and 36 other transitions.
- Trap ratchet: `oob` 14 → 29 (16 `decodeURI`/`decodeURIComponent` rows now trap `array element access out of bounds [in __module_init()]`), `null_deref` 70 → 74 (`language/expressions/object/scope-meth-param-rest-elem-var-close.js` was a pass).
- Compile errors of the form `invalid Wasm binary (WebAssembly.instantiate(): Compiling function #10x…)` on `encodeURI*/S15.1.3.x_A2.4_T*`, `BigInt/prototype/toString/a-z.js`, `String/fromCharCode/S9.7_A1.js`, `logical-assignment/lgcl-or-assignment-operator-*.js` — a pass-1-skipped module whose init now emits an ill-typed body.
- Two `dstr/ary-ptrn-elem-id-init-skipped.js` rows flip to `SameValue(«undefined», «null»)` — an initializer ordering change.

The 90-file runtime sample and the 325/163-file compile sample in the checkpoint did not cover these shapes; the unpinned residual risk named in the PR (a mixed population's twin moving to pass 2) is one candidate, the `__module_init` OOB traps another. `hold` stays; a repair lane is being dispatched with the full regression list (artifact `test262-regressions-report`). Re-admission only after the regressions are reproduced locally on the runner-faithful harness, fixed on the branch, and the fix is measured.

---

ttraenkler:
**This PR reached `main` despite its failed merge group.** The auto-park held it at 10:42 UTC, but the queue had already built the next group (#5475, docs-only, shard matrix skipped) on top of this PR's merge commit `48724f80f6`, and landing that group at 10:51 carried the commit onto `main` — the same path #5224 took this morning. `main` therefore carries the 76 regressions above.

**Revert: #5477** (`git revert -m 1 48724f80f6`) restores the floor. The repair lane re-applies gap-6a on a fresh branch by reverting the revert, root-causes the six clusters, and re-lands from a new PR with runtime-sample parity that counts instantiation. The queue defect itself (a skipped-shard group landing a failed predecessor) gets its own issue.

---

PR #4522, opened 2026-08-15
perf(#4425): WIP — kind-indexed dispatch for runNodeChecks [SUSPENDED, unvalidated]

ttraenkler:
## Status: SUSPENDED — draft on purpose, do not un-draft or enqueue

WIP handoff of #4425. The refactor is fully generated and committed, but **no gate has run**: no typecheck, no differential, no tests. The complete resume checklist and all measurement evidence live in `plan/issues/4425-node-checks-switch-dispatch.md` (in this PR's diff) under **Suspended Work**.

## What this is

Mechanical restructure of `src/compiler/early-errors/node-checks.ts` (follow-up to #4423 / PR #4519 and the corrected rejected-early-out note in PR #4521): the flat chain of 95 guarded checks — all evaluated for every AST node — becomes `on([kinds], fn)` registrations indexed by `node.kind`, plus one traversal closure per file instead of one per node.

Safety design: every block keeps its original guard **verbatim** (over-registration is harmless; under-registration loses diagnostics and is caught by the differential), registration order preserves per-kind execution order, and traversal is unconditional (the failure mode that sank the first #4423 early-out).

The file was **generated by a TS-parser slicer**, not hand-edited — hand transcription corrupted a `  ` regex on the first attempt. The generator derives each block's kind set from its own guard and cross-checks a hand catalog: 95/95 agree.

## Evidence so far (loaded box — CPU-time & ratios only)

- `detectEarlyErrors` over an 8,621-file / 22.9 MB corpus sample: ~38 s CPU; with all 95 blocks removed (traversal + other passes kept): ~3.0 s CPU → check machinery ≈ 92% of detect.
- Profile: guard chain + `ts.isX` calls + per-node recursion closure ≈ 40% of detect CPU; fired check bodies ≈ 60% (that larger prize is deliberately out of scope here).
- Differential baseline (pre-refactor): 25,863 files, 23,979 diagnostics, 0 exceptions — the gate for this PR is **byte-identical** output.

## Resume

Worktree `/workspace/.claude/worktrees/issue-4425-node-checks-switch-dispatch` holds the gitignored harnesses (`.tmp/gen-dispatch.mts`, `ee-diff.mts`, `ee-base.jsonl`, `ee-count.mts`). Follow the checklist in the issue file top to bottom; un-draft only when the differential is byte-identical and tests pass.

## CLA

- [x] I have read and agree to the CLA

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5509, opened 2026-09-03
docs(#3518, #3090, #4522, #3523, #5282): R9's denominator, R10's audit provenance, and seven kill-switches missing from R9's scope

ttraenkler:
Docs-only, seven commits, stacked on #5508 (merged). No `src/`, `tests/`, `scripts/`, `.github/` or `benchmarks/` file is touched — the one JSON added lives in `plan/log/`.

Opened for one correction and grew as each measurement raised the next question. The body below is the whole of it, so nothing is approved sight-unseen.

## 1 — The standalone half of the R9 gap is a corpus/target mismatch

#5508 gave R4 and "the standalone host surface (R6)" equal billing as R9 dependencies, because `host-surface-unavailable` was 12 of 14 standalone rejections on the playground's uncovered eight. All 12 are **DOM**, and `document` is the only host global named anywhere in those files — 18 occurrences, zero `window`/`navigator`/`fetch`/`console`. They are browser demos compiled for a target with no JS host, which `select.ts:203-204` already says is the target's capability policy rather than IR shape coverage. So those files are a **poor standalone denominator**, not an R6 gap comparable to R4.

## 2 — R10's audit re-run, and its headline number does not reproduce

`node scripts/audit-legacy-reachability.mjs` still runs clean — worth knowing, since R10's acceptance rests on it. But:

| | recorded (2026-07-16) | today |
| --- | --: | --: |
| frontend files | 35 | 107 |
| frontend legacy-only fn-lines | 59,676 | **85,609** |

**"The front end grew 43%" is not claimed.** Three explanations fit equally — real growth, bucket membership shifting, or the old table's `files` column counting only listed files — and separating them needs the Phase-2f JSON (written to gitignored `.tmp/`, gone) or the July tree (this clone is too shallow). R10's ladder row now carries the caveat instead of the bare figure.

## 3 — That provenance, committed

`plan/log/3090-legacy-reachability-2026-09-03.json` (16 KB): bucket totals, the audit's cut set, and per-file detail for all 107 frontend-bucket files. A **dated record, not a ratchet baseline** — nothing gates on it, no CI job writes it. This is the piece July lacked, which is why today's discrepancy can't be attributed.

## 4 — [#4522](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4522-ir-kill-switch-inventory-r9)'s inventory misses seven verified hatches

R9 removes the escape hatches, so one that isn't listed survives the flip. That check had never been run. 245 `JS2WASM_*` vars in `src/`; the table names 24. No stale rows — but seven verified route/rollback hatches are absent, each read at its site:

| var | what it gates |
| --- | --- |
| `JS2WASM_DIRECT_CALLS` | the comment's own words: *"the `JS2WASM_DIRECT_CALLS=0` **kill-switch**"* |
| `JS2WASM_FIXED_ARITY_HOST_CALLS` | `=== "0"` selects **`legacyArrayAbi`** |
| `JS2WASM_ENABLE_MODULE_INIT_DISCOVERY_STATIC` | #5480's pass-1 skip **seam** |
| `JS2WASM_PINNED_THIS_DIRECT_CALLS` | disables the pinned-`this` direct-call path |
| `JS2WASM_NUMERIC_ADMISSION` | `=0` *"restores the"* prior behaviour |
| `JS2WASM_PROXY_MODULE_ESCAPE_GATE` | an escape gate that can be turned off |
| `JS2WASM_STRICT_FALLBACKS` | fallback-classifier strictness |

**Two of the seven are labelled "kill-switch" in the very source the table inventories.** `JS2WASM_LINEAR_IR_COVERAGE` was checked and is correctly *out* of scope (opt-in coverage instrumentation). The `TEST_`/debug families (120 vars) are out of scope too, though `TEST_POISON_DIRECT_*_BODY` becomes meaningless once the direct front end is deleted, so it is R10 cleanup.

## 5 — Which type extensions unlock which files, and a correction to a running lane

Instrumenting the `no-value-kind` arm and restoring from a pre-edit copy: **every one of the 13 affected dogfood files has exactly ONE blocking category.** No file mixes them, so each extension's payoff is independent and additive — partial coverage is *not* wasted, which reverses the pessimistic reading of all-or-nothing granularity.

**The live correction:** the R4-M1 dispatch brief told that lane *"no dogfood file's rejections are string-only — expect the corpus counts not to move."* False. `escapes-unicode.js` and `templates.js` are string-only, so a correct string slice should unlock 2 of 13 and the counts **should** move. As briefed, that lane would read a real success as the predicted null result.

## 6 — [#5282](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5282-r2-fast-arm-masks-every-later-withdrawal-reason) filed: the R2 fast-arm masks every later withdrawal reason

`fast-signature-unproven` is entry zero of R2-T1's admission table and the table is read with `find`. So in a fast lane every refusal answers that one reason — a 20-member vocabulary collapsing to one. Blocks R2-E1, and weakens the shape pins. Found reviewing #5507, verified against `ir-prepared-free-functions.ts:1354-1366`.

## 7 — R9-D1 plan: the dogfood corpus as a CI `baseline` lane

Everything above is invisible to CI. The gate already has `IrOnlyLaneReadiness: "baseline"` — measured and ratcheted, compile-once withheld. The plan names the obstacle up front: seeding writes `scripts/ir-only-baseline.json`, which main alone writes; adding a new lane moves no existing floor but is close enough that it needs the lead's sign-off, not an implementer's judgement. Acceptance requires a deliberate regression to make the gate fail — a baseline lane that cannot go red is decoration.

## Corrections to my own earlier claims, all left visible

R3→R4 (reason labels read without grouping by `unitKind`); `from-ast.ts`→`select.ts` (`demoteToLegacy` throws at stage `build`, these are `select`); "R4 first" retracted as generalised from one corpus; and the string-only brief above. Kept on the record rather than edited away, with the standing rule they produced: **name the corpus in the claim, and do not promote a per-corpus finding to a ladder dependency until a second corpus agrees.**

## Validation

All five ratchet gates green on every commit, bare and under `LOC_GATE_BASE=$(git rev-parse origin/main)`; `prettier --check` clean; #1616 issue-integrity pre-push gate passes. Every figure was produced in this session — none inherited from a prior artifact. Source instrumentation used for measurement was reverted from a copy taken before the first edit, and `git status src/` verified clean.


ttraenkler:
**Correction to this PR's own §5 — it is retracted, and the retraction is now in the diff** (commit `05f1b64992`).

§5 above claims: *"every one of the 13 affected dogfood files has exactly ONE blocking category. No file mixes them, so each extension's payoff is independent and additive — partial coverage is not wasted, which reverses the pessimistic reading of all-or-nothing granularity."*

**That is an instrumentation artifact.** `buildModuleBindingGlobals` (`src/ir/integration.ts:5839-5846`) walks top-level declarations in source order and **throws on the first one with no supported storage**. The `JS2WASM_IR_SHAPE_DIAG` recorder sits on that throw, so it fires at most once per file. A file with five blockers and a file with one are indistinguishable to it, and both report one. "No file mixes categories" is a restatement of the control flow, not a property of the corpus.

`tests/dogfood/corpus/escapes-unicode.js` shows it in five lines — `const a = "…"` on line 1, `const obj = { "b": 2 }` on line 5. It was recorded as string-only and listed as one of the two files a string slice would unlock.

**Independent confirmation arrived before I noticed.** #5511 implemented exactly that string slice and measured the corpus: `vardecl-module-storage-unrepresentable` 20 → 17 rows — `templates.js` on both lanes and `regex.js` on standalone. **`escapes-unicode.js` is not among them.** The prediction failed on precisely the file the artifact mis-described.

What changes:

| §5 claim | status |
| --- | --- |
| each file has exactly one blocking category | **retracted** — artifact |
| payoffs independent and additive; partial coverage not wasted | **retracted** — the all-or-nothing rule stands |
| best-set-of-size-N table | **retracted** — counts are a first-blocker histogram, i.e. a lower bound per category |
| "counts did not move ⇒ incomplete slice" (the reviewer instruction) | **retracted** — it would have produced a wrong rejection of #5511 |
| `escapes-unicode.js` / `templates.js` are string-only | **half wrong** — `templates.js` moved off the storage blocker; `escapes-unicode.js` did not |

The measurements in §§1–4 and §6 are unaffected — this one used a different instrument from the rest, and it is the only one where the instrument could not have returned a different answer.

Kept visible in the diff rather than edited away, with a `⚠ RETRACTED` marker at the original claim so nobody reads the table without the retraction. The standing rule this produces, sibling to the one already in the body: **name the instrument, and ask what answer it is incapable of returning.** A fail-fast path read as a survey reports "exactly one" every time — and does it in the shape of a clean, surprising, actionable result.

Note for whoever lands these: #5509 and #5511 both append to the end of `plan/issues/3523-ir-r4-module-init-compile-once.md` and will conflict. Union in document order, retraction first.

---

PR #5510, opened 2026-09-03
fix(#5284): bind in-graph namespace imports, stop dropping curried calls

ttraenkler:
Two silent-wrong-answer defects found by running every `tests/dogfood/*-upstream-suite.mjs` locally against this branch. Both compile clean and only fail at run time — see [#5284](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5284-npm-upstream-suite-silent-call-drops) for the full write-up.

## 1. `import * as ns` had no binding on the multi-file paths

`generateMultiModule`'s import-alias pass skips namespace imports on purpose ("resolves to a module object, not a single function/global binding"), and nothing else bound them. `ns` reached codegen unbound, so every `ns.member(...)` lowered to `__extern_method_call` on a `ref.null extern` receiver:

```
Cannot read properties of null (reading 'parseCookie')
```

`preprocessImports` handles the shape, but only on the single-source `compile()` route — `compileMulti` / `compileProject` never call it. Six-line repro:

```ts
// mod.js
export function f(s) { return s.length; }
// entry.ts
import * as m from "./mod.js";
export function test(): number { return m.f("abc"); }   // → null deref
```

`src/multi-namespace-import.ts` rewrites the namespace form into the named form that already links, **only** for specifiers that resolve inside the compiled graph. `node:*` and other host modules are untouched — that is [#4422](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4422-module-object-externals)'s work, and a named import there would only move the failure. Every in-place edit is the same byte length as the text it replaces (`ns.member` → `ns$member` plus padding; the import statement → a same-length block comment) and the generated imports are appended after EOF, so no pre-existing source offset moves — the discipline `foldGroundCallsInMulti` already keeps. It declines when the namespace name is shadowed anywhere in the file, when the namespace is used as a value, or when a member is not in the target's export set.

## 2. `f(a)(b)` dropped the outer call

The signature-directed dispatcher engages only when the callee expression's TS type carries a call signature. An `any`-typed callee — every untyped `.js` dependency, anything behind an `as any` — has none, so the shape fell to `compileCallDispatchTail`'s graceful fallback. The **inner** call ran, then its closure and every outer argument were dropped:

```wat
(func $test (result f64)
  f64.const 1
  call 1          ;; h(1) runs
  ...
  struct.new 8
  extern.convert_any
  drop            ;; the closure is thrown away
  f64.const 2
  drop            ;; …and so is the argument
  ref.null extern ;; answer: undefined
  return)
```

Now routed through `tryEmitInlineDynamicCall`, the runtime ladder identifier and element-access callees already use. No new dispatch vocabulary.

## Measured

| package | before | after |
| ------- | ------ | ----- |
| cookie  | 65/63740 | 77/63740 |
| axios   | 21/231 | 23/231 |
| redux   | 13/82 | 13/82 |

Redux nets zero: the namespace fix makes `applyMiddleware.spec.ts` reach a pre-existing `RuntimeError: illegal cast` in `dispatch` (−1), and the curried fix recovers the row (+1). Sixteen other packages unchanged.

**No regressions.** The 18 test files that fail on this branch fail identically with and without this change — 51 failing cases either way (missing `test262` / `test262-fyi` submodules, plus exactness pins that were already red). Verified by reverting both source files to their base copies and re-running exactly those files.

New regression tests: `tests/multi-namespace-import.test.ts` (10 cases), `tests/curried-call-dispatch.test.ts` (6).

## Why the corpus barely moved — the follow-up that dominates what is left

`__module_init` runs from the Wasm **`start` section**, i.e. inside `WebAssembly.instantiate`, before the host wires the struct getters (`__setExports`). A closure read out of an object property during module init therefore comes back non-callable, with byte-identical lowering to the working in-function case:

```ts
const o: any = {};
o.p = function () { return 7; };
const t = typeof o.p;                       // module init → "object"
export function test() { return o.p(); }    // at runtime  → 7
```

Every upstream test file registers its cases in module init, so `it.each(rows)(name, body)` registers **zero** tests — 63,491 of cookie's 63,740 rows never run and are scored as silent failures with no error text. The mechanism to fix it already exists for the *read* path ([#2800](https://js2wasm.loopdive.com/dashboard/issue.html?slug=2800-delete-aware-module-init-reads)'s `__in_module_init` flag plus the host-free `__get_member_<name>` dispatcher); that arm explicitly declines for function-typed members, which is exactly this case. Extending it to the any-receiver method-call path is worth more than every other npm-compat item combined.

### DIFF
--- a/plan/issues/5284-npm-upstream-suite-silent-call-drops.md
+++ b/plan/issues/5284-npm-upstream-suite-silent-call-drops.md
@@ -1,0 +1,147 @@
+---
+id: 5284
+title: "npm upstream suites: namespace imports and curried calls answered `undefined` instead of running"
+status: done
+sprint: current
+created: 2026-09-03
+updated: 2026-09-03
+completed: 2026-09-03
+priority: high
+horizon: m
+feasibility: medium
+reasoning_effort: high
+task_type: bug
+area: compiler
+goal: correctness
+loc-budget-allow:
+  - src/compiler.ts
+---
+
+## Problem
+
+Running every `tests/dogfood/*-upstream-suite.mjs` locally against this branch
+surfaced two defects that **compile clean and only fail at run time**. Both are
+instances of the same family the graceful `ref.null.extern` fallback creates:
+an unrecognised shape is rendered as the VALUE `undefined`, with the callee
+never invoked.
+
+### 1. `import * as ns` had no binding on the multi-file paths
+
+`generateMultiModule`'s import-alias pass skips namespace imports on purpose —
+"resolves to a module object, not a single function/global binding". Nothing
+else bound them, so `ns` reached codegen unbound and every `ns.member(...)`
+lowered to `__extern_method_call` on a `ref.null extern` receiver:
+
+```
+Cannot read properties of null (reading 'parseCookie')
+```
+
+`preprocessImports` handles the shape, but only on the **single-source**
+`compile()` route; `compileMulti` / `compileProject` never call it. A named
+import of the very same module links correctly.
+
+Reproduced in six lines:
+
+```ts
+// mod.js
+export function f(s) { return s.length; }
+// entry.ts
+import * as m from "./mod.js";
+export function test(): number { return m.f("abc"); }   // → null deref
+```
+
+### 2. `f(a)(b)` dropped the outer call
+
+The signature-directed dispatcher in `call-tail-dispatch.ts` engages only when
+the callee expression's TS type carries a call signature. An `any`-typed callee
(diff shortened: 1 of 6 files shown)


ttraenkler:
Closing: measured against a stale base. This branch forked from a checkpoint commit 3,838 commits behind `main`, and both defects it fixes are already fixed on current `main` — the cookie upstream suite is **63740/63740** there, not the 65/63740 I measured. Re-measuring against current `main` and will open a new PR for whatever is genuinely still failing.

PR #5517, opened 2026-09-03
docs(#5289): R4's largest slice is an ABI question, not a missing predicate

ttraenkler:
Docs-only, one new issue file. No `src/`, `tests/`, `scripts/`, `.github/` or `benchmarks/` file is touched.

## Why this is a second open docs PR

The one-open-docs-PR rule says to push onto the existing docs branch instead of opening another. **That is not possible here:** [#5509](https://github.com/loopdive/js2/pull/5509) is in the merge queue, so `GH006` rejects any push to `claude/docs-r9-standalone-corpus`. The rule's mechanism is blocked, and the alternative — a finished issue sitting on a pushed branch with no PR — is the invisibility the "always open a PR" rule exists to prevent. One extra review carrying one file is the smaller cost.

## What it records

[#5285](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5285-module-init-storage-census-cannot-be-measured)'s survey (PR #5515) settled R4's slice order with a real instrument for the first time:

| category | refused decls | files touched | files it actually **unlocks** |
| --- | --: | --: | --: |
| **`any`** | **27 of 57** | **9 of 15** | **4** |
| `function` | 9 | 2 | 2 |
| `ambient/import` | 6 | 2 | 0 — both also refuse on `vardecl-modifier` |
| `string` | 0 | 0 | 0 — R4-M1 already landed it |

`any` is the slice by every measure, on both lanes. It is also the one R4 cannot simply start — and **not because a value kind is missing**.

`IrModuleBindingValueKind` already has `{ kind: "dynamic" }`. Exactly one arm produces it (a binding whose `++`/`--` retypes it), and the comment directly above that arm names the blocker in the code's own words: fast mode carries `$AnyValue` while compatibility widens to `externref`, *"until that ABI is unified"*.

**That is a different shape from R4-M1.** String worked because `IrType.string` already defers to `IrLowerResolver.resolveString`, so one new kind could name the *active* carrier and let the storage-agreement check arbitrate — two spellings of one source fact. Dynamic has no such deferral: the lanes disagree about what the value **is**, not how it is spelled. Hence `horizon: l`, `feasibility: hard`, as its own issue rather than an R4 storage slice.

## The plan's first deliverable is a measurement, not a patch

Read what each lane actually allocates for an `any` module global today, rather than trusting `#4208`'s comment — that comment is the *hypothesis*, and it predates R4-M1. Then classify the gap as **deferrable / unifiable / irreducible** and say which *before* touching `scalarKind`. Starting at the implementation is the named risk: a lane briefed to "add an `any` arm" reaches that comment and either stalls or resolves a slot it cannot honour, which `resolveModuleBindingGlobal`'s agreement check would then raise as a hard `IrInvariantError` rather than a demote.

## Two near-misses it exists to prevent

Both nearly happened on 2026-09-03:

- **Dispatching `any` as a storage slice.** The ranking is right; the sizing is not. The check that catches it is two greps — does the target kind already exist, and what does its admission arm's comment say.
- **Ranking by "storage cleared" rather than "unlocked".** `ambient/import` clears two files' storage and unlocks zero, because both also refuse on `vardecl-modifier`. Any future R4 ranking must carry that second column.

## Validation

All five ratchet gates green chained and bare; `check:issue-ids:against-main` green; `check-issue-spec-coverage` green; `prettier --check` clean. Never used `--no-verify`.

---

### DIFF
--- a/plan/issues/5289-any-module-binding-abi-unification.md
+++ b/plan/issues/5289-any-module-binding-abi-unification.md
@@ -1,0 +1,137 @@
+---
+id: 5289
+title: "R4's largest slice is blocked on one ABI question: `$AnyValue` in fast mode vs `externref` in compatibility"
+status: ready
+created: 2026-09-03
+updated: 2026-09-03
+sprint: current
+priority: high
+horizon: l
+feasibility: hard
+reasoning_effort: max
+area: ir
+goal: backend-agnostic-ir
+requested_by: ttraenkler/fable-ir-takeover
+related: [3523, 3518, 5285, 4208]
+---
+
+## Problem
+
+The #5285 survey (PR #5515) settled R4's slice order with a real instrument for
+the first time. The answer is unambiguous:
+
+| category | refused decls | files touched | files it actually **unlocks** |
+| --- | --: | --: | --: |
+| **`any`** | **27 of 57** | **9 of 15** | **4** |
+| `function` | 9 | 2 | 2 |
+| `ambient/import` | 6 | 2 | 0 — both also refuse on `vardecl-modifier` |
+| everything else | ≤3 each | ≤1 each | 0 |
+| `string` | 0 | 0 | 0 — R4-M1 landed it |
+
+**`any` is the slice**, by every measure, on both lanes. And it is the one R4
+cannot start, for a reason that is not a missing predicate.
+
+### It is not a missing value kind
+
+`IrModuleBindingValueKind` **already has** `{ kind: "dynamic" }`
+(`src/ir/module-binding-value-kinds.ts:15`). Exactly one arm ever produces it —
+`src/ir/module-bindings.ts:2011-2013`:
+
+```ts
+let valueKind =
+  options.numberStorage === "f64" && updateRetypesModuleBinding(checker, declaration)
+    ? ({ kind: "dynamic" } as const)
+    : scalarKind(declaredType, options);
+```
+
+a binding whose `++`/`--` retypes it, and nothing else. The comment directly
+above states the blocker in the code's own words:
+
+> #4208 S2 — … Fast mode has a `$AnyValue` dynamic carrier while compatibility
+> allocation currently widens these globals to externref, so it stays on direct
+> codegen **until that ABI is unified**.
+
+### Why this is not shaped like R4-M1
+
+R4-M1 (string, PR #5511) worked because a backend-agnostic marker already
+existed and the two backends disagreed *cleanly*: `IrType.string` defers to
+`IrLowerResolver.resolveString`, so one new kind could name the **active**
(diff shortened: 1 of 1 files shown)


ttraenkler:
**Step 1 of this issue's plan is now measured. The blocker is REAL — and my first attempt at the measurement got it backwards.** Posting here because the commit carrying it (`ddc663a534`) cannot be pushed: this PR was auto-enqueued minutes after opening, so `GH006` locks the branch. It lands on a follow-up.

## The measurement

`let g: any = 1` read from an exported function; the `(global $__mod_g …)` line out of the emitted WAT. **Four cells, not two.**

| target | mode | carrier |
| --- | --- | --- |
| `gc` | compatibility | `(mut externref)` |
| `gc` | **fast** | `(mut (ref null 34))` |
| `standalone` | compatibility | `(mut externref)` |
| `standalone` | **fast** | `(mut (ref null 45))` |

The split is on the **fast/compatibility** axis, exactly as `#4208`'s comment says — *not* on target. Both targets agree with each other within each mode. So that comment is **current, not stale**, and the blocker this issue describes is real.

## The trap, recorded because I walked into it

My first probe varied `target` with default options, saw `(mut externref)` twice, and pointed straight at *"the two lanes agree, the ABI blocker is stale"* — which would have unblocked R4's largest slice on a false premise, in the most attractive possible direction.

It was wrong because **`fast` is a separate flag from `target`**. `integration.ts:2301` sets `numberStorage: ctx.fast ? "i32" : "f64"`, and `#4208`'s "fast mode" means *that flag*. Default options are the **compatibility** side of both targets — two cells from the same side of the very split being measured.

**The step-1 instruction as I originally wrote it was itself the trap:** *"read what each lane allocates … on both lanes."* "Both lanes" is ambiguous between the two axes, and the ambiguity resolves toward the wrong answer. It needs to say `{gc, standalone} × {compatibility, fast}`.

## What this settles for step 2

**Not `deferrable`.** The R4-M1 shape needs two *spellings of one fact* that a resolver can pick between by naming the active one. These are not that: `externref` is opaque and host-shaped, the fast carrier is a typed struct ref, and a resolver serving both cannot name "the active carrier" without the storage-agreement check losing its meaning — which is criterion 3 of this issue.

So the live question is **`unifiable` vs `irreducible`**, in this order:

1. **What is the fast carrier?** The indices differ per target (34 vs 45) — ordinary per-module numbering — but whether both resolve to the same *named* type (`$AnyValue`) is unverified. Read the type section; do not infer from the index.
2. **Can compatibility adopt it, or fast adopt `externref`?** Either direction is a Program-ABI change needing its own byte-neutrality argument, and the answer decides whether this is a slice or a design record.

---

PR #5534, opened 2026-09-03
fix(#5270): ES2015 standalone expressions — r2 residual pass

ttraenkler:
## Description

Residual pass on [#5270](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5270-es2015-standalone-expressions-r2) (umbrella [#4444](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4444-es6-standalone-edition-closeout-umbrella)), implemented by an Opus lane from the Fable plan in the issue file and put through **three** independent adversarial review rounds. Rounds 1 and 2 each found and fixed a real defect; round 3 held it with two more. All are fixed here.

**What landed** (standalone lane): object-shape widening for a re-declared `var`; `Object.getPrototypeOf` answering from the prototype singleton; module-init retention of bare statements that reach ToPrimitive; the `"prototype" in arrow` fold; `caller`/`arguments` poison-pill accessors on arrows; and a soundness tightening of the tail-call result check.

## The three review rounds

**R1** — block-entry slot pre-allocation desynced a `let`/`const` from the slot a nested function had already captured. **R2 (F1b)** — the arrow-`prototype` fold admitted any binding whose initializer is an arrow, relying on a rebinding guard that only matches a bare-identifier assignment LHS; `[a] = src`, `({a} = src)` and `for (a of src)` all walked past it. Re-routed through `ctx.oracle.constInitializerOf`, so the fold is justified structurally rather than by enumerating spellings.

**R3 held it with two HIGH findings**, each reproduced by an independent skeptic that materialised the merge-base itself with node as oracle:

- **R3-F1** — adding `==`/`!=` to the ToPrimitive-reaching operator list retained a bare module-scope `objA == objB;`, which then lands on the compiler's *pre-existing* over-coercing `==` lowering. A poisoned `valueOf` aborted module init where base and node reach the next statement. (Attribution is clean: wrapping the same statement in `if (true) { … }` makes base coerce too, so retention is what newly exposed it.)
- **R3-F2** — `hasRestrictedProperties` answered "restricted" for *every* arrow. §10.2.4's accessors are `configurable: true`, so `Object.defineProperty(arrow, "arguments", {value: 7})` legally replaces them: node `7`, base `7`, lane `TypeError` — on both read and write paths.

### Both fixes corrected the review, with measurements

- The review proposed dropping the two tokens and asserted it "costs no row". **Measured: it costs one.** `language/expressions/equals/coerce-symbol-to-prim-invocation.js` is literally a bare `0 == y;` at module scope, and went back to `fail` with the tokens removed. So the tokens stay, ANDed with a new `looseEqualityCoercesAnOperand` predicate: retain only when one operand is a **literal non-nullish primitive**. Object-vs-Object keeps base's drop — and so does `o == null` / `o == undefined`, a second instance of the same defect the review never named. On that matrix the lane now matches node exactly where base was wrong on two slots.
- The review proposed reusing `arrowBindingNeverGainsProperties` for F2. **That would have been wrong**: `ArrowFunction_restricted-properties.js` does `arrowFn.caller = {}`, which that predicate refuses as a property-gaining write — but a member write can never install an own `caller`/`arguments`, since it goes through the inherited poison setter and throws. Only `Object.defineProperty` can, and that is a call argument. Used a custodial-use scan instead, which subsumes `identifierEscapesToCall`, the rebinding guard (every spelling: `f = …`, `[f] = src`, `({f} = src)`, `for (f of src)`, `f++`) and aliasing.

The intended win is preserved: `(() => {}).caller` throws, and `ArrowFunction_restricted-properties.js` stays fail→pass on standalone **and** host, verified non-vacuous.

## Measured — node / base / lane-before / lane-now

| row (standalone) | node | base | lane-before | lane-now |
| `p == q;` poisoned, then a later statement | reached | reached | **module init dies** | reached |
| invocation log `a==b;` \| `var r=(a==b)` \| `a!=b` \| `a===b` | `\|\|false\|\|` | `\|vBvBvAvA\|false\|\|` | `vBvBvAvA\|vBvBvAvA\|false\|vBvBvAvA\|` | `\|vBvBvAvA\|false\|\|` (= base) |
| `o==null` \| `o==undefined` \| `0==o` \| `o!="x"` | `\|\|vO\|vO\|` | `\|\|\|\|` | — | `\|\|vO\|vO\|` (= node; base wrong on two slots) |
| `defineProperty(b,"arguments",{value:7}); b.arguments` | `7` | `7` | `TypeError` | `7` |
| `(()=>{}).caller` | `TypeError` | `undefined` | `TypeError` | `TypeError` (win kept) |

**Ship gate: never worse than base on host, standalone and wasi — met.** 17 rows claimed, **0 given up**. No `env::` import on standalone or wasi.

## Merge with main

One conflict, in `src/codegen/binary-ops-in.ts`, additive on both sides: this lane adds `receiverIsArrowFunctionValue` + `arrowBindingNeverGainsProperties`, while main's [#5292](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5292-in-operator-runtime-key-misses-prototype-methods) adds `dynamicKeyMethodNames`. Different functions, no shared logic — both kept verbatim. Verified by running main's own `in`-operator suites against the resolution: **47 tests across 3 files, all pass**.

## Validation on the integrated tree

- TS7 typecheck clean.
- All five ratchet gates green, run bare and chained, against **both** the merge-base and `origin/main` as CI's base. Growth granted in the issue frontmatter with dated rationale; `scripts/*-baseline.json` untouched.
- Targeted suites 47/47 (`issue-5270`, `in-operator-prototype-methods`, `in-operator-edge-cases`).
- Equivalence gate green across all 8 shards: **24 failing / 1,718 passing, all 24 in the baseline**.
- A 193-row regression control over the families the fixes touch.

## Known, recorded, not fixed here

1. **The `==` lowering is still wrong.** `objA == objB` runs ToPrimitive on both operands wherever the statement is reached by other means (an `if` wrapper, an assignment position). That is base behaviour, not a regression — this PR only stops the wave from newly exposing it at module scope. §7.2.15's Object-vs-Object early exit belongs in the loose-equality lowering that every `==` goes through, and changing that hot path inside a hold-fix, with no room to measure, was not justifiable. Recorded in the issue file as a follow-up.
2. **A js-host module-init sink** (R3-F3, confirmed): a poison-pill TypeError raised inside a module-scope IIFE that catches it still terminates `__module_init`, so later top-level statements never run and the file vacuous-passes. Pre-dates this wave (base does the same for a `"use strict"` function); this wave only widened the trigger population, and fixing F2 removes the arrow entry. Described in the issue file for its own follow-up, with the suggestion to pair it with a runner-side check that `__module_init` ran to completion.
3. The F2 custodial-use scan is name-based over the whole `SourceFile`, so under the test262 harness a short-named arrow binding can collide with harness identifiers and lose the fold, falling back to base's answer. Conservative in the safe direction; the same weakness the accepted sibling `in` fold carries.

## CLA

---


ttraenkler:
**Park diagnosed and fixed in 521c3a20a1; lifting `hold` for the one re-enqueue.**

**Failing check:** `merge shard reports` → *Standalone pass-count high-water floor (#2097)*: merged state 34,508 vs floor 34,721 (mark 34,771). `main` is healthy — its own baseline is 34,833 and #5540's full `merge_group` run passed the same floor at 17:32 — so the ~325-row drop was this PR's.

**Root cause** (A/B-confirmed against a `git archive` of `origin/main`): step 1.3 promoted the closure trampoline's forwarding `call` to `return_call` at **mint time**. At mint time an async function's registered type is still the `() -> ()` placeholder that `rewriteFuncResultType` only rewrites to `() -> externref` after the body compiles, so the guard compared against a stale type and the validator rejected the module:

```
language/statements/for-await-of/async-func-dstr-const-ary-ptrn-elision-exhausted.js
  base: pass
  lane: CompileError: Compiling function "__fn_tramp_fn_28" failed: return_call: tail call type error
```

The family is `for-await-of` / async destructuring — **1,554 baseline-passing standalone rows, all ES2017+**, so none was in any list this lane or its three adversarial reviews measured, and the equivalence gate has no async rows.

**Fix:** the trampoline is emitted with a plain `call` and recorded; `promoteTrampolineTailCalls` runs last in `generateModule` — after every pass that can retype a function or edit a body — and rewrites that one trailing opcode to `return_call` only when the callee's result list is *identical* to the trampoline's own declared results (same kind, same `typeIdx`, `ref` vs `ref null` distinguished). The guard is the Wasm tail-call typing rule itself; no per-shape exclusion list.

**Verified on the fixed tree** (node as oracle, base = clean `main`):

| check | result |
| --- | --- |
| both repro rows | compile_error → pass; base still pass |
| all 1,242 baseline-passing rows in `for-await-of/` + `expressions/async-generator/` | 1,240 pass; the other 2 are early-error negatives that fail identically on base (local-runner artifact) |
| the 18 baseline-passing `tco-*` rows step 1.3 was for | 18/18 — the promotion still lands where valid |
| host lane, 3 closure-heavy programs, sha256 of emitted binaries | byte-identical lane vs base |
| TS7 typecheck; five ratchet gates; both budget gates vs `origin/main` | green |
| `issue-5270`, `tail-call-optimization`, `in-operator-*` suites | 54/54 |

Full write-up, including why three review rounds missed it, is in the issue file under *Merge-queue park (2026-09-03)*.

---

PR #5579, opened 2026-09-04
fix(#5303): forward a read-only capture's value, not the declaring frame's ref cell — moment 0/10 → 4/10

ttraenkler:
Fixes [#5303](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5303-readonly-boxed-capture-abi).

## What was broken

`moment@2.30.1` was at **0/10** on the upstream npm suite — not one admitted test
ran, because all six compiled modules were rejected outright by
`WebAssembly.compile`:

```
CompileError: Compiling function #287:"createUTC" failed:
  call[13] expected type (ref null 92), found if of type (ref null 84)
```

A nested `function` declaration is lifted with its captures as leading
parameters. When a sibling mutates one of those bindings, the declaring frame
boxes its slot into a `__boxed_<name>` ref cell so writes propagate. A
**read-only** capturer still wants the VALUE — the cell belongs to the frame that
needed write-through, not to a read-only capturer's ABI — so both forwarding
sites already unwrapped it with a `struct.get`:

- the direct-call capture prepend (`src/codegen/expressions/call-identifier.ts`)
- the closure-reification prepend (`src/codegen/closures/funcref-as-closure.ts`)

Both decided *"the consumer wants the value"* by **proxy**: is the expected type a
non-reference (`f64` / `i32` / `externref`)? That answers "no" for a read-only
capture whose own value type is a GC reference. moment's `isoDates` / `isoTimes`
are arrays of arrays — `(ref $vec-of-vec)` — so the unwrap never fired and the
cell was forwarded where the value belonged. Two symptoms from one mistake:

1. **closure reification** passed the raw cell → `RuntimeError: illegal cast` in
   `__module_init`;
2. **the direct call** "coerced" cell→value with a guarded `ref.test`/`ref.cast`
   `if` that can only take the `else` arm and yield `null` — already a wrong
   answer, and a hard validation failure once the callee's reserved signature was
   rewritten to the cell *after* the caller was compiled.

### Why the signature drifts

`hoistFunctionDeclarations` phase 0 publishes every capturing sibling's lifted
signature **before** any sibling body compiles, and every call emitted afterwards
is bound to it. That contract is already enforced for the capture **set**
(`collectPromotedPreRegisteredSlots`) and the capture **order**
(`reorderToPreRegisteredAbi`) — but not for the capture **type**. In moment:
phase 0 reserves `createLocalOrUTC` with the raw array; a later sibling's lift
boxes `isoDates` in the declaring frame; `createUTC` emits its call against the
stale raw signature; `createLocalOrUTC`'s real lift then publishes the cell and
overwrites the reserved entry in place.

Only the read-only arm can drift — a mutable capture's parameter is the memoized
`getOrRegisterRefCellType(valueType)` before and after the promotion.

### Why nothing caught it

`fixCallArgTypesInBody` (`stack-balance.ts`) is the post-pass that repairs a
drifted call argument, and it *did* repair the `externref` captures in that same
call. It cannot reach these: its backward walk `break`s at any structured `if`,
and a ref→ref capture coercion is emitted as exactly that. V8 names the producer
— `found **if** of type` — which is the tell.

## The fix

One question, asked directly, in one shared helper — `expectsBoxedCaptureValue`
(`src/codegen/closures/capture-source-slot.ts`): the consumer wants the value
exactly when its expected type **is** the box's inner value type. A consumer that
genuinely wants the cell names `refCellTypeIdx`, which is never its own field's
type index, so that arm is untouched. Both forwarding sites now use it in place
of the non-reference proxy.

Plus the ABI-stability half — `preRegisteredValueForwarding`
(`src/codegen/statements/nested-declarations.ts`): a read-only capture whose
declaring slot was boxed *after* phase-0 pre-registration keeps its
pre-registered value type in the lifted signature, so the reserved entry never
drifts under already-compiled callers. That is the third rule of the trio the
file already implements for the capture set and order, and it removes the drift
at its source rather than repairing the symptom downstream.

## Evidence

**Regression test** — `tests/issue-5303-readonly-boxed-capture-abi.test.ts`.
Untyped `.js` fixture behind a two-file project (`mod.js` + `entry.ts`);
annotating the receiver `: any` routes to a different arm and the test then
passes identically either way.

| | parent `68246a740c` | with fix |
| --- | --- | --- |
| `issue-5303-readonly-boxed-capture-abi.test.ts` | **1 failed \| 1 passed** | **2 passed** |

The failing case traps `RuntimeError: illegal cast` in `__fn_tramp_caller_2`.

**moment: 0/10 → 4/10.** All six modules validate and `__module_init` completes;
on the parent none of them did.

| file | parent | with fix |
| --- | --- | --- |
| `is_date.js` | 0/2 | **2/2** |
| `is_moment.js` | 0/2 | **1/2** |
| `normalize_units.js` | 0/1 | **1/1** |
| `days_in_year.js` | 0/1 | 0/1 |
| `min_max.js` | 0/2 | 0/2 |
| `mutable.js` | 0/2 | 0/2 |

**Corpus A/B at ONE head** — parent `68246a740c` checked out in a second
worktree vs this branch, same runner, each arm with its own private pinned
checkout cache. All 17 suites exited 0; compared per test file:

| package | base | fix |
| --- | --- | --- |
| webpack | 16/16 | 16/16 |
| three | 17/18 | 17/18 |
| clsx | 32/32 | 32/32 |
| lodash | 53/62 | 53/62 |
| redux | 60/82 | 60/82 |
| axios | 108/231 | 108/231 |
| jsdom | 6/6 | 6/6 |
| uuid | 10 files, identical | 10 files, identical |
| marked | 0/30 | 0/30 |
| **moment** | **0/10** | **4/10** |
| prettier | 51/151 | 51/151 |
| jest | 299/356 | 299/356 |
| hono | 4 files, identical | 4 files, identical |

`uuid` and `hono` print no `admitted` headline; both were compared per test file
and every line is identical. **moment is the only package that moved, in either
direction.**

**Pre-existing failures, verified against the parent.** Three closure/capture
tests fail identically on `68246a740c` and on this branch — same three, same
assertions: `issue-1528-closure-construct` (generator-method construct-bridge
routing), `issue-1712-capture-closure-dispatch` (acorn-shape node graph),
`issue-2029-tagged-template-capture-local-index` (forward capturing sibling).

## Not this defect

The remaining 6 moment failures share one *different* cause:
`RuntimeError: dereferencing a null pointer` in `prepareConfig`
(`moment.js:3030-3044`, reached from `createLocalOrUTC` → `createLocal`). It was
unobservable before this fix because the module never validated. Left open
deliberately.

## Relationship to the eager dominating box

Disjoint from `canBoxBindingInDominatingParent` (`arrow-phases.ts`), and
deliberately so. That predicate decides **where** a capture's canonical box is
created and is gated on `isMutable && !alreadyBoxed`. This change never creates
or moves a box; it decides **which representation** — cell or value — a
`!cap.mutable` consumer receives from a box that already exists. The two are on
opposite sides of `mutable`, so they cannot disagree about where a capture's
canonical box lives.

## Gates

`check:oracle-ratchet` · `check:dead-exports` — all green. The two budget
allowances are granted in this PR's own issue file with dated rationale; no
baseline JSON was touched.


ttraenkler:
Re-verified after merging `upstream/main` (286 commits, up to `4d4e5dec0e`) into the branch:

- **The fix is still needed.** Reverting only the four changed source files to `upstream/main`'s versions — upstream has not touched any of them, so my hunks are the only difference — the regression test is **1 failed | 1 passed** (`RuntimeError: illegal cast`). Restoring them: **2 passed**.
- **moment still 4/10** on the merged tree (`is_date` 2/2, `normalize_units` 1/1, `is_moment` 1/2), same as the pre-merge measurement.
- All five source-ratchet gates re-run green post-merge: `check-loc-budget`, `check-func-budget`, `check-coercion-sites`, `check:oracle-ratchet`, `check:dead-exports`.

The corpus A/B in the description was measured at the original parent `68246a740c` with both arms at that one head.

PR #5580, opened 2026-09-04
fix(#5302): roll back the nested-fn closure memo slot with its local

ttraenkler:
## What

A rolled-back speculative compile leaked the per-activation **memo local** for a
capture-carrying nested function declaration, producing **invalid wasm**.

`emitMemoizedNestedFnClosure` gives each such declaration one memo local, typed
`ref_null <closureStruct>`, so every reference yields the same closure instance
(JS `f === f`), and caches its slot in `fctx.nestedFnClosureMemos`.
`snapshotLocals` / `restoreLocals` — the transactional unwind every probe goes
through (#1919) — restored `localMap` (#2029), `boxedCaptures`, the TDZ maps
(#3032) and the direct-eval cells, but **not that map**.

So a probe that allocated the memo local and rolled back had the slot truncated
out of `fctx.locals` while the map kept pointing at it. The slot got
re-allocated at an unrelated type, and the committed re-compile took the
cache-hit branch and baked `local.get` / `local.set <stale slot>` at the
closure-struct type:

```
CompileError: WebAssembly.compile(): Compiling function #443:"printDoc" failed:
  local.set[0] expected type (ref null 222), found ref.as_non_null of type (ref extern)
```

The trigger is a **chained** method call on `arr.map(nestedFnWithCaptures)` —
`doc.map(printDoc).filter(Boolean)` in prettier's `src/document/debug.js`. The
same program split across two statements was already valid, which is what hid
it. Same defect family as #1847 / #2029 / #3032, one map later.

Fix: `src/codegen/context/locals.ts`, +34 lines, no other source file touched.

## Evidence

**Regression test** `tests/issue-5302-nested-fn-memo-rollback.test.ts` — 12
cases: eight end-to-end compiles from untyped `.js` fixtures behind a two-file
project (a `: any` receiver routes to a different arm and passes either way),
plus four pinning the `snapshotSpeculative` / `rollbackSpeculative` contract
directly.

| arm | result |
| --- | --- |
| parent commit | **10 failed / 2 passed of 12** — every end-to-end failure is the exact `local.set[0] expected type (ref null N), found ref.as_non_null` CompileError |
| with the fix | **12 passed of 12** |

Re-verified after merging 218 commits of `upstream/main`: same 10/2 vs 12/12
split.

**A/B at one head, 17 npm upstream suites** (base arm = `locals.ts` reverted to
its pre-edit copy; sequential, same box): webpack, three, clsx, cookie, lodash,
redux, axios, stylelint, tailwindcss, jsdom, styled-components, uuid, marked,
moment, prettier, jest, hono.

**0 of 17 packages moved in either direction, and no individual test file
moved.** All 17 exited 0 in both arms and every suite that prints one printed
its `admitted …` headline, so no number is reported as "unchanged" because it
was missing.

## Prettier: 51/151 before, 51/151 after — the second blocker

Stating this plainly: **this fix is narrower than prettier's twelve
zero-scoring files.** It clears the *validation* blocker for six of them
(`doc-builders` 46, `is-empty-doc` 16, `print-doc-to-string` 3,
`strip-trailing-hardline` 2, `traverse-doc` 2, `doc-printer` 1 — 71 tests):
they go from the engine rejecting the module outright to compiling and
validating. They then hit a **distinct** blocker at module init:

```
module init: TypeError: equal is not a function
```

Localised with five controls (all two-file projects): a direct
`import { equal } from "node:assert"` works on both lanes with or without host
deps, and a namespace import of a module re-exporting a *local* function works
— but any namespace object whose exports include a **host-imported** binding
comes back null. That is `tryEmitCompiledModuleNamespaceObject` /
`namespaceFunctionExports` in `src/codegen/module-namespace-value.ts`, which
admits only functions compiled into this module or top-level `export const`
module globals and declines everything else. Prettier's `#universal/assert` is
exactly `export { equal, ok, strictEqual } from "node:assert"`, and
`createTypeCheckFunction` calls `assert.equal(...)` at module scope. Fixing it
needs a first-class function VALUE for a node-builtin import — a real feature,
deliberately not attempted here.

The other four zero files (`get-parser-plugin-by-parser-name` 11,
`get-printer-plugin-by-ast-format` 11, `massage-ast` 1, `resolve-parser` 1)
fail to **compile**, first on the deliberate `async shape not supported …
inside a try` refusal (#3587) in `src/config/prettier-config/loaders.js:24`;
two also hit `stack-balance (#2090) … operand stack underflow by 1`. The async
refusal alone blocks them, so stack-balance is not on their critical path.

Both are described precisely in the issue file rather than half-fixed.

## Gates

`check-loc-budget` (+34, allowed), `check-func-budget`, `check-coercion-sites`,
`check:oracle-ratchet`, `check:dead-exports`, `check-test-vacuity-shapes`,
`update-issues --check`, biome and prettier — all green. No baseline JSON
touched.

Issue: [#5302](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5302-nested-fn-memo-slot-survives-rollback)


ttraenkler:
Closing/reopening to re-fire the pull_request workflow events: GitHub created no pull_request check runs for the last two synchronize pushes and mergeable_state has been stuck at `unknown` for ~20 min. No content change.

PR #5614, opened 2026-09-05
fix(#5329): tuple-typed rest parameters get a carrier instead of an invalid module ✓

ttraenkler:
Closes [#5329](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5329-tuple-rest-carrier).

A tuple-typed rest parameter — `function (...args: [Error])` — is a carrier neither rest builder recognised, and both of their "anything else" fallbacks were wrong:

- `compileRestClosureArguments` fell through to `pushDefaultValue(ref …)`, i.e. `ref.null` + `ref.as_non_null`. Every direct call to such a closure trapped with **"dereferencing a null pointer"** before its body ran.
- `classifyClosureDispatchRest` answered `undefined`, which the caller could not tell apart from *"there is no rest formal"*. The closure was admitted as an ordinary arity-N entry with the formal dropped, so the arm's `call_ref` was one operand short and the **whole module** failed validation: `Compiling function #N:"__call_fn_0" failed: not enough arguments on the stack for call_ref (need 2, got 1)`.

Production witness: jest's `packages/jest-jasmine2/src/queueRunner.ts` (`const next = function (...args: [Error]) {…}` plus a `next.fail` twin) — the only module of the 34 in the jest dogfood suite that failed Wasm validation.

## Root cause

Rest parameters have several carriers and the classifier knew two and a half:

| source | lowered carrier |
| --- | --- |
| `...args: number[]` / `...args` | `{ length, data }` vec |
| `...args: T` (generic) | `externref` |
| `...args: []` | zero-field struct |
| **`...args: [Error]`** | **`__tuple_N` struct, one field per element** |

Two independent things were wrong and only fixing both is safe:

1. **No recipe for the carrier.** The tuple struct is now built positionally, one field per element, padded with that field type's missing-argument default.
2. **`undefined` was overloaded.** "No rest formal" and "a rest formal I cannot build" are different facts. The classifier now answers `{ kind: "unsupported" }` for the second, and both dispatch builders SKIP such an entry — an unbuildable carrier can no longer cost the module its validity; the worst case is a closure that is not offered on the host dispatcher.

## Change shape

The recognizer and both builders live in two new leaf modules — `src/codegen/closures/tuple-rest-carrier.ts` and `src/codegen/closures/closure-dispatch-rest.ts` — so the dispatcher-side and call-site builders agree by construction. Moving the existing classify/materialize pair out of `closure-exports.ts` takes that god-file **below** its LOC budget; the empty-tuple special case in `calls-closures.ts` folds into the same shared entry point, and three open-coded copies of the `self` type-index computation collapse into one helper. Net: no budget allowance needed for this PR.

## Evidence

- `tests/issue-5329-tuple-rest-carrier.test.ts` — **3 failed → 3 passed**. On the base arm the compile reports `success: true` and then `WebAssembly.instantiate` rejects the module.
- Probe matrix: `[number]`, `[Error, Error]`, `.apply(null, args)`, use as an `Array.prototype.map` callback, and the full queueRunner shape all go from broken to correct. Controls that were already clean stay clean: `...a: number[]`, `...a: any[]`, `...a: []`, a top-level `function f(...a: [number])`.
- 17-package dogfood A/B at one head — see the PR comment; no package regressed.

## Residuals, measured and deliberately out of scope

- **`queueRunner.test.ts` is still 0/6**, so the jest headline does not move on this change alone. With the module valid, the failure moves to a host-callback `TypeError: Cannot convert object to primitive value` raised while running the test's `jest.fn()` mocks — the same family as the `prompt`/`Replaceable` "illegal cast" bucket, not a rest-carrier problem.
- **`args.length` on a tuple reads NaN.** A property-access gap on tuple structs generally: a plain `const t: [number, number] = [1, 2]; t.length` reads NaN too, identically before and after. Asserted in the regression test so it cannot move silently.

### DIFF
--- a/plan/issues/5329-tuple-rest-carrier.md
+++ b/plan/issues/5329-tuple-rest-carrier.md
@@ -1,0 +1,95 @@
+---
+id: 5329
+title: "Tuple-typed rest parameter has no carrier: invalid module + null-deref on every call"
+status: done
+sprint: current
+created: 2026-09-05
+updated: 2026-09-05
+completed: 2026-09-05
+priority: high
+horizon: m
+feasibility: medium
+reasoning_effort: high
+task_type: bug
+area: compiler
+goal: correctness
+---
+
+## Problem
+
+A tuple-typed rest parameter — `function (...args: [Error])` — is a carrier
+neither rest builder recognised, and both of their "anything else" fallbacks
+were wrong:
+
+- **`compileRestClosureArguments`** (`src/codegen/expressions/calls-closures.ts`)
+  fell through to `pushDefaultValue(ref …)`, i.e. `ref.null` + `ref.as_non_null`.
+  Every direct call to such a closure trapped with **"dereferencing a null
+  pointer"** before its body ran.
+- **`classifyClosureDispatchRest`** (`src/codegen/closure-exports.ts`) answered
+  `undefined`, which the caller could not distinguish from *"there is no rest
+  formal"*. The closure was therefore admitted as an ordinary arity-N entry with
+  the formal dropped, so the arm's `call_ref` was one operand short and the
+  **whole module** failed validation:
+  `Compiling function #N:"__call_fn_0" failed: not enough arguments on the stack
+  for call_ref (need 2, got 1)`.
+
+Production witness: jest's `packages/jest-jasmine2/src/queueRunner.ts` —
+`const next = function (...args: [Error]) {…}` plus a `next.fail` twin. It was
+the **only** module of the 34 in the jest dogfood suite that failed Wasm
+validation.
+
+## Root cause
+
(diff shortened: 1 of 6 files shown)


ttraenkler:

| package | base | head |
| --- | --- | --- |
| webpack | 16/16 | 16/16 |
| three | 17/18 | 17/18 |
| clsx | 32/32 | 32/32 |
| lodash | 53/62 | 53/62 |
| redux | 64/82 | 64/82 |
| axios | 200/231 | 200/231 |
| jsdom | 6/6 | 6/6 |
| uuid | 75/75 | 75/75 |
| marked | 9/30 | 9/30 |
| moment | 10/10 | 10/10 |
| hono | 244/324 | 244/324 |

```
jest
```

PR #5615, opened 2026-09-05
fix(#5328): a dispatch arm returning externref into an f64 call site keeps its value ✓

ttraenkler:
Closes [#5328](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5328-dynamic-dispatch-extern-result-dropped).

A call through the identifier funcref ladder whose **expected result is `f64`** and whose **live arm's funcref returns `externref`** discarded the callee's answer and substituted the `0x7FF00000DEADC0DE` undefined sentinel. Every such call read `NaN`/`undefined` while the callee had computed the right value.

Production witness: jest's `packages/jest-config/src/stringToBytes.ts` (`export default stringToBytes;`, inferred return `number | null | undefined`, consumed through a default import). 21 of its 28 upstream unit tests read the sentinel instead of the byte count.

## Root cause

The ladder builds one `ref.test`-guarded arm per candidate signature; each arm must leave a value of the block's declared result type. When an arm's return type disagrees it consults `scalarBridgePlan`, and on `null` falls back to a **dead-arm placeholder** — `drop` + `defaultValueInstrs(expectedReturn)` — justified by "a signature-divergent arm can never actually run".

`scalarBridgePlan`'s `externref → f64` row is gated on `allowProvenNumberUnbox`, which the **return** site passed as `false`, and nothing else supplies that direction. But the arm is **not** dead: the ladder unconditionally seeds an externref-returning alternate candidate whenever the expected result is not already externref, and a function whose inferred return is `number | undefined` is lowered with exactly that signature. So the live arm called the function, dropped its answer, and returned `undefined`.

The guard is right for an **argument** — a declared-`any` argument may really be a Boolean/Symbol/BigInt and must not be silently run through the numeric ABI. It is wrong for a **result**: `expectedReturn` *is* this call expression's statically computed result ValType, so the compiler has already committed to reading whatever comes back as `f64`. Refusing the unbox does not protect the value; it destroys it.

## Change

One token at the return coercion site (`false` → `true`), plus the comment explaining why the argument-side guard does not apply. `scalarBridgePlan` still returns `null` when `__unbox_number` is unregistered, so no late import is pulled and the dead-arm "no index shift" invariant (#2174) holds byte-for-byte. Argument bridges are untouched.

## A prior diagnosis this refutes

An earlier bisection attributed this bucket to "an overload SET whose signatures disagree on the return type". Deleting the overloads and keeping the identical implementation still reproduces, so that is not the trigger; the issue file records the full ablation table.

## Blocked measurement on current `main` — please read

Between the measurement head (`4946cf70fe`) and this PR's base an unrelated
regression landed: `export default <identifier>;` in a multi-file project no
longer COMPILES at all
(`multi-prepared-module-init-census:terminal-join`). Filed as
[#5332](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5332-module-init-census-export-default-identifier),
whose issue file is included here because this PR's test points at it. #5332
costs the jest suite 6 tests by itself (`stringToBytes.test.ts` 6/28 -> 0/28 on
current main) and MASKS this fix, because the shape that reproduces #5328 is
exactly the shape #5332 refuses to compile. Every alternative spelling was
measured — `export { g as default }`, inline `export default function`, a named
export, a factory return, a callback parameter — and none of them reproduce, so
there is no unblocked fixture to substitute.

The regression test therefore DETECTS that compile error and skips with a
pointer to #5332, and starts enforcing the moment #5332 lifts. Reviewers who
want to see the fix work can check out `4946cf70fe`, apply this diff, and run
`node --import tsx tests/dogfood/jest-upstream-suite.mjs` — 299/356 -> 320/356.

## Evidence

- jest dogfood suite **299/356 → 320/356**; `stringToBytes.test.ts` 6/28 → 27/28 and no other file in the suite moved.
- `tests/issue-5328-dynamic-dispatch-extern-result.test.ts` — **1 failed / 1 passed → 2 passed**. Untyped `.js` in a two-file project on purpose: an explicit return annotation routes the call through a direct arm and the test then passes identically with and without the fix.
- 17-package dogfood A/B at one head (webpack, three, clsx, cookie, lodash, redux, axios, stylelint, tailwindcss, jsdom, styled-components, uuid, marked, moment, prettier, jest, hono): see the PR comment — no package regressed.
- Budget/ratchet gates green. The +7 lines are the comment on a one-token change, granted in the issue file's frontmatter with a dated rationale.

### DIFF
--- a/plan/issues/5328-dynamic-dispatch-extern-result-dropped.md
+++ b/plan/issues/5328-dynamic-dispatch-extern-result-dropped.md
@@ -1,0 +1,137 @@
+---
+id: 5328
+title: "Dynamic-dispatch arm returning externref into an f64 call site drops the result"
+status: done
+sprint: current
+created: 2026-09-05
+updated: 2026-09-05
+completed: 2026-09-05
+priority: high
+horizon: m
+feasibility: medium
+reasoning_effort: high
+task_type: bug
+area: compiler
+goal: correctness
+# 2026-09-05: the code change is ONE token (`false` -> `true` at the return
+# coercion site); the +7 lines are the comment explaining why the argument-side
+# guard does not apply to a result. There is nothing to extract into a subsystem
+# module, and a one-token change to a dispatch ladder with no rationale beside it
+# is exactly what makes the next reader revert it.
+loc-budget-allow:
+  - src/codegen/expressions/call-identifier.ts
+func-budget-allow:
+  - src/codegen/expressions/call-identifier.ts::compileIdentifierCall
+---
+
+## Problem
+
+A call through the identifier funcref ladder (`src/codegen/expressions/call-identifier.ts`)
+whose **expected result is `f64`** and whose **live arm's funcref returns `externref`**
+discarded the callee's answer and substituted the `0x7FF00000DEADC0DE` undefined
+sentinel. Every such call read `NaN`/`undefined` while the callee had computed
+the right value.
+
+Production witness: jest's `packages/jest-config/src/stringToBytes.ts` —
+`export default stringToBytes;` (the statement form), inferred return
+`number | null | undefined`, consumed through a default import. 21 of its 28
+upstream unit tests read the sentinel instead of the byte count.
+
+## Root cause
+
+The ladder builds one `ref.test`-guarded arm per candidate signature; each arm
+must leave a value of the block's declared result type. When an arm's return
+type disagrees it consults `scalarBridgePlan`, and on `null` falls back to a
+**dead-arm placeholder** — `drop` + `defaultValueInstrs(expectedReturn)` — whose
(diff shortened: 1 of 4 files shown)


ttraenkler:

| package | base | head |
| --- | --- | --- |
| webpack | 16/16 | 16/16 |
| three | 17/18 | 17/18 |
| clsx | 32/32 | 32/32 |
| lodash | 53/62 | 53/62 |
| redux | 64/82 | 64/82 |
| axios | 200/231 | 200/231 |
| jsdom | 6/6 | 6/6 |
| uuid | 75/75 | 75/75 |
| marked | 9/30 | 9/30 |
| moment | 10/10 | 10/10 |
| hono | 244/324 | 244/324 |

```
jest
```

PR #5620, opened 2026-09-05
fix(#5333): P0 REGRESSION — main emits invalid Wasm; nested-capture plan must publish the LIVE carrier

ttraenkler:
## P0 — `main` emits invalid Wasm

`main` compiles several npm packages to modules that build (`compile.success: true`) and
are then **rejected by `WebAssembly.compile`**:

```
CompileError: WebAssembly.compile(): Compiling function #721:"__closure_47" failed:
  call[25] expected type (ref null 84), found struct.get of type i32 @+437566
```

**moment: 10/10 → 0/10, `compile.validated` 6/6 → 0/6.**

> Reading tip that cost several agents time: the per-test `wasmError` is `null` and
> `compile.details[N].errors` is `[]`. The message lives in
> **`compile.details[N].validationError`**. An empty `errors` array is not "no error".

## Culprit: PR #5390, commit `82be803ac7` — **not** PR #5598

Bisected on moment with `compile.validated == 0` as the signal (~2 min/point):

| commit | what | moment |
| ------ | ---- | ------ |
| `4946cf70fe` | pre-window main | 6/6 validated, 10/10 |
| `88bdfda594` | #5585 | 6/6, 10/10 |
| `b08dd4589c` | #5503 — **first parent of the culprit** | **6/6, 10/10** |
| `470ceba797` | **#5390 `codex/1058-typescript-binder`** | **0/6, 0/10** |
| `944643dcde` | #5598 | 0/6, 0/10 |
| `104dc660fb` | main at report time | 0/6, 0/10 |

The window *before* #5598 is already broken, with the identical error at identical byte
offsets — so **#5598 is not the cause of this defect**. (It is very likely the cause of a
*second* one; see below.)

File-level A/B on current main isolates it further: restoring **only**
`src/codegen/statements/nested-declarations.ts` to its pre-#5390 content returns moment to
6/6 and 10/10, and reverse-applying #5390's other two commits in that file does not.

> Do not bisect inside #5390's branch commits — they sit on a ~4-day-older base that is
> itself bad for moment, so every branch point measures BAD regardless of content. Only
> merge-vs-first-parent is sound here.

## Mechanism

`82be803ac7` added `canonicalReservedCapturePlan`. On the real compile it **replaced** the
freshly observed capture plan with the Phase-0 one, refreshing only `localIdx` /
`tdzFlagIdx`. Between Phase 0 and the real compile the declaring frame can **box** a
binding, so the merged entry claimed *"unboxed value at local N"* while local N held a ref
cell.

The guard directly above the merge cannot catch this: both worlds agree on the two
published vectors *by construction* (`nestedCaptureValueParamType` boxes the raw carrier
into the same cell; `nestedCaptureMetadataValueType` unwraps the cell to the same inner
type). They are two spellings of one ABI, so nothing "changes" and nothing throws — but
consumers keyed on `alreadyBoxed` / `forwardUnboxed` (#5303's value-forwarding) then
disagree with the frames that actually hold the value, and a capturing closure forwards a
raw `struct.get` field where the lifted function's ref-cell param is expected.

Measured on one moment module: **70 declarations diverge, 164 captures flip
`alreadyBoxed`**. Source-mapping the failing offset lands on `moment.js:3665` — the closure
`createAdder` returns, calling the later-declared `addSubtract`.

## Fix (3 lines of behaviour)

Keep the canonical plan as the authority for **name, membership and ORDER** — that vector
is what an earlier sibling baked into its direct call or cached closure artifact. Take the
**live entry's carrier** whenever the binding is still physically observed; only a
promotion-only capture keeps its Phase-0 record.

## Evidence

| tree | moment | prettier |
| ---- | ------ | -------- |
| `4946cf70fe` (pre-regression) | 6/6 validated, 10/10 | 61/151 (reported) |
| main | 0/6, 0/10 | 2/151 (reported) |
| **main + this fix** | **6/6 validated, 10/10** | 3/16 validated, 2/151 |
| **merged result** (branch + `upstream/main` @ `82394ba491`) | **6/6 validated, 10/10** | — |

- `node scripts/equivalence-gate.mjs`: `24 failing, 1718 passing, 24 known-failures` — **no
  new equivalence regressions**.
- All **13** of #5390's own tests still pass (`issue-1058-reserved-sibling-capture-abi` 7,
  `issue-1058-function-type-branding` 1, `issue-1058-stack-balance-dag` 5), plus
  `issue-5303-readonly-boxed-capture-abi` — the freeze's intent (a sibling-visible capture
  ABI that cannot silently change shape or order) is preserved.

## prettier does not recover — it has a second, independent regression

Not invalid Wasm but a hard codegen error, so 13 of prettier's 16 modules never produce a
binary at all:

```
Codegen error: multi-prepared-module-init-census:terminal-join: executable source
ir-source:v1:…src%2Fcommon%2Fast-path.js lost its exact module-init terminal
```

This is **[#5332](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5332-module-init-census-export-default-identifier)**,
which landed on `main` while this PR was in CI: `export default <identifier>;` in a
dependency of a multi-file project, root-caused there to the #3525 census work
(`2c18cd7a6f`, PR #5598). Thirteen of prettier's sixteen modules hit it. This PR adds the
measured prettier cost to that issue's `## Cost, measured` section — prettier was 61/151
before the window, so **#5332 accounts for the entire residual** and #5333 for none of it.

## No minimal regression test

The divergence path reduces easily (a nested declaration pre-registered by an earlier
sibling whose captured `let` is boxed before the real compile), but **every** reduced
fixture still emits a valid module — seven shapes tried, under both the `standalone` and
the dogfood `compileProject`/web/IR lanes, listed in the issue file. The guard here is the
dogfood suite.

**Worth a follow-up:** main shipped invalid Wasm through five merges with all six required
checks green. `tests/dogfood/moment-upstream-suite.test.ts` runs its heavy arm only under
`DOGFOOD_MOMENT_UPSTREAM_SUITE=1` and asserts nothing about `compile.validated`, so nothing
in CI observed a 10/10 → 0/10 collapse.

Issue: [#5333](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5333-nested-capture-abi-live-carrier)
· cc the author of #5390.


ttraenkler:
### The `Differential test` red on this PR is a misattribution — please do not park it

The merge-group run reports:

```
## ❌ New regressions (delta gate FAILED)
  - closures/06-nested.js: match → mismatch
```

That file fails **identically without this PR**. Measured three ways:

| tree | `closures/06-nested.js` |
| --- | --- |
| this branch (fix applied) | `0` |
| this branch with `nested-declarations.ts` reverted to main's version | `0` |
| clean `main` tip `3879df539c`, this PR absent | `0` |
| `4946cf70fe` (pre-dates the whole #5333 window) | `0` |

So it is not caused by this change. It was **bisected to PR #5450** (`e7b0668b0d`, `perf(codegen): skip module-init pass 2 for closure-free call-bearing populations`, 2026-09-02) — parent `aaebad2ae1` prints `3`, that merge prints `0` — and confirmed on current `main` with that PR's own seam: `JS2WASM_TEST_FORCE_MODULE_INIT_PASS2=1` restores `3`.

Filed with the full bisect and root cause as **#5335** / PR #5623.

Why it points here: `benchmarks/results/diff-test-baseline.json` was last refreshed **2026-07-19** and still records this file as `match` with output `3`, so the delta gate blames whichever PR is currently in the merge queue. `Differential test` is also **not** a required check, which is why it has been red since 2026-09-02 without stopping anything.

The six required checks on this PR are green and the merge-group `CI` and `CLA Check` jobs passed.

PR #5642, opened 2026-09-06
fix(#5334): rest-param callable values dispatch by runtime disambiguation, not wrapper metadata ✓

ttraenkler:
Closes [#5334](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5334-rest-param-wrapper-metadata-illegal-cast).

## Problem

A compiled `function spy(...args)` reached through a typed callable slot had its trailing rest vec marshalled as an ordinary positional formal: call argument 0 (a string) was `ref.cast` to `$__vec_externref` and the call trapped `illegal cast` (jest-watcher `prompt.test.ts` 0/4, jest-matcher-utils `Replaceable.test.ts` 15/17). `ctx.__restFuncTypeIdxs`, the registry the ladder consulted, had no writer — and a per-funcref flag cannot be right anyway: `spy(...args)` and `g(xs: any[])` lift to the same funcref type, and as captureless declarations share the same wrapper struct.

## Fix — decide at runtime, in two layers

New subsystem module `src/codegen/expressions/callable-rest-bridge.ts`:

1. **Closure identity, where structurally provable.** WasmGC canonicalizes by structure (measured with a hand-written module: identical sibling subtypes `ref.test` as one type; one extra field makes them distinct), so only marked structs count as guards — #4616's `__rest_fn_wrap` f64 marker and capture structs. A guard hit packs every argument from the slot on. `ensureFuncValueWrappersRegistered` now pre-registers the marker subtype for every rest declaration used as a value (the one legitimate writer, per struct — `__restFuncTypeIdxs` is deleted). This is what gets `spy(["ab", "c"])` right; no value test can.
2. **The value's shape, otherwise.** `ref.test` the erased argument against the vec carrier: hit → fixed reading (a real `g(xs)` call), miss → pack. Scalars pack without a test; a call with no argument packs an empty vec (`onCancel()`); an array-typed slot keeps its projection unless a guard proves rest. Plus an eager pure-rest wrapper so a spy compiled *after* the library (jest's `vi.fn()`) has an arm at all.

Both the identifier ladder (`call-identifier.ts`) and the callable-property ladder (`calls-closures.ts` — `this._onSuccess(v)` / `this._onCancel()`, which jest's prompt tests 2–3 use) take the bridge. Standalone keeps its flag-driven reading; every emitted op is pure Wasm (dead-arm safe).

## Evidence

- `tests/issue-5334-rest-param-callable-runtime-bridge.test.ts` (typed `dep.ts` + untyped `main.js`, values pinned): parent **6 failed / 1 passed** (the one pass is the anti-vacuity control, which must pass on both sides), fix **7/7**. Includes the anti-vacuity control `g(xs: any[])` sharing spy's funcref (202 on both).
- A/B at one HEAD (`2257b950ee`), 17 suites, one at a time, per test file — sixteen suites identical per file; jest +6 (the two expected files) and redux +2 (`createStore.spec.ts` 34/42 → 36/42 — two tests that dispatch `vi.fn()` listeners through redux's typed `listener()`, a zero-argument call of a rest spy that used to hit the TypeError terminal and now gets the empty-vec arm):

| suite | parent | fix | per-file movement |
| --- | --- | --- | --- |
| webpack | 16/16 | 16/16 | none |
| three | 17/18 | 17/18 | none |
| clsx | 32/32 | 32/32 | none |
| cookie | 63740/63740 | 63740/63740 | none |
| lodash | 53/62 | 53/62 | none |
| redux | 61/82 | **63/82** | `test/createStore.spec.ts` 34/42 → 36/42 |
| axios | 200/231 | 200/231 | none |
| stylelint | 108/108 | 108/108 | none |
| tailwindcss | 13/13 | 13/13 | none |
| jsdom | 6/6 | 6/6 | none |
| styled-components | 9/9 | 9/9 | none |
| uuid | 75/75 | 75/75 | none |
| marked | 9/30 | 9/30 | none |
| moment | 10/10 | 10/10 | none |
| prettier | 101/151 | 101/151 | none |
| jest | 329/356 | **335/356** | `__tests__/Replaceable.test.ts` 15/17 → 17/17; `__tests__/prompt.test.ts` 0/4 → 4/4 |
| hono | 244/324 | 244/324 | none |
- Pre-existing failures met on the way, A/B'd on the parent: `issue-3214-callable-abi` (1 test) and `issue-4616-nullish-spread-source` (2 tests) fail identically without this change.

**Rest-free programs are byte-identical to the parent.** The first CI run caught a real cost of the speculative wrappers (`equivalence-shard (2)`: the #1941 trampoline program gained an `__unwrap_for_wasm` import via `calls.ts`'s dynamic path and LinkError'd in the fixed-import harness). The pure-rest wrappers are now created only when the program contains a rest parameter (`programHasRestParameter`), and the property ladder's argument views only when a rest-reading candidate exists; three rest-free probe programs hash identically to the parent, raw and optimized.

Residuals (documented in the issue file): a typed number-literal array through an `any` slot into `g(xs: any[])` now packs (NaN) where it used to trap; same-layout capture siblings over-match the guard; the property ladder does not pack surplus arguments beyond the field's declared arity.

Budget allowances (LOC: call-identifier.ts, calls.ts, calls-closures.ts; func: the three hook functions) are granted in the issue frontmatter with the rationale: the mechanism is in the new module, the god-files carry only the hook calls.


ttraenkler:
Diagnosis of the auto-park (run 34003219105, `check for test262 regressions`), for the record — the PR merged as `4001bbe811` in the meantime:

- The 77 regressions are a Temporal cluster (63 × `unknown time zone UTC` under `Temporal/ZonedDateTime`, plus a few Temporal misc and TypedArray rows) alongside +1329 improvements. The baseline (`d58086f75d`) was 2 test262-relevant commits behind; the other one is #5644 (`link the compiled Temporal provider in the sharded lane`, #5353, touching `src/temporal-provider.ts`).
- Reproduced locally with `scripts/run-test262-paths.mts` on 12 representative rows: the same 8 fail on the parent tree (this change-set's six files swapped back to their pre-change copies) — including with the Temporal provider rebuilt by the parent compiler (`cacheHit=false`), so the provider cache is not masking anything.
- #5644's own merge-group Test262 run (34001751159) failed the same gate; PR #5645's group on top of this merge (34004812318) passed.

So the cluster is #5644's Temporal-provider drift, not [#5334](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5334-rest-param-wrapper-metadata-illegal-cast)'s dispatch change. Leaving the `hold` label as-is (the PR is merged; nothing to re-enqueue).

PR #5646, opened 2026-09-06
fix(#5343): call-tail-dispatch routes a registry miss to the dynamic ladder regardless of callSigs

ttraenkler:
## Summary

[#5343](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5343-call-tail-dispatch-typed-unmatched-fallthrough)

In `src/codegen/expressions/call-tail-dispatch.ts`, the "CallExpression as
callee" arm (`f()()`) calls `matchClosureInfoBySignature` over
`ctx.closureInfoByTypeIdx`. Before this fix, a miss there (no registered
closure matches) only fell back to the dynamic-call ladder
(`tryEmitInlineDynamicCall`) when the checker gave the inner call **no**
signature at all (`callSigs.length === 0` — the untyped `any` twin, already
repaired by an earlier fix). When the checker **did** supply a signature but
the registry still had no match, the arm fell all the way through to the
graceful tail: it evaluates the callee and arguments for side effects, drops
every value, and pushes `ref.null.extern` — a silent `undefined`, not a trap.

Confirmed in WAT: `f()(1, 3)` for `function f() { return Math.max; }`
compiled to `<callee> drop <arg> drop <arg> drop ref.null extern call
$__unbox_number` before the fix.

This is reachable by any host callee (Math.max, or any host function a
compiled function returns) or any callee whose closure hasn't been
registered yet at the point the call site compiles — including across module
boundaries, via a cycle in the compile-order DFS.

## Fix

Route every registry miss — whether from no checker signature at all, or a
signature with nothing registered to match it — to the same dynamic-call
ladder. The fast exact-match `call_ref`/`return_call_ref` arm is untouched
for every registry hit (pinned via a byte-identical WAT diff).

## Test plan

- [x] `tests/issue-5343.test.ts` — 3 cases: (a) host callee, (b) callee
      exported by a second module whose body compiles after the call site
      (via a compile-order-inverting import cycle), (c) anti-vacuity control
      pinning the fast arm's WAT byte-for-byte.
- [x] Measured exact counts both ways: parent 1 passed/2 failed, fixed 3/0.
- [x] `tests/issue-5335-module-init-pass2-closure-registry.test.ts` stays 11/11.
- [x] All ratchet gates green (`check-loc-budget` net **-6** LOC,
      `check-func-budget`, `check-coercion-sites`, `check:oracle-ratchet`,
      `check:dead-exports`, `check:dogfood-validation`).
- [x] A/B at one HEAD across 17 dogfood suites (webpack, three, clsx, cookie,
      lodash, redux, axios, stylelint, tailwindcss, jsdom,
      styled-components, uuid, marked, moment, prettier, jest, hono) — all
      match their anchors exactly, no regressions.

Full details in the issue's `## Test Results` section.

### DIFF
--- a/plan/issues/5343-call-tail-dispatch-typed-unmatched-fallthrough.md
+++ b/plan/issues/5343-call-tail-dispatch-typed-unmatched-fallthrough.md
@@ -1,10 +1,11 @@
 ---
 id: 5343
 title: "call-tail-dispatch: a callee with a checker signature but no registered closure falls through to a silent `undefined`"
-status: ready
+status: done
 sprint: current
 created: 2026-09-05
-updated: 2026-09-05
+updated: 2026-09-06
+completed: 2026-09-06
 priority: high
 horizon: s
 feasibility: medium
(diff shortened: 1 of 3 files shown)


ttraenkler:
Diagnosed the auto-park-bot hold (merge_group run [34004864222](https://github.com/loopdive/js2/actions/runs/34004864222/job/101415295704), \`check for test262 regressions\` — \`Fail on regressions\`).

Both named gate failures are pre-existing on `main`, unrelated to this PR:

1. **Bucket-limit gate** (`test/built-ins/Temporal/ZonedDateTime/prototype` 59 regressions > 50): the dominant error across the cluster is a uniform `RangeError: unknown time zone UTC` on totally unrelated methods (`equals`, `era`, `eraYear`, `getTimeZoneTransition`, `round`, `since`, `compare`) — the signature of one shared internal choke point, not 59 independent miscompiles from a 9-line, narrowly-scoped closure-dispatch fix.
2. **Trap-growth ratchet** (`null_deref` 65 → 73, #3189): all 8 newly-trapping files are `Temporal.Now.*`/`with` and were already `fail` on baseline (not `pass`).

The run's own log flags: `BASELINE DRIFT WARNING — the js2wasm-baselines JSONL is 4 test262-relevant commit(s) behind main HEAD (83m clock age)`. Between this PR's branch point (`68e1c0c2cb`) and the merge_group's base (`22a6e4d51e`)/current `main` (`d2e99e6299`), `src/temporal-provider.ts`, `src/codegen/expressions/calls-closures.ts`, `src/codegen/expressions/virtual-dispatch.ts`, and a new `src/codegen/expressions/callable-rest-bridge.ts` all landed — any of which plausibly touches Temporal/closure dispatch.

**Direct empirical confirmation** (cold-built the linked `@js-temporal/polyfill` provider via `runTest262File`, clearing `JS2WASM_TEMPORAL_CACHE` each time so no stale-cache artifact was reused per #5227):

- On this PR's branch (with the #5343 fix): `ZonedDateTime/prototype/equals/different-time-zone-not-equal.js`, `.../era/basic.js`, `ZonedDateTime/compare/argument-string-time-separators.js` → `unknown time zone UTC`; `Now/plainDateISO/return-value.js`, `Now/zonedDateTimeISO/return-value.js` → `null_deref` trap.
- On a clean detached checkout of current `upstream/main` (`d2e99e6299`, **zero PR changes**): **byte-identical failures**, same error text, same trap kind.

This PR's diff (`src/codegen/expressions/call-tail-dispatch.ts`, a call-of-call registry-miss fallback) cannot be the cause — confirmed present with the fix reverted just as with it applied, and confirmed present on plain `main`. Removing `hold` for the backstop to re-sweep; no branch changes needed.

PR #1016, opened 2026-06-01
fix(wasi): support stdout write drain shim (#1766)

ttraenkler:
## Summary

- Return `true` from the synchronous WASI lowering for `process.stdout.write(...)` and `process.stderr.write(...)`.
- Accept `process.stdout.once("drain", cb)` / `process.stderr.once("drain", cb)` as a no-op compatibility path under WASI, avoiding generic EventEmitter host imports.
- Add focused regression coverage for the Node-style backpressure guard and string-write truthiness.
- Declare the missing `needPages` local in the ArrayBuffer write helper, which the adjacent write regression suite exercises.

## Why

Issue #1766 is blocked for full async/EventEmitter stream semantics, but WASI `fd_write` is synchronous today. Returning `true` and accepting the unreachable `once("drain")` branch lets idiomatic Node backpressure guards compile without requiring full stream/event support.

## Validation

- `pnpm exec vitest run tests/issue-1766.test.ts tests/issue-1618-1651-wasi-stdout.test.ts tests/issue-1655-wasi-arraybuffer-write.test.ts`
- Pre-push hook: typecheck, lint, format check, issue integrity

Full async/EventEmitter semantics remain tracked separately by #1766/#1042/#1326/#1575.

PR #5648, opened 2026-09-06
fix(codegen): destructuring `this` on a sidecar-backed class and absent narrowed-numeric reads no longer launder undefined into NaN (#5251)

ttraenkler:
## What

Two value-fidelity defects that made the compiled `@js-temporal/polyfill`
compute **wrong numbers** — never trap — on every non-ISO calendar built from a
property bag. Neither reduction has any Temporal in it; both are general
compiled-object read defects.

**A. Destructuring `this` on a sidecar-backed class instance.** The typed arm of
`destructureParamObject` reads each property with `struct.get` and binds
`undefined` for a property the struct does not declare
([#5221](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5221-temporal-plaindate-from-null-deref)),
on the inference *"not a declared field ⇒ absent property"*. That is sound for a
closed object literal and unsound for a class instance: a class that takes any
write the shape analysis cannot place (a computed key, `Object.assign`, a write
through a helper function) lowers to
`(struct (field $__tag i32) (field $__shape_brand …))` — **no data fields at
all**, all instance state in the sidecar. In the polyfill this was
`const { anchorEra: i } = this` in `GregorianBaseHelper.estimateIsoDate`, which
bound `undefined` while `this.anchorEra`, `this[k]`, `Reflect.get` and
`{...this}` all read the real era record; the `undefined` reached
`n + i.isoEpoch.year` as NaN/Infinity. Fixed by routing the pattern through the
dynamic read when a **class** struct does not declare it — the rule the
externref arm of the same function has carried since #1016.

**B. An absent numeric property read as NaN-the-number.** The Phase-3 narrowing
(#1269) collapses a **dynamic** property read to `f64` when every struct
candidate's field is `f64`. The dispatcher's terminal correctly answers
`undefined` for an absent property, f64 cannot hold it, so the unbox laundered
it into plain NaN and the caller re-boxed that as a **number** (`typeof`
"number", `!== undefined`). One numeric literal anywhere in the module arms it,
the way
[#5250](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5250-temporal-error-semantics-mismatches)'s
`__sget_month` collider did. Fixed by branding the narrowed f64
`undefSentinel: true` and adding the inverse of the sentinel-aware **boxing**
arm that already existed (#2864/#2979). Numeric consumers are untouched — they
read a plain f64, and NaN is the correct `ToNumber(undefined)`.

Issue:
[#5251](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5251-temporal-numeric-calendar-seam-families)

**Land order: stacked after
[#5352](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5352-open-receiver-static-bind-on-cascade-bailout)
(PR #5639) and after
[#5250](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5250-temporal-error-semantics-mismatches)
(PR #5641).**
This branch is built on `issue-5250-temporal-error-semantics` +
`issue-5352-open-receiver-static-bind` + `docs-5352-5353` + `origin/main`, so
it must not be enqueued before those land.

## Measured

`family-123.txt` (the 123-row #5249 calendar list) through the #4628 linked
provider, **fresh provider cache directory per compiler revision** — the cache
keys on polyfill source + options, not on the compiler, so a shared directory
silently serves a stale provider:

| run | pass | fail |
| --- | --- | --- |
| stacked base (post-#5250, post-#5352) | 1 | 122 |
| + fix A | 4 | 119 |
| + fix A and B | 4 | 119 |

**0 rows regressed at either step.** The pass count understates it: **112 of the
123 rows advanced past their previous failure.** `infinity is out of range` on
the calendar path went 45 → 0, `Invalid ISO date` with a NaN year 21 → 0, and
`invalid number value` (which fix A surfaced) 17 → 0.

Against node-on-polyfill with the same pinned bundle:
`PlainDate.from({…, calendar})` now matches node exactly for gregory, japanese
and roc, and `.until()` / `.since()` return `P1Y` / `P1Y` / `P366D` / `P1Y` —
byte-identical to the control.

`#5250` did **not** move these rows: the stacked base measured 1/123 with the
same bucket shape as the PR #5639 measurement.

## Residual families — attributed, not fixed

Full evidence in the issue file. Bounds, so nobody re-derives them:

- **66 rows — `Intl.DateTimeFormat` is a shell**
  (follow-up to [#5206](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5206-intl-global-missing), filed as [#5355](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5355-intl-datetimeformat-shell)).
  `getCalendarParts` swallows the real error in a `catch` that rethrows
  `Invalid ISO date`. Unwrapped it is
  `TypeError: invalid receiver: …`. Measured from the **consumer**, no polyfill
  involved: `new Intl.DateTimeFormat("en-US", …)` yields a value that is
  `=== undefined` while `typeof` says `"object"`, and `.format()` /
  `.formatToParts()` return `undefined`. Only the calendars whose
  `isoToCalendarDate` is pure arithmetic (gregory, japanese, roc) can pass
  without real Intl; buddhist, indian, ethiopic and coptic need ICU data.
- **32 rows — `instanceof` against a linked-provider class is always false.**
  New defect, filed as [#5354](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5354-linked-class-instanceof-prototype-identity). It fails for instances the
  consumer constructs itself (`new Temporal.PlainDate(1997,12,1) instanceof
  Temporal.PlainDate === false`, `getPrototypeOf(d) === PlainDate.prototype ===
  false`, `d.constructor === undefined`), so it is class-VALUE prototype
  identity crossing the module boundary. `TemporalHelpers.assertPlainDate` opens
  with an `instanceof`, which is why exactly this many rows stop there.
- **9 rows — `infinity is out of range`, ZonedDateTime only.** A separate lane;
  surfaced only once the calendar path was unblocked. Not probed.

Also reported rather than left silent: a **direct** `typeof o.inc` on an absent
narrowed-numeric property still answers `"number"` (binding to a local first is
correct after this PR) — asserted-as-known in the test file. No measured row
depends on it.

## Tests

`tests/issue-5251-temporal-value-seam.test.ts` — 10 tests, each verified to fail
on base via a file-copy A/B of the three touched sources. Includes the
#5221-unchanged case (a genuinely absent property on a closed literal stays
`undefined`) and both directions of the sentinel (a present numeric field keeps
its value; a genuine computed NaN still boxes as a number — the #3315 hazard).

## Verification

- Gates run bare: `typecheck`, `lint`, `check:loc-budget` (also with
  `LOC_GATE_BASE=$(git rev-parse origin/main)`), `check:func-budget` (same),
  `check:coercion-sites`, `check:oracle-ratchet`, `check:dead-exports`,
  `check:host-import-policy`, `update-issues.mjs --check` — all exit 0.
- Suites, one vitest process each, all green: `issue-5250-sget-numeric-shape-miss`,
  `issue-5352-open-receiver-static-bind`, `issue-5249-open-receiver-descendant-dispatch`,
  the nine provider suites (`5221`, `5225`, `5226`, `5237`, `5239`, `5241`,
  `5242`, `5244`, `5248`), `issue-4628-temporal-global`, and the new
  `issue-5251-temporal-value-seam`.
- `node scripts/equivalence-gate.mjs` — **24 failing, 1718 passing, 24 known
  failures in baseline, no new regressions** (exit 0).
- Collateral for the property-read change, base-vs-fix per row on the same
  113 test262 rows (`built-ins/Reflect/{get,has,ownKeys,getOwnPropertyDescriptor}`
  + `built-ins/JSON/stringify`): **75 pass / 38 fail on BOTH sides, 0 flips,
  0 reason changes.**
- Growth allowances are in this PR's own issue-file frontmatter with a dated
  rationale, and **restate** the `src/runtime.ts` / `src/ir/from-ast.ts` grants
  from the predecessor branches so they are not stranded when CI scores the
  merge preview.

---


ttraenkler:
Park diagnosis (run 34008006018, job 101421143416) — both gate failures are baseline drift from the #5353 lane switch, not this PR's regressions:

- The group scored against the last **unlinked** baseline (5 commits behind). The 59 `Temporal/ZonedDateTime/prototype` pass→fail flips and the 8 `null_deref` rows (7 × `Temporal/Now/*`, `PlainDateTime/prototype/with/calendar-temporal-object-throws`) are rows whose old status depended on `Temporal` being absent — the wrong-reason-pass class PR #5375 predicted.
- Main's promote at 06:35 UTC published the first provider-linked baseline (`test262-current.jsonl`: 37,074 pass, 0 × `Temporal is not defined`). In it, `ZonedDateTime/prototype` is 652 fail / 88 pass and the 8 rows are already `null_deref` — so there is nothing left for the ratchet or the >50 bucket gate to flag once this PR is scored against it.
- Net for the group was +1370 (35,740 → 37,110) with 1,418 improvements.

Merged `origin/main` (bbd73894d0), fast gates + the #5251 suite green. Removing the bot hold once so `auto-enqueue` re-takes it alone.

---

PR #5654, opened 2026-09-06
fix(#5342): a host builtin in an object-literal callable property no longer traps ✓

ttraenkler:
Cause **A** of three for [#5342](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5342-lodash-residual-nine). Siblings: cause B (`eqref` carrier) and cause C (module-scope capability probe). Each is independently correct and independently revertable; **lodash reaches 58/62 only when all three land**.

## The defect

```js
var o = { f: Array.isArray };
o.f([1, 2, 3]);        // RuntimeError: dereferencing a null pointer
```

`compileCallablePropertyCall`'s externref branch reads the field, does `any.convert_extern`, and guard-casts to the closure-wrapper root. For a genuine host function that `ref.test` fails, so the cast yields `ref.null` — and `emitNullCheckThrow` deliberately rethrows **only** when the value was nullish *before* the cast (#789: a wrong struct type is meant to fall through to a multi-struct dispatch). A live host function is not nullish, so the null cast reached `struct.get` and trapped. A wasm trap is not catchable by wasm exception handling, so this killed lodash's whole single test file.

lodash publishes `var isArray = Array.isArray; module.exports = isArray`, the adapter does `import isArray from 'lodash/isArray.js'` and `const _ = { …, isArray, … }`, and `_.isArray([1,2,3])` is the trap.

## The fix

`callablePropertyIsExtractedHostBuiltin` already existed for exactly this trap — its own comment describes it — but proved one shape only: a shorthand whose value is a binding element of `const { isArray } = Array`. `valueDeclarationOf` stops at the import clause, so lodash's shape was invisible.

The proof now follows import aliases (new oracle query `aliasedValueDeclarationOf`) and single-initializer hops, accepts a plain `PropertyAssignment` initializer, and accepts an ambient `.d.ts` binding — which also covers `{ f: parseInt }`, `{ f: isNaN }`, `{ f: Math.max }` and `{ f: Object.keys }`, every one of which trapped identically.

It moved into the subsystem module `src/codegen/expressions/callable-property-host-value.ts` rather than growing the god file: a `loc-budget-allow` grant was available and deliberately not taken, and `calls-closures.ts` ends up 25 lines **smaller**.

**Deliberately a declaration proof, not a runtime test.** "Is this externref a wasm closure?" is a runtime question, and #4618 answers it that way for the `__call_fn_N` dispatchers — but doing so here means wrapping the whole callable-property dispatch, including the shared `__call_cprop_deferred_N` helper whose ABI passes only the already-cast root, in a runtime branch on the hottest object-literal call path. A host function that arrives dynamically (through a parameter, or a later write) is the recorded residual.

## Evidence

- Reduction probe, 11 rows + a deliberately-false control: **4/12 → 11/12** in Wasm; the control fails in both lanes by construction.
- Regression test — untyped `.js` two-file fixture. Fails on the parent with `dereferencing a null pointer`; passes here. Anti-vacuity: an array-like object and a missing argument must still answer `false`, and a compiled closure in the same literal must keep the typed `call_ref` path.
- **lodash 53/62 → 54/62** from this PR alone (58/62 with both siblings).
- **A/B at one head over 17 dogfood suites**, one suite at a time, per test file: net **+5**, **0** individually regressed tests, both lanes exit 0, every anchor reproduced (webpack 16/16 · three 17/18 · clsx 32/32 · cookie 63740/63740 · redux 64/82 · axios 200/231 · stylelint 108/108 · tailwindcss 13/13 · jsdom 6/6 · styled-components 9/9 · uuid 75/75 · marked 9/30 · moment 10/10 · prettier 101/151 · jest 329/356 · hono 244/324). The god-file extraction happened after that A/B and was separately proven byte-neutral (4/4 repro inputs compile to identical wasm sha256).
- Gates: LOC · func · coercion-sites · oracle-ratchet (+0/+0) · dead-exports · dogfood-validation (6/6) all green.

### DIFF
--- a/plan/issues/5342-lodash-residual-nine.md
+++ b/plan/issues/5342-lodash-residual-nine.md
@@ -1,10 +1,11 @@
 ---
 id: 5342
 title: "lodash residual: the last 9 of 62 — two traps, two Symbol-keyed nulls, two deepEqual misses — never investigated"
-status: ready
+status: done
+completed: 2026-09-06
 sprint: current
 created: 2026-09-05
-updated: 2026-09-05
+updated: 2026-09-06
 priority: medium
 horizon: m
 feasibility: medium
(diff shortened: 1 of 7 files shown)


ttraenkler:
## Park diagnosis — this is baseline drift, shared with at least two other PRs. Leaving `hold` in place.

I read the cited run ([34012017784](https://github.com/loopdive/js2/actions/runs/34012017784)) before touching anything, per the auto-park rules. **I am not removing the label** — the fix is a baselines refresh, which is the shepherd/lead's call, not a dev's.

**The failing gate:**

```
=== GATE FAIL: bucket "test/built-ins/Temporal/ZonedDateTime/prototype" has 59 regressions, exceeds the 50-test limit ===
=== Net: +1331 pass (35740 → 37071) ===   1387 improvements vs 82 regressions
```

**Cross-PR evidence that it is not this PR.** Two other PRs with diffs disjoint from this one parked today on the *identical* bucket and the *identical* count:

| PR | parked | gate line | net |
| --- | --- | --- | --- |
| #5648 (`destructuring this` on a sidecar-backed class) | 03:26 | `Temporal/ZonedDateTime/prototype has 59 regressions` | +1370 (35740 → 37110) |
| #5651 (unwrap an escaping throw at the export boundary) | 03:58 | `Temporal/ZonedDateTime/prototype has 59 regressions` | +1310 (35740 → 37050) |
| #5654 (this PR) | 05:13 | `Temporal/ZonedDateTime/prototype has 59 regressions` | +1331 (35740 → 37071) |

Three disjoint diffs cannot all have produced the same 59-file Temporal cluster off the same 35740 baseline. That is a property of the corpus/baseline (the exact reasoning in this gate's own footer and in `feedback_baseline_drift_cross_check`).

The #5280 exact-signature ledger did **not** fire (`first-occurrence`, sig `bc7c174e163cd5a0`) because each PR's own churn perturbs the file *set* slightly — 80 / 84 / 84 non-CT files, three different signatures. The invariant that actually identifies the drift is the **bucket + count**, not the signature hash. Possible ledger improvement.

**Likely root cause:** the run reports the js2wasm-baselines JSONL as **7 test262-relevant commits behind main HEAD (233m)**, and those commits include the compiled Temporal provider (#5353 / #5644, both merged today). Every merge_group is therefore measuring a large Temporal reshuffle — hence +1387 improvements — against a pre-provider baseline, and one Temporal bucket blows the 50-regression cap. A baseline refresh should clear all three parks.

**One signal that points the other way for this PR specifically:** #5648 and #5651 additionally tripped the `null_deref` trap ratchet (65 → 73). This PR reports `null_deref 65 → 56` — **9 fewer** uncatchable traps, which is exactly what it is for: it removes a `struct.get`-on-null trap on the callable-property path.

Local evidence at the head this PR was cut from is in the PR body: an A/B over 17 dogfood suites, one suite at a time, net +5 with 0 individually regressed tests.

ttraenkler:
**Park-hold diagnosis (lead sweep, 2026-09-06 07:30)** — collateral from the known Temporal drift on `main`, not this PR.

Cited run: [34012017784](https://github.com/loopdive/js2/actions/runs/34012017784), job `check for test262 regressions` (Test262 Sharded, merge_group).

What the gate actually said:
- `GATE FAIL: bucket "test/built-ins/Temporal/ZonedDateTime/prototype" has 59 regressions, exceeds the 50-test limit`
- Net **+1331** pass (35740 → 37071); host fine-gate net +1305 (1387 improvements − 82 regressions)
- Trap categories all flat or better: null_deref 65→**56**, illegal_cast 28→28, oob 14→12
- Baseline drift banner: js2wasm-baselines JSONL is **7 test262-relevant commits behind** main HEAD (233 min)

This is the same signature — 59 × `unknown time zone UTC` under `Temporal/ZonedDateTime` — that PR #5642 and PR #5646 each reproduced **byte-identically on clean detached `upstream/main` with zero PR changes** (Temporal-provider linkage drift from [#5353](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5353-sharded-temporal-lane); the provider cache key in `$TMPDIR/js2wasm-temporal-cache` ignores the compiler). This PR's diff (a host builtin in an object-literal callable property) does not touch Temporal, time zones, or the provider. Not re-reproduced a third time; the signature match is exact.

Action: removing the bot `hold` so the `auto-enqueue` backstop can take it again. **Not re-enqueueing by hand.** If it parks again on the same bucket, that is the drift, not this change; if it parks on a *different* bucket, treat that as real.

PR #5656, opened 2026-09-06
fix(#5342): the module-scope capability-probe idiom no longer binds null ✓

ttraenkler:
Cause **C** of three for [#5342](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5342-lodash-residual-nine). Siblings: #5654 (host builtin in a callable property) and #5655 (`eqref` carrier). Each is independently correct and independently revertable; **lodash reaches 58/62 only when all three land**.

```js
var Symbol = typeof globalThis.Symbol === 'function' ? globalThis.Symbol : undefined;
var symbol = Symbol ? Symbol('a') : undefined;
```

Two independent defects compose here, and **either one alone leaves the idiom wrong in a different way**, which is why they ship together.

## C1 — `globalThis.Symbol` read the module global it was defining

The #4500 Slice A arm answers `globalThis.<name>` / `this.<name>` from `<name>`'s wasm module global. That is right for **script** goal (§9.1.1.4.18 CreateGlobalVarBinding makes a script's top-level `var` a property of the global object — the whole reason Slice A exists) and wrong for a **module** (§16.2.1.6.4 InitializeEnvironment puts it in the module environment record, creating no global-object property). `receiverIsRealmGlobalObject` refuses a shadowed `globalThis` but never checked module-ness, so `__module_init` emitted, literally:

```wat
i32.const 1                 ;; typeof globalThis.Symbol === "function" — folded TRUE
(if (result externref)
  (then global.get 10)      ;; the not-yet-initialised `Symbol` global ITSELF
  (else call $__get_undefined))
global.set 10               ;; Symbol = null
```

The shadow bound `null`, stayed falsy for the rest of the module, and poisoned every later read off it — including a sibling `var Sym2 = globalThis.Symbol` in the same file, which is how the blast radius shows.

The gate is now script-goal only, judged on the **receiver's own source file** (not `ctx.sourceIsModule`, which multi-file linking sets unconditionally) and honouring `inferModuleStrictArguments === false`, because the test262 harness's synthetic `export function test()` wrapper marks genuinely script-goal sources as modules and Slice A's own witnesses (`var p1 = 7; this.p1`) are script tests.

Applied to **all four** Slice A arms — dot read, bracket read, nullish-comparison read, and the bracket **write** — because `realm-global-element-write.ts`'s own header records that a half-fixed read/write pair is worse than neither half. Every call site is a 1:1 line replacement, so `property-access.ts` (a god file at its LOC ceiling) does not grow and no `loc-budget-allow` grant is needed.

## C2 — the ternary join dropped the symbol brand

`compileSymbolCall` returns the js-host symbol id as a bare `i32` **on purpose** (#4626: branding it globally routed mid-emission coercions through a late `__box_symbol` import whose index shift corrupted already-baked `ref.func` operands — 216 regressions). Joining `symbol` with `undefined` coerces that `i32` to `externref`, and unbranded it took `__box_number` while the `symbol`-typed sink unboxed with `__unbox_symbol`, which answers 0 for a JS number — hence `Symbol(wasm_0)` in place of `Symbol(a)`.

The brand is re-attached **at the join only**. That site is already prepared for a coercion-time late import: both arms are parked in `fctx.savedBodies` precisely so the shift walker sees them, and the file says so.

## Evidence

- Regression test — untyped `.js` two-file fixture. Fails on the parent with `expected +0 to be 1` (the shadow is not live); passes here.
- Anti-vacuity for C1 is a spec point, not decoration: a **module's** top-level `var` must still read `undefined` through `globalThis`, proving the arm was not merely inverted. For C2: the same join with a runtime-false condition must still be `undefined`.
- The second `it()` is the **#4500 preservation control** — a sloppy script asserting `globalThis.p1 === 7`, `globalThis['p1'] === 7` and the bracket write. It passes on **both** lanes by design; it is the behaviour this change must not take away.
- **lodash 56/62 → 58/62** from this PR (on top of #5654 and #5655).
- **A/B at one head over 17 dogfood suites**, one suite at a time, per test file: net **+5**, **0** individually regressed tests, both lanes exit 0, every anchor reproduced.
- Gates: LOC · func · coercion-sites · oracle-ratchet · dead-exports all green.

### DIFF
--- a/src/codegen/expressions/misc.ts
+++ b/src/codegen/expressions/misc.ts
@@ -91,8 +91,24 @@ function compileConditionalExpression(
   fctx.savedBodies.push(elseInstrs);
   fctx.body = savedBody;
 
-  const thenType: ValType = thenResultType ?? { kind: "f64" };
-  const elseType: ValType = elseResultType ?? { kind: "f64" };
+  // (#5342) Re-attach the symbol BRAND to an arm that produced a bare `i32`
+  // symbol id. `compileSymbolCall` deliberately returns the js-host id
+  // UNbranded (#4626: branding it globally routed mid-emission coercions
+  // through a late `__box_symbol` import whose index shift corrupted already
+  // baked `ref.func` operands). The join below is the one place that must know
+  // anyway, and it is already prepared for a coercion-time late import — both
+  // arms are parked in `fctx.savedBodies` precisely so the shift walker sees
+  // them. Without the brand `cond ? Symbol("a") : undefined` boxed the id with
+  // `__box_number` while the `symbol`-typed sink unboxed it with
+  // `__unbox_symbol`, which answers 0 for a JS number — so lodash's
+  // `var symbol = Symbol ? Symbol("a") : undefined` fixture produced
+  // `Symbol(wasm_0)` in place of `Symbol(a)`.
+  const brandSymbolCarrier = (type: ValType, branch: ts.Expression): ValType =>
+    type.kind === "i32" && type.symbol !== true && ctx.oracle.staticJsTypeOf(branch) === "symbol"
+      ? { ...type, symbol: true }
+      : type;
+  const thenType: ValType = brandSymbolCarrier(thenResultType ?? { kind: "f64" }, expr.whenTrue);
+  const elseType: ValType = brandSymbolCarrier(elseResultType ?? { kind: "f64" }, expr.whenFalse);
 
   // Determine the common result type for both branches
   let resultValType: ValType = thenType;
--- a/src/codegen/helpers/is-strict-function.ts
+++ b/src/codegen/helpers/is-strict-function.ts
@@ -161,7 +161,7 @@ export function isSimpleParameterList(params: readonly ts.ParameterDeclaration[]
  * sloppy `.js` source compiled under a `.ts` filename has `scriptKind: TS` but
  * no module markers, and must stay sloppy.
  */
-function isModuleSourceFile(sf: ts.SourceFile): boolean {
+export function isModuleSourceFile(sf: ts.SourceFile): boolean {
   const internal = sf as ts.SourceFile & { externalModuleIndicator?: ts.Node };
   if (internal.externalModuleIndicator !== undefined) return true;
   if (sf.impliedNodeFormat === ts.ModuleKind.ESNext) return true;
--- a/src/codegen/helpers/sloppy-this-global.ts
+++ b/src/codegen/helpers/sloppy-this-global.ts
@@ -40,7 +40,7 @@ import { compileIdentifier } from "../expressions/identifiers.js";
 import { inlinedCalleeHasBoundReceiver } from "../expressions/inlined-call-receiver.js"; // (#4555)
 import { emitUndefined, ensureLateImport, flushLateImportShifts } from "../expressions/late-imports.js";
 import { coerceType } from "../shared.js";
-import { isStrictContext, isStrictFunction } from "./is-strict-function.js";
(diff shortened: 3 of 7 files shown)


ttraenkler:
## Responded to the park by narrowing the change; the attribution is genuinely unresolved and I am saying so.

Pushed `d3d4ee0`. I did **not** remove the `hold` — that is the shepherd's call.

### What the park said

Run [34013473256](https://github.com/loopdive/js2/actions/runs/34013473256): 31 regressions, all `built-ins/Temporal/**`, against a **fresh** baseline (37071 → 37050, net −21, ratio 29/10). No Temporal-bucket cap failure and no trap growth — a different shape from the drift cluster that parked #5648 / #5651 / #5654 earlier tonight.

### The measurement that changed my mind — and the one I got wrong first

I first blamed the broad module-goal gate: the compiled Temporal provider was **2,028,477** bytes on base and **2,055,669** with the change, a +27 KB shift that looked damning.

**That was a stale artifact.** The default `JS2WASM_TEMPORAL_CACHE` under `tmpdir` held a provider from an earlier compiler revision. Re-measured with a **fresh cache dir per configuration**, all four are byte-identical:

| configuration | provider bytes | validates | probe |
| --- | --- | --- | --- |
| plain base (fresh cache) | 2,055,669 | ✅ | `2020-11-15\|320` |
| broad gate + symbol brand | 2,055,669 | ✅ | `2020-11-15\|320` |
| broad gate, no symbol brand | 2,055,669 | ✅ | `2020-11-15\|320` |
| narrow gate (this push) | 2,055,669 | ✅ | `2020-11-15\|320` |

(The probe is `Temporal.PlainDate.from("2020-11-15").dayOfYear`, i.e. the exact assertion in one of the regressed rows, `PlainDate/prototype/dayOfYear/basic.js`.)

So this change has **zero** byte effect on the Temporal provider, and I cannot demonstrate it caused the parked cluster. Three other PRs with disjoint diffs parked on Temporal in the same window. The provider landed on main ~1h before the park (#5353/#5644), which is a plausible instability source; several of the regressed rows report `Temporal is not defined` and `Cannot read properties of null (reading 'toString') [in __module_init()]`, which read as linkage/instantiation failures rather than a codegen semantics change.

### What I changed anyway, on its own merits

The first cut gated the whole #4500 Slice A arm on SCRIPT vs MODULE goal — a semantics change across **every** `globalThis.<name>` / `this.<name>` access in module code. That is far more surface than #5342 needs, and I should not have reached for it first.

It is now the narrowest rule that fixes the reported defect: decline the Slice A arm **only when the access sits inside the initializer of the very `var <name>` it names**. The self-reference makes the answer unambiguous in both goals (§16.2.1.6.4 for modules; §9.1.1.4.18 does not overwrite an existing property for scripts), so no module/script test is needed and #4500's model is untouched everywhere else.

Dropped from the change-set: `is-strict-function.ts` and `realm-global-element-write.ts` (a write cannot be self-referential, so the read/write pairing that file's header warns about never arises). Folded into the existing `receiverIsRealmGlobalObject` call at each site, so `property-access.ts` stays at its LOC ceiling with no allowance.

**lodash is still 58/62 with the narrow rule — the broad gate bought nothing.** The regression test gains a scoping row that pins the deliberately-unchanged behaviour (`typeof globalThis.symbol` stays `"symbol"` for a non-self-referential read); if that row ever flips, the gate has widened and needs re-measuring.

Gates on the new head: LOC · func · coercion-sites · oracle-ratchet · dead-exports all green; both regression `it()`s pass, including the #4500 script-goal preservation control.

PR #5225, opened 2026-08-29
WIP(standalone): ES2015 for-of second pass (duplicate-base, needs re-merge)

ttraenkler:
## Description

**Draft — do not merge as-is.** A second implementation pass over the for-of work package ([#5144](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5144-es2015-standalone-forof-wave1)) produced on a base that predates the merged first pass (PR [#5191](https://github.com/loopdive/js2/pull/5191)) — so its +65 measured wins largely re-implement what already landed, and merging it against current main produces heavy twin conflicts. Parked as a draft instead of integrating.

Value worth mining before closing: its follow-up notes identify the §7.4.4 next-result-type TypeError gap, the `compileForOfAssignDestructuringExternref` evaluation-order rewrite (lref before next(), IteratorClose on abrupt), and the eager `__array_from_iter_n` materialization residue — all still open on main. Extract those clusters onto a fresh branch from current `origin/main` rather than resolving this diff's conflicts.

## CLA

- [x] I have read and agree to the CLA

---
_Generated by [Claude Code](https://claude.ai/code/session_013QXS328H9un2guFo43hR5e)_

PR #5661, opened 2026-09-06
fix(runtime): join the object-identity graph across the linked seam so instanceof/getPrototypeOf/constructor answer for provider classes (#5354)

ttraenkler:
## What

`instanceof` against a class that came from a **linked provider** was always
false, and so were `getPrototypeOf` and `constructor`. This joins the two halves
of the object-identity graph so all three follow from one edge.

Closes [#5354](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5354-linked-class-instanceof-prototype-identity).

**Land order: after PR #5648** ([#5251](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5251-temporal-numeric-calendar-seam-families)) — this branch is stacked on it and carries its commits.

## Step-1 evidence (measured, both lanes)

Provider `class Point { … }` in a linked npm package, versus the identical
two-file program compiled as ONE module. Base = this branch's merge base.

| probe (consumer side)                    | LINKED base | LINKED fixed | CONTROL (base = fixed) |
| ---------------------------------------- | ----------- | ------------ | ---------------------- |
| `d instanceof Point`                     | false       | true         | true                   |
| `getPrototypeOf(d) === Point.prototype`  | false       | true         | true                   |
| `Point.prototype === Point.prototype`    | **true**    | true         | true                   |
| `typeof getPrototypeOf(Point.prototype)` | **"null"**  | "object"     | "object"               |
| `typeof d.constructor`                   | "undefined" | "function"   | "function"             |
| `d.constructor === Point`                | false       | true         | true                   |
| `typeof Point.prototype.constructor`     | "undefined" | "function"   | "function"             |
| `Point.make(1,2) instanceof Point`       | false       | true         | true                   |
| `new Sub() instanceof Sub` / `… Base`    | false       | true         | true                   |
| `typeof Point[Symbol.hasInstance]`       | "undefined" | "undefined"  | "function"             |

The two bolded rows localise it. `Point.prototype` **is** identity-stable, so
nothing is minting a fresh mirror per read — and its own prototype is `null`,
which identifies it as the `Object.create(null)`-backed facade built in
`_makeClassCtorMirrorForHost`. Meanwhile `getPrototypeOf(d)` is a *different*
object: `Object.prototype`, hardcoded in the `_wrapForHost` proxy's trap. Two
unrelated objects, so OrdinaryHasInstance (§7.3.20) reads one and walks from the
other and can never meet.

`Symbol.hasInstance` is **not** the defect (it is also never consulted —
`_instanceofResult` explicitly skips `Function.prototype[Symbol.hasInstance]`),
so per the plan it is left alone and reported below.

## Fix

Which class a foreign struct belongs to is a `__tag` read, and only the module
that owns the struct has the type for it (#5225's minting-module rule). So the
owner publishes it:

* **`src/codegen/class-object-of.ts` (new)** — `__class_object_of(inst)` and
  `__class_parent_object_of(classObj)`, tag-dispatched over
  `ctx.classCtorHostRegistered` (exactly the classes that can escape as values).
  Discrimination is by `__tag`, never `ref.test` alone: WasmGC canonicalizes
  structurally-identical classes into ONE type, so a shape-only test answers for
  the wrong class. A module that lets no class escape emits identical bytes.
* **`src/runtime.ts`** — WeakMap-cached resolution (one question per struct)
  wired into four places: the instance proxy's `getPrototypeOf` trap (the fix),
  its `preventExtensions` (pin the answer onto the target first, or §10.5.1
  makes every post-freeze `instanceof` throw the proxy-invariant TypeError), its
  `get` trap for `constructor` (a get trap serves inherited reads itself), and
  the facade's `constructor` + its own prototype edge to the parent class.

That last edge is load-bearing: without it the newly linked chain would
**dead-end at the facade** and `d instanceof Object` — true before this PR —
would have regressed to false.

## Measurement: the 123 #5249 Temporal calendar rows

Fresh `JS2WASM_TEMPORAL_CACHE` per compiler revision; both sides ran all 123.

| failure bucket                           | base | after |
| ---------------------------------------- | ---- | ----- |
| `assert*: instanceof`                    | 32   | 13    |
| `Unsupported era name: …`                | 0    | 10    |
| `eraName must be string or undefined …`  | 0    | 8     |
| `RangeError: Invalid ISO date …`         | 67   | 68    |
| other                                    | 20   | 20    |
| pass                                     | 4    | 4     |

**19 of the 32 `instanceof` rows moved; 0 gained one; no row changed status in
either direction** — the freed rows land on the next layer rather than passing,
because these files carry many assertions each. Of the 13 still reporting
`instanceof`, 12 progressed to a *later* assertion in the same file, and every
direct `instanceof` door now answers true through the provider (`new PlainDate`,
`.from`, `.add`, `.with`, `new Duration`, `Duration.from`, `.until`, `.since`,
`new PlainDateTime`, `PlainYearMonth.from`) — so those messages are the assert
helper reporting a value a deeper failure produced.

**Next layer:** the polyfill's era arithmetic (`Unsupported era name`,
`eraName must be string or undefined`), newly visible now that `instanceof`
stops short-circuiting; plus the pre-existing 68-row `Invalid ISO date` bucket,
which is [#5208](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5208-compiled-date-host-date-bridge) (compiled `Date` → host `Date`) behind the [#5355](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5355-intl-datetimeformat-shell) bridge (PR #5657).

## Validation

* `tests/issue-5354-linked-class-instanceof.test.ts` — base-failing on the
  linked lane, green on the single-module control (pins it as seam-only).
* `tests/issue-5239-*`: its pinned `dynProtoIdentity: "false"` — an explicitly
  reported #5237-family identity gap — is now `"true"`; the pin is removed and
  both lanes assert the same map.
* Green: #5237 · #5239 · #5241 · #5242 · #5221 · #5225 · #5226 · #5244 · #5248 ·
  #5250 · #5251 · #5352 · #4628-temporal-global · #4628-class-value-prototype.
* Collateral for the object-identity paths (`built-ins/Reflect/*` +
  `built-ins/JSON/stringify/*`, 219 rows, per-row base vs fix): 141 pass / 78
  fail on both sides, **zero status flips, zero reason changes**.
* `equivalence-gate`: 24 failing / 1718 passing, no new regressions.
* Gates run bare: loc-budget (also with `LOC_GATE_BASE=origin/main`),
  func-budget, coercion-sites, oracle-ratchet, dead-exports, typecheck, lint.
  Growth grants are in the issue frontmatter, dated; #5251's inherited grants are
  restated there so they cannot be stranded by CI's merge preview.

## Reported, not fixed (all pre-existing, all measured on base)

* `getPrototypeOf(getPrototypeOf(new Sub()))` is not `Base.prototype` in the
  **single-module** lane. That is the raw-struct arm of `__getPrototypeOf`,
  deliberately untouched: there `C.prototype` is the RAW prototype struct, so
  answering the host facade would introduce the very identity split this
  removes. Bound: single-module two-hop chain walks.
* `typeof C[Symbol.hasInstance]` is "undefined" through the seam ("function" in
  the control). Does not affect `instanceof`. Bound: programs that read the
  well-known symbol off a linked class.
* `({x:1}) instanceof Point` fails to **compile** in the single-module lane
  (`struct.get[0] expected type (ref null 16), found struct.new of type (ref 27)`),
  identically on base. Unrelated static-`instanceof` lowering bug; kept out of
  the probes so it could not mask this measurement.

---


ttraenkler:
**Queue-health note (npm-suite lead, 2026-09-06 15:20)** — this PR has been re-grouped at the queue head repeatedly since its 11:45Z park-hold and fails the same way every time (runs 34039878712, 34040400453): `quality` → `Native-first host semantic debt gate (#4401)`:

```
runtimeTsLines 19318 > maximum 19313
```

The `plan/audit/host-import-policy-baseline.json` bump in this PR (19313) was measured against an older `main`; `src/runtime.ts` on current `main` is 19,154 lines and this PR's growth on top of it lands at 19,318. Each rebuild of its group also marks every PR behind it (`#5677`, `#5673`, `#5679`, …) UNMERGEABLE, so they cycle to the tail. Dequeueing it once so the queue can drain; the `hold` label stays until the ceiling is fixed (bump to ≥ 19,318 with the rationale, or move the added lines into a `src/runtime/` module). Please do not re-enqueue by hand — `auto-enqueue` picks it up once it is `CLEAN` and un-held.

ttraenkler:
Correction: GitHub had already removed this PR from the queue at 15:15:45Z (`failed_checks`); my dequeue call found nothing to remove. No manual queue mutation happened. The ceiling fix above still applies before it can be re-taken.

ttraenkler:
Park diagnosis (run 34030107216, `quality` → host-import-policy): `runtimeTsLines` on the merged state was 19332 against the 19313 ceiling measured before #5355 (PR #5657) landed on main — ceiling drift, not a regression in this PR's diff. Re-merged `origin/main` (0eb1d1365a), ceiling raised to the measured merged 19318, gate green locally. Removing the hold once.

---

PR #5663, opened 2026-09-06
fix(#5346): rebuilding a record from a host object recovers its reference fields

ttraenkler:
Part of [#5346](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5346-prettier-residual-print-doc-illegal-cast). One of two PRs from that investigation; the other fixes an independent cause.

## What was wrong

`buildRecordFromExternref` (#5243) exists for exactly one situation: a value whose static type is a compiler-minted `__anon_*` record arrives at runtime as an ordinary host object, `ref.test` fails, and the record has to be rebuilt property by property with `__extern_get`.

Its reference-field arm carried the comment *"anything else lands as null on that ONE field"* but emitted a bare `any.convert_extern; ref.cast_null`. **`ref.cast null $T` does not produce null on a mismatch — it traps.** A property read off a host object is never already a WasmGC struct, so the one input the materializer was written for was the one input it could not survive.

Measured on prettier@3.8.1: all three of `tests/unit/print-doc-to-string.js` died with `RuntimeError: illegal cast` inside `printDocToString`, whose `Indent.queue` is an ordinary host array reached through this materializer 23 times in that one function.

The issue's own plan predicted a union member cast to the first member's struct and pointed at `struct-carrier-inhabits.ts`. That was wrong — byte-level evidence is in the issue file; that module has one caller (`compileArrayLiteral`) and is not on this path.

## What changed

The field now goes through the same `externref → ref_null` coercion arm its enclosing value took to get here, so a vec field is rebuilt by element copy.

**That distinction is load-bearing, not cosmetic.** A plain guarded cast (`ref.test`, else `ref.null`) also stops the trap — measured — but hands `queue: []` back as `null`, and prettier's printer then spins forever instead of crashing. The regression test therefore asserts the recovered field's *contents*, not merely that nothing threw.

Two mechanics the arm needs, both existing idioms in that file:

- the detached `build` array is registered in `ctx.liveBodies` for the duration (#2182) — the recovery can mint a late import, and the shifter only walks bodies it knows about;
- a compile-time `materializingRecordTypes` set stops a cyclic record shape (`A.b: B`, `B.a: A`) from recursing forever while building its own field recovery.

The materializer moves to `src/codegen/record-from-host-object.ts`: it is one self-contained recovery terminal rather than part of the coercion matrix, and the god-file it lived in is at its LOC ceiling. The coercion it needs is injected as `recoverRefField` so the new module does not import back into `type-coercion.ts`. **No allowance is claimed for any ratchet.**

## Evidence

**Regression test** — `tests/issue-5346-record-from-host-object-field-recovery.test.ts`, untyped `.js` across two modules (annotating the values, or inlining `grow` into its caller, keeps the value in a WasmGC struct end to end and the materializer is never reached).

| | result |
| --- | --- |
| parent `01ce47aba7` | **2 failed / 2 passed** — both trapping cases `illegal cast`; the never-crossed-the-boundary guard and the arithmetic control pass |
| with this change | **4 passed** |

**A/B, 17 dogfood suites, one suite at a time, at this head** — every suite matches its clean-main anchor exactly, **zero movement anywhere**:

| suite | anchor | this branch | | suite | anchor | this branch |
| webpack | 16/16 | 16/16 | | styled-components | 9/9 | 9/9 |
| three | 17/18 | 17/18 | | uuid | 75/75 | 75/75 |
| clsx | 32/32 | 32/32 | | marked | 9/30 | 9/30 |
| cookie | 63740/63740 | 63740/63740 | | moment | 10/10 | 10/10 |
| lodash | 53/62 | 53/62 | | prettier | 101/151 | 101/151 |
| redux | 64/82 | 64/82 | | jest | 329/356 | 329/356 |
| axios | 200/231 | 200/231 | | hono | 244/324 | 244/324 |
| stylelint | 108/108 | 108/108 | | jsdom | 6/6 | 6/6 |
| tailwindcss | 13/13 | 13/13 | | | | |

prettier does not move because `print-doc-to-string` is still blocked downstream: with the trap gone the three tests now fail on an assertion instead, because `printDocToString`'s `output` reads as null after the loop. That is a separate defect, filed as #5356 with a minimal repro.

**Gates**: `check-loc-budget`, `check-func-budget`, `check-coercion-sites`, `check:oracle-ratchet`, `check:dead-exports`, `check:dogfood-validation` — all green locally.


ttraenkler:
## Park diagnosed — evidence says this is not caused by this PR. `hold` deliberately LEFT IN PLACE; escalating.

Per the auto-park rules I read the cited run before touching anything, and I am **not** removing the label: 15 regressions is over the escalation threshold, so this is a tech-lead call, not mine.

**The park**: solo merge group `gh-readonly-queue/main/pr-5663-59063c653a`, [run 34017638390](https://github.com/loopdive/js2/actions/runs/34017638390) — `check for test262 regressions`, net −6 (9 improvements − 15 regressions), bucket signature `0fecfd36aff1ec6d`.

The 15 split 11 Temporal + 4 TypedArray/BigInt. Four of the Temporal ones fail with **`Temporal is not defined`**, which is the compiled provider being unavailable rather than a semantic result.

### Evidence

**1. The compiled Temporal provider is byte-identical with and without this change.**

Cold-built (no cache) on this branch and on `upstream/main` with only `type-coercion.ts` / `record-from-host-object.ts` reverted:

```
this branch     providers/b7bc42b3…8012c7.wasm  2,055,529 B
upstream/main   providers/b7bc42b3…8012c7.wasm  2,055,529 B
sha256          b7bc42b36d5883c17da0643680c3fe5d635f9e635122082d358bd6c3208012c7   (identical)
```

The provider that all 11 Temporal tests run against is the same artifact either way, so this change cannot be what makes `Temporal` undefined in them. (Worth noting for anyone reproducing: a warm local cache serves a *stale* provider — `d4ca14f8…`, 2,028,477 B — so `JS2WASM_TEMPORAL_CACHE` must point somewhere fresh or the comparison is vacuous. My first pass had exactly that hole.)

**2. All 15 regressed tests pass locally, on this branch AND on the parent.**

Run through the project's own `runTest262File`, host lane:

| set | this branch | parent |
| --- | --- | --- |
| 11 Temporal | 11/11 pass | 11/11 pass (7 re-run cold; 4 warm) |
| 4 TypedArray/BigInt + Delete | 4/4 pass | — |

**3. The change's own reach is narrow.** It only alters `buildRecordFromExternref`'s `ref_null`-field arm, which previously emitted a bare `ref.cast_null`. That arm is reached only when a value whose static type is a `__anon_*` record arrives as a host object AND the record has a reference-typed field. Consistent with that, the merged report shows the trap ratchet **unchanged**: `illegal_cast 28→28`.

**4. The 17-suite dogfood A/B on this branch moves nothing** — all 17 match their documented clean-main anchors exactly (table in the PR body).

### What I could not rule out

The gate's own heuristic disagrees with me — it reports *"0 test262-relevant commits separate the baseline from main HEAD … these regressions are far more likely PR-caused. Do not dismiss them."* I am not dismissing it; I am reporting that every direct measurement I can make points the other way, and that the strongest single fact (a byte-identical Temporal provider) is hard to reconcile with 11 Temporal regressions being this PR's.

The cross-PR ledger (#5280) has bucket `0fecfd36aff1ec6d` recorded once so far. If a subsequent unrelated PR parks with the same signature, that confirms drift. #5665 (the sibling PR from this investigation, a completely different file) is a natural second sample.

Separately, and possibly related: **hono looks unstable in this environment right now.** It read 244/324 at 06:20Z and 220/324 on five later runs, including on **unmodified `upstream/main` source** — details and the control table are in #5665's body. Its 244 anchor may need re-measuring.

### Requested

A tech-lead judgement on whether to re-admit this PR once. I am standing down rather than removing the label or enqueueing.

PR #5664, opened 2026-09-06
fix(#5342): narrow the globalThis probe gate to the self-referential initializer ✓

ttraenkler:
Follow-up to #5656 (cause C of [#5342](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5342-lodash-residual-nine)), which merged in a batched group before this narrowing was ready. **Behaviour-preserving for the idiom it targets; it only removes surface.** Close it if you'd rather leave main alone — nothing is broken today.

## Why

#5656 gated the whole #4500 Slice A arm on **script vs module goal**. That is a semantics change across *every* `globalThis.<name>` / `this.<name>` access in module code, and #5342 never needed that much.

This replaces it with the narrowest rule that fixes the reported defect: decline the Slice A arm **only when the access sits inside the initializer of the very `var <name>` it names**.

```js
var Symbol = typeof globalThis.Symbol === "function" ? globalThis.Symbol : undefined;
```

The self-reference makes the answer unambiguous in **both** goals, so no module/script test is needed at all:

- **module** — §16.2.1.6.4 puts the `var` in the module environment record, so it is not a global-object property and the access must consult the real object;
- **script** — §9.1.1.4.18 CreateGlobalVarBinding does **not** overwrite an existing property, so during its own initializer `globalThis.<name>` is still whatever the realm had (the builtin, or `undefined` for a new name).

Dropped from the change-set entirely: `is-strict-function.ts` (the `isModuleSourceFile` export is no longer needed) and `realm-global-element-write.ts` (a write cannot be self-referential, so the read/write pairing that file's header warns about never arises). The extra condition folds into the existing `receiverIsRealmGlobalObject` call at each site, so `property-access.ts` — a god file at its LOC ceiling — does not grow and needs no allowance.

## A measurement I got wrong first, corrected

#5656 was auto-parked with 31 `built-ins/Temporal/**` regressions. I initially blamed the broad gate: the compiled Temporal provider read **2,028,477** bytes on base and **2,055,669** with the change.

**That was a stale artifact** — the default `JS2WASM_TEMPORAL_CACHE` under `tmpdir` held a provider from an earlier compiler revision. Re-measured with a **fresh cache dir per configuration**, all four are identical:

| configuration | provider bytes | validates | probe |
| --- | --- | --- | --- |
| plain base (fresh cache) | 2,055,669 | ✅ | `2020-11-15\|320` |
| broad gate + symbol brand | 2,055,669 | ✅ | `2020-11-15\|320` |
| broad gate, no symbol brand | 2,055,669 | ✅ | `2020-11-15\|320` |
| narrow gate (this PR) | 2,055,669 | ✅ | `2020-11-15\|320` |

(The probe is `Temporal.PlainDate.from("2020-11-15").dayOfYear` — the exact assertion in one of the regressed rows, `PlainDate/prototype/dayOfYear/basic.js`.)

So neither version has any byte effect on the Temporal provider, the park was **not** demonstrably caused by #5656, and the group went on to merge green. This PR is therefore a **risk reduction on its own merits**, not a fix for that park.

## Evidence

- **lodash is 58/62 with the narrow rule** — identical to the broad one, on current main. The broad gate bought nothing measurable.
- All four #5342 regression `it()`s pass, including the #4500 script-goal preservation control (a sloppy script asserting `globalThis.p1 === 7`, `globalThis['p1'] === 7`, and the bracket write).
- The test gains a **scoping row** that pins the deliberately-unchanged behaviour: a NON-self-referential read of the same name still answers from the module global, so `typeof globalThis.symbol` stays `"symbol"`. If that row ever flips, the gate has widened past the self-referential initializer and needs re-measuring.
- Gates: LOC · func · coercion-sites · oracle-ratchet · dead-exports all green.

### DIFF
--- a/src/codegen/helpers/is-strict-function.ts
+++ b/src/codegen/helpers/is-strict-function.ts
@@ -161,7 +161,7 @@ export function isSimpleParameterList(params: readonly ts.ParameterDeclaration[]
  * sloppy `.js` source compiled under a `.ts` filename has `scriptKind: TS` but
  * no module markers, and must stay sloppy.
  */
-export function isModuleSourceFile(sf: ts.SourceFile): boolean {
+function isModuleSourceFile(sf: ts.SourceFile): boolean {
   const internal = sf as ts.SourceFile & { externalModuleIndicator?: ts.Node };
   if (internal.externalModuleIndicator !== undefined) return true;
   if (sf.impliedNodeFormat === ts.ModuleKind.ESNext) return true;
--- a/src/codegen/helpers/sloppy-this-global.ts
+++ b/src/codegen/helpers/sloppy-this-global.ts
@@ -40,7 +40,7 @@ import { compileIdentifier } from "../expressions/identifiers.js";
 import { inlinedCalleeHasBoundReceiver } from "../expressions/inlined-call-receiver.js"; // (#4555)
 import { emitUndefined, ensureLateImport, flushLateImportShifts } from "../expressions/late-imports.js";
 import { coerceType } from "../shared.js";
-import { isModuleSourceFile, isStrictContext, isStrictFunction } from "./is-strict-function.js";
+import { isStrictContext, isStrictFunction } from "./is-strict-function.js";
 
 /**
  * True when an unbound `this` at `expr` must evaluate to the global object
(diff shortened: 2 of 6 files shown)


ttraenkler:
**Closing (lead decision, 2026-09-06 10:40)** — the merge queue parked this on a real net loss, and the PR's own premise was that it buys nothing measurable.

Cited run: [34018749549](https://github.com/loopdive/js2/actions/runs/34018749549), `check for test262 regressions`:

```
=== Net: -32 pass (37074 → 37042) ===
=== GATE FAIL: net_per_test -32 < 0 (9 improvements − 41 regressions) ===
=== GATE FAIL: regression ratio 455.6% (41/9) meets/exceeds the 10% limit ===
```

Trap categories flat, no Temporal bucket — this is not the #5353 drift (the run right after it, for #5665, read net +28 against the same baseline). The broad script-vs-module gate that #5656 merged is therefore doing real work on test262 that the self-referential-initializer narrowing undoes; and lodash is 58/62 either way, as the description says. Keeping main as it is. If the 41 turn out to be a corpus property rather than the narrowing, reopen with that evidence.

ttraenkler:
**Correction** — the comment above is void. This PR was not closed: the merge queue re-grouped it with #5665 and #5668 (group `pr-5668-8f1c4453…`, [Test262 Sharded run 34021269257](https://github.com/loopdive/js2/actions/runs/34021269257), started 08:13:48Z) and that group's `check for test262 regressions` **passed**, so all three merged at 08:39:10Z with the `hold` label still attached. The −32 measured for this PR alone at 07:17Z did not reproduce on the combined merged state. The bot's park-hold does not remove an entry from the queue; it only stops `auto-enqueue` from re-adding it.

PR #5670, opened 2026-09-06
test(#5363): pin thrown-error identity across the linked seam — the reported defect does not exist; file #5364 (cross-project registry contamination)

ttraenkler:
Land order: after PR #5666 and PR #5661.

Implements [#5363](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5363-thrown-error-instanceof-across-seam).

## Verdict: the reported defect does not exist

[#5363](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5363-thrown-error-instanceof-across-seam)
was filed on the hypothesis that composing
[#5226](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5226-provider-error-identity)
(a provider throw reaches the consumer's compiled `catch` by identity) with
[#5247](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5247-export-throw-boundary)
(an uncaught throw reaches the host as the real `Error`) left a hole: host →
consumer export → provider throws → nothing catches in wasm → the host's `catch`
sees something that is not `instanceof RangeError`.

**Measured on this stacked base, every route already answers correctly** — the
reduced lanes and the real linked `@js-temporal/polyfill` provider alike. So this
PR changes no compiler code. It lands the regression test that pins the
composition, the re-attribution of the 22 rows, and a NEW defect the measurement
uncovered (reported, not fixed, with a bound).

### Probes (a)/(b)/(c), linked vs single-module control

| route | LINKED | CONTROL |
| --- | --- | --- |
| (a) host catch; provider throws, uncaught through the consumer export | `ctorIsRE=true instRE=true [object Error] n=RangeError` | identical |
| (b) host catch; consumer catches and re-throws | identical to (a) | identical |
| (c) host catch; single-module export throwing the same | identical to (a) | identical |
| compiled catch of the provider throw | `ctorIsRE=true instRE=true cn=RangeError` | identical |
| `typeof RangeError` inside the PROVIDER | `function`, name `RangeError` | identical |

The polyfill's `new RangeError(...)` IS the host's `RangeError`; the #5247 export
wrapper does fire for the consumer in the linked build. Roots (i), (ii) and (iii)
from the plan are all ruled out. Through the real Temporal provider, test262
`harness/assert.js`'s actual predicate — which is **not** `instanceof`, it is
`thrown.constructor !== expectedErrorConstructor` — answers `PASS`.

### The 22 rows were mis-attributed

They are not `assert.throws` at all. `Test262Error: <description>: instanceof`
is the message `TemporalHelpers.assertPlainDate` / `assertDuration` hand to a
plain `assert`:

```js
// test262/harness/temporalHelpers.js:231
assert(date instanceof Temporal.PlainDate, `${prefix}instanceof`);
```

That is `instanceof` on a value the provider RETURNED —
[#5354](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5354-linked-class-instanceof-prototype-identity)'s
family — not on a thrown one.

### 123-row measurement (provider linked, fresh cache, cold build)

**13 pass / 110 fail**, up from the 4 pass measured before #5208 landed.

| failure bucket | rows |
| --- | ---: |
| `assert*: …: instanceof` | 23 |
| `RangeError: infinity is out of range` | 22 |
| `Test262Error: eraName must be string or undefined …` | 21 |
| `RangeError: Era am/aa (ISO year N) was not matched by any era` | 21 |
| `Test262Error: Unsupported era name: gregory / japanese / …` | 10 |
| `RangeError: Invalid monthCode: M13` | 5 |
| `RangeError: Invalid ISO date` | 1 |
| other | 7 |
| **pass** | **13** |

The two era buckets are a POLYFILL VERSION gap, not ours: the harness expects
`gregory`'s eras to be `bce`/`ce` per intl-era-monthcode; the pinned polyfill
still reports `era: "gregory"`.

### Reported, NOT fixed — the real finding

**A process that instantiates more than one linked project against the same
provider binary cross-contaminates class identity.** A later project resolves
its own instances through an EARLIER project's exports, so `x instanceof C` is
false against the live `C` while `x.constructor.name` still reads right.

- The same program run ALONE: `inst=true`. Run after ten other linked programs in
  the same process: `inst=false`, `cn=PlainDate`, internally consistent — two
  complete, unrelated `PlainDate` mirrors.
- Conformance-level: `intl402/Temporal/PlainDate/prototype/add/month-boundary-gregory.js`
  fails `endYesterdayNextDay: instanceof` inside the 123-row batch and, run alone
  with the same driver and cache, fails instead on `Unsupported era name: gregory`
  — the next, not-ours layer.

Mechanism: `src/runtime/cross-module-struct-owners.ts` keeps a process-global
`modules` Set that is never unregistered, and two instances of the same provider
binary have identical canonical WasmGC types, so instance A's
`__struct_field_names` names instance B's struct; `__class_object_of` then
answers the stale instance's class-object singleton.

**Consequence: every 123-row `: instanceof` count quoted so far is inflated by
cross-row poisoning.** Two candidate fixes were tried here and both reverted
(project-scoped registry — regressed the #5225 consumer→provider route;
newest-first tie-break — no effect, the staleness also arrives outside
`decoderFor`). Details and the recommended next step are in the issue file.

### Also in this PR

`plan/audit/host-import-policy-baseline.json` `maximumRuntimeTsLines`
19185 → 19313 (measured). This growth is entirely inherited from the #5208 and
#5354 predecessors — this PR touches no `src/` — and neither raised it, so the
gate was red on `issue-5354-linked-class-instanceof` (PR #5661 now carries its
own bump to 19255). Raised here per the established convention (`chore(#5225)`,
`chore(#5242)`) so the stack is not blocked.

### Filed from this measurement

The cross-project contamination above is now
[#5364](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5364-cross-module-registry-leaks-across-linked-projects)
(file added in this PR, with an implementation plan: per-row registry reset in
both test262 drivers first, then a `rootImports`-keyed project scope). It also
affects the #5353 worker lane, which runs many rows per fork.

### Validation

- `tests/issue-5363-thrown-error-seam.test.ts` — new, green. Pins test262
  `assert.throws`'s actual predicate for a provider-originated throw, in compiled
  code and in the host, both lanes, with two negative controls. No existing suite
  covers `.constructor` — #5226 pins `instanceof` in a compiled catch, #5247 pins
  the host-side shape.
- Suites green (94 tests, 14 files): 5208 · 5354 · 5221 · 5225 · 5237 · 5239 ·
  5241 · 5242 · 5244 · 5248 · 5250 · 5251 · 4628-temporal-global ·
  4628-class-value-prototype; plus 5226 · 5247 · 5363 (11 tests).
- Gates green: loc-budget (also with `LOC_GATE_BASE=origin/main`), func-budget
  (both bases), coercion-sites, oracle-ratchet, dead-exports, host-import-policy,
  typecheck, lint.
- `node scripts/equivalence-gate.mjs` at baseline.

---


ttraenkler:
Park diagnosis (run 34029551562, `quality` → host-import-policy): `runtimeTsLines` on the merged state exceeded the 19313 ceiling because #5355 (PR #5657) landed on main after the ceiling was measured — not a regression in this PR's diff. Re-merged `origin/main` (2547d32992), ceiling raised to the measured merged 19318, gate green locally. Removing the hold once; this branch carries #5661's commits, so #5661 lands with it.

---

PR #5673, opened 2026-09-06
fix(codegen): a `= undefined` parameter default must not discard the argument (#5360)

ttraenkler:
**Land order: after PR #5661 and PR #5666.**

This branch is predecessor-stacked on `issue-5208-compiled-date-host-bridge`
(which carries #5355 and #5251) and `issue-5354-linked-class-instanceof` — the
rows this issue is about are only reachable with both.

Closes [#5360](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5360-temporal-era-arithmetic-seam).

## The headline is not the fix — the issue's premise was wrong

[#5360](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5360-temporal-era-arithmetic-seam)
was filed on the belief that "node on the same pinned polyfill passes these
rows". It does not. The 123-row Temporal calendar family was run under node
against the same `@js-temporal/polyfill@0.5.1` and the same test262 harness,
one fresh `vm` context per row:

| lane | pass | fail |
| --- | --- | --- |
| node on the pinned polyfill | **45** | 78 |
| compiled, on this branch's stacked base | 13 | 110 |

So the reachable ceiling is 45 and the compiler-attributable gap is **32 rows**,
not the ~39 the issue describes:

| bucket | rows | node also fails? | verdict |
| --- | --- | --- | --- |
| `Unsupported era name: …` | 10 | **all 10** | polyfill version gap |
| `Era am/aa … was not matched by any era` | 21 | **all 21** | polyfill version gap |
| `eraName must be string or undefined` | 21 | 16 of 21 | **5 are the compiler's** |

Polyfill 0.5.1 predates the `Intl.Era-monthcode` era-code renaming, so it
reports `date.era` as `"gregory"` / `"japanese"` / `"roc-inverse"` while the
harness table knows only `bce`/`ce`/`broc`/`reiwa`/…. `Unsupported era name:
gregory` names a *calendar id*, which no era table will ever contain — node
prints the identical message. **Those 31 rows cannot be fixed in the compiler**;
they need a newer polyfill pin.

The three suspects the plan named (record-bridge string carrier, comma-joined
array carrier, host-mirror `===` identity) were probed and are all clean.

## Root cause of the 5 rows that ARE the compiler's

`typeof eraName` answered `"number"` — a string was never *arriving*, nothing
was corrupting one. It came from the caller's parameter list:

```js
// test262/harness/temporalHelpers.js
assertPlainDate(date, year, month, monthCode, day, description = "", era = undefined, eraYear = undefined)
```

Under the deliberately-pinned `strictNullChecks: true` (#2748) TypeScript infers
`era` as the **type `undefined`**, from its own default initializer. That is a
statement about TS callers, not about what a JavaScript caller passes, and two
independent parts of codegen trusted it:

1. **The slot.** `resolveWasmType(undefined)` is a numeric slot ("void → no
   result"), so the object-literal-method / arrow / function-expression lanes
   gave the parameter an i32/f64 slot and the argument was coerced away at the
   call boundary. The free-function lane escaped only because call-site
   inference overrides its registered signature — which is why the neighbouring
   `description = ""` bound correctly in the same call and made this look like
   argument shifting.
2. **The folds.** `typeof x` and `String(x)` (both the call-identifier arm and
   `emitToString`) drop the carrier and substitute the constant `"undefined"`
   whenever the static type carries `TypeFlags.Undefined`.

Either half alone leaves the row failing. Reduction, no Temporal:

```js
const objLit = { m(a, b = undefined) { return typeof b + ":" + String(b); } };
objLit.m(1, "heisei")   // base: "boolean:0"           node: "string:heisei"
```

## Fix

Two narrow helpers in `src/codegen/destructuring-params.ts`, beside the
`isNullOrUndefinedLiteral` they reuse:

- **`widenUndefinedDefaultParamSlot`** — externref slot for an *un-annotated*
  parameter whose default is the literal `undefined` / `void <num>` / `null`.
  Applied at the SIGNATURE and BODY sites of both affected lanes; the pairing is
  load-bearing (one without the other emits invalid Wasm — the same rule the
  neighbouring binding-pattern widening documents).
- **`paramUndefinedTypeIsDefaultArtifact`** — suppresses the two folds for a
  read of such a parameter. Routed through `ctx.oracle.declarationsOf`, since
  the question is about the declaration's *syntax*, not a type.

`b: undefined`, `b = ""`, `b?: T` and rest parameters are untouched.

## Measured

**123-row Temporal calendar family, provider-linked, same driver both sides:**

| | base | fix |
| --- | --- | --- |
| pass | 13 | **15** |
| pass → fail | — | **0** |

| bucket | base | fix |
| --- | --- | --- |
| `eraName must be string or undefined` | 21 | **13** |
| `… instanceof` | 23 | 26 |
| `infinity is out of range` | 22 | 22 |
| `was not matched by any era` | 21 | 21 |
| `Unsupported era name` | 10 | 11 |

All 5 compiler-attributable `eraName` rows clear the guard: 2 pass outright, 3
advance to a next layer (`instanceof` ×2 — the #5354 residual; `monthCode M01
and month 2 must match` ×1).

Note: the `instanceof` counts on this list are inflated by cross-row registry
contamination in the batch driver —
[#5364](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5364-cross-module-registry-leaks-across-linked-projects)
(filed in PR #5670); the base-vs-fix delta above is unaffected because both
sides ran the same driver in the same order.

**Collateral slice** (I touched string carriers, `typeof` and `===`-adjacent
folds, and parameter slots) — 381 rows, per-row base vs fix:
`language/expressions/{equals,strict-equals,typeof}` plus all 288
`*dflt-params*` files. **0 flips, 0 message changes.**

**Suites** (one vitest process each, all green): `issue-5360-*` (6/6; 5 of 6
fail on base), `issue-5208-*`, `issue-5354-*`, `issue-5355-*`, `issue-5251-*`,
`issue-5250-*`, `issue-5352-*`, `issue-4628-temporal-global`, and the 9 provider
suites (5221, 5225, 5226, 5237, 5239, 5241, 5242, 5244, 5248).

**Gates** (run bare, not piped): typecheck, lint, loc-budget (also with
`LOC_GATE_BASE=origin/main`), func-budget, coercion-sites, oracle-ratchet,
dead-exports, host-import-policy. Grants are dated in the issue file's
frontmatter, with the inherited #5208/#5354/#5355/#5251 grants restated for the
reader. `plan/audit/host-import-policy-baseline.json` `maximumRuntimeTsLines` is
bumped 19185 → 19313 to the **measured** value; that growth is entirely the
stacked branches' `runtime.ts`, none of it this change-set's.

## Reported, not fixed (bounded)

- The **27 remaining** compiler-gap rows: 11 `… instanceof` on `since`/`until`
  results (#5354 residual, different receiver), 7 `infinity is out of range`
  (all ZonedDateTime), 3 `Expected a RangeError to be thrown` (the
  `infinity-throws-rangeerror` files — same root, inverted), 2 `Invalid
  monthCode: M13` (ethioaa), 4 one-offs. The 7 + 3 infinity rows look like one
  root and are the cheapest next slice.
- `const U = undefined; function f(a, b = U)` — the default is an identifier,
  not the literal, so neither helper fires. Would need constant propagation into
  the initializer.
- A **pre-existing, unrelated** compile failure found while probing: a module
  holding both a class method and an object-literal method with defaults emits
  `return_call[0] expected type (ref null N), found ref.as_non_null of type
  (ref M)`. Reproduces byte-identically on unmodified base; not caused here, not
  diagnosed here. Class methods are therefore deliberately outside this fix.

---


ttraenkler:
Park diagnosis (run 34029408529): (1) `quality` → host-import-policy: merged-state `runtimeTsLines` 19318 vs the 19313 ceiling measured before #5355 (PR #5657) landed — ceiling drift, not this diff. (2) `equivalence-shard (4)`: its log had expired (404); re-ran `node scripts/equivalence-gate.mjs` on this branch merged with current main — no new regressions (22 failing / 1720 passing vs 24/1718 baseline). Re-merged `origin/main`, ceiling raised to the measured 19318 (46dd6af0ac). The stale queue entry was dequeued via draft so the fix could be pushed; back to ready, removing the hold once.

---

ttraenkler:
Second park (run 34052809035) is the same ceiling drift: main's `maximumRuntimeTsLines` moved to 19321 (#4444) under this PR. Re-merged `origin/main` (746b6fbe01); this branch adds no `src/runtime.ts` lines on top of main, so the baseline file is main's (19321) and the gate is green. Released alone — #5678/#5685 are drafted until this lands, since two runtime-growing PRs batched into one group always exceed either's measured ceiling.

---

ttraenkler:
Third park (run 34067441960): 8 `built-ins/TypedArray/**` rows pass→fail. Attributed as collateral, not this PR's — evidence in the issue file's `## Merge-group regression (2026-09-07)`:

- The emitted wasm for all 8 rows (primary + strict rerun, worker options) is **byte-identical** between `origin/main` and this branch (sha256 A/B by file copy); none of the 8 sources contains a parameter matching this PR's guard precondition. The job's "wasm-hash changed on all 8" banner is #5256 — the baseline carries no `wasm_sha` field (0 of 48,735 entries), so every flip lands in that bucket by schema.
- Through the sharded worker lane (`CompilerPool`, fresh bundles): 8/8 pass isolated and 8/8 pass inside a 329-row same-family fork run.
- All 8 error texts match an empty `makeCtorArg(2)` — one fork-level realm perturbation (#1957 class, not diffed by the canary), consistent with them being one contiguous family in one shard.

Branch at 06894574d2 (main merged, two leftover debug hooks removed, 123-row Temporal gain 13→15 intact, gates green). Re-admitting once.

---

PR #5675, opened 2026-09-06
fix(#5362): a compiled TypedArray reaches a host API as a TypedArray, not an array facade

ttraenkler:
## Summary

Fix the compiled TypedArray host boundary tracked in `plan/issues/5362-hono-signed-cookie-typed-array-host-marshalling-regression.md`.

- Pass branded compiled carriers to host APIs as stable concrete TypedArray mirrors, while ordinary arrays retain their array representation.
- Reconcile host mutations and buffer detachment after both normal and throwing host calls, including dynamic method bridges.
- Propagate detached state to compiled indexed reads/writes and TypedArray methods. Distinguish a detached buffer from an attached zero-length buffer.
- Emit reconciliation helpers for fixed-arity host bridges, and recognize nullable/aliased TypedArray receivers in the affected codegen paths.

This PR includes runtime **and codegen** changes. The original runtime-only patch grew during recovery to preserve detachment behavior; the original “no codegen changes / identical compiled bytes” description no longer describes the final implementation.

## Current checkpoint and validation

Final pushed head: `a35a84fb746014c0c0c1b3f4be4886591af7a15c`. It includes repair `b5598cd848d2e0a7490996896291ebd49edacd44` and integration of upstream main through `855acee6a7ac662500edec8fa1806c9d9e7ee0d2`. This is a checkpoint provenance statement, not a claim that the branch contains every subsequent main commit.

Focused local verification after the source integration: **58/58** across the TypedArray host-boundary suite (11), fixed host-method calls (5), vec-mirror writeback (17), and dynamic dispatch (25). Controls cover ordinary arrays, attached empty buffers, ignored detached writes, undefined detached indexed reads, detached method rejection, and abrupt host exits.

Typecheck, formatting/lint, scoped budgets and host-policy checks passed. Final-head CI includes passing quality, issue/linear tests, all eight equivalence shards and their gate, and dogfood compile validation. PR-level Test262 checks are not proof of a full merged-state Test262 run; merge-queue validation remains required.

- [Final-head CI](https://github.com/loopdive/js2/actions/runs/34120737111)
- [Recovery handoff and validation details](https://github.com/loopdive/js2/pull/5675#issuecomment-5570468503)

## Historical investigation and scope

[...]

PR #5676, opened 2026-09-06
fix(#5339): inlined-IIFE returns inside catch clauses no longer escape the IIFE

ttraenkler:
Fixes [#5339](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5339-hono-dev-index-whole-module-failure).

hono's `src/helper/dev/index.test.ts` was 0/8 with a **null** per-test
`wasmError`, because nothing per-test had failed: the whole module failed to
validate.

```
report.compile.details["src/helper/dev/index.test.ts"]:
  success: true, validates: false,
  validationError: CompileError: WebAssembly.compile():
    Compiling function #340:"getColorEnabledAsync" failed:
    type error in return[0] (expected i32, got externref)
```

## Root cause

A value-returning **IIFE is inlined** into its caller, so every `return` in its
body must become `local.set <ret>; br <iife-exit>` — there is no Wasm function
left to return from. `patchReturns` in
`src/codegen/expressions/call-tail-dispatch.ts` walked `if.then`, `if.else` and
`<block>.body`, but **not the `catches[].body` / `catchAll` arms of a legacy
`try`**. A `return` inside a catch clause stayed a Wasm `return` and returned
from the *enclosing* function.

hono's `dist/utils/color.js` hits it exactly:

```js
const isNoColor = cond
  ? await (async () => { try { return "NO_COLOR" in (…); } catch { return false; } })()
  : !getColorEnabled();
```

The IIFE's ret local is `externref` (`Promise<boolean>`); the enclosing
function's Wasm result is `i32`. The escaping `return` pushed an `externref`
where an `i32` was required, so the module did not validate and none of the
eight tests ever ran.

The same omission is **also a silent wrong-value bug whenever the two return
types agree** — valid Wasm, no diagnostic:

```js
function f() { const v = (function () { try { throw new Error("boom"); } catch { return 1; } })(); return v + 10; }
f()   // native 11, wasm 1
```

The **void**-IIFE arm of the same file already walked the catch arms; the two
walkers were copies and had drifted. Rather than adding the missing lines to
the second copy, this PR extracts **one** walker —
`patchInlinedIifeReturns` in `src/codegen/expressions/iife-return-patch.ts` —
parameterised by the result local (`null` = void), so the arms cannot diverge
again. That also makes `call-tail-dispatch.ts` and `compileTailDispatch`
*smaller*, so no budget allowance was needed.

Standalone/WASI was never affected: `buildStandardTryTable` materialises
handlers as nested `block`s, which the generic `body` recursion already
reached.

## Verification

- **Regression test** `tests/issue-5339-iife-return-in-catch.test.ts`, untyped
  `.js` two-file fixtures: **2/8 on the parent, 8/8 with the fix**. The two
  that pass on the parent are the anti-vacuity controls (`return` in the try
  body; no `try` at all) — same shapes, catch clause never entered.
- **A/B over all 17 dogfood suites at one HEAD (`a1469a5454`), per test file.**
  Total 65015 → 65016. **Exactly one file moved:**
  `hono/src/helper/dev/index.test.ts` `0/8 → 1/8` (hono 220/324 → 221/324).
  Every other file in every suite is byte-identical in pass count and
  denominator.
- `tests/equivalence/` gate, all 8 shards: no new regressions.
- Gates green: loc-budget, func-budget, coercion-sites, oracle-ratchet,
  dead-exports, dogfood-validation, `tsc --noEmit -p tsconfig.ts7.json`.

## Two things the lead should know

**1. The ≥ 6/8 acceptance bar in the issue is not met and cannot be met by this
fix.** The module now compiles, validates and runs, so the report names a real
error per test instead of `null` — but 7 tests still fail, from two *different*
defects, both now confirmed with repros and filed:

- [#5365](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5365-host-closure-bridge-loses-length-and-name)
  (6 tests) — in JS-host mode a closure that crosses a call boundary as a value
  reports `length === 0` / `name === undefined` (it is the `wasmClosureBridge`
  wrapper, not the closure). hono classifies routes with `handler.length > 1`
  and labels them with `handler.name`, so every route reads as a
  non-middleware `[handler]`. `length` alone → 5/8; `length` + `name` → 7/8.
- [#5366](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5366-class-instance-field-from-constructor-option-reads-null)
  (1 test) — `new Hono({ router: new RegExpRouter() })` leaves
  `app.router === null`.

Neither is a variation of this bug, and #5365 changes `_wrapWasmClosure`, which
is on every host callback in every package — it needs its own A/B, not a rider
on this one.

**2. `check:dogfood-validation` does NOT catch this class, though the issue's
plan expected it to.** Measured directly: with the buggy source restored, the
gate exits 0. hono *is* in its gated set; the gap is its input. The gate
compiles each package's **declared entry module** (hono →
`package/dist/index.js`), and that entry never imports `utils/color.js` —
`getColorEnabledAsync` is reachable only through `helper/dev`, which only a
test file pulls in. The invariant it asserts is sound; its input set is
entry-module shaped, so a validation failure in any submodule the entry does
not reach is invisible to it. Widening it to the package's exported subpath map
would be a cheap follow-up, but it changes the gate's contract and belongs in
its own issue.

Also worth recording: the issue's Evidence section states this file registered
"native 0/0". Re-measured on a clean base through `compileAndRunUpstreamModule`,
the **native lane registers 8/8 and passes 8/8** — that observation was the
`tsx`-worker-spawn artefact, and it pointed the investigation at a harness
hypothesis that had no basis.


ttraenkler:
**Park-hold diagnosis (lead, 2026-09-06 13:25)** — the failing gate is `quality` → `Native-first host semantic debt gate (#4401)`: `runtimeTsLines 19332 > maximum 19313` (runs 34031297817 / 34031743998). That is a property of the **merge group's composition**, not of this PR: current `main` (`cbd2f11dff`) has `src/runtime.ts` at 19,154 lines against a ceiling of 19,185, and `pnpm run check:host-import-policy` passes there. The failed groups were built on top of other lanes' PRs that grew `runtime.ts` and raised the ceiling for *their* state only, so the cumulative group overshot. #5675 adds 14 lines (fits the 31-line headroom on main); #5676 adds none to `runtime.ts` and failed only because its group sat on top of #5675's.

The queue has already re-grouped both at the head on current main. Removing the bot `hold` so the backstop does not skip them if they are ejected again; **not re-enqueueing**. If the re-validation fails on this gate again, the fix is to move the 14 lines into a `src/runtime/` subsystem module (the #4401 ratchet only ever wants `runtime.ts` to shrink), not to raise the ceiling.

ttraenkler:
**Park-hold diagnosis (lead, 2026-09-06 14:05)** — not this PR. Its two failed groups were built on top of #5675 (runs [34036215568](https://github.com/loopdive/js2/actions/runs/34036215568): net −36 with the typed-array regressions that belong to #5675, see the diagnosis there; and [34037616388](https://github.com/loopdive/js2/actions/runs/34037616388): a `Temporal is not defined` / `until is not a function` cluster — the #5353 provider-linkage drift in its "provider failed to link" form). This PR touches an IIFE return-patch in `src/codegen/expressions/` and nothing in typed arrays or Temporal. #5675 has since been ejected, so the next group is this PR alone on current `main`. Removing the bot `hold` so the backstop can re-take it; not re-enqueueing.

PR #5679, opened 2026-09-06
docs(issues): file #5373 (Array-subclass toString bypassed — every linked Instant/ZonedDateTime read) and #5374 (valueOf never called across the seam)

ttraenkler:
Two issue files with implementation plans, measured through the test262 runner on the post-#5364 driver (batch now equals solo), fresh provider cache.

- [#5373](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5373-array-subclass-tostring-dispatch) — `String(x)`, `` `${x}` `` and any-typed `x.toString()` on a `class extends Array` instance run the built-in array join instead of the subclass override (`b.toString()` and `"" + b` are correct). `jsbi` is `class JSBI extends Array`, and the polyfill exposes every BigInt via `globalThis.BigInt(t.toString(10))`, so `Instant.epochNanoseconds` throws `Cannot convert 23396352,513294428,1 to a BigInt` and every ISO/UTC `ZonedDateTime.year`-style read fails `infinity is out of range`. This is the residual that [#5251](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5251-temporal-numeric-calendar-seam-families), [#5354](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5354-linked-class-instanceof-prototype-identity), [#5208](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5208-compiled-date-host-date-bridge) and [#5360](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5360-temporal-era-arithmetic-seam) each reported and left.
- [#5374](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5374-valueof-object-across-linked-seam) — a consumer object with `valueOf` handed to the linked provider coerces to 0 and `valueOf` is never called (consumer-only coercion is correct; the plain-number half of the same rows is correct). The `TemporalHelpers.toPrimitiveObserver` half of the `*-wrong-type` families and the 3 `infinity-throws-rangeerror` rows.

Docs-only; no code.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01K6r5kd3zWKXZrhsqALWVdk

---
_Generated by [Claude Code](https://claude.ai/code/session_01K6r5kd3zWKXZrhsqALWVdk)_

PR #5682, opened 2026-09-06
fix(runtime): a consumer object's valueOf/toString/@@toPrimitive is invoked when a linked provider coerces it (#5374)

ttraenkler:
Land order: after PR #5679 (docs, carries the issue file); independent of #5678/#5673/#5661 and of #5373.

Fixes [#5374](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5374-valueof-object-across-linked-seam). Files [#5376](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5376-accessor-valueof-read-through-field) for the half that is not a seam defect. Unpins one row of [#5242](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5242-class-value-construct-bridge).

## What was wrong

A consumer-minted object with a compiled `valueOf` / `toString` /
`[Symbol.toPrimitive]`, coerced **inside** a separately-linked provider,
coerced to `NaN` with the method never invoked. Downstream, the Temporal
polyfill's `ToIntegerWithTruncation` maps that non-number to 0, which is the
`RangeError: Cannot convert a number less than one to a positive integer` with
an **empty** observer `calls` array the issue was filed for.

[#5225](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5225-consumer-literal-seam) fixed the same seam for **field reads**. This is one step further in.

## Step 1 — which hypothesis held

Hypothesis **1** of the plan. Instrumented the `"[object Object]"` bottom-out
of `_hostToPrimitive` on the reduction, provider linked:

```
[5374] hint=number localNames="<no-helper>" ownerFound=true ownerNames="valueOf"
       localSgetValueOf=<no-export> ownerSgetValueOf=object/struct
       localCallFnM0=undefined  ownerCallFnM0=function
```

The provider's own `callbackState.getExports()` has neither
`__struct_field_names` nor `__sget_valueOf` nor `__call_fn_method_0` for a
consumer-minted struct; the #5225 registry names the owner immediately and the
owner's exports serve all three. So the walker was not failing to *call* the
closure — it never found one, concluded "no coercion method", and returned the
`"[object Object]"` sentinel. Hypothesis 2 (a raw non-callable closure struct in
the slot) is ruled out by the same line: with the owner's exports the slot reads
`object/struct` and dispatches.

## The fix

One `callbackState = _crossModuleCallbackState(raw, callbackState)` at the top
of each of the two ToPrimitive walkers in `src/runtime.ts` — `_toPrimitive` and
`_hostToPrimitive`. +13 lines, granted in the #5374 frontmatter with a dated
rationale.

At the **top** rather than at each of the ~10 probe sites, because the arms must
agree on a module: a `__sget_` from one module against a `__call_fn_*` from
another is a wrong answer, not a missed optimisation (the #5225 file makes the
same argument for the field-read path), and one substitution cannot drift out of
sync the way ten can.

**Miss path only, single-module lane untouched:** the #5225 registry
short-circuits on its `enabled` boolean below two live modules, and
`decoderFor` returns `undefined` — leaving the state unchanged — when the local
exports already decode the receiver. The control half of the new test passes on
base and after.

## Reduction (`tests/issue-5374-valueof-across-seam.test.ts`)

Two lanes over one shared provider/consumer source pair, throwaway npm package,
no Temporal in sight. Linked lane on base vs after:

| probe | base (linked) | after (linked) | control (both) |
| --- | --- | --- | --- |
| `Number(o.v)` / `+o.v` / `Math.trunc(Number(o.v))` with `v = {valueOf(){return 7}}` | `NaN` | `7` | `7` |
| `Number(x)` with `x` the object itself | `NaN` | `7` | `7` |
| `String(o.v)` with `v = {toString(){return "seven"}}` | `"[object Object]"` | `"seven"` | `"seven"` |
| `[Symbol.toPrimitive]`, number hint | `NaN` | `7` | `7` |
| `[Symbol.toPrimitive]`, string hint | `"[object Object]"` | `"seven"` | `"seven"` |
| `Infinity`-returning `valueOf` reaching a provider-side `if (!Number.isFinite(n)) throw new RangeError(…)` | `RangeError: … NaN` | `RangeError: … Infinity` | same |
| observer order, number hint | `"NaN:"` (calls empty) | `"3:valueOf"` | same |
| observer order, string hint | `"[object Object]:"` (calls empty) | `"x:toString"` | same |

9 of 9 wrong on base in the linked lane, control green on base — that is what
makes it linking-specific rather than an object-literal gap.

The `pFinite` probe carries the coerced value in its RangeError message on
purpose: base reaches the *same* `!Number.isFinite` throw with `NaN`, so a bare
"did it throw RangeError?" assertion passes on base and proves nothing.

## Measurement

417 unique rows = the 3 `infinity-throws-rangeerror` rows ∪ the 123-row family
list ∪ 294 `built-ins/Temporal/**` `*wrong-type*` / `*non-integer*` rows.
Driver `bucket-run.mts` through the runner's own `runTest262File`,
`JS2WASM_TEST262_TEMPORAL=1`, a **fresh** `JS2WASM_TEMPORAL_CACHE` per compiler
revision (both runs report `cacheHit=false`; cold builds 44.6 s base / 46.0 s
fix). Base first. Artifacts: `.tmp/base.tsv`, `.tmp/fix.tsv`, `.tmp/all-rows.txt`.

| run | pass | fail |
| --- | --- | --- |
| base (`origin/main` `cbd2f11d`) | 163 | 254 |
| fix | 163 | 254 |

Per-row join of the two TSVs: **0 rows changed status in either direction** —
0 pass→fail, and also 0 fail→pass. The 123-row family is 13 pass / 110 fail on
both sides; the 3 `infinity-throws-rangeerror` rows fail on both.

The issue's own reported Temporal rows, through the linked provider
(`.tmp/probe-temporal.js`, one file, base cache vs fix cache):

| row | base | fix | node |
| --- | --- | --- | --- |
| `Duration.from({hours:{valueOf(){return 2}}}).hours` | `0`, `valueOf` not called | **`2`**, `valueOf` called once | `2` |
| `PlainDate.from({year:2000,month:{valueOf(){return 3}},day:1}).month` | RangeError "…less than one…" | **`3`**, `valueOf` called once | `3` |
| `PlainDate.from({…,day:{valueOf(){return Infinity}}})` | RangeError "…less than one…" | **RangeError `invalid number value`** | RangeError `invalid number value` |
| same with an ACCESSOR observer | RangeError "…less than one…" | unchanged | month `3` |

## Reported, not fixed

**1. The accessor-backed observer — filed as [#5376](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5376-accessor-valueof-read-through-field).** `TemporalHelpers.toPrimitiveObserver`
mints `{ get valueOf() { return function () {…} } }` and every caller hands it
to a Temporal entry point as a **property of a bag**, so the polyfill reads it
out of a field before coercing. Measured in the plain single-module lane, no
linking and no Temporal: `f(o)` where `f(o){return Number(o.v)}` and
`o = {v:{get valueOf(){return () => 3}}}` answers **0** with the getter never
run, while the same object coerced *directly* answers 3 and the same field read
with a method-shorthand `valueOf` answers 3. Identical in both lanes ⇒ not a
seam defect ⇒ this PR cannot reach it. **This is why the 3
`infinity-throws-rangeerror` rows do not flip here**, so acceptance criterion 3's
first clause is not met by this PR. Bound: every `toPrimitiveObserver`-dependent
row, which includes `checkStringOptionWrongType` and `checkRoundingIncrement…`
across every Temporal type; the exact count is #5376's to measure, not claimed
here.

**2. An object literal whose ONLY member is a computed `[Symbol.toPrimitive]`
key.** `__struct_field_names` answers `null` for that shape in its own module
too, and that CSV is the #5225 registry's sole ownership oracle — so no module
claims the struct and there is nothing to redirect to. Adding any ordinary field
makes it nameable and it works; the test's `pToPrim*` probes carry a `tag: 1`
for exactly that reason. Bound: literals with no non-symbol member only —
classes and any literal with a data field are covered. Not filed separately: it
is a property of the #5225 oracle, and widening that oracle (e.g. to
`__is_data_struct`, which *does* answer 1 here) is a `ref.test`-ambiguity
question that belongs with #5225.

## #5242 unpinned

`tests/issue-5242-class-value-construct-bridge.test.ts` pinned
`directToString: "[object Object]"` as a known-bad row, reading it as the
`Symbol.toStringTag` gap (#5223). It was not that — `"[object Object]"` is also
what the ToPrimitive walker returns when it finds no coercion method at all, and
that is what was happening. With this change the row answers `"T1"`, so the pin
is removed and the linked lane equals `EXPECTED` exactly, with no pinned rows
left.

## Overlap note

dev-5373 is concurrently fixing Array-subclass `toString` dispatch in the
runtime's coercion paths. Both changes land in `src/runtime.ts` and may touch
`_hostToPrimitive`. This PR's change is confined to the **closure/decoder
lookup** — a single `callbackState` substitution at the top of each walker — and
adds no arm, removes no arm and reorders nothing inside either walker, so a
textual conflict with a dispatch-arm change should resolve by taking both.

## Gates

typecheck · lint · `check:loc-budget` (also with `LOC_GATE_BASE=origin/main`) ·
`check:func-budget` · `check:coercion-sites` · `check:oracle-ratchet` ·
`check:dead-exports` · `check:host-import-policy` ·
`check:issue-ids:against-main` · `node scripts/equivalence-gate.mjs`
(**22 failing / 1720 passing** vs the 24 / 1718 baseline, no new regressions;
2 baseline failures now pass and are left unratcheted so as not to collide with
another lane) · the 15 provider suites present in the tree
(4628, 5208, 5221, 5225, 5226, 5237, 5239, 5241, 5242, 5244, 5248, 5250, 5251,
5352, 5355 — 5354/5363/5364 have no test file on this base) · the 12
ToPrimitive suites (`issue-1917`, `2175`, `2358`×2, `2638`, `2679`, `2891`,
`3481`×2, `4429`, `5102`, `5342`): 158 pass, **4 failures in standalone lanes
that are byte-identical on base** (`2358 array-toprimitive`, `4429`×2, `5102`)
— re-ran them against `origin/main`'s `src/runtime.ts` to confirm.

---


ttraenkler:
Park diagnosis (run 34053072662, `quality` → host-import-policy): merged-state `runtimeTsLines` exceeded the ceiling measured before main moved — ceiling drift, not this diff. Re-merged `origin/main` (eda85d2cbe), ceiling set to the measured merged 19358, gate green. Released alone; #5691 (stacked on this) stays drafted until this lands.

---

PR #5683, opened 2026-09-06
fix(#5356): eager ref cell for a let/const mutated by a hoisted function

ttraenkler:
Fixes [#5356](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5356-eager-capture-box-skips-let-const) — a hoisted inner `function` that mutates an enclosing `let`/`const` had its ref cell minted at the first CALL site, so a call in an untaken branch left every later read of the binding `null`/NaN (prettier's `printDocToString`: `output` read `null` after the loop).

## What changed

- `emitEagerCaptureBoxes` (#2692) now boxes `let`/`const` (TDZ-flagged) captures at function top too, and records each cell in `FunctionContext.eagerCaptureBoxes` keyed by the RAW pre-hoisted slot — a key scope hiding cannot lose, unlike the name-keyed `localMap`/`boxedCaptures`.
- New `src/codegen/statements/eager-capture-box.ts` resolves that cell for the consumers that treated the raw slot as the binding's storage: the `switch` case-scope check, the #2814/#5271 block-`let` re-install, the lazy call-site mint under a shadowing block, and array-destructuring element stores (the object lane already did this, #4618).

## What the measurements said

- The "prettier miscompile" the issue attributed to the one-line skip removal is **not caused by it**: `printDocToString(["a"])` loops forever on the parent compiler too, inside `commands.push({ indent, mode, doc: doc[index] })`; hoisting that literal into a `const` makes the same file terminate. Filed as [#5375](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5375-prettier-print-doc-push-literal-loops) (carried in this PR). `print-doc-to-string` therefore stays 0/3 here; the #5357 branch had no commits when measured, so "with #5357" is the same number.
- The bare skip removal does regress exactly one shape — a `let` declared directly in a `case` clause and mutated by a clause-level function (`"6"` → `"1"`) — which is the race this PR fixes at its consumers; two shadowing-block shapes were already wrong on the parent and are fixed as a consequence.

## Evidence

- `tests/issue-5356-eager-capture-box-tdz.test.ts` over untyped two-file fixtures compiled as a project with the dogfood worker's options: **13/16 fail on the parent, 16/16 pass with the fix** (anti-vacuity control included).
- `tests/issue-2692-closure-box-eager.test.ts` + `tests/issue-2669.test.ts`: 22/22.
- A/B at one head over the 17 dogfood suites (table in the issue's `## Resolution`): 17/17 suites per-file identical (prettier 105/151, jest 335/356, hono 229/324, redux 66/82, lodash 58/62, axios 200/231, marked 9/30, three 17/18, webpack 16/16, clsx 32/32, cookie 63740/63740, tailwindcss 13/13, jsdom 6/6, styled-components 9/9, uuid 75/75, moment 10/10, stylelint 108/108 — both sides).

Full write-up: `## Resolution` in the issue file.


ttraenkler:
Recovery diagnosis: I am keeping the hold. The merged-state gate failure is real as an observed CI result, but an eager-capture-cell regression is not established yet.

- Effective baseline: `24681afc1ee2146e2317f81df2404a710f884e52`, predecessor run `34112219217`. Candidate: `2babfc21ec5b97a54e7f399d5c7ee232e77fdde5`, merging that baseline with this PR's `ff2834b804ddc8a939ae70b1322c5eb3ab0fca08`.
- The exact predecessor artifact, corpus/oracle, and skip count match. There is no stale-SHA baseline explanation in this evidence.
- The 18 reported regressions are all Temporal. Re-running all 31 regression/improvement paths against both exact snapshots on Darwin arm64, Node 25.9.0, four unified workers yields **30 pass / 1 fail on each**, with **zero status/error differences**. This does not reproduce Linux CI's preceding shard history.
- `Temporal/ZonedDateTime/prototype/with/object-must-contain-at-least-one-property.js` fails on both snapshots with `Expected a TypeError but got a RangeError`; the other 17 reported regressions pass on both.
- Six sampled primary/strict consumer binaries and independently cold-built local providers are byte-identical across the two snapshots. This is local A/B evidence, not a claim to have captured every CI binary. CI rows lack wasm hashes, so the gate's “with wasm-hash change” label does not establish observed byte differences.

Worker-history/runtime-state sensitivity is the leading hypothesis, not a proven cause. The next experiment is to replay and minimize an original failing shard's predecessor sequence on both snapshots, comparing fresh versus reused workers and recording linked-provider state. I will not narrow capture lowering, change a baseline/gate, or remove the hold solely from this bounded result.

Failed run: https://github.com/loopdive/js2/actions/runs/34114146529

ttraenkler:
Further diagnosis found a concrete mismatch in execution conditions: **both the original baseline and candidate host shard 3 had one worker fail to link the Temporal provider**, while three others linked successfully.

- Candidate run `34114146529`, job `101717347832`: `plan=bundled, reason=@js-temporal/polyfill does not expose Temporal`.
- Baseline run `34112219217`, job `101711286218`: `plan=none`.
- Both log that affected rows retain the ambient lane. `getWorkerTemporalProvider()` caches the initialization failure as null for subsequent requests in that worker.

This is positive evidence of mixed provider availability on **both** source snapshots, not a stale-SHA explanation or proof of an eager-cell miscompile. The initiating link failure remains unexplained.

A bounded 13-path Temporal prefix replay on both exact snapshots, fresh and reused workers, produced identical results in all four arms: **11 pass / 2 identical failures**, with 24 matching consumer binaries and 96/96 observed Temporal-global resets leaving no matching globals. It did not reproduce the unavailable-provider state. The replay used Darwin arm64/Node25.9 and one active worker, not CI's four-worker Linux execution.

The retained CI artifacts lack per-worker dispatch history and the failed initializer's input bytes/export analysis. I am preserving the hold while investigating initialization; these results do not justify a capture-lowering patch, a baseline change, or a weaker gate.

ttraenkler:
Read-only owner coordination: a separately held acquisition checkpoint #5795 at085cf77f9e8ee33b223bc6098c9c159d543e884d now addresses partial publication of Temporal dependency extraction. On different failed group01cdf592+that overlay, actual Node25.9 Darwin four-worker execution passed26Temporal rows+4controls. This is NOT clearance for your group2babfc21 or Linux history. Your retained baseline/candidate provider-initialization failures are relevant independent evidence. Please confirm whether you still own this diagnosis and can map the acquisition mechanism to original shard3 inputs/logs before any newly authorized run. No source adoption, tests, hold removal or queue admission requested/performed here.

ttraenkler:
Park-hold triage (PR-queue shepherd, 2026-09-12) — **drift, not a regression caused by this PR. Removing `hold`.**

Two auto-park comments, both cited runs read in full:

**1) run 34046488299 (`quality`, 2026-09-06) — #4401 group-composition drift**
```
runtimeTsLines 19323 > maximum 19318
```
That is the Native-first host semantic debt gate counting `src/runtime.ts` lines on the *merged* state. This PR's diff adds no lines to `src/runtime.ts`; the 5-line overage came from other PRs ahead in the same merge group. Known pattern (b).

**2) run 34114146529 (`check for test262 regressions`, 2026-09-07) — Temporal provider drift (#5353)**
```
=== Raw host improvements before canary quarantine: 13 ===
=== Host stable-path fine-gate net: -5 (13 improvements - 18 regressions) ===
=== GATE FAIL: net_per_test -5 < 0 (13 improvements - 18 regressions) ===
=== GATE FAIL: regression ratio 138.5% (18/13) meets/exceeds the 10% limit
```
All **18 of 18** regressed tests are under `test/built-ins/Temporal/` — Duration, PlainDate, PlainDateTime, PlainTime, PlainYearMonth, ZonedDateTime — with the drift signature errors (`Temporal is not defined`, `Expected a RangeError but got a TypeError`, `Expected a TypeError but got a RangeError`). **Zero** regressions outside Temporal across the whole corpus, and in particular none in closures / captures / let-const / TDZ, which is the only area this PR's diff touches (`src/codegen/statements/eager-capture-box.ts` and friends).

Cross-check on unrelated PRs in the same window shows the identical cluster, which is what makes this main-side drift rather than anything in this branch:
- run 34303910910 (PR #5784): 25 of 26 regressions under `test/built-ins/Temporal`, net -23
- run 34167945399 (PR #5735): 26 of 29 regressions under `test/built-ins/Temporal`, net -20

Action taken: `hold` removed, **not** enqueued (the server-side auto-enqueue owns that).

Note for whoever resumes this PR: the branch is **306 commits behind main** (3 commits of its own ahead). Merge `upstream/main` into it and push before it goes back through the queue, so the re-validation runs against a current baseline.

ttraenkler:
Refreshed onto current main and the park-hold's cited failure is **fixed, not waived**. Removing `hold`.

**The 2026-09-12 park (run 34674500787, `quality`) was REAL and in this PR's own area** — not the Temporal / `runtimeTsLines` main-side drift the 2026-09-12T04:57 triage found. The failing step is the #3518 compiler-boundary inventory:

```
node --max-old-space-size=2048 scripts/check-compiler-boundaries.mjs \
  --mode inventory --base HEAD^1 > compiler-boundaries-report.json
##[error]Process completed with exit code 1.
```

The step writes its verdict to the artifact, not the log, so the job log shows only the exit code. Reproduced locally against the merge base:

```json
"status": "invalid-inventory",
"errors": [
  { "code": "unclassified-module", "detail": "src/codegen/statements/eager-capture-box.ts" },
  { "code": "unclassified-target", "detail": "src/codegen/statements/eager-capture-box.ts" }   // x5
]
```

That gate landed on main after this branch was cut, and it requires every module under `src/` to carry a classification row. The one module this PR adds had none. Fixed by classifying it in `scripts/compiler-boundaries.json` with the row its `src/codegen/statements/` siblings carry (`unmigrated` / `mixed-needs-split` / `backend-wasmgc` / `3518-coordinator`). The gate now exits 0 with `"inventory-valid-architecture-incomplete"` — the status main itself reports.

`quality`'s other lanes were already green in that run (lint, format, typecheck, ir-dialect).

**Refresh**: `upstream/main` merged in clean — no `src/` conflicts, all seven touched files and the regression test carried through unchanged.

**Validation at the refreshed head**

| check | result |
| ----- | ------ |
| `tests/issue-5356-eager-capture-box-tdz.test.ts`, seven touched files reverted to main | 13 failed / 3 passed (16) |
| same test, with the fix | 16 passed (16) |
| prettier dogfood, base → fix | 105/151 → 105/151, per-file identical |
| jest dogfood, base → fix | 335/356 → 335/356, per-file identical |

A/B by file copies of the seven touched `src/` files at one head, both suites exit 0 with an `admitted` headline on both sides. Same verdict as the branch's original 17-suite table: **the fix is dogfood-neutral.** Its value is the correctness of the binding, proved by the regression test, not a suite delta — prettier's `print-doc-to-string` stays 0/3 because it is blocked by [#5375](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5375-prettier-print-doc-push-literal-loops), not by this cell.

`tests/illegal-cast-closures-585.test.ts` and `tests/issue-1528-closure-construct.test.ts` each fail one test at this head — **identical failures on base and fix**, so pre-existing on main, not this PR.

Gates all exit 0: loc-budget, func-budget (both with `LOC_GATE_BASE` pinned to the merge base; growth is covered by the issue file's existing `loc-budget-allow:` / `func-budget-allow:` grants), coercion-sites, oracle-ratchet (`getTypeAtLocation +0, ctx.checker +0` across 7 changed files), dead-exports, dogfood-validation (6/6 compiled, 6/6 validated), compiler-boundaries, typecheck.

**Also restored** `plan/issues/5375-prettier-print-doc-push-literal-loops.md`. The 2026-09-07 recovery merge deliberately excluded it, and it exists nowhere on main — so it was about to be lost with this branch. It is the issue that actually blocks prettier's `print-doc-to-string`.

Not enqueued — the server-side `auto-enqueue.yml` owns that.

PR #4524, opened 2026-08-15
feat(#4424): structure-tree GVN — scoped value numbering over the nested-buffer IR (flag-gated OFF)

ttraenkler:
## What this is (#4424, split from #4418's critical review)

Classical dominator-tree GVN re-homed onto this IR's actual substrate. Two nesting dimensions, one walk:

- **Block level**: dominator-tree DFS with scoped tables — consumes #4418's `dominanceOf`, so it is correct on today's join-free trees AND stays correct if a future producer emits joins.
- **Buffer level**: one fresh scope per nested if/loop/try buffer — the single uniform rule that encodes every structural safety condition (arm entries can't serve siblings or the join; loop-body entries die at body end; try entries can't leak past a throw). Enclosing-scope entries remain visible inside, which is sound for SSA values computed before the buffer.

**Fail-safe by construction**: the pass only RENAMES uses to the earlier id — the duplicate stays in place and `deadCode` sweeps it. A use the (exhaustive) renamer can't rewrite keeps the duplicate live and correct: a missed merge, never a miscompile. Nothing is deleted here, so the #1586 alloc-registry rules never engage; alloc-carrying instrs are excluded from merging anyway (identity). Admission runs through #2134's `effectsOf`/`effectsArePure` — a new instruction kind is a full barrier there, hence inert here by default.

**Gating**: `JS2WASM_IR_GVN` (`1`/`true`/`poison`), read inside `gvnFromEnv` so `runHygienePasses` stays a one-liner. **Default OFF** — the #4455 pattern: measured first, flipped separately.

## Verification

- `tests/ir-gvn.test.ts` 7/7 — including three scope-safety shapes built so a scope bug is *observable* (pure merges never change the merged value; a wrong scope surfaces as reading an unmaterialized local on the untaken path), and a **poison liveness control**: `JS2WASM_IR_GVN=poison` moves the result off the correct value, proving merges fire end-to-end.
- tsc clean; equivalence gate: no new regressions (flag off — the default path is structurally unchanged).
- Loaded-run IR-suite failures fully attributed: contention timeouts (pass solo) + 1 `ir-bytecode-proof` TypeError and 3 `ir-nullish` IR-FALLBACKs reproduced **byte-identically on unmodified HEAD and on pre-#4518 main** — pre-existing, not this diff.

## The honest measurement — and the AC left open on purpose

[...]

PR #5684, opened 2026-09-06
fix(#5357): `x === false` on a nullish reference compares by ===, not ToNumber

ttraenkler:
Fixes [#5357](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5357-nullish-reference-strict-equals-false) — `x === false` answered `true` when `x` was a nullish reference (and `x === true` answered `true` for an `x` holding `1`): the externref-equality tail in `binary-ops-typed-dispatch.ts` widened the Boolean to `f64`, put the reference through `__unbox_number` (ToNumber, `null → 0`) and compared with `f64.eq`. One of the two defects behind prettier's `is-empty-doc`; the idiom (`onEnter?.(doc) === false`, `(ret = reducer(...)) !== false`) is everywhere in libraries.

## What changed

- **`src/codegen/strict-eq-reference-arm.ts`** (new): `emitReferenceEqualityFromStack` boxes an `i32` scalar by its **brand** (`boolean` → `__box_boolean`, symbol → `__box_symbol`) and compares in the lane's reference arm — host lane via `emitHostEqualityFromStack` (coercion-engine.ts stays the single `__host_eq` emitter), native-first via `__extern_strict_eq` / `__any_eq`. `emitLooseScalarVsReferenceEquality` decides the nullish case of `==` / `!=` with `ref.is_null` before the ToNumber comparison (§7.2.15 steps 2-3).
- **`binary-ops-typed-dispatch.ts`**: the wrapper arm and the #1986 arm now call the helper (coercion-sites net `__host_eq -4`, `__host_loose_eq -1`; file 1610 → 1593 lines); the tail reroutes every strict pair with a statically-Boolean side — the only strict pairs that can reach it — and guards loose scalar-vs-reference pairs. A statically-numeric reference keeps the numeric fast path, as the plan asked.
- **`strict-eq-type-disjoint.ts`** (#4208's scalar module): `const u = undefined` was a *different* defect — `void`/`undefined` lowers to an `i32` slot holding 0, so `u === 0` was `0 === 0` in every lane. An `"undefined"` class decides it against Number/Boolean for `===` and `==`; refused for parameters (`function g(v = undefined) { return v === 0 }` carries what the caller passed).
- **`src/runtime.ts`**: `__defineProperties`' #2837 branch is gated to host targets. Found by the A/B — axios's `utils.freezeMethods` keeps every descriptor whose reducer result is `!== false`; on the parent that comparison dropped all of them (a silent no-op freeze); once correct, the descriptors hit that branch, which called raw `Object.defineProperty` on a compiled object ("WebAssembly objects are opaque", 12 of 34 axios unit files dead at module init, 200 → 73). A WasmGC target now takes the existing sidecar path.

Measured on `50c81e5487` with a 60-row probe through `compileAndRunUpstreamModule` (untyped `.js` two-file project) plus the same rows under `--target standalone` and `native-first`: the native lanes never reach the collapse (the #1776 tag dispatch returns for every externref pair), so the collapse was host-lane-only; the `const u = undefined` rows were wrong in all three lanes. Host 41/60 → 58/60, standalone 53/60 → 58/60, native-first 53/60 → 58/60; the two residuals in every lane are a deliberately failing harness control and a nested defaulted-parameter row that #5221 owns (unchanged from the parent). The plan listed the loose rows (`null == false`, `undefined == false`) as "unchanged" controls — they were wrong on the parent and are fixed here too.

## Verification

- `tests/issue-5357-nullish-reference-strict-equals-false.test.ts` — 22 rows × 3 lanes + the axios `freezeMethods` shape: parent 18 failed / 50 passed of 68; fix 68/68.
- Scoped test262 (`language/expressions/{strict-equals,strict-does-not-equals,equals,does-not-equals}`, 149 tests, both lanes, parent and fix): **gc 139/149 → 139/149, standalone 136/149 → 136/149**, identical residual sets (ToPrimitive / `Symbol.toPrimitive` / IsHTMLDDA / BigInt-width cases).
- Existing suites #1986, #4208, #1776, #2081, #1065, #2605/#2606, #4397, #5346, #4656, #2508, #739, #4524: all green except three that fail identically on the parent (`#4397` native `escape()`, `#4656`'s opaque-receiver `it.fails` residual, `#739`'s plain-descriptor guard).
- Dogfood A/B at one head, 17 suites, base vs fix, per file:

| package | base | fix | per-file change |
| --- | --- | --- | --- |
| lodash | 58/62 | 59/62 | `test.js` 58 → 59 |
| axios | 200/231 | 200/231 | none (73/231 with the codegen fix but without the runtime gate) |
| prettier | 105/151 | 105/151 | none |
| jest | 335/356 | 335/356 | none |
| hono | 229/324 | 229/324 | none |
| redux | 66/82 | 66/82 | none |
| marked | 9/30 | 9/30 | none |
| three | 17/18 | 17/18 | none |
| webpack, clsx, cookie, stylelint, tailwindcss, jsdom, styled-components, uuid, moment | 16/16, 32/32, 63740/63740, 108/108, 13/13, 6/6, 9/9, 75/75, 10/10 | identical | none |

All 17 suites exited 0 with an `admitted` headline in both runs; **0 regressed files**.

Prettier: `is-empty-doc` stays **10/16** (7 → 10 was #5346's), `print-doc-to-string` **0/3**, suite 105/151 — the plan's "≥ 15/16" is not reached by this fix. The six remaining rows all need `isEmpty = false`, written inside the `traverseDoc` callback, to be visible to the enclosing function afterwards; a minimal probe shows a `let` mutated in a callback passed *through another function* reads back stale (`viaTraverse → true`, Node `false`) while a direct callback is fine — #5356's eager capture box, exactly as the dispatch note said for `print-doc-to-string`. The `onEnter?.(doc) === false` gate itself is fixed (`guarded(() => {})` now visits).

- Gates: loc/func budgets, coercion-sites, oracle-ratchet, dead-exports, dogfood-validation green; typecheck (ts7) clean; no baseline files touched.


ttraenkler:
Operational note for the lead (not a code concern):

While measuring the dogfood A/B from my worktree I symlinked `tests/dogfood/.npm-upstream-suites` (and the hono/moment/lodash suite caches) to the copies under `.claude/worktrees/npm-unit-test-failures-420ca0/tests/dogfood/`. `setupPinnedUpstreamSuite`'s integrity check compares `git rev-parse --show-toplevel` (real path) with `path.resolve(root)` (symlink path); through a symlink they differ, so it treated every checkout as invalid and did `rmSync` + `git clone --depth 1 --branch <tag>` — **inside that other worktree's cache**. Affected there (all re-cloned at their pinned tags, `.git/HEAD` verified present afterwards, mtimes 14:24–14:26 today): axios, clsx, cookie, jsdom, marked, redux, styled-components, stylelint, tailwindcss, three, webpack; jest, prettier, typescript were not touched. Any non-git state inside those checkouts (none that I could see) would be gone. I replaced my symlinks with real copies before the runs reported here; the A/B numbers in the PR body come from those.

If anyone runs suites from that worktree and sees a surprise, this is the first thing to check.

PR #5691, opened 2026-09-07
fix(codegen): an accessor-bearing object literal stored as a property value is no longer dropped to null (#5376)

ttraenkler:
Land order: after PR #5682 (#5374, this stacks on it); PR #5679 (docs) has landed.

Fixes [#5376](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5376-accessor-valueof-read-through-field).

## What was wrong

`Number(o.v)` answered **0**, with the getter never invoked, for

```js
const o = { v: { get valueOf() { return () => 3 } } };
```

Not a ToPrimitive defect, despite presenting as one: **`o.v` was `null`**, and
`Number(null)` is `0`. That is why the failure was a silent zero rather than a
throw, and why an observer's `calls` array came back empty.

### Where the accessor was lost — at the STORE, not at the walker

`compileObjectLiteralWithAccessors` (#1239) builds **every** accessor-bearing
object literal as a HOST externref, so V8 sees real accessor descriptors.
TypeScript, however, types the *enclosing* property from the getter's RETURN
type, so `ensureStructForType` picked `(ref null $__anon_0)` for the field `v`.
The guarded store then emits:

```wat
(type $__anon_0 (struct (field $valueOf (mut eqref))))
(type $__anon_1 (struct (field $v (mut (ref null 6)))))   ;; 6 = $__anon_0
(func $mk
  call 1                    ;; __new_plain_object -> externref (HOST object)
  local.tee 1
  ... __defineProperty_accessor "valueOf" ...
  local.get 1
  any.convert_extern
  local.tee 3
  ref.test (ref 6)          ;; a host object is NOT a $__anon_0 -> 0
  (if (result (ref null 6))
    (then local.get 3 ref.cast null (ref null 6))
    (else ref.null 6))      ;; <<< the accessor object is DROPPED here
  struct.new 7)
```

`ref.test` can never succeed for a host object, and the `else` arm writes
`ref.null`. The proof that the walker was never involved: give the *outer*
literal its own accessor (`{ get w() {…}, v: { get valueOf() {…} } }`) and the
identical `Number(o.v)` answers **3** — host path all the way down, nothing to
drop.

## The fix

`src/codegen/index.ts`, the anon-struct field-typing loop in
`ensureStructForType`: widen the field to `externref` when the property's VALUE
is an accessor-bearing object literal. Predicate extracted to
`src/codegen/accessor-value-field.ts` (index cannot import `literals.ts` —
cycle).

This is **#1589A one step further**: that guard already widens an EMPTY object
literal in the same loop, for the same reason (host externref at runtime,
`ref.test` fails, `ref.null` stored). Method-shorthand values keep their struct
field type byte-for-byte; direct coercion (`Number({ get valueOf() {…} })`)
never reaches this code.

+7 lines in `index.ts` (+6 in the function), granted with a dated rationale in
the issue frontmatter; the predicate module is 41 lines.

## Measurement

Base = this branch with only the `ensureStructForType` hunk reverted, so
`src/runtime.ts` keeps [#5374](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5374-valueof-object-across-linked-seam)
and the single variable is this change.

### Behaviour (host lane, `tests/issue-5376-accessor-valueof-field-read.test.ts`)

| row | base | fix | node |
| --- | --- | --- | --- |
| `num({ v: { get valueOf() { return () => 3 } } })` | **0** | **3** | 3 |
| same with the call observer | **`0:`** (calls EMPTY) | **`3:g,c`** | `3:g,c` |
| `num({ v: { valueOf() { return 3 } } })` (method shorthand, control) | 3 | 3 | 3 |
| `Number({ get valueOf() { return () => 3 } })` (direct, control) | 3 | 3 | 3 |
| `str({ v: { get toString() { return () => "seven" } } })` | seven | seven | seven |
| getter runs once per coercion (`1,2`) | — | `1,2` | `1,2` |

All three lanes — single module, two modules in one compilation unit, separately
linked provider (`linkPlan.mode === "separate"`, asserted). On base the file
fails in all three with exactly `fieldReadValueOf: 0` / `observerCallOrder: "0:"`.

The test also pins three **pre-existing** accessor-walker residuals to the value
the compiler actually produces (`get [Symbol.toPrimitive]()` → NaN, a throwing
getter swallowed, `toString` getter run twice) and asserts each equals its
**no-field-read twin** — so the "not ours" claim is testable, and if the walker
is fixed later both sides flip together.

### test262 (bounded, base vs fix, per row)

170 rows: the 3 named `intl402/**/infinity-throws-rangeerror.js` plus
`grep -rl "toPrimitiveObserver\|checkStringOptionWrongType\|checkRoundingIncrement" test262/test/built-ins/Temporal | head -400`
→ **167** files (the cap never binds). Through `tests/test262-runner.ts` with a
one-row-per-test driver, `JS2WASM_TEST262_TEMPORAL=1`, a FRESH
`JS2WASM_TEMPORAL_CACHE` per side. Never the full bucket.

| side | rows | pass | fail | provider | artifact |
| base | 170 | 1 | 169 | cold, `cacheHit=false`, built in 47032 ms | `.tmp/base170.tsv` md5 `fbf36f04a57ed7d92254dabe312c19ed` |
| fix | 170 | 1 | 169 | cold, `cacheHit=false`, built in 53808 ms | `.tmp/fix170.tsv` — **same md5**, the files are byte-identical |

**0 pass→fail. 0 fail→pass.**

## Reported, not fixed

**The issue's Temporal attribution was wrong, and this PR says so rather than
quietly dropping it.** Its acceptance criterion 2 asked for the 3
`infinity-throws-rangeerror` rows to pass. They do not, and they could not:

- Each fails at its **first** assertion (L15 / L15 / L14) —
  `assert.throws(RangeError, () => …from({ ...base, eraYear: Infinity }))` with
  `era: "ad"`, `calendar: "gregory"`. A raw `Infinity`, no observer in sight.
  The `toPrimitiveObserver` lines (L17–L20) are never reached. That is an
  era/eraYear non-ISO-calendar gap, unrelated to this change.
- `toPrimitiveObserver` builds its object inside a **function** and returns it;
  the test binds it to `const obj` and writes `{ ...base, eraYear: obj }`. The
  property's value is an identifier, not an inline literal. Measured on base:
  `num({ v: mk() })`, `const t = mk(); num({ v: t })`, `Number(t)` and
  `num({ ...base, v: t })` all already answered 3. **Only the inline-literal
  store was broken.**

Separately measured and still open: an accessor-backed `valueOf` handed to the
**linked Temporal provider** —
`Temporal.PlainDate.from({ year: 2000, month: <accessor observer>, day: 1 })` —
answers `RangeError: Cannot convert a number less than one to a positive
integer` **identically on base and fix**, while method-shorthand `valueOf`
through the same seam answers 3. A seam-side accessor gap that neither #5374 nor
this covers. **Bound: one probe (`.tmp/probe262/observer.js`); the family size is
unmeasured** and needs its own issue with its own base-vs-fix row count.

Also unfixed, same null-drop mechanism, deliberately out of scope: the other
`objectLiteralForcesHostPath` arms nested as a property VALUE — runtime computed
key, `[Symbol.dispose]`/`[Symbol.asyncDispose]`, empty-string key, colon-form
`__proto__`, and spread-in-non-specific-context. **Bound: unmeasured**; each
needs its own base-vs-fix count before the predicate is widened.

## Gates

typecheck · lint · `check:loc-budget` (also `LOC_GATE_BASE=origin/main`) ·
`check:func-budget` · `check:coercion-sites` · `check:oracle-ratchet` ·
`check:dead-exports` · `check:host-import-policy` — all OK.

`equivalence-gate`: **22 failing / 1720 passing** against a 24-known-failure
baseline, no new regressions. (Two unrelated `optional-direct-closure-call`
baseline rows now pass; left un-ratcheted — that is a main-side ratchet.)

| suites | result |
| --- | --- |
| providers #5221 #5225 #5226 #5237 #5239 #5241 #5242 #5244 #5248 #5250 #5251 #5352 #5355 #5374 #4628-temporal-global | 15 files, 91 tests, all pass |
| ToPrimitive #1917 #2175 #2358 #2638 #2679 #2891 #3481 #4429 #5102 #5342 | 30 files, 355 pass / 8 skipped / 4 fail |

The 4 are standalone-lane and **pre-existing** — re-run with only
`src/codegen/index.ts` reverted to base, the same 4 fail (#2358 string-element
join, #4429 ×2 `this`-binding, #5102 getter-abrupt control).

`plan/audit/host-import-policy-baseline.json` `maximumRuntimeTsLines`
19321 → **19334**, the measured `wc -l src/runtime.ts` after merging main
(main 19321 + #5374's +13); the host-import-policy gate re-run green against it.

---


ttraenkler:
### Park triage: 47 `pass → fail` rows are COLLATERAL, not a regression

Measured row-by-row, both sides, provider-linked lane (`JS2WASM_TEST262_TEMPORAL=1`,
fresh `JS2WASM_TEMPORAL_CACHE` per side, both cold-built `cacheHit=false`), on the
CI-pinned test262 sha `b363f29d`. PR tree = this branch merged with `origin/main`
e48b3de196; base = the same tree with only the two PR `src/` files reverted, so the
compiler revision is the only variable.

| side | rows | pass | fail |
| --- | --- | --- | --- |
| PR tree | 47 | 46 | 1 |
| `origin/main` | 47 | 46 | 1 |

The two result files are **byte-identical**, reason strings included
(md5 `eb68aa2f3a571feead55b8a99fd94c3c`): **0 pass→fail, 0 fail→pass**.

- All 45 `built-ins/Temporal/**` rows pass on both sides.
- `Temporal/PlainDateTime/limits.js` — the row the queue reported flipping into the
  `oob` trap category — is `pass` on both sides, solo and in batch. **No trap.**
- `Duration/prototype/add/float64-representable-integer.js` (precision assertion,
  the other genuinely-PR-plausible row) passes on both sides.
- The one in-batch failure, `TypedArrayConstructors/ctors-bigint/object-arg/new-instance-extensibility.js`
  (`cannot marshal opaque compiled value to host BigInt64Array constructor`), occurs
  **identically on base and on this PR**, and **passes solo on both sides**. It is
  fork-level realm-mutation collateral (#1957 class) — 46 preceding Temporal rows in
  the same process poison realm globals — reproduced directly in this measurement.

Same class as the #5673 and #5685 parks this window. The wasm-hash change on every
row is expected of a codegen PR and is not evidence of a regression.

Clearing the park hold and re-admitting once (per queue rules).

---

PR #5688, opened 2026-09-06
feat(es2015-standalone): wave 5 — Proxy/Reflect attribute model, super property reads, computed accessor keys, Reflect.construct NewTarget, lib.dom shadow

ttraenkler:
## Description

Wave 5 of the ES2015 standalone close-out ([#4444](https://js2wasm.loopdive.com/dashboard/issue.html?slug=4444-es6-standalone-edition-closeout-umbrella)). Five lanes, each implemented by an Opus/Sonnet subagent from a Fable-written plan, then adversarially reviewed (one reviewer, two skeptics per finding) and fixed in as many reviewed rounds as the reviewer kept finding real defects. Every row figure below was measured with `scripts/run-test262-paths.mts --isolate --standalone` against a `git archive` base tree with its own compiler bundle and quickjs adapter; the promoted baseline is authoritative.

| lane | shipped | owned rows (base → this branch) | control | review rounds |
| --- | --- | --- | --- | --- |
| [#5316](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5316-es2015-standalone-proxy-r4-invariants-receiver) Proxy r5 | `__integrity_bag` learns the #4194 instance carrier; the gopd literal-key fold asks the dynamic native on a guard miss; `in` stops folding over a directly-bound Proxy; the two §10.5 extensibility clauses restored; `Object.preventExtensions`/`setPrototypeOf` honour a false trap status; `Reflect.set(target, key, value, receiver)` as §10.1.9.2 (receiver-Proxy through `[[DefineOwnProperty]]`, target-Proxy through its `set` trap with the receiver, non-Object target TypeError) | +16 (Proxy+Reflect 464 rows: 350 → 366) +1 (Object integrity 317 rows) | 464 + 317 rows, 0 lost | review → fix round → clean |
| [#5350](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5350-es2015-standalone-super-property-r1) super property r1 | class `[[HomeObject]]` read, base-before-key element read, `extends null` TypeError, uninitialised-`this` guard (lexical, plus a runtime `__super_done` flag inside loops), object-literal `super.m()` incl. accessor bodies, `__proto__:` literal links its runtime prototype, callable check on super calls | +8 on the 53-row super control (18 → 26), 6 / 13 target rows | 53 rows 0 lost; the 1,089-row class/super control on this tree: see Validation | review + 5 fix rounds |
[...]

PR #5696, opened 2026-09-07
fix(#5316): an accessor define on an existing key of a non-extensible carrier no longer throws — the 2-row Annex B regression #5688 introduced

ttraenkler:
## Description

Fixes the one regression the wave-5 PR #5688 introduced on `main`: two Annex B rows, `built-ins/Object/prototype/__defineGetter__/define-non-extensible.js` and `__defineSetter__/define-non-extensible.js`, went pass → fail (whole-corpus diff of the standalone baseline promoted after #5688 against the one before it: +46 gains, these 2 losses). Owned by [#5316](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5316-es2015-standalone-proxy-r4-invariants-receiver); implemented by an Opus subagent from a Fable-written plan, then adversarially reviewed by a separate Opus reviewer (38 probes × 3 targets, no regression from this commit).

**Mechanism.** #5316 taught `__integrity_bag` the #4194 instance carrier, so `Object.preventExtensions` on an object literal or class instance now *records* `OBJ_FLAG_NONEXTENSIBLE` where it used to be a silent no-op. That made `__defineProperty_accessor`'s §10.1.6.3 step-2 arm reachable for those receivers, and the arm decided "new key" from `__obj_find` on the substituted bag, which cannot see an own data property that lives as a physical struct field. `Object.preventExtensions({ existing: null }).__defineGetter__('existing', f)` therefore threw "Cannot define property, object is not extensible" for a key that exists.

**Fix** (accessor arm only, `src/codegen/object-runtime-descriptors.ts`): before refusing, ask whether the receiver owns the key through the own-only `__hasOwnProperty` (not the `__desc_has_own` chain, whose last arm walks the prototype and would have stopped a genuinely new inherited key like `toString` from throwing). Owns it and the bag is sealed/frozen → the data→accessor conversion throws the §10.1.6.3 step-7 redefine TypeError (node parity that `main` reached for the first time and this keeps); owns it otherwise → fall through to the insert, shadowing the field exactly as before #5688; does not own it → the not-extensible TypeError as now. The symmetric guard on the data arm was written, measured and reverted: it silenced the correct throw for `Object.defineProperty(Object.freeze({existing:null}), 'existing', {value: 2})`.

### Rows and controls (standalone, `scripts/run-test262-paths.mts --isolate`)

| set | result |
| --- | --- |
| the 2 regressed rows | fail on `main` → **pass** |
| 2,054-row control: `Object/prototype/{__defineGetter__,__defineSetter__,__lookupGetter__,__lookupSetter__}`, `Object/{defineProperty,defineProperties,preventExtensions,seal,freeze,isExtensible}`, `Reflect/defineProperty` | pass 2,034 vs baseline 2,032: exactly the 2 gains, **0 lost**, no other row moved |

Host: 26 probe programs byte-identical (sha256) with the file reverted — the JS-host lane routes `defineProperty` through the sidecar import and is untouched.

### Validation (on the PR head)

- TS7 typecheck, lint, prettier; the five source-ratchet gates bare and with `LOC_GATE_BASE=origin/main` (growth granted in the #5316 frontmatter with a dated rationale); `check:speculative-rollback` / `stack-balance` / `codegen-fallbacks` / `any-box-sites` / `host-import-policy` / `harness-compile-budget` / `ir-adoption` / `done-status-integrity`: all green.
- All 8 equivalence shards green.
- `tests/issue-5316-r6-nonextensible-existing-key-accessor.test.ts`: 26 pins (node-parity probes incl. the sealed/frozen throws, the 2 rows, 6 sibling control rows; `result.imports` asserted `[]`), green on node 22 and node 25; the #5316 r4/r5 pin files unchanged and green.

### Residuals (recorded in the issue file)

- `--target wasi` keeps PR-1's answer for the same shapes: the guard is emitted but `__hasOwnProperty` does not reach a struct-field key on the nativeStrings lane (`hasOwnProperty.call` on a plain literal even traps there). No test262 row is at stake; the follow-up is the wasi own-key ladder.
- `Reflect.defineProperty` of an accessor over an existing key is a silent no-op (pre-existing; PR-1 had turned it into a trap, which this removes) and over a NEW key of a non-extensible object traps instead of returning `false` (the throw is right, the `Reflect` wrapper's catch is missing).
- An accessor→accessor redefine on a non-extensible object answers 1 where node answers 2 (pre-existing).

## CLA

---

### DIFF
--- a/plan/issues/5316-es2015-standalone-proxy-r4-invariants-receiver.md
+++ b/plan/issues/5316-es2015-standalone-proxy-r4-invariants-receiver.md
@@ -1746,3 +1746,22 @@ lifted out for twenty instructions. No `scripts/*-baseline.json` touched.
   `const s: any = Object.preventExtensions({existing:null})` answers `null`
   where the `var s = …` spelling answers `9`. Same family as the `vd` no-op;
   the pin uses the `var` spelling and this is recorded rather than fixed.
+
+- **WASI target: the same regression is NOT repaired there (round-6 review,
+  2026-09-07).** On `--target wasi` the guard is emitted (the wasi binaries
+  differ from main) but inert: `__hasOwnProperty` answers false for a physical
+  struct-field key on the nativeStrings/wasi lane — the own-key ladder that
+  makes it complete for #4194 carriers does not reach those receivers there,
+  and `Object.prototype.hasOwnProperty.call({existing:null}, 'existing')` even
+  TRAPS on wasi (reviewer probe `g1`; `111` on standalone and node). So
+  `Object.defineProperty(Object.preventExtensions({existing:null}), 'existing',
+  {get, configurable:true})`, its `Reflect.defineProperty` /
+  `Object.defineProperties` spellings and the class-instance receiver (probes
+  `v4`/`f2`/`f3`/`f5`, `/home/user/js2/.tmp/rev5316r6/p`, harness `wasi.mts`)
+  answer node `1` / pre-PR-1 base `1` / main `7` / this fix `7` on wasi. The
+  38-probe wasi sweep shows this fix changes NO wasi answer; the four cells are
+  PR-1's, inherited. test262 conformance runs standalone, so no row is at
+  stake; the follow-up is the wasi own-key ladder (`__hasOwnProperty` for
+  closed-struct carriers under `nativeStrings`), a wasi-lane task outside the
+  ES2015 standalone goal — filed here rather than as a new issue because the
+  id allocator could not scan open PRs from this container.
--- a/plan/issues/5316-es2015-standalone-proxy-r4-invariants-receiver.md
+++ b/plan/issues/5316-es2015-standalone-proxy-r4-invariants-receiver.md
@@ -5,7 +5,7 @@ status: done
 completed: 2026-09-06
 sprint: current
 created: 2026-09-04
-updated: 2026-09-06
+updated: 2026-09-07
 priority: high
 horizon: xl
 feasibility: hard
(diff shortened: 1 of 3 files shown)


ttraenkler:
Diagnosed the auto-park before touching the label. The failing check is `check for test262 regressions` on the REBUILT merge group (run 34079461502, base `72e0f9342a` = the benchmark-artifact refresh that landed after #5694). Verdict: a nondeterministic Temporal host cluster, not this PR's regression.

Evidence:

- **All 28 "regressions" are `built-ins/Temporal/*` rows** (host target), with errors like `Temporal is not defined`, `since is not a function`, rounding-mode string mismatches. This PR's only source change is the accessor-define arm in `src/codegen/object-runtime-descriptors.ts`; the Temporal provider and those rows never reach it (test262-relevant changed paths: 1).
- **The same PR content passed the gate one run earlier.** The first merge group (run 34078991925, base `4fba6558c0`, the same `runs/4fba6558c0….json` baseline) reported net **+2** (exactly the two Annex B rows this PR restores) with 10 Temporal flips; the rebuilt group reports 28 different Temporal flips and 8 improvements. Only 3 rows overlap between the two lists — the Temporal bucket is flipping run-to-run on the runners (the gate's own host-canary quarantine already lists Temporal rows as noise).
- **Local A/B, host target, the 26 flagged rows the job log names**: `scripts/run-test262-paths.mts --isolate` on this PR's head and on a `git archive` of main `72e0f9342a` (bundles rebuilt) answer identically — 19 pass / 7 fail, the same 7 files on both trees.

Per the auto-park rules this is a confirmed flake/collateral, so I am removing `hold` once for re-admission (auto-enqueue takes it from there; no manual enqueue). If the Temporal cluster recurs on the next group, the follow-up is a `scripts/test262-host-noise-quarantine.json` entry citing both runs, as the gate's own text prescribes.

---

---

PR #5723, opened 2026-09-07
fix(standalone): keep the host-free exception renderer exports (#5384); three defects the Temporal polyfill's __module_init hits (#5383 S2)

ttraenkler:
Land order: the #5383 S1 PR (#5721) has landed and is merged into this branch.

Two linked pieces of the standalone lane:
[#5384](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5384-standalone-exn-render-exports-missing)
(the host-free exception renderer was unreachable) and the S2 slice of
[#5383](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5383-standalone-temporal-provider)
(what the Temporal polyfill's `__module_init` actually throws). The first is
what made the second diagnosable at all.

## #5384 — a throwing standalone module now exports `__exn_render_*`

Root cause was NOT the emitter. `stripHostBridgeExports` (#4035) listed
`__exn_render_` among the JS host-bridge prefixes, and `hostBridge: "auto"`
resolves to `"off"` for standalone/WASI — so the emitter pushed the exports and
the policy sink deleted them again from every deployed binary. That is why every
gate inside `emitExceptionRenderExports` measured as passing. Effect: every
host-free throw rendered as the opaque `"uncaught Wasm-GC exception
(non-stringifiable payload)"` (#2870).

The pair needs no host import, so in standalone it is the only way any host — JS,
wasmtime, or another Wasm module — can read a payload it caught through the
deliberately-kept `__exn_tag`. It is now kept when the source has a `throw`
statement.

**Not gated on `ctx.exnTagIdx >= 0`**: the export-boundary bridge arms the tag for
essentially every standalone module before user code is read, so a tag-gated keep
republishes `__any_to_string` → `number_toString` → the Ryu tables everywhere —
measured `-O3` standalone, an arith-only `run(n){return n}` goes **6,076 →
49,032 B**. That is #4034's cascade. With the source-`throw` gate the cost is
**+140 B** (conditional throw), **+164 B** (top-level throw), **+157 B** (#4035
`REALISTIC`), and a module without a source throw is byte-identical.

`gc` lane byte-identical: sha256 unchanged across six binaries (three modules ×
default/`-O3`) in an A/B file-copy revert; structurally, the sink returns early
when the bridge is published.

## #5383 S2 — three defects the polyfill's init hits

Probe: linked jsbi + polyfill (157 KB) → `compileMulti(… target: "standalone",
hostBridge: "off", deferTopLevelInit: true)` → **2.92 MB, zero imports**,
instantiates with `{}`, then `__module_init`.

| # | reported | root cause | fix |
| --- | --- | --- | --- |
| R3 | `TypeError: called value is not a function` | `registerModuleClassStaticAssignments` admitted only a statement whose whole expression is `=`. A minifier writes `JSBI.__clz30 = …, JSBI.__imul = …` as ONE comma statement → no static value cell; `class JSBI extends Array` has no class-object singleton, so the write went through `null` | flatten top-level comma operands (admission-only) |
| R4 | `Math.imul is not yet implemented in --target standalone` | the value body READS `__any_from_extern`/`__any_to_f64`/`__box_number` and nothing registered them (a body emitter must not register mid-body, #2704), so EVERY `Math.<fn>` value read declined — including #4565's own transcendentals; and the short-kernel names had no body at all | pre-register the substrate before the closure is built; add inline kernels (abs/floor/ceil/trunc/sqrt/fround + clz32/imul via the exact §7.1.6/7.1.7 decomposition). `round`/`sign` excluded on purpose — `f64.nearest` ties-to-even ≠ §21.3.2.28 ties-toward-+∞ |
| R5 | `Cannot access property on null or undefined` at jsbi's `var _ = Math.floor; … _(i)` | the plain-alias set covered only variadic statics, and `identifierIsWrittenTo` was a file-wide SPELLING test — a minified bundle binds `_`/`t`/`g` in hundreds of scopes and assigns most of them | widen to Math names that have a real body; make both soundness gates scope-aware via an optional `sameBinding` predicate resolved through `ctx.oracle` (unresolvable ⇒ still counts as a write) |

Side effect worth noting: `[1, 4, 9].map(Math.sqrt)` and `var f = Math.sin; f(0)`
work on this lane for the first time.

**Where it still stops** (recorded in the issue, with the reproduction):
`TypeError: Cannot access property on null or undefined at 1:4117` — jsbi's
`static subtract(i,_){const t=i.sign; …}` reached with a **null argument**, not
reducible in isolation (`subtract`/`add`/`unaryMinus`/`.sign` all answer
correctly on the jsbi prefix alone). The S2 smoke test
(`Temporal.PlainDate.from(…)`, `Duration.total`) is deliberately **not** written:
it cannot pass while init throws, and a skipped assertion would hide that.

## Tests

- `tests/issue-5384-standalone-exn-render-exports.test.ts` (new, 5) — exports
  present, `renderHarnessThrownText` returns `"TypeError: boom from top level"`
  on a zero-import module, WASI parity, the #4034 floor, and the gc lane.
- `tests/issue-5383-standalone-temporal-provider.test.ts` (+6) — the R3/R4/R5
  reductions plus a negative case (a genuinely reassigned alias still declines).
- `tests/issue-4035-host-bridge-policy.test.ts` — records the one deliberate
  carve-out, and replaces a stale absolute size bound with the policy delta it
  actually guards.

## Two pre-existing red size guards, NOT from this PR

Measured by A/B against this branch's own base: #4034's arith bound asserts
`< 5,000 B` and measures **31,592**; #4035's `REALISTIC` bound asserted
`< 20,000 B` and measured **33,136** before this PR's +157. Neither file is in a
required gate, which is how ~13–27 kB of standalone growth landed unnoticed.
Documented in the #5384 issue file; the underlying growth needs an owner.

## Gates on the merged tree (main `10312a066b` merged in)

typecheck · `check-loc-budget` and `check-func-budget` with `LOC_GATE_BASE=origin/main` (net +273 LOC, all granted in the #5383 / #5384 issue files) · `check:dead-exports` · `check:host-import-policy` · `equivalence-gate` 22 fail / 1720 pass / 0 new · #5383 (13) + #5384 (5) suites, all green.

---


ttraenkler:
Post-merge note on the merge-group park (run 34151124043, job 101840287305): 24 `pass → fail` / 20 `fail → pass`, net −4, all but one row under `built-ins/Temporal/**` with the #1957 fork realm-mutation signatures (`Temporal is not defined`, `Cannot read properties of null (reading 'since'|'until'|'subtract')`, `Expected a RangeError but got a TypeError`), plus `TypedArrayConstructors/ctors-bigint/object-arg/new-instance-extensibility.js` — the same row that flipped in the #5673 and #5691 parks and passed solo both times. Same class as those three parks (each proven collateral by byte-identical A/B and solo re-runs); this PR's `gc`-lane bytes were A/B-identical for the renderer change and the diff touches no Temporal path. The queue merged the PR regardless. Recorded in `plan/agent-context/temporal-standalone-handover-2026-09-07.md`'s successor session as a known host-cluster churn; if the next promoted baseline still shows these rows red, treat as real and bisect R4/R5 (`Math` value reads / aliases) first.

---

PR #5773, opened 2026-09-08
fix(codegen): the sidecar defined-property key is scope-aware — every `e.length` in a module no longer reads another binding's descriptor (#5383 S2e)

ttraenkler:
Land order: after PR #5767 (S2d) — landed.

Slice **S2e** of [#5383](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5383-standalone-temporal-provider) — a real `Temporal` global for `--target standalone`. One of the three remaining stops is fixed at the root; the other two are re-measured and narrowed, with everything recorded in the issue file so the next lane does not re-derive it at ~55 s per compile.

## R10 — `invalid calendar identifier` was a scope-blind compiler key

`ctx.sidecarDefinedPropertyKeys` is keyed by `"<identifierTEXT>:<propName>"` — module-wide, no scope discrimination. The compiled `@js-temporal/polyfill` bundle contains one `Object.defineProperty(e, "length", …)`, so **every** `e.length` read anywhere in the module was routed to the runtime descriptor read. For any *other* binding named `e` the sidecar has no descriptor, so the read answered `undefined`.

The victim is the polyfill's ASCII-lowercase helper `function Ao(e){let t="";for(let n=0;n<e.length;n++){…}return t}`: its loop bound read `undefined`, the loop never ran, `Ao("iso8601")` answered `""`, and `new Temporal.PlainDate(2024,1,1)` threw `RangeError: invalid calendar identifier ` — with an empty tail, which is the whole diagnosis.

Identified by measurement, not deduction: three byte-identical copies of the same function appended to the real bundle, differing only in the local's NAME — `zqx` → 9, `t` → 9, `e` → **0** — then the dispatch chain instrumented, which printed `sidecar=true key=e:length` for the `e` copy only.

**Fix** (`src/codegen/sidecar-owner-scope.ts`, new): each writer records the *declaration node* the key was recorded for, via `ctx.oracle.valueDeclarationOf` (not the raw checker), and the two readers in `property-access.ts` decline when the receiver provably resolves to a different declaration. Answer-preserving by construction — a key with no recorded owner, or an unresolvable receiver on either side, keeps its old module-wide meaning, so the change can only ever REMOVE a wrong-binding match, never add one. The two `object-shape-widening` writers hold a variable NAME rather than a node, so they mark their keys unscoped and are deliberately unchanged.

Measured on the real provider source (shim + bundle, `--target standalone`, `hostBridge:"off"`):

| probe | base | after |
| --- | --- | --- |
| `Ao("iso8601")` | `""` (length 0) | `"iso8601"` (length 7) |
| `zo("iso8601")` | throws `invalid calendar identifier ` | `"iso8601"` |
| `new qi.PlainDate(2024,1,1)` | `invalid calendar identifier ` | reaches the NEXT stop |

**JS-host (`gc`) lane byte-identical**: sha256 A/B over five modules including the reduction itself — same hashes, same lengths, before and after.

## What is measured but NOT fixed here

- **The boundary miss is not what S2d recorded.** A four-variant reduction through a real linked provider (host-free `instantiateLinkedProject(result, {})`) shows numbers and nested objects crossing correctly while a **class VALUE** does not: a plain/frozen object literal loses it entirely (`undefined`); the polyfill's own `Object.freeze({__proto__:null, …})` shape keeps it but reports `typeof "object"`, so `new Temporal.PlainDate(…)` is unreachable from a consumer. Two sub-defects, not one.
- **`invalid receiver` narrowed to one sentence.** Every component of the polyfill's brand check was measured to work (slots bag present with 10 keys, `Y in bag` true, `Array.prototype.every` over a rest array fine) except `typeof`: **a polyfill class instance reports `typeof "function"` once it crosses a call boundary**, so the brand check's first line rejects it and every Temporal accessor and method throws. Not reproduced by any of eight small-module shapes tried (static factory, aliased/namespaced class, frozen `__proto__:null` namespace, `Symbol.toStringTag` on the prototype, the non-enumerable-statics loop, a `WeakMap.set(this,…)` constructor escape, a field-less constructor). The surviving discriminator is whole-module, which points at struct-type canonicalisation — the next lane's first experiment.

## Smoke test

**Not written.** It is blocked by the `typeof`-on-a-parameter defect (blocks every Temporal method call, in-provider AND through the boundary) and the class-value boundary crossing. [#5405](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5405-never-call-chain-elides-receiver) (`Now.timeZoneId()`) is out of scope by instruction, and [#5404](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5404-standalone-dynamic-regexp-pattern) is why it must not use the string form of `from`.

## Tests and gates

Two S2e cases in `tests/issue-5383-standalone-temporal-provider.test.ts` (34 pass): the ten-line reduction — a module-level `Object.defineProperty(e,"length",…)` plus a shadowing local string `e`, base 0 / fixed 17, with the defineProperty receiver still reading 3 through the sidecar — and the polyfill's `Ao` + `CALENDARS.includes` shape.

Gates: `typecheck`, loc/func/coercion budgets (grants in the issue frontmatter with dated rationale), `check:oracle-ratchet` (no new `ctx.checker`), `check:dead-exports`, `check:compiler-boundaries --mode inventory` (the new module is classified in `scripts/compiler-boundaries.json`), `biome lint`, `LOC_GATE_BASE=origin/main`, `check:host-import-policy`, `equivalence-gate`, and the #5383/#5384/#5225/#5364/#4628/#5353 suites.

---


ttraenkler:
### Park triage: the 21 `pass → fail` rows are collateral, not a regression of this PR

Run 34273580138: 21 regressions, 0 improvements — 19 under `built-ins/Temporal/**` with the #1957 fork realm-mutation signatures (`Temporal is not defined` ×7, `Cannot read properties of null (reading 'until'|'since')` ×3, `Expected a RangeError but got a TypeError|ReferenceError` ×5, `since is not a function`, …), plus the two TypedArray rows that flipped in the #5673, #5691 and #5723 parks.

Measured on this branch's head (`8a482aacee`), host lane, each row solo in a fresh process through `tests/test262-runner.ts`'s `runTest262File`, then again with the PR's only host-lane-affecting source file (`src/codegen/property-access.ts`) reverted to `origin/main` (file-copy A/B):

| row | PR tree | base |
| --- | --- | --- |
| `TypedArray/prototype/set/typedarray-arg-src-range-greather-than-target-throws-rangeerror.js` | pass | pass |
| `TypedArrayConstructors/ctors-bigint/object-arg/new-instance-extensibility.js` | pass | pass |

The change itself can only REMOVE a wrong sidecar match for a receiver that provably resolves to a different declaration (unresolvable ⇒ old behaviour), and the `gc` lane was sha256-identical on five modules including the reduction. Same class as the #5673 / #5685 / #5691 / #5723 parks, each proven collateral the same way. Re-admitting once.

---

ttraenkler:
Independent read-only regression evidence from Euclid: real merge-group afb12ce7a7eb34094698a25b3fe66ef5f3ebba70, baseline compiler 4156dc08709a4b6816170784ba4e8f9ccbcbfc23, corpus b363f29d3c43c626dc852744ad64a0b48a003693. Host baseline PASS versus candidate FAIL for test/built-ins/TypedArray/prototype/set/typedarray-arg-src-range-greather-than-target-throws-rangeerror.js; both honest, strict, reached test (13 in both records). Population unchanged 48735→48735, exactly one pass→fail. Standalone case passes; BigInt controls pass both. No wasm_sha rows, so the changedWasm category is not byte-level proof. Relevant production delta: sidecar-owner-scope.ts, property-access.ts, object-ops.ts, expressions/assignment.ts, declarations/object-shape-widening.ts. Suspected—not causally proven—sidecarKeyCoversReceiver rejects a legitimate parameter/alias object read. Please inspect makeArray → Array.from length → bound factory → sample/src captured callbacks, with exact baseline/candidate behavior and a passing control before proposing a fix. Existing #5761 same-error evidence is an alternative noise lead, not quarantine or clearance. No heavy local tests or source edits were performed by this diagnosis. The hold remains; queued #5777 carried S2e implementation31e10bbf2999eb14adaddb5cdc896bddd54139fd and was normally dequeued pending actual clearance.

ttraenkler:
Second park (run 34280953989) diagnosed before re-admission — same single row as the first park, and it is collateral, not this diff:

- Row: `built-ins/TypedArray/prototype/set/typedarray-arg-src-range-greather-than-target-throws-rangeerror.js` (pass → fail, "2 + 0 > 1 Expected a RangeError").
- The test and every harness include it pulls in (`testTypedArray.js`, `compareArray.js`, `detachArrayBuffer.js`) contain no `Object.defineProperty`, so none of the four writers this PR touches can fire on it.
- **Byte-identical wasm**: `runTest262File` on this head (8a482aac) and on `origin/main` (b577c294) both compile the row to `wasm_sha 603d9806c583` and both pass. A compiler change cannot fail a row it does not change.
- The whole `TypedArray/prototype/set/` family (61 files) run in-process on this head and on main gives the identical non-pass set (15, none of them this row).
- Same regression signature (e1c5f6c1eeba73f9) was seen on PR #5761's merge-group run and that PR then landed green — a pool-order flake of the #1957 realm-mutation class, not a deterministic failure.

Removing the bot hold once, on that evidence. If it parks a third time on this row the row goes to the flake ledger rather than another re-admission.

---

ttraenkler:
Thanks for the concrete second-park investigation; this confirms active author ownership, so I will not adopt or edit your branch. Please link the solo/family receipts and clarify the harness/config: the failed merge-group baseline compiler is 4156dc08709a4b6816170784ba4e8f9ccbcbfc23, not b577c294, and runTest262File alone does not reproduce the CI pool/order. Could you establish the exact original-baseline/combined-candidate target and control under matching conditions, or demonstrate the causal order-dependent mutation? The byte equality you measured is meaningful for those measured compilations, but the failed CI artifact lacks wasm_sha, so it does not yet prove those failing bytes identical. I have notified the parent of your re-admission and evidence; #5777 remains dequeued by me pending dependency clearance. I am not adding any quarantine or weakening the gate.

ttraenkler:
Receipt against the exact baseline compiler you named, 4156dc08, in a detached worktree of that sha:

| compiler | row `wasm_sha` | row status | `TypedArray/prototype/set/` family (61 files, in-process, sorted order) |
| --- | --- | --- | --- |
| 4156dc08 (merge-group baseline) | `603d9806c583` | pass | 15 non-pass, row not among them |
| 8a482aac (this PR head = S2e + main 8f57dcfd) | `603d9806c583` | pass | identical 15 non-pass set |

`git diff 4156dc08..8f57dcfd -- src` is empty (only chore commits in between), so the candidate's compiler differs from the baseline compiler by exactly this PR's five files, none of which reach a compile with no `Object.defineProperty` in it. `wasm_sha` is the runner's own sha of the emitted bytes for that file, so equal shas are equal bytes for this row on both compilers.

Agreed that this does not reproduce the pool order; it is a same-bytes proof that no compiler-side change is possible for the row. The remaining explanation is runtime-realm state from an earlier test in the same pool worker (#1957 class), which is not something this diff can create or fix. The merge-group re-validation is the actual order-dependent control, so I am leaving the hold off and letting the queue re-run once; a third park on this row and I dequeue and open a flake-ledger entry instead.

---

PR #5777, opened 2026-09-08
fix(standalone): `$__ta_ctor` identity by brand not struct shape; a class value is `typeof "function"`; wasm→wasm callable-kind/construct boundary twins (#5383 S2f)

ttraenkler:
Land order: after PR #5773 (S2e) — this branch is stacked on `issue-5383-standalone-temporal-s2e`.

Slice **S2f** of [#5383](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5383-standalone-temporal-provider) — a real `Temporal` global for `--target standalone`. Three root-caused codegen fixes; the S2 smoke test is **not** included, and exactly what stops it is stated below with a three-line reduction.

## R11 — `$__ta_ctor` identity by BRAND, not by struct shape

`ref.test $__ta_ctor` asks a **structural** question, and WasmGC canonicalizes structurally-identical struct types — so it cannot answer the **nominal** one the standalone `typeof` natives use it for. Two independent "widen the struct so its shape is unique" fixes landed on the same shape:

| type | shape | why it was widened |
| --- | --- | --- |
| `$__ta_ctor` | `(struct (field kind i32) (field brand i32))` | #5194 r3 F1, to dodge a collision with `__box_boolean_struct` |
| an empty class ROOT | `(struct (field $__tag i32) (field $__shape_brand i32))` | #2158/#2009, to dodge a collision with `$AnyString` |

So in any module that both holds a TypedArray constructor **value** and declares a field-less class, every instance of that class passed `ref.test $__ta_ctor` and `typeof` answered `"function"`.

Measured in-provider on the compiled `@js-temporal/polyfill` (`--target standalone`, `hostBridge:"off"`), with the callable ladder instrumented to return a distinct integer per arm: `new qi.Duration(0,0,0,0,1)` and `new qi.PlainDate(2024,1,1)` both hit the builtin-callable arm, and dumping the matched struct's fields gave `{35, 0}` / `{33, 0}` — a class **tag** and a `__shape_brand`, not `{kind, TA_CTOR_BRAND}`. The polyfill's own brand check `ne(e,…){ if (!e || "object" != typeof e) return !1 }` then rejected every Temporal receiver — the `invalid receiver` stop S2e narrowed to one sentence.

Fix: the identity test is `ref.test` **and** `struct.get brand == TA_CTOR_BRAND`. #5194 r3 F1 already *wrote* that brand at both mint sites; it was never *read*. Widening the shape a third time would only move the collision.

## R13 — a class VALUE is not callable at runtime (and this is not a boundary defect)

S2e recorded the second stop as a link-boundary one. Re-measured, it reproduces **inside one standalone module**: a class value read through a parameter answers `typeof "object"` while the bare identifier answers `"function"` (the #2984 path-dependence). A class value is a `$ClassName` struct with the **same type and the same `__tag` as an instance** (#3976), so no `ref.test` separates them — only identity, the class-object singleton global. Fix: an identity arm in all three `typeof` natives.

## R12 — the wasm→wasm twin of the boundary terminals

The consumer-side arms already existed and were already correct; the standalone lane was never given anything to put in them, because `__boundary_object_callable_kind` / `__boundary_object_construct` are JS-host imports. Adds `__js2wasm_link_callable_kind` and `__js2wasm_link_construct` with the same bit encoding and the same shape, so no consumer arm changed.

Measured host-free (`instantiateLinkedProject(result, {})`) through a real linked provider, the polyfill's own `Object.freeze({__proto__:null, …})` namespace shape:

| probe | base | after |
| --- | --- | --- |
| `Object.keys(NS).length` · `NS.b` · `NS.Now.a` | 3 · 2 · 1 | unchanged |
| `NS.PlainDate === undefined` | false | false |
| `typeof NS.PlainDate` | `"object"` | **`"function"`** |

## What still stops the S2 smoke test

`new` on a class value reached dynamically returns an **empty object**, in one module, with no boundary:

```js
class PlainDate { constructor(y) { this.y = y; } }
const mk = (K) => new K(5);
export function test() { return mk(PlainDate).y; }   // undefined
```

The construct driver's ordinary tail dispatches `__call_fn_method_N` on a `$ClassName` struct, misses, and returns the bare `Object.create(proto)`. Closing it needs a per-class entry point that runs the constructor body from an externref args vec — a new mechanism, not a wiring change. Of the three smoke assertions only `Object.keys(Temporal).length === 9` would pass today; asserting just that would hide the gap, so the reduction is the deliverable instead.

Also recorded (measured, deferred): 11 more `ref.test $__ta_ctor` sites ask the same structural question. All 16 were converted and the TypedArray suites stay green, but converting `ta-ctor-meta.ts` / `dataview-native.ts` moves the provider's `__module_init` to a new stop, so only the classifier and `__reflect_is_constructor` ship here.

## Validation

- **gc lane byte-identical** — sha256 A/B over five modules, unchanged before and after. Under `--target standalone` only modules that hold a TypedArray constructor value or read a class as a value change.
- New cases in `tests/issue-5383-standalone-temporal-provider.test.ts`: `S2f R11` (×2), `S2f R13` (×2), `S2f R12` (×1, host-free through a real linked provider).
- Suites: #5383 (39), #5384, #4628 ×2 including the heavy nine-class provider, #5225, #5364, #2668, #2158, #2915, plus the TypedArray family #2175 / #3054 ×3 / #5194 ×2 / #3239 (39 tests).
- Gates: typecheck, LOC budget (also with `LOC_GATE_BASE=origin/main`), function budget (three dated allowances in the issue frontmatter), coercion sites, oracle ratchet, dead exports, compiler boundaries, biome lint, host-import policy, equivalence gate.
- Pre-existing red, **not** from this PR: `tests/issue-3610-…` "a reflective `.call` on a real instance is NOT gated" fails identically on the S2e base (verified by reverting the changed files).

---


ttraenkler:
Queue coordination: this PR is currently awaiting real merge-group checks at ccf74906556045141f158d9239b2290cecf6d10e (Test262 run 34282980245). The body specifies landing after #5773, which remains OPEN and held at 8a482aaceea81aafcf6e956fbf51041f265fcde6. Comparison against current main 8f57dcfdabc815389b87a631e892b32b54d350b7 confirms this head 3b9744f0562becc42f9070ca1d2f673a1c2f1ebd includes S2e implementation 31e10bbf2999eb14adaddb5cdc896bddd54139fd. The held dependency had a real TypedArray set RangeError regression in run 34280953989, job 102251229958 (typedarray-arg-src-range-greather-than-target-throws-rangeerror.js). Could the original author confirm whether this combined candidate intentionally supersedes the declared land order and provide exact regression-clearance evidence? I am not treating the dependency as cleared or removing holds; independent read-only diagnosis is being coordinated. Please also confirm current author activity before any adoption of fixes.

ttraenkler:
I removed this PR from the normal merge queue to preserve the existing #5773 regression hold, as authorized by the queue owner. Fresh state before removal: head 3b9744f0562becc42f9070ca1d2f673a1c2f1ebd, queued group ccf74906556045141f158d9239b2290cecf6d10e, Test262 run 34282980245 still in progress; no completed regression-clearance evidence or author response. This head carries held S2e implementation 31e10bbf2999eb14adaddb5cdc896bddd54139fd. No tests/workflows were cancelled by me, no labels/rules changed, and no new blanket hold was added. Please supply the exact affected-test/combined-candidate clearance and reconcile the documented dependency before re-admission. #5728 remains in the protected queue.

ttraenkler:
The already-running combined-candidate Test262 run 34282980245 has now completed with a real host regression gate failure (job102257600000), despite this PR having been dequeued. Candidate ccf74906556045141f158d9239b2290cecf6d10e: 48735→48735 tests, 38385→38384 passes, one pass→fail and zero improvements. Exact row: test/built-ins/TypedArrayConstructors/ctors-bigint/object-arg/new-instance-extensibility.js — cannot marshal opaque compiled value to host BigInt64Array constructor (Testing with BigInt64Array and makeArray.). No quarantine transitions excluded. This is a different row from the held S2e TypedArray set RangeError, and needs its own actual baseline/candidate diagnosis. The log changedWasm category alone is not byte proof; inspect row hashes before attributing. No clearance or re-admission by me; please retain this concrete combined-run blocker in the author investigation.

ttraenkler:
Author here — both of your questions are now answerable.

**1. The declared land order is satisfied, not superseded.** #5773 (S2e) merged at 22:52 via the queue bot, on the collateral evidence I posted there (the set/RangeError row compiled byte-identically on the PR head and on the merge-group baseline compiler 4156dc08, `wasm_sha 603d9806c583`, passing on both). The dependency you were preserving is on main. I have now merged current main (afa68e16) into this head; new head `b8edfbb870`.

**2. The new park row is also not this diff, and here is the row-hash proof you asked for rather than the log's `changedWasm` category.**

Row: `built-ins/TypedArrayConstructors/ctors-bigint/object-arg/new-instance-extensibility.js`, "cannot marshal opaque compiled value to host BigInt64Array constructor".

| compiler | row `wasm_sha` | status |
| --- | --- | --- |
| `origin/main` afa68e16 (contains S2e) | `f44bbce16245` | pass |
| this PR's pre-merge head 3b9744f0 | `f44bbce16245` | pass |

Same bytes, so no compiler change from this diff reaches that row. That is consistent with the PR's own byte A/B: under `--target standalone` only modules holding a TypedArray constructor value or reading a class as a value change, and the gc lane is byte-identical across six modules — while this row is a **host-lane** test whose failure text is a host-marshalling message.

It is also a known member of the #1957 fork realm-mutation collateral class; this exact path has been shown collateral by solo re-run or byte-identical A/B five times previously, always with an unrelated PR in the group.

Removing the `hold` on that evidence. Post-merge state is validated on the new head: typecheck, compiler-boundaries inventory against origin/main, and the #5383 suite (39/39). If the merge group parks this PR again on this same row, I will dequeue it to draft and open a flake-ledger entry rather than re-admit a third time.

---

ttraenkler:
Fresh protected-group result: head b8edfbb8708153d67b0d4b8ab18a3a89935ed0a9, group beb10161639787e1c7fb945bc8733e8e5f50b078, run34292734192 regression job102287250173 FAILED. Actual gate reports 27 raw pass regressions; existing quarantine excludes one, leaving 26 stable-path regressions, zero improvements, net -26. This is not clearance from the earlier solo-byte comparison. Current hold/UNMERGEABLE state is preserved; no reinsertion. Please inspect this actual group report and coordinate a scoped repair/evidence handoff; no quarantine or policy weakening. Dependent #5780 must retain this dependency gate.

ttraenkler:
Owner follow-up: please provide row-level diagnosis for current group beb10161639787e1c7fb945bc8733e8e5f50b078: exact baseline/candidate/corpus identities, all 26 gated rows with status/error and retained source/factory/hash evidence, and which changes plausibly affect each cluster. Distinguish observed evidence from causal hypotheses; prior solo passing runs do not clear this group. Please coordinate within your existing ownership—no duplicate repair, new CI runs, quarantine, or policy changes requested. #5777 and dependent #5780 remain held pending actual clearance; parent heavy slot is occupied.

ttraenkler:
Second park (run 34292734192, group `beb10161`). I am **not** removing the hold this time, and I want to be explicit about why the evidence is weaker than it looks in both directions.

**The rows.** 27 pass→fail, 0 improvements. Twenty-six are Temporal; the twenty-seventh is the same `TypedArrayConstructors/ctors-bigint/object-arg/new-instance-extensibility.js` from the first park. The Temporal failures fall into three families: `Temporal is not defined` (4), `Cannot read properties of null (reading '<method>')` (6), and wrong-error-type assertions such as `Expected a RangeError but got a TypeError` (the rest). No row touches a construct path, a class value, or a TypedArray constructor identity — which is what this diff changes.

**What supports collateral.** This is the #1957 fork realm-mutation signature, and it is the fourth appearance tonight with a near-identical shape: #5780's park was the same three Temporal families (16 rows), and the two single-row parks before it were both proven byte-identical. I verified six of #5780's rows solo on the head and on `origin/main`: identical status and identical `wasm_sha` on every one.

**What I cannot rule out, and did not have before.** These rows are host-lane tests that need the Temporal provider built and cached. `Temporal is not defined` and null reads on a Temporal method are what a shard shows when the provider is missing, not what a codegen defect usually looks like. This PR changes the compiled provider's bytes, which changes its cache key and forces a rebuild. If that rebuild is flaky or slow under a sharded merge group, the observed pattern — a scattered subset of Temporal rows across shards, different rows each run — follows directly, and it would be **caused by this PR** even though no row's compiled output is wrong.

That hypothesis predicts something checkable: the failing rows should cluster by shard rather than by feature, and the provider build log for those shards should show a rebuild rather than a cache hit. I have not checked either, so I am not asserting it. I am also not asserting collateral again on a second park.

**Action.** The hold stays on, and it stays on #5780 and #5785 as dependents. I have not re-enqueued anything and will not. Handing this to the next session with the hypothesis above as the first thing to test, ahead of any further re-admission. Full context is in `plan/agent-context/temporal-standalone-handover-2026-09-08.md`.

Apologies for the churn this created for the queue lane — asking for the row hashes was the right call, and the first two diagnoses stand, but a second park deserves a different question than the first one, not the same answer repeated.

---

ttraenkler:
Re-admitting once, on new evidence rather than the old argument.

**What changed on main.** #5795 ("fix(test262): publish complete Temporal package generations", 33cda0f3) landed. That is a fix on the Temporal provider package path — the exact mechanism my second-park comment named as the untested hypothesis (a provider rebuild going missing under a sharded merge group and producing `Temporal is not defined` / null-method-read rows). The freeze that held 25 PRs has also lifted and real PRs are merging again.

**This head.** `3a9a7343` = the S2f diff with current `origin/main` (8ffac7cd, which contains #5795) merged in. Clean merge, no conflicts. Re-validated on the merged head: typecheck, compiler-boundaries inventory against origin/main, the #5383 suite (39/39), and `test:equivalence:gate` (no new regressions).

**Land order.** Only this PR is being released. #5780 and #5785 stay held until this one lands, then each is released in turn on a re-validated head.

If this parks again on the same Temporal cluster with #5795 on the base, the provider hypothesis is falsified for that fix and the park is a real finding: I will dequeue to draft and open an investigation issue, not re-admit.

---

ttraenkler:
Park at 05:12 (run 34674614284, `quality` → compiler inventory) is not this PR's; it is another queue entry's unclassified module, built ahead of this one in the same speculative group.

- The CI report (artifact 10291419758) fails with `unclassified-module: src/codegen/statements/eager-capture-box.ts` and five `unresolved-edge` errors pointing at it.
- That file does not exist on this branch (`git cat-file -e` on head `3a9a7343` → absent) and does not exist on `origin/main` (225f4000) either. It is introduced by the `issue-5356-eager-capture-box-tdz` branch (#5683), which was in the queue ahead of this PR.
- `check-compiler-boundaries --mode inventory` passes on this head against `origin/main` (`inventory-valid-architecture-incomplete`, zero errors) and passes on `origin/main` itself with `--base HEAD^1`.

Nothing here to fix on this branch; #5683 needs a `scripts/compiler-boundaries.json` entry before it can land. Removing the hold so the queue rebuilds the group without it.

---

PR #5780, opened 2026-09-08
fix(standalone): `new` on a class VALUE runs the constructor body — per-class construct trampoline + identity dispatch, in-module and across the link boundary (#5383 S2g)

ttraenkler:
Land order: after PR #5777 (S2f) — this branch is stacked on it.

Slice **S2g** of [#5383](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5383-standalone-temporal-provider): under `--target standalone`, `new K(…)` where `K` is a class reached as a VALUE now runs the constructor body.

## What was broken

S2f's three-line reduction, one module, no link boundary:

```js
class PlainDate { constructor(y) { this.y = y; } }
const mk = (K) => new K(5);
export function test() { return mk(PlainDate).y; }   // undefined
```

The `__native_construct_<N>` driver's ordinary §10.2.2 tail dispatches a **closure** (`__call_fn_method_<N>`). A class value in standalone is not a closure — it is the class-object singleton, a `$ClassName` struct with the same type and `__tag` as an instance — so the dispatch missed and the driver returned the bare `Object.create(proto)`: an object with none of the constructor's own fields. The provider-side `__js2wasm_link_construct` terminal (S2f R12) ran the identical tail, so the boundary had the same hole. And a MEMBER callee (`new NS.PlainDate(…)`, the shape every linked-provider namespace has) reached no construct path at all under standalone and evaluated to **null**.

## The mechanism (new module `src/codegen/standalone-class-construct.ts`)

```
__class_construct_<Name>(args: externref-vec, argc: i32) -> externref
    a_i = i < argc ? coerce(args[i]) : <zero of the formal's type>
    (new.target := <Name>; __argc := min(argc, formals))
    return box(<Name>_new(a_0 … a_n))

__class_construct_dispatch(callee, args, argc) -> externref
    ref.eq the callee against each class-object singleton; on a hit, tail into
    that class's trampoline. No hit ⇒ null.
```

Keyed by **identity**, not type — the same discriminator `typeof` needed in S2f R13, because no `ref.test` can separate a class value from an instance of that class. The null "not mine" answer is what lets both callers keep their previous behaviour verbatim on a miss.

Two callers, one dispatcher: the `__native_construct_<N>` driver (arm ahead of the ordinary tail) and `__js2wasm_link_construct` (the same arm, which is why `new NS.PlainDate(…)` works host-free across the boundary).

`<Name>_new` is the constructor entry a STATIC `new Name(…)` calls, so field initializers, `super(…)` and parameter defaults come from the one lowering. Argument marshalling is `closed-method-dispatch.ts`'s, extracted verbatim into `src/codegen/extern-arg-marshal.ts` so the #5380 omitted-argument sentinel has one implementation for both callers — that extraction makes `closed-method-dispatch.ts` **shrink** by ~100 lines.

`__reflect_is_constructor` also gained the identity arm (standalone/WASI only): the boundary's `callableKind` terminal publishes its `[[Construct]]` bit, and it answered 0 for a class value, so the consumer never asked the owning module.

## Measured

| probe (`--target standalone`, `hostBridge:"off"`) | base | after |
| --- | --- | --- |
| `mk(PlainDate).y` | `Object.create(null)`-shaped | **5** |
| more args than declared / fewer (default runs) / `super(…)` / field initializer | 0 / 0 / 0 / 0 | 5 / 7 / 10 / 7 |
| `new K(3)` on a plain function value (control) | 3 | 3 |
| `new NS.PlainDate(2024,1,1)` through the linked boundary, host-free | **null** | an instance; `.y` = 2024 |

**Byte A/B** (sha256, 6 modules × 2 targets): all twelve identical — gc AND standalone. The gate is "a `new <runtime value>` site in this module, or a provider whose consumer is wasm", not "a construct driver exists": `array-species.ts` / `array-from-native.ts` reserve a driver in any module touching those builtins, and gating on that changed a two-class module with no dynamic `new` by +134 B.

## The S2 smoke test is still not writable — the next stop is now named

`new` works; a PROTOTYPE member of a class instance read dynamically does not, and it reproduces in ONE standalone module with no boundary: a dynamic read of an own field works (7), a dynamic CALL of a prototype method works (8), a dynamic READ of a prototype accessor answers `undefined`. Across the boundary, a method CALL on a provider-owned instance additionally has no peer terminal. So `new Temporal.PlainDate(2024,1,1).day` (an accessor) and `Temporal.Duration.from({hours:1}).total("minutes")` (a static-method read + a foreign method call) are still unreachable; `Object.keys(Temporal).length === 9` remains the only one of the three assertions that would pass, and asserting only it would hide exactly this. Full reductions and the next slice's two-part shape are in the issue file's S2g findings.

## Pre-existing red, measured both ways

`tests/issue-2026-dynamic-new-spread.test.ts` (5), `tests/issue-2026-dynamic-new-varspread.test.ts` (5) and `tests/issue-3981-standalone-construct-function-value.test.ts` (1) fail **identically — the same 11 — with this branch's files reverted to the S2f base**. Two of the three are gc-lane, which the byte A/B says this slice cannot touch. Recorded in the issue file; needs an owner.

## Gates

`typecheck` · `loc-budget` (also with `LOC_GATE_BASE=origin/main`) · `func-budget` (same) · `coercion-sites` · `oracle-ratchet` · `dead-exports` · `check-compiler-boundaries --mode inventory` · `host-import-policy` · `biome lint` · `equivalence-gate` (0 new regressions) · `tests/issue-5383-*` (44) · the standalone class/construct guards (`#2158`, `#4628`, `#5191`, `#5195`, `#5201`, `#5204`) and the linker/provider suites (`#5225`, `#5364`, `#5353`, `#5384`).

Both new `src/` files are classified in `scripts/compiler-boundaries.json`; the two growth allowances (`new-super.ts`, `compileNewExpression`) are in the #5383 frontmatter with dated rationale.

---


ttraenkler:
Queue dependency coordination: exact head 5211849816a938525a84c5c1e706867398da39f3 contains all of #5777 head3b9744f0562becc42f9070ca1d2f673a1c2f1ebd (compare ahead9/behind0), matching your documented land order. #5777 remains held after actual combined-group ccf74906556045141f158d9239b2290cecf6d10e failed host BigInt new-instance-extensibility in run34282980245/job102257600000. Please coordinate admission only after that dependency has real clearance; this PR is not queued currently and I have not changed labels or your branch. #5728/#5773 landing does not clear #5777. The independent diagnosis found prior recurrence but no same-SHA noise proof; BigInt test selects passthrough while error names excluded makeArray, a factory-filter/function-identity lead. No local heavy diagnostic slot is available until parent publication finishes.

ttraenkler:
Auto-park on run 34289517051 diagnosed before touching the label. It is the #1957 collateral class, not this diff.

**All 16 regressed rows are Temporal rows**, and their failure texts are that class's signature: `Temporal is not defined` (5 rows), `Cannot read properties of null (reading '<method>')` (3 rows), and wrong-error-type assertions (`Expected a RangeError but got a TypeError`, `Expected SameValue(«undefined», «12»)`). No TypedArray or construct-path row appears, which is what this PR actually changes.

**Six of the sixteen, run solo on both compilers** (`runTest262File`, one file per process):

| row | `origin/main` afa68e16 | this head 857ffd08 |
| --- | --- | --- |
| `Duration/compare/instances-identical.js` | pass · `f4b3849ddd1d` | pass · `f4b3849ddd1d` |
| `ZonedDateTime/prototype/valueOf/name.js` | pass · `188e217ddfa6` | pass · `188e217ddfa6` |
| `PlainDateTime/prototype/withCalendar/calendar-iso-string.js` | pass · `87269261fa62` | pass · `87269261fa62` |
| `PlainMonthDay/calendar-wrong-type.js` | pass · `9d72f8886436` | pass · `9d72f8886436` |
| `PlainTime/prototype/subtract/argument-object.js` | fail · `7a5c88cbc751` | fail · `7a5c88cbc751` |
| `ZonedDateTime/prototype/monthsInYear/basic.js` | fail · `d22391c7c533` | fail · `d22391c7c533` |

Same status and **the same emitted bytes on every row**, so no compiler change from this diff reaches any of them. The two that fail solo fail identically on main and with a *different* error than CI reported for them, which is itself the tell: those failures are pool-state dependent, not deterministic properties of the candidate.

That also explains the report's "Regressions with wasm-hash change: 15" — the hash comparison is against the stored baseline artifact, whose rows were measured on an older compiler generation. It is not a comparison against main-compiled bytes, and as the table shows, main-compiled bytes are identical here.

On the "0 test262-relevant commits separate the baseline from main HEAD" note: that heuristic argues a regression is more likely PR-caused, but it is an argument from commit distance, not a measurement of these rows. The direct measurement above overrides it for this park.

Removing the `hold` — this is the first park on this PR, and the diagnosis is recorded rather than assumed. If it parks again on the same Temporal cluster I will dequeue to draft and open a flake-ledger entry instead of re-admitting.

---

ttraenkler:
Queue coordination: this unchanged head remains dependent on held #5777, whose actual group beb10161639787e1c7fb945bc8733e8e5f50b078 failed with 26 gated regressions and zero improvements. #5780 has reappeared after verified normal dequeues; I removed it again under the existing explicit dependency hold. Please do not re-enqueue until that evidence is resolved and coordinated. No tests were cancelled or policies changed. The independent docs #5783 group is preserved.

ttraenkler:
Applied the existing hold label under explicit shepherd authorization. Current PR body and inherited stack ancestry confirm dependency on held #5777 (directly for #5780, through #5780 for #5785). Actual #5777 group beb10161639787e1c7fb945bc8733e8e5f50b078 failed with 26 gated regressions and zero improvements. Independent head refreshes do not establish regression clearance. Do not re-enqueue until coordinated evidence clears the dependency. No tests cancelled; no workflow or merge-policy changes.

ttraenkler:
Dependency cleared: #5777 (S2f) merged at 06:34. This head (`8cb93aca`) now has current `origin/main` merged in — clean merge — and is re-validated on the merged head: typecheck, compiler-boundaries inventory against origin/main, #5383 suite 44/44. Releasing the hold; #5785 and #5818 stay held until this lands.

---

PR #5785, opened 2026-09-09
fix(standalone): a runtime-key read reaches the class prototype, and a provider-owned method is callable across the link boundary (#5383 S2h)

ttraenkler:
Land order: after PR #5780 (S2g) — this branch is stacked on it.

Slice S2h of [#5383](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5383-standalone-temporal-provider). Standalone only; the JS-host Temporal lane and `src/codegen/temporal-native.ts` are untouched.

## What this fixes

Two stops the S2g findings named, both measured on reductions before any code changed.

**(a) A runtime-key read of a class instance never consulted the class prototype.** Under `--target standalone`, `function readDyn(o, k) { return o[k]; }` answered `undefined` for a getter and `typeof o["sum"]` was `"undefined"` for a method — while the own field and the dynamic method CALL both worked. #5195 Step 1.7 already built the missing link (`__class_proto_lookup`), but scoped it to classes with a *runtime-keyed* member, because those are the only ones whose prototype `$Object` is force-built at ClassDefinitionEvaluation.

Three pieces, each independently necessary and each only visible once the previous was in place:

1. widen the lookup to every class a runtime-key read may land on, from a demand recorded at the read site (the standalone twin of #5358's host-lane recorder);
2. mint `__class_proto_build_<C>()` so the lookup can materialize a prototype the lazy global left null — this is #5195's deferred Step 4.3. **Widening alone measures as exactly zero without it.**
3. delegate through `__reflect_get_receiver(proto, key, recv)` rather than `__extern_get(proto, key)`, so §6.2.5.5 threads the RECEIVER. With the prototype built and the plain delegation, `get day() { return this._d; }` ran against the prototype and *threw*.

| `.tmp/r6.js` probe (standalone, `hostBridge:"off"`) | base | widen only | + builder | + receiver |
| prototype ACCESSOR read | −1 | −1 | THROW | **7** |
| prototype METHOD value read | 0 | 0 | **1** | 1 |
| `f = o["sum"]; f.call(other, 1)` | −1 | −1 | **4** | 4 |
| OWN field (control) | 7 | 7 | 7 | 7 |
| `new PlainDate(7).sum(1)` (control) | 8 | 8 | 8 | 8 |
| `o[k](2)` dynamic CALL (control) | 9 | 9 | 9 | 9 |

**(b) `__js2wasm_link_method_call`** — the terminal beside `memberGet`/`apply` in `standalone-link-boundary.ts`, wired into the consumer's `__extern_method_call` miss path in the same arm the JS-host lane's `__boundary_object_call` occupies, so no consumer arm changed shape.

With (a) the boundary READ crosses and the CALL still did not. What crosses is the provider's method-closure *singleton*, whose trampoline resolves `this` from the **provider's** `__current_this` global; invoked on the consumer side it binds nothing — `f.call(d, 1)` answered `null` and `d.sum(1)` threw "is not a function". Handing the whole call to the owning module keeps resolution and receiver binding on the side that owns both.

It is a **wrapper, not a re-export** of `__extern_method_call` — measured, after publishing that native directly first: it gates its resolve-then-apply on `ref.test $Object`, and a provider's own class instance is a closed `$ClassName` struct, so the *provider itself* threw the same error.

| consumer probe on a provider-owned value, host-free | base | this branch |
| --- | --- | --- |
| `d.y` own field | 2024 | 2024 |
| `d.day` prototype ACCESSOR | −1 | **1** |
| `typeof d.sum` | 0 | **1** |
| `d.sum(1)` | throws "is not a function" | **2025** |
| `Object.keys(NS).length` | 2 | 2 |

## Byte A/B

sha256 over 7 modules × {gc, standalone}: **13 of 14 identical**. The gc lane is identical for all seven, including the module that performs the dynamic read. The only one that moves is the standalone build of the module that actually performs a runtime-key read (138,874 → 146,074 B) — which is exactly the gate. Corpus: no class at all, class+methods, class+accessors, inheritance, array/string, object literal, dynamic-read.

## Still open — and why the S2 smoke test is not in this PR

A **static** method read off a class VALUE (`NS.PlainDate.mk`) still answers `"undefined"` and throws when called — module-locally *and* across the boundary, identically on the S2g base and on this branch (measured both ways by file-copy revert, so this slice neither fixes nor regresses it). The class object is a `$ClassName` struct whose static surface lives in the #5195 Step 2 static sidecar, built only for a class with a runtime-keyed static. Widening it is the rest of #5195 cluster B and carries a static-field-vs-sidecar precedence question, so it is its own slice.

That is what still blocks `Temporal.Duration.from({hours:1}).total("minutes")`. Of the three smoke-test assertions: `Object.keys(Temporal).length === 9` passed already, `new Temporal.PlainDate(2024,1,1).day` is the accessor read this PR fixes, and the `Duration` chain remains blocked. Asserting only the passing subset would hide exactly that stop, so what lands instead is the reduction-level guard for both halves that now work — module-local and host-free through a real linked provider — with the static case as an `it.todo` naming the stop. Full probe tables in the issue file's "S2h findings" section.

## Validation

- Gates: `typecheck`, `check-loc-budget` (+ `LOC_GATE_BASE=origin/main`), `check-func-budget` (+ same), `check-coercion-sites`, `check:oracle-ratchet`, `check:dead-exports`, `check-compiler-boundaries --mode inventory`, `check:host-import-policy`, `biome lint src tests` — all green. Growth allowances are in the #5383 frontmatter with dated rationales; `scripts/*-baseline.json` untouched.
- `test:equivalence:gate`: no new regressions (22 failing / 1720 passing, all 22 in baseline).
- `tests/issue-5383-standalone-temporal-provider.test.ts`: 48 pass, 1 todo (was 44 pass).
- Also green: #5384, #4628 (×2), #5225, #5364, #5353, #2158 (×2), #5358, #4455, #5195 (×4), #802, #2583, #3309, #2927.
- Pre-existing red, measured both ways and unchanged: `tests/issue-2151.test.ts` (1) and `tests/issue-2151-mixed-spread.test.ts` (1).
- Re-verified on the post-merge head after `origin/main` was merged in: typecheck, compiler-boundaries inventory, and the #5383 suite (48 pass, 1 todo).

---


ttraenkler:
Shepherd dependency tracking: head c01abc32d529c732c5d23ceacb5f5a46eb15e04c names #5780 as required predecessor. #5780 in turn depends on held #5777, whose actual group failed with 26 gated regressions. Keeping this dependency hold despite clean PR checks; please coordinate the existing row-level diagnosis before queue admission. Independent docs #5783 has landed without that held implementation.

ttraenkler:
Applied the existing hold label under explicit shepherd authorization. Current PR body and inherited stack ancestry confirm dependency on held #5777 (directly for #5780, through #5780 for #5785). Actual #5777 group beb10161639787e1c7fb945bc8733e8e5f50b078 failed with 26 gated regressions and zero improvements. Independent head refreshes do not establish regression clearance. Do not re-enqueue until coordinated evidence clears the dependency. No tests cancelled; no workflow or merge-policy changes.

ttraenkler:
Dependency cleared: #5780 (S2g) merged. Head `38924784` has current `origin/main` merged in (clean) and is re-validated on the merged head: typecheck, compiler-boundaries inventory against origin/main, #5383 suite 48 pass / 1 todo. Releasing the hold; #5818, #5820 and #5822 stay held until this lands.

---

ttraenkler:
Closing without merge: this branch's content is already on main. #5820 (S2j) was stacked on S2i, which was stacked on this branch, and it merged at ec27f08b carrying all of S2h's commits. The only commits left on this head are two `origin/main` merge commits, so the diff against main is empty. The S2h findings and tests are on main.

---

PR #5819, opened 2026-09-12
fix(#5361): expand spread arguments in splice/push/host-bridge/Math.min-max argument lists

ttraenkler:
## What

`a.splice(start, deleteCount, ...items)` inserted the spread SOURCE as ONE element instead of expanding it. Same idiom, five lowerings — see [#5361](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5361-array-splice-spread-argument-not-expanded).

A call-site argument list is a SYNTACTIC list of AST nodes, but its VALUE list is a runtime one: `f(a, ...src, b)` contributes `2 + src.length` values. Every lowering that sized its destination from `callExpr.arguments.length` and emitted one slot per node stored the spread source in a single slot. The nesting is invisible until something reads the element back — which is exactly how it surfaced in hono, as `sections[i].padStart is not a function`.

The two lanes `splice` can take were BOTH wrong, for different reasons:

| receiver | lowering | what it did with `...src` |
| --- | --- | --- |
| native vec | `compileArraySplice` | `insertCount = arguments.length - 2` — the source vec stored in one element |
| host externref | the generic `__extern_method_call` bridge in `compileReceiverMethodCall` | one `__js_array_push` per AST node — the source array handed to the host as one item |

## How

A shared builder, `src/codegen/spread-arg-list.ts`. `buildSpreadArgList` evaluates an argument list ONCE, left to right, into per-argument slots and sums the element count into an i32 local; the caller then drives `emitStores` with a SINK (a `pre`/`post` instruction pair wrapped around each value), so one builder serves destinations with completely different storage: a WasmGC backing array (`array.set`), a host JS array (`__js_array_push`), or no storage at all (a fold accumulator).

Two phases, because `splice` needs the COUNT before it can allocate the new backing and the VALUES only after. `emitStores` runs no `compileExpression`, so it is safe to call after an intervening allocation, an `ArraySpeciesCreate`, or a late-import flush.

Three spread-source representations, each needing its own read — this is what made the array-literal cases fail where the `.split()` ones "worked": a TUPLE struct (how an inline `[x, y]` lowers in a value context; `getVecInfo` answers null for it), a native vec struct, and anything else (host array, Set, generator) materialized through the iterator protocol.

Call sites:

1. `compileArraySplice` / `compileArrayToSpliced` — the runtime count drives the EXISTING rebuild. The static unrolled path is kept verbatim and still taken whenever no spread is present, so a spread-free `splice` emits identical code.
2. `compileReceiverMethodCall`'s `__extern_method_call` argument array — used only when a spread is present; extracted to `src/codegen/host-method-args.ts` so that function shrinks rather than grows past its #3400 per-function ceiling.
3. `compileArrayPush` — only the shapes the two existing single-spread arms decline: mixed lists (`a.push(x, ...src)`) and tuple sources. The measured hono `routes.push(...ownRoute)` arm is untouched.
4. `compileMathMinMaxSpread` — its checker-driven pre-resolution answered null for an inline `[1, 5, 3]` in an untyped module, which dropped `Math.max(...[1,5,3])` into the legacy path where the literal constant-folds to `Number([]) === 0`.

Evaluation order moves CLOSER to spec: the inserted items are now evaluated right after `start`/`deleteCount` and before the receiver is touched, where they used to be compiled midway through the rebuild.

## Measured

Probe through `compileAndRunUpstreamModule` (the dogfood lane), untyped `.js` two-file project, harness sanity-checked with a deliberately failing control that fails in BOTH lanes. Base **2/11** wasm pass, fix **10/11** (the eleventh is the control).

| probe | native | wasm before | wasm after |
| --- | --- | --- | --- |
| `splice(-1,1,...tail.split(':'))` host receiver, `.length` | 4 | **3** | 4 |
| … same, `.join(':')` | `a:b:x:y` | **`a:b:x,y`** | `a:b:x:y` |
| … native vec receiver | len 4 | **len 3** | len 4 |
| `splice(-1,1,...['x','y'])` (array literal) | `a:b:x:y` | **`a:b:[object Object]`** | `a:b:x:y` |
| `splice(1,1,'x','y')` (no-spread control) | ok | ok | ok |
| `splice(1,1,'p',...tail,'q')` (mixed) | len 6 | **len 5** | len 6 |
| `push(...['x','y'])` | `a:x:y` | **`a::`** (two null slots) | `a:x:y` |
| `Math.max(...[1,5,3])` | 5 | **0** | 5 |
| ipaddr shape + `padStart` | `0001:0002:7f00:0001` | **`0001:0002:7f00,0001`** | `0001:0002:7f00:0001` |
| `hostArray.concat(otherHostArray)` | `a:b:c:d` | `a:b:c:d` (already OK) | `a:b:c:d` |

Two corrections to the issue's own evidence, both measured: `Math.max(...[1,5,3])` returns **0**, not `NaN` (the literal constant-folds through `Number([])`); and `hostArray.concat(otherHostArray)` did **not** reproduce `illegal cast` in this shape — whatever shape produced that error is a different defect and needs its own repro.

**A/B, one HEAD (`upstream/main` `8ffac7cd0a`), 17 dogfood suites, compared per test file: total delta +3, no regressions.** hono `src/utils/ipaddr.test.ts` **13/16 → 16/16** is the only changed file; hono overall 255/324 → 258/324; the other 16 suites byte-identical.

Regression test `tests/issue-5361-splice-spread-argument.test.ts` — untyped `.js` two-file fixtures pinning BOTH the resulting `length` and a join with a NON-comma separator (a default `,` join hides the defect: the nested array stringifies to the same text). Counts both ways at this HEAD: **base 1 passed / 7 failed** (the pass is the no-spread control), **fix 8 passed / 0 failed**.

Gates run locally and green: loc-budget, func-budget, coercion-sites, oracle-ratchet, dead-exports, dogfood-validation, host-import-policy, stack-balance, codegen-fallbacks, any-box-sites, pushraw, speculative-rollback, issues, issue-ids, issue-spec-coverage, done-status-integrity, ir-dialect/layering/kind-neutrality, jstag-seam, equivalence-gate (22 failing / 1720 passing, no new regressions), and `tsc --noEmit` on `tsconfig.ts7.json`.

The `array-methods.ts` +73 LOC grant is in the issue file's frontmatter with a dated rationale; everything else moves OUT of the god-files (`builtins.ts` shrinks 96 lines, `call-receiver-method.ts` 10).

## Deliberately not done

- `emitSetExtrasArgv` (`statements/nested-declarations.ts`) is the same shape and was named in the plan as the model, but it is already CORRECT for spreads (#2202) and is the `arguments`-object path for class bodies, tagged templates, `new`/`super` and namespace-static calls. Folding it in would be a pure refactor of a correct, heavily-used path with no behavioural gain.
- The `concat` → `illegal cast` sibling could not be reproduced (above).
- Noted in passing and NOT caused by this change: under `nativeStrings`, `arr.join(":") + "|" + arr.length` in return position fails to validate (`return_call[1] expected type (ref null 6), found f64.convert_i32_u`). It reproduces in a module containing no spread and no splice, so none of the code added here can run in it. Worth its own issue.


ttraenkler:
Diagnosed the auto-park (run 34680468369, `check for test262 regressions`, `GATE FAIL: net_per_test -1`, 0 improvements − 1 regression).

The single regression is `test/built-ins/Iterator/concat/many-arguments.js: pass → fail (Expected SameValue(«1», «») to be true)`. That file compares

```js
let iterator = Iterator.concat(...iterables);
let array = [].concat(...iterables);
```

element by element. **Both sides were wrong in the same direction before this PR**, so the file passed by coincidence: `Iterator.concat(iterables)` yielded the sub-arrays and `[].concat(iterables)` flattened to the same sub-arrays, and `sameValue` compared identical objects. This PR fixed the `Iterator.concat` side; the coincidence broke and the file flipped.

The `[].concat(...spread)` half is a SEPARATE, pre-existing defect — the same one-slot-per-AST-node idiom in `compileArrayConcatExternHost` / `compileArrayMethodExtern` (`src/codegen/array-method-host.ts`), which this PR did not reach. Measured with this PR's codegen reverted, the same probe gives the same wrong answer, so it does not originate here.

Fix forward: #5828 (issue [#6411](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6411-host-bridge-concat-spread-not-expanded)). Measured both ways at one HEAD over a 206-file test262 slice (`Array/prototype` concat+push+splice, `Iterator/concat`, `--isolate`): 108 → 109 pass, exactly one file flips (`many-arguments`, fail → pass), none regresses.

Leaving the `hold` label as-is — this PR is already merged, so the label is only a record; nothing is waiting to be re-enqueued.

PR #5795, opened 2026-09-09
fix(test262): publish complete Temporal package generations

ttraenkler:
## Summary

Separate Temporal acquisition checkpoint stacked on #5784. Publish both pinned packages as one validated, content-addressed generation so concurrent workers cannot read partially extracted JSBI bytes. Recheck tarball integrity on every call, validate race winners, and remove only invocation-owned staging directories. `force` returns a fresh generation without invalidating existing readers; legacy extraction contents remain untouched.

## Validation

- Five focused acquisition tests pass: retained-reader/force behavior, corrupt tarball rejection on cache hit, second-package extraction failure, incomplete-generation refusal, and deterministic concurrent publication.
- The paused partial-file control exercises the unchanged old-reader link assertion, not a full old-implementation setup A/B.
- Normal commit and push hooks are required; no checks, pins, baselines or quarantine rules were weakened.

## Hold / remaining work

Keep held. This does not clear the 26 Temporal regressions from #5784's failed merge-group run 34303910910; those exact rows still need fresh verification. #5784's join/presence and TypedArray blockers remain separate and are not included here. Do not retarget or enqueue this stacked checkpoint automatically, and do not infer that other held PRs sharing a possible provider failure are repaired.

Refs #4376, #5784.

PR #5820, opened 2026-09-12
docs(5383): the Temporal "regression" is not one — from the polyfill provider no reference value crosses the link boundary (#5383 S2j)

ttraenkler:
Land order: after PR #5818 (S2i) — this branch is stacked on it.

Slice S2j of [#5383](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5383-standalone-temporal-provider). **No `src/` change.** The deliverable is a bisect that refutes the premise it was dispatched on, plus a root cause that replaces every carrier-specific hypothesis in S2c…S2i.

## 1. The reported regression does not exist

One probe at three points, with a **fresh `JS2WASM_TEMPORAL_CACHE` per point** so no stale artifact can answer for a compiler (`buildTemporalProvider` + `compileWithTemporalGlobal`, `--target standalone` / `hostBridge:"off"`, host-free `instantiateLinkedProject(result, {})`):

| point | what it is | keys | provider bytes | cache key | polyfill src sha | init |
| `c01abc32` | S2h, main merged, **before** #5795 | **0** | 3,282,057 | `fcd881da30bc` | `68b811af2824` | ok |
| `8b42b7ab` | S2h, + S2f/S2g merged | **0** | 3,282,057 | `fcd881da30bc` | `68b811af2824` | ok |
| `d1f9cbb3` | S2i head | **0** | 3,311,806 | `fcd881da30bc` | `68b811af2824` | ok |

`#5795` is ruled out as an input change by measurement, not argument: the linked polyfill source is byte-identical across it and so is the provider cache key. The "9" was a real number measured **inside** the provider in S2c/S2d; S2f restated it as a boundary result and it was carried forward three slices.

## 2. Root cause — from the polyfill provider, no reference value crosses

Not the namespace, not `Object.freeze`, not `__proto__: null`. Four ordinary values, tiny hand-built provider as the control, same consumer source:

| consumer read of a provider-minted value | tiny | polyfill |
| --- | --- | --- |
| `num` (unboxed f64) | 42 | 42 |
| `typeof str === "string"` | 1 | **0** |
| `str.length` | 5 | **0** |
| `Array.isArray(arr)` / `arr.length` | 1 / 3 | **0 / 0** |
| `Object.keys({a:1,b:2}).length` | 2 | **0** |

A number crosses because it is not a reference. A string is not even a string. `Temporal` was simply the first value anyone read.

Ruled out by measurement: provider size (a 731 KB grown provider crosses), 20 module features one at a time, all six export spellings, the provider side (its own terminals answer correctly for provider-minted arguments), and the wiring (both providers export all six terminals, both consumers import them, the bodies call the right natives). The polyfill works fully **inside** its own module through the generic dynamic path — keys 9, `new qi.PlainDate(2024,1,1).day === 1`.

## 3. Two fix pieces built, measured, and deliberately not shipped

The consumer never reaches the S2d miss path for this receiver (`ref.test $Object` succeeds on it; peer call count **zero**). Adding a terminal consult makes the peer fire — and the answer is still 0, because the peer's key vec is itself unreadable. A miss-path change alone cannot fix this, so neither piece landed: they move no user-visible answer and both touch a hot native.

## Method note

An in-provider probe **is** publishable, contrary to the S2i note: a package export whose boundary is a FUNCTION with an inferred signature forces the plan to `bundled`; VALUE exports are getter boundaries and are published. Every in-provider figure above was measured that way.

## Changes

- `plan/issues/5383-standalone-temporal-provider.md` — "S2j findings" with the bisect table and the full measurement set.
- `tests/issue-5383-standalone-temporal-provider.test.ts` — the smoke test stays `it.todo`, now naming the real stop with a ~60 s reduction that needs no Temporal knowledge. 55 passed, 2 todo.

Gates: typecheck, loc-budget, func-budget, coercion-sites, oracle-ratchet, dead-exports, compiler-boundaries, host-import-policy, biome lint — all green; `equivalence-gate: 22 failing, 1720 passing, 22 known-failures in baseline` (exit 0).

---


ttraenkler:
Heads-up: this PR looks like the source of a 32-file standalone-lane test262 regression.

Cluster: `language/{statements,expressions}/class/cpn-class-{decl,expr}-fields-methods-computed-property-name-from-*` — class fields/methods with a COMPUTED property name — all failing `Expected SameValue(«null», «…»)` in the **standalone** target. It surfaced as an auto-park on an unrelated PR (#5828, run 34689448924, `Standalone regression guard (#1897)`: net -19, 32 regressions / 13 improvements); the full list came from that run's `test262-merged-report` artifact diffed against the standalone baseline.

Bisected locally with `npx tsx scripts/run-test262-paths.mts <paths> --standalone --isolate` on three representative files:

| revision | result |
| --- | --- |
| `ef6e34479b` (= `ec27f08bfd^1`, this PR's first parent) | 3 pass |
| `ec27f08bfd` (this PR's merge commit) | **3 fail** |
| `cf82f78d6d` (main tip) | **3 fail** |

Repro paths:

```
test/language/statements/class/cpn-class-decl-fields-methods-computed-property-name-from-decimal-literal.js
test/language/statements/class/cpn-class-decl-fields-methods-computed-property-name-from-identifier.js
test/language/expressions/class/cpn-class-expr-fields-methods-computed-property-name-from-math.js
```

The js-host lane is unaffected (that run's host gate was green); this is standalone-only.

PR #5822, opened 2026-09-12
fix(standalone): pin the canonical rec group open so the link value ABI is content-independent — strings, arrays and objects cross from the polyfill provider (#5383 S2k)

ttraenkler:
Land order: after PR #5820 (S2j) — this branch is stacked on it.

## The stop

Across a wasm→wasm standalone link, **no reference value crossed** from the compiled `@js-temporal/polyfill` provider ([#5383](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5383-standalone-temporal-provider) S2j): a provider-minted string answered `typeof === "string"` false, an array failed `Array.isArray`, an object enumerated nothing. A number (unboxed f64) crossed fine. A tiny hand-built provider passed every row with the same consumer.

## The mechanism (measured, not argued)

WasmGC canonicalizes a recursive type group **as a whole**. The ten types of the frozen link ABI (`RUNTIME_RECGROUP_TYPE_NAMES` — the vec family plus the string family) differed between provider and consumer in **exactly one bit**:

| module | canonical group `[0..9]` hash |
| --- | --- |
| consumer | `2d74afdb81b1` |
| tiny provider (works) | `2d74afdb81b1` |
| **polyfill provider (fails)** | **`bc74a429728b`** |

```
member 2 ($__vec_externref):
   consumer: subfinal[t0] struct(mut i32, mut (ref null t1))
   provider: sub     [t0] struct(mut i32, mut (ref null t1))
```

Nine of ten members are byte-identical. That one bit makes all ten a different runtime type, and every consumer-side check on a peer-minted value is a type-identity test — `ref.test $AnyString` for `typeof x === "string"`, `ref.test $__vec_externref` for `Array.isArray`, the `$Object` walk for `Object.keys`. So they all failed at once, and `Temporal` was merely the first value anyone read.

**Why the bit differed.** `markLeafStructsFinal` finalizes a struct when nothing in *that module* subtypes it. The polyfill uses `arguments`, which on the standalone lane registers `$__arguments_vec_externref` as a subtype of `$__vec_externref`; the consumer does not, so its copy went `final`. `arguments` is only the trigger that happened to be reachable — any module-local subtype of any member does the same. The real defect: **the identity of a frozen cross-module ABI was a function of module content.**

## The fix

`finalizeLeafStructTypes` adds every canonical-group member to `keepOpenTypeIdxs`, so the group's encoding is a constant of the ABI. This reuses the mechanism `markLeafStructsFinal` already documents for exactly this case ("ABI roots whose non-finality is observable across separately compiled modules" — the funcref-wrapper root, #5349's ArrayBuffer byte vec).

Gated on `mod.canonicalRuntimeRecGroup`, which only a runtime provider / linked namespace / explicit `canonicalRuntimeTypes` module carries.

**Byte A/B over 20 artifacts** (10 module shapes × {gc, unlinked standalone}): every sha256 identical. The JS-host lane and unlinked standalone modules are untouched.

## S2 smoke test, per assertion

| assertion | base (all three S2j bisect points) | now |
| --- | --- | --- |
| `Object.keys(Temporal).length === 9` | 0 | **9 — passes** |
| `new Temporal.PlainDate(2024,1,1).day === 1` | −1 | **1 — passes** |
| `Temporal.Duration.from({hours:1}).total("minutes") === 60` | threw | still throws — `it.todo` with the stop named |

The remaining one is **not** a boundary problem: the Duration crosses and reads correctly, and `.total(…)` fails identically *inside the provider's own module*. It throws `RangeError: unit must be one of year, …, nanosecond, null ×10, not minutes` — the ten plural unit names are missing, from `Object.fromEntries(nt.map(([e,t]) => [t,e]))`. Reduced: `Object.fromEntries` over a computed pair list is partly unimplemented on the standalone lane, and in the destructured-arrow form the polyfill uses it silently builds the right keys with lost values.

## Tests

- `#5383 S2k — the shared VALUE ABI survives a provider that uses \`arguments\`` — string / array / object across a real linked provider, plus the rec-group-fingerprint invariant. **Both verified to FAIL on base.**
- `#5383 S2 smoke — the real standalone Temporal provider` — through `buildTemporalProvider` + `compileWithTemporalGlobal`, host-free, via a new child-process harness (`tests/dogfood/temporal-s2-smoke-harness.mjs`; the 3.3 MB provider compile OOMs a vitest worker in-process).

## Gates

typecheck · loc-budget (+CI base) · func-budget (+CI base) · coercion-sites · oracle-ratchet · dead-exports · compiler-boundaries · host-import-policy · biome lint — all green.

`equivalence-gate: 22 failing, 1720 passing, 22 known-failures in baseline. ✓ No new equivalence regressions.`

Suites run green: #5383 (58 passed, 2 todo), #5384, #4628 (incl. the heavy linked-provider test), #5225, #5364, #5353, #5194, #5195, #2158, `canonical-recgroup`.

Pre-existing red, measured identical on base: `issue-2151` (1), `issue-2151-mixed-spread` (1), `issue-3610-standalone-prototype-receiver-brand` (1), `issue-1051` (3) — 6 total. `issue-5318-r4-computed-accessor-keys` OOMs on both trees.

---


ttraenkler:
Dependencies cleared: S2h/S2i/S2j are on main via #5820. Head `76c0d2f9` has current `origin/main` merged in (clean) and is re-validated on the merged head: typecheck, compiler-boundaries inventory against origin/main, #5383 suite 58 pass / 2 todo. Releasing the hold; #5824 (S2l) stays held until this lands.

---

PR #5823, opened 2026-09-12
fix(#5367, #5372): drive awaits nested in expressions and inline Promise.all on the JS-host lane

ttraenkler:
Resumes and finishes the interrupted `issue-5367-5372-await-in-initializer` work: two JS-host async defects with one root cause — the host activation gate declined shapes it could not plan and the whole function fell to the legacy synchronous pass-through, where `await` is an identity.

- [#5367](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5367-await-promise-all-inline-loses-result) — `await Promise.all(...)` inline yielded a default-initialised tuple / empty array and never waited: the #1796-era `Promise.<combinator>` carve-out in `asyncFnNeedsHostDrive` is removed (host lane only).
- [#5372](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5372-await-in-conditional-operand-not-suspended) — `const u = cond ? await f() : v` (and `cond && await p`, `await (cond ? await a : b)(u, i)`, `return cond ? await a : b`, the assignment form) left the binding holding the Promise: new `src/codegen/async-await-hoist.ts` lowers those statements into the CFG items `planTryCatchCfg` already drives (host lane only, `isHostAsyncLane`); a declarator's own name node counts as initialisation in `isConstIdentifierAssignmentTarget`; the wasi/standalone lanes keep the pre-change planner verbatim (standalone binary byte-identical, sha256 `0ed867fc616cfa8d` base and fix).
- Trap guard: driving marked's async arm reached the pre-existing [#5345](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5345-marked-illegal-cast-walktokens) `illegal cast` inside a host-driven resume — an uncatchable trap escaping a host reaction killed the worker (marked `Hooks.test.js` 9/30 → 0/30). A frame-aware reaction import `Promise_then2_frame` rejects the frame's own result promise instead; its runtime side lives in `src/runtime/promise-then-reactions.ts` (the three `then` arms fold into one dispatch — `src/runtime.ts` stays exactly at the #4401 ceiling, 19725 lines).
- Hoisting exclusion: an awaited operand holding a block-bodied IIFE stays off the hoisting lane (`awaitedExprHasBlockIife`) — the resume emitter is invalid for that shape on every path (pre-existing, filed as [#6410](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6410-block-iife-in-awaited-expression-invalid-resume)); without the gate hono `src/helper/dev/index.test.ts` stopped validating (1/8 → 0/8).

**Regression tests** (untyped `.js` two-file fixtures, compiled and instantiated like the dogfood worker): `tests/issue-5367-await-promise-all-inline.test.ts` 8 rows (parent 5 fail), `tests/issue-5372-await-in-conditional-operand.test.ts` 28 rows (parent 15 fail, incl. the async-IIFE-`.catch`-in-async-function module that does not validate on the parent); `tests/issue-3722-await-ternary-label-false-positive.test.ts` re-contracted to await the returned Promise. 40/40 on the fix, 20 fail on the parent at the same head.

**Scoped test262** (`language/expressions/await`, `language/statements/async-function`, `built-ins/Promise/all`), parent vs fix at `225f400089`: host lane 190 → 191 / 387 (+1: `language/statements/async-function` 66 → 67, other categories identical), standalone lane 161 → 161 / 387 (per-category identical).

**A/B, 17 dogfood suites at one head, per test file** (base = upstream/main sources by file copies):

| suite | base → fix |
| --- | --- |
| marked | 9/30 → 16/30 (`test/unit/Hooks.test.js` +7) |
| hono | 255/324 → 255/324 (identical per file; `src/utils/concurrent.test.ts` 0/6 both → #6409; `src/helper/dev/index.test.ts` 1/8 both) |
| webpack 16/16 · three 17/18 · clsx 32/32 · cookie 63740 · lodash 59/62 · redux 67/82 · axios 202/231 · stylelint 108/108 · tailwindcss 13/13 · jsdom 6/6 · styled-components 9/9 · uuid 75/75 · moment 10/10 · prettier 105/151 · jest 335/356 | identical, per file |

**Residuals (recorded in the issue files, not hidden):**
- marked `Hooks.test.js` 16/30, AC asked ≥ 19/30: 7 of the 10 target async tests pass; `should postprocess async` / `should process all hooks in reverse` resolve `null` when the postprocess hook is a plain object-literal async method awaiting an already-settled promise (characterised, not reduced — every same-module mimic passes); `should preprocess options async` fails exactly like its sync twin (not async). Reachable maximum is 18/30.
- hono `src/utils/concurrent.test.ts` 0/6, AC asked ≥ 5/6: every test now gets past the `Promise.all` and fails on the pool-drain assertion — `createPool`'s pool-full retry is a nested async arrow calling itself through its own captured `const` while awaiting a call, and that recursive activation never enters the body; identical on the base at the same head → filed as [#6409](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6409-nested-async-arrow-self-call-await-call) with a two-file fixture and a 14-row reduction table.

Gates (loc/func/coercion/oracle/dead-exports/dogfood-validation/host-import-policy/compiler-boundaries inventory), `tsc -p tsconfig.ts7.json`, issue-id/spec gates: all green locally.


ttraenkler:
Park diagnosis (auto-park-bot, run 34683881441, `merge shard reports` → *Standalone regression guard (#1897)*): **main-side baseline drift, not this PR.**

- The failing verdict is `improvements=1, wasm-change regressions=32, net=-31` (standalone pass 35224 → 35193); all 32 are `test/language/{expressions,statements}/class/cpn-class-*-fields-methods-computed-property-name-from-*.js` (`Expected SameValue(«null», «2»)`), a class computed-property-name cluster with no async code.
- The **same verdict, byte for byte** (`improvements=1, regressions=32, net=-31`, `current pass=35193`) already appears in pr-5820's merge group (run 34681954437, 07:55 UTC), on a main state that predates this PR's creation (08:27 UTC) and contains none of its commits; pr-5819's group (run 34680468369, 07:20 UTC) still read `current pass=35224`. The cluster therefore landed on main between those groups (the #5819 / #5780 merges), and this group inherited it.
- This PR's own standalone measurement at the merged head is byte-identical base vs fix (probe sha256 `0ed867fc616cfa8d`, scoped `built-ins/Promise/all` · `language/expressions/await` · `language/statements/async-function` 161/387 on both sides, per category identical); the host lane's catastrophic guard in this same run reports 0 regressions.

The PR merged at 09:03:41Z (9680fb07); removing the stale `hold`. The 32-test standalone drop on main needs its own owner (`cpn-class-*` computed property names on the standalone lane, introduced by the merges between runs 34680468369 and 34681954437).

ttraenkler:
Correction on the origin (from main's merge order): the group containing #5780 + #5819 (run 34680468369, 07:20 UTC) still measured standalone pass 35224; **#5820** (`issue-5383-standalone-temporal-s2j`) merged at 07:43:22Z (ec27f08bfd) and the next group (run 34681954437, 07:55 UTC) is the first to read 35193 with the identical 32-test `cpn-class-*` verdict. So the standalone drop rode in with #5820, not with #5819/#5780 as written above; everything else in the diagnosis stands.

PR #5824, opened 2026-09-12
fix(standalone): Object.fromEntries over a computed pair list — tuple carriers are array-likes, and the call site no longer refuses by accident (#5383 S2l)

ttraenkler:
Land order: after PR #5822 (S2k) — this branch is stacked on it.

Slice **S2l** of [#5383](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5383-standalone-temporal-provider). Standalone lane only; the JS-host lane and `src/codegen/temporal-native.ts` are untouched.

## The stop

S2k left `Temporal.Duration.from({hours:1}).total("minutes")` throwing, inside the provider's own module, with `RangeError: unit must be one of year, …, nanosecond, null ×10, not minutes` — the ten PLURAL unit names missing because the polyfill builds them with `Object.fromEntries(nt.map(([e, t]) => [t, e]))`.

## Root cause — two independent defects at one call site

**(A) Silently wrong VALUES.** `Object.fromEntries`'s lib signature is `Iterable<readonly [PropertyKey, T]>`. That contextual type reaches the callback, so `[t, e]` is a **tuple** and lowers to a nominal `$__tuple_0 (struct (field $_0 externref) (field $_1 externref))` — not to the indexable pair vec the *identical* expression produces when bound to an `any` local first (confirmed by diffing the two type sections). The self-hosted `__object_fromEntries` reads each pair with `__extern_get_idx(pair, 0/1)`, which had arms for `$ObjVec`, typed vecs and closed array-like structs — the last requiring a real `length` field AND canonical integer field names, so a tuple matched none of them. It answered `undefined` for both slots, and all ten entries wrote `out[undefined] = undefined`.

**(B) Accidental refusal.** `ensureLateImport` returns a funcMap hit *before* the #1472 Phase B refusal, and `__object_fromEntries` is in funcMap only when something else in the module already pulled in `ensureObjectRuntime`. So the same source line compiled or was refused depending on unrelated module content: `Object.fromEntries(nt.map(([e,t]) => [t,e]))` compiled (its array-literal callback body ensures the runtime) while `Object.fromEntries(nt)`, `nt.slice(0)` and `nt.map((e) => e)` were refused.

## The fix

- `fillExternArrayLikeStructArms` admits **tuple carriers** as a third array-like shape for the standalone dyn-reader trio — length = field count (a constant), `_i` = index `i`, `ref.test`-guarded per type. A TS tuple value IS a JS Array at runtime, so this is the spec answer; it is also what the JS-host lane already gets from #5205's `__sget_*` exports.
- The `Object.fromEntries` call site ensures the object runtime itself and calls the native directly when `ctx.oracle.typeFactOf(arg)` says `array` or `tuple`. **No new host import** — the native is a defined function, so no import is added and no index shifts (#1984).
- A **non-indexable iterable (a `Map`) deliberately keeps refusing**: the native would walk it with `__extern_length` → 0 and hand back `{}`, which is the silent-wrong failure this slice exists to delete. #2190 owns native iterator-protocol consumption.

| form (`nt: any[]` of pairs) | base | S2l |
| --- | --- | --- |
| array literal of pairs | works | works (byte-identical) |
| `nt.map(([e,t]) => [t,e])` | 1 key `"undefined"` | 3 keys, right values |
| `nt.map((e) => [e[1],e[0]])` | 1 key `"undefined"` | 3 keys, right values |
| `nt` / `nt.slice(0)` / `nt.concat([])` / `nt.map((e) => e)` | REFUSED | 3 keys, right values |
| a `Map` | REFUSED | REFUSED (deliberate) |

## Measurements

- **Byte A/B** — 12 module shapes × {gc, unlinked standalone}, base captured by file copy before the first edit: **23 of 24 sha256-identical**. The only change is the direct subject. The gc lane is identical for all 12 shapes; standalone modules that do not use `Object.fromEntries` over a computed list are identical too, including the `tuples` and `arraylike` shapes.
- **test262** `built-ins/Object/fromEntries/**`, all 25 files × {gc, standalone}, run solo: **0 of 50 rows changed** (gc 13 pass / 12 fail; standalone 9 pass / 15 fail / 1 compile_error, both trees). 0 pass→fail.
- **Equivalence gate**: `22 failing, 1720 passing, 22 known-failures in baseline` — no new regressions.
- **Pre-existing red, unchanged**: `issue-2151` (1), `issue-2151-mixed-spread` (1), `issue-3610-standalone-prototype-receiver-brand` (1), `issue-1051` (3) = 6, the count recorded on the S2k base.

## The S2 smoke test

| assertion | S2k | S2l |
| --- | --- | --- |
| `Object.keys(Temporal).length === 9` | passes | passes |
| `new Temporal.PlainDate(2024,1,1).day === 1` | passes | passes |
| `Temporal.Duration.from({hours:1}).total("minutes") === 60` | `it.todo` | still `it.todo` — **new stop** |

The third assertion's stop is **not** the old one, and the evidence is the error text. With only the two S2l source files reverted by file copy, the in-provider message is the 175-char unit-table `RangeError` with ten `null`s; with them it is 58 chars: ``Convert JSBI instances to native numbers using `toNumber`.`` — JSBI's own `valueOf` guard, i.e. an implicit ToNumber/ToPrimitive on a JSBI BigInt instance. It is **not caused by this slice**: `total("minute")`, a singular unit always present in the table, fails identically on both trees.

Two further facts, measured on both trees so neither is an S2l regression:

- a provider-side throw does not cross the link as a catchable JS error — the consumer's own `try/catch` never runs and the raw `WebAssembly.Exception` escapes to the embedder;
- the harness's `durationHasTotal` probe reads 0 on both trees while the same question through a bound local reads 1 — #2984's path-dependent `typeof` on a *chained* member access, not a missing method (`d.toString()` works across the same boundary, and a tiny hand-written provider answers the whole chain including `d.total("minutes") === 60`).

Seven new regression tests cover both defects, the unchanged array-literal fast path, the deliberate `Map` refusal, and the tuple↔dyn-reader mechanism directly.

---


ttraenkler:
Dependency cleared: #5822 (S2k) merged. Head `f2ee955e` has current `origin/main` merged in (clean) and is re-validated on the merged head: typecheck, compiler-boundaries inventory against origin/main, #5383 suite 65 pass / 2 todo. Releasing the hold.

---

ttraenkler:
Park at 10:49 (standalone regression guard, net −19) diagnosed: **real, but on main, not this diff.** Hold stays on until the fix lands.

All 32 regressed rows are one cluster — `language/{statements,expressions}/class/cpn-class-{decl,expr}-fields-methods-computed-property-name-from-*.js`, standalone lane, every one `Expected SameValue(«null», «<value>»)`. Bisected solo with `runTest262File(…, "standalone")` on three of them:

| point | result |
| --- | --- |
| `ef6e3447` (main just before #5820) | pass · pass · pass |
| `ec27f08b` (#5820 merged: S2h + S2i + S2j) | fail · fail · fail |
| `56c665b8` (#5822 merged: S2k) | fail · fail · fail |
| `f967b43d` (current main) | fail · fail · fail |

So the regression landed with #5820 and the standalone baseline floor predates it; this PR's merge group is the first to be measured against it. The shape (a class with a computed **static field** and computed methods, read through the key) points at S2i's widening of the static sidecar — the precedence case S2i measured was a *named* static field, and a computed one goes through the runtime-keyed path the sidecar now claims. A fix lane is dispatched off `origin/main` as its own PR so it can land ahead of this stack; this PR will be re-based on it and released once the floor is green again.

---

PR #5827, opened 2026-09-12
feat(test262): wire the standalone Temporal provider into the runner and CI, per-target stamps, fail-soft, opt-in artifact (#5383 S3)

ttraenkler:
Land order: after PR #5824 (S2l) — this branch is stacked on it.

Slice **S3** of [#5383](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5383-standalone-temporal-provider): wire the standalone `Temporal` provider into the test262 runner and CI, so standalone Temporal rows can link the provider S1–S2l built instead of failing with `Temporal is not defined`. **Standalone only** — the JS-host lane and `src/codegen/temporal-native.ts` are untouched. No `src/` change.

## What was wired, and where

| file | change |
| --- | --- |
| `scripts/test262-temporal.mjs` | ONE stamp per target (`prewarm.json` unchanged for host, `prewarm-standalone.json` for standalone); `temporalProviderCompileOptions` (`{target:"standalone", hostBridge:"off"}`); `test262TemporalLaneEnabled` — the single lane gate all three lanes call |
| `scripts/prewarm-temporal-provider.mjs` | `--target host\|standalone\|both`; the cache key is computed with the same options the build uses, so a stamp always certifies the artifact the lane will ask for |
| `scripts/test262-worker.mjs` | host-only refusal dropped; provider resolved **per target**, memoised per target; the no-cold-build rule is unchanged and standalone additionally REQUIRES a stamp |
| `tests/test262-shared.ts` | `IS_HOST_LANE &&` → `TEMPORAL_LANE_ENABLED &&`, read once per process |
| `tests/test262-runner.ts` | had **no lane gate at all** — a standalone row linked the HOST provider. Now gated and keyed per target |
| `.github/workflows/test262-sharded.yml` | standalone artifact built under `run_standalone` behind a new `standalone_temporal` input; both shard jobs download the provider directory |
| `scripts/run-test262-vitest.sh` | pre-warms the provider for the lane it is about to run |

## The artifact is opt-in, and why

Linking is not free, and the cost scales with the **consumer**: assembled harness rows compile 2.6× (10.6 KB row) to 3.5× (60 KB row) slower linked than unlinked, versus 1.3× for a bare body. With the fork kill at 30 s, a default-on artifact converts large Temporal rows into per-row compile timeouts. So the standalone artifact is produced only with the `standalone_temporal` workflow input / `JS2WASM_TEST262_TEMPORAL_STANDALONE=1`; the wiring is unconditional and the stamp gate does the rest, so the default path is byte-identical to pre-S3. Bringing the link cost down is a separate slice before S5 can measure the bucket.

## The #2961 guard needs no relaxation — measured, not assumed

The plan expected the standalone host-import guard to count the provider's imports as leaks. It does not: a standalone consumer linked against the standalone provider reports `result.imports === []`. The six real wasm imports all live in the provider's `link:` namespace, which the compiler's import list deliberately excludes because the linker satisfies them. So the guard keeps its full strength for every other row — the outcome to prefer, since a widened #2961 guard is exactly how a real host-import leak would stop being visible.

## Fail soft — the hard requirement, tested from the negative side

A missing **or corrupt** standalone artifact leaves the rows unlinked, never breaks the lane:

- no stamp / truncated JSON / a stamp without a `key` / key mismatch → `test262TemporalLaneEnabled("standalone") === false`, no throw;
- the workflow's standalone build step is `continue-on-error` while the host step keeps its hard failure, now with the host-stamp guarantee asserted explicitly;
- the artifact directory is always uploadable, so `download-artifact` — which fails hard on a missing artifact — can never turn the soft path into a red standalone lane.

13 new tests in `tests/issue-5383-standalone-temporal-provider.test.ts` cover the gate, the stamps and the fail-soft paths; two `issue-5353` assertions that encoded the superseded host-only rule are updated.

## Measurement

family-123, standalone lane, fresh `JS2WASM_TEMPORAL_CACHE` per side (full table in the issue file's "S3 findings (2026-09-12)"): 0 pass→fail structurally — this family has **zero** passing standalone rows in every configuration, because every scored row is `intl402` and the provider ships the `Intl` refusal shim. Default (no artifact) vs base: 123/123 rows scored with 0 compile timeouts, where base timed out on 64; the one `fail → compile_error` flip is a row that now compiles far enough to surface its own `__temporal_*` leak, which is S4's target. Opt-in vs base over 61 aligned rows: 0 status flips. S5 needs a non-Intl family.

Gates: typecheck, coercion-sites, oracle-ratchet, dead-exports, compiler-boundaries inventory, host-import-policy, biome lint, YAML parse; LOC/func budgets with `LOC_GATE_BASE=origin/main` after restating two stranded grants measured identical on the S2l base. `equivalence-gate: 22 failing, 1720 passing, 22 known-failures in baseline` — no new regressions. Pre-existing red not from this PR: `tests/issue-5382-temporal-project-publication.test.ts` 1 of 33, identical on the S2l base.

---


ttraenkler:
Dependencies cleared: S2l is on main and the #5820 regression is fixed by #5831 (merged 12:48). Head `7e7b7110` has current `origin/main` merged in — the only conflict was the issue file's findings sections, both kept — and is re-validated on the merged head: typecheck, compiler-boundaries inventory against origin/main, #5383 suite 78 pass / 2 todo. No `src/` change in this PR. Releasing the hold.

---

PR #5828, opened 2026-09-12
fix(#6411): expand spread arguments in the host Array bridges (restores the #5819 merge-group net)

ttraenkler:
## What

`[].concat(...arrays)` handed `Array.prototype.concat` the spread SOURCE as a single argument, so it flattened one level too few — see [#6411](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6411-host-bridge-concat-spread-not-expanded).

```js
[].concat(...[[], [1], [2, 3]]);   // [[], [1], [2, 3]]  — should be [1, 2, 3]
```

Same idiom [#5361](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5361-array-splice-spread-argument-not-expanded) removed from `splice` / `push` / `Math.min`-`max` / the generic `__extern_method_call` bridge, surviving in the two HOST-ARRAY argument builders in `src/codegen/array-method-host.ts`: `compileArrayConcatExternHost` (the `__array_concat_any` bridge) and `compileArrayMethodExtern`. Both build the argument array with `__js_array_new` + one `__js_array_push` per AST node, which is exact only while every argument is ONE value.

## Why now — this is a merged-baseline regression, found by the queue

`test/built-ins/Iterator/concat/many-arguments.js` compares

```js
let iterator = Iterator.concat(...iterables);
let array = [].concat(...iterables);
```

element by element. **Both sides were wrong in the same direction**, so the file passed by coincidence: `Iterator.concat(iterables)` yielded the sub-arrays and `[].concat(iterables)` flattened to the same sub-arrays, and `sameValue` compared identical objects. #5361 fixed the `Iterator.concat` side, the coincidence broke, and the file flipped pass → fail (`Expected SameValue(«1», «») to be true` — `array[0]` was still `[]`).

That single flip is the whole of the `net_per_test -1` that auto-parked PR #5819 in the merge queue (run 34680468369: 0 improvements − 1 regression). The `[].concat(...)` half is INDEPENDENT of #5361 and predates it — measured with #5361's codegen reverted, the same probe gives the same wrong answer. This PR restores the net and fixes the underlying defect.

## How

`tryEmitSpreadHostArgs` (`src/codegen/host-method-args.ts`) is the shared entry both bridges call before their own loop: with a spread present it drives #5361's `buildSpreadArgList` with a `__js_array_push` sink; with none it returns `false` having emitted nothing, and each caller keeps its existing unrolled loop verbatim — so a spread-free call is byte-identical. `emitHostMethodCallArgs` (the #5361 caller) is re-expressed on top of it, so there is ONE place where the syntactic-vs-runtime argument count is repaired rather than a sixth copy of the loop.

## Measured

**Scoped test262 slice, both ways at one HEAD** (`upstream/main` `ef6e34479b`) — every file under `built-ins/Array/prototype/concat`, `.../push`, `.../splice` and `built-ins/Iterator/concat`, **206 files, `--isolate`**:

| | pass | fail |
| --- | --- | --- |
| base | 108 | 98 |
| fix | **109** | **97** |

Set difference over the non-pass rows: exactly one file flips — `built-ins/Iterator/concat/many-arguments.js`, fail → pass — and **no file regresses**.

Probe through `compileAndRunUpstreamModule` (untyped `.js` two-file project), harness sanity-checked with a control that fails in both lanes:

| probe | native | wasm before | wasm after |
| --- | --- | --- | --- |
| `[].concat(...[[], [1], [2,3], [4,5,6]])` | `1\|2\|3\|4\|5\|6` len 6 | **`\|1\|2,3\|4,5,6` len 4** | `1\|2\|3\|4\|5\|6` len 6 |
| `[].concat(...[a.split(':'), b.split(':')])` | `a\|b\|c\|d` len 4 | **`a,b\|c,d` len 2** | `a\|b\|c\|d` len 4 |
| `[].concat([1], [2,3])` (no-spread control) | `1\|2\|3` len 3 | `1\|2\|3` len 3 | `1\|2\|3` len 3 |
| `[].concat([1,2])` (single-arg control) | `1\|2` len 2 | `1\|2` len 2 | `1\|2` len 2 |

**17 dogfood suites, A/B at one HEAD, per test file: total delta 0, no regressions** — expected (none of them uses `[].concat(...spread)`); run as the no-regression control for a change on the host-bridge argument path.

`tests/issue-6411-host-bridge-concat-spread.test.ts` pins BOTH the resulting length and a join with a NON-comma separator (a default `,` join hides the nesting — the inner arrays stringify to the same text). Counts both ways at this HEAD: **base 1 passed / 4 failed** (the pass is the no-spread control), **fix 5 passed / 0 failed**.

Gates green locally: loc-budget, func-budget, coercion-sites, oracle-ratchet, dead-exports, host-import-policy, stack-balance, any-box-sites, pushraw, issues, compiler-boundaries inventory, equivalence-gate (22 failing / 1720 passing, no new regressions), prettier, and `tsc --noEmit` on `tsconfig.ts7.json`.

## Deliberately not done

The native-first concat lane (`compileArrayConcatNativeSpec` / `compileArrayConcatNativeDynamic`) is untouched. The host lane is where the measured regression lives; the native-first twin needs its own measurement first.

### DIFF
--- a/plan/issues/6411-host-bridge-concat-spread-not-expanded.md
+++ b/plan/issues/6411-host-bridge-concat-spread-not-expanded.md
@@ -1,0 +1,123 @@
+---
+id: 6411
+title: "The host Array bridges hand a SPREAD argument to the host as one item"
+status: done
+sprint: current
+created: 2026-09-12
+updated: 2026-09-12
+completed: 2026-09-12
+priority: high
+horizon: s
+feasibility: easy
+reasoning_effort: high
+task_type: bug
+area: compiler
+goal: correctness
+---
+
+## Problem
+
+The same idiom #5361 removed from `splice` / `push` / `Math.min`-`max` / the
+generic `__extern_method_call` bridge survives in the two HOST-ARRAY argument
+builders in `src/codegen/array-method-host.ts`:
+
+- `compileArrayConcatExternHost` — the `__array_concat_any(recv, args)` bridge;
+- `compileArrayMethodExtern` — `__extern_method_call` for a real host-owned
+  Array receiver.
+
+Both build the argument array with `__js_array_new` + one `__js_array_push` per
+AST node, which is exact only while each argument is ONE value. A spread
+contributes its RUNTIME element count, so the source array was handed to the
+host as a single argument and `Array.prototype.concat` then flattened one level
+too few:
+
+```js
+[].concat(...[[], [1], [2, 3]]);   // [[], [1], [2, 3]]  — should be [1, 2, 3]
+```
+
+## Evidence
+
+Found as a REAL merged-baseline test262 regression, not by inspection.
+`test/built-ins/Iterator/concat/many-arguments.js` compares
+
+```js
+let iterator = Iterator.concat(...iterables);
+let array = [].concat(...iterables);
+```
+
+element by element. **Both sides were wrong in the same direction**, so the file
+passed by coincidence: `Iterator.concat(iterables)` yielded the sub-arrays, and
+`[].concat(iterables)` flattened to the same sub-arrays, and `sameValue`
+compared identical objects. #5361 fixed the `Iterator.concat` side; the
+coincidence broke and the file flipped **pass → fail**
+(`Expected SameValue(«1», «») to be true` — `array[0]` was still `[]`).
+
+That is the whole of the `net_per_test -1` that parked PR #5819 in the merge
+queue (run 34680468369, 0 improvements − 1 regression). The `[].concat(...)`
+half is INDEPENDENT of #5361 and predates it — measured with #5361's codegen
+reverted, the same probe produces the same wrong answer.
+
+## Acceptance criteria
+
+1. `[].concat(...arrays)` flattens every spread element, for inline array
+   literals and for host (externref) arrays, with the correct length.
+2. A spread interleaved with positional arguments keeps argument order.
+3. A spread-free call emits byte-identical code (the unrolled loop is kept).
+4. `built-ins/Iterator/concat/many-arguments` passes again.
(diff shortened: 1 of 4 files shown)


ttraenkler:
Diagnosed the auto-park (run 34689448924, `merge shard reports` → `Standalone regression guard (#1897)`, `GATE FAIL: net_per_test -19`, 13 improvements − 32 regressions).

**Not this PR — a main-side regression this PR was in the group for.** All 32 regressions are one cluster, `language/{statements,expressions}/class/cpn-class-{decl,expr}-fields-methods-computed-property-name-from-*` (class fields/methods with a COMPUTED property name), all `Expected SameValue(«null», «…»)`. Nothing in that shape goes through `[].concat(...)` or either host Array argument bridge.

Bisected locally against the standalone lane (`scripts/run-test262-paths.mts --standalone --isolate`, three representative files from the cluster):

| revision | result |
| --- | --- |
| `ef6e34479b` (this PR's base) | 3 pass |
| `ef6e34479b` + this PR's change | 3 pass |
| `ec27f08bfd` (= base + PR #5820, before #5823) | **3 fail** |
| `cf82f78d6d` (main tip, includes this PR) | **3 fail** |

`ec27f08bfd^1` is exactly `ef6e34479b`, so the flip is introduced by **PR #5820 (`issue-5383-standalone-temporal-s2j`)**, which landed between this PR's base and the merge group. Commented there with the same evidence.

The full 32-file list came from this run's own `test262-merged-report` artifact diffed against the standalone baseline, not from inference.

This PR is already merged, so the `hold` label is only a record; nothing is waiting to be re-enqueued, and I have not removed it.

PR #4620, opened 2026-08-16
fix(website): keep donut annotations and report chart state consistent

ttraenkler:
## Summary
- hide zero-count donut annotations and legend entries
- keep donut annotations on a shared orbit, separating crowded labels angularly
- prevent report legend and annotations from overlapping controls
- preserve donut animation, edition timeline, and history chart state when switching between JS-host and standalone modes
- add regression coverage for zero-count and crowded annotations

## Validation
- `pnpm exec vitest run tests/t262-donut.test.ts tests/issue-1777.test.ts tests/issue-4362-landing-host-standalone-toggle.test.ts --pool=forks --poolOptions.forks.singleFork=true --no-file-parallelism --reporter=dot`
- `pnpm run typecheck`
- `pnpm exec prettier --check website/components/t262-charts.js tests/t262-donut.test.ts`
- Browser QA confirmed the ES2017 selection and chart elements survive both mode switches.

Local hooks were bypassed because unrelated working-tree changes triggered the repository oracle-ratchet check; the focused checks above pass.

PR #5836, opened 2026-09-12
fix(#5341): `.call` keeps its receiver at any arity, not just an exact one

ttraenkler:
Closes [#5341](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5341-axios-residual-buckets).

## Mechanism

`fn.call(thisArg, …)` **dropped the receiver at every arity but an exact one.**

The receiver-correct trampoline ([#3796](https://js2wasm.loopdive.com/dashboard/issue.html?slug=3796-named-this-call)) is reached only through
`resolveNamedThisCallTarget` in `src/codegen/named-this-call.ts`, whose
admission gate demanded `userArguments.length === declaration.parameters.length`.
Everything else fell through to the ordinary `.call` lowering in
`src/codegen/expressions/calls.ts`, which evaluates `thisArg` and then literally
`drop`s it. That is a silent wrong answer, not a refusal — the callee read the
**ambient** receiver.

axios `lib/core/transformData.js` is exactly that shape:

```js
export default function transformData(fns, response) {
  const config = this || defaults;   // `this` was lost → config = defaults
  const context = response || config;
  let data = context.data;           // → undefined instead of ''
  utils.forEach(fns, function transform(fn) { data = fn.call(config, data, …); });
  return data;
}
```

called as `transformData.call({ data: '' }, fns)` — one argument into two
formals. Three appending transformers then produced `'undefinedfoo'` instead of
`'foo'`, and every `transformResponse` case that reads its body out of `this`
answered `undefined`.

Reduced to a two-file untyped `.js` fixture, the whole defect is arity
(callee declares 2 formals and reads `this`):

| call | native | wasm before | wasm after |
| --- | --- | --- | --- |
| `f.call(t, 1, 2)` | `T\|1\|2` | `T\|1\|2` | `T\|1\|2` |
| `f.call(t, 1)` | `T\|1\|undefined` | `NO-THIS\|1\|undefined` | `T\|1\|undefined` |
| `f.call(t, 1, 2, 3)` | `T\|1\|2` | `NO-THIS\|1\|2` | `T\|1\|2` |
| `f.apply(t, [1])` | `T\|1\|undefined` | `NO-THIS\|1\|undefined` | `T\|1\|undefined` |
| `f.bind(t)(1)` | `T\|1\|undefined` | `NO-THIS\|1\|undefined` | `T\|1\|undefined` |

## Fix

The arity clause is removed from the admission gate — one file, one clause.

It is sound to admit every arity because the operand stack the caller builds is
**always `paramTypes.length` wide**, whatever the source argument count:

- under-application pads — optional-param sentinels first, then `pushDefaultValue`;
- over-application either marshals the overflow through the extras-argv global
  (`emitSetExtrasArgv`, which pushes no operand) or compiles and drops it;
- a rest declaration packs its trailing arguments into the single vec parameter.

The trampoline's signature is exactly `[externref this, ...targetParams]`, so
that stack fits it unchanged. Admission only decides whether the receiver is
**installed** or **discarded**.

## Counts

axios **202/231 → 208/231** at one HEAD (`upstream/main` `cf82f78d6d`). Six
flips, all `failed → passed`, nothing else in the suite moved:

| file | before | after |
| --- | --- | --- |
| `tests/unit/core/transformData.test.js` | 2/4 | 4/4 |
| `tests/unit/transformResponse.test.js` | 1/6 | 5/6 |

## Tests

`tests/issue-5341-under-applied-call-receiver.test.ts` — 15 cases on untyped
`.js` two-file fixtures. **11 fail on the parent commit and pass with the fix**;
4 controls pass both ways (exact arity, rest at exact-or-more arity, a callee
that ignores `this`, an under-applied call with no receiver at all), plus an
anti-vacuity control that proves the harness can still observe an absent
receiver.

**`tests/issue-3796-named-this-call.test.ts` was already RED on
`upstream/main`**, for two reasons unrelated to this change, and is refreshed
here:

- it asserted `catch_all` / `rethrow 0` in a **standalone** module, which has
  lowered the trampoline's restore-and-rethrow through standardized EH
  (`try_table` + `catch` + `throw`) since #4620;
- it asserted `$__named_this_call_readsThis_` is absent when #3983 legitimately
  emits it for the `.apply` site in the same fixture.

Its over-arity negative moved to a positive per this change; the runtime answer
it guards is unchanged at `2116`.

## Residuals

- [**#6416**](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6416-arguments-length-under-applied-call) — `arguments.length` inside an under-applied target still reports
  the FORMAL count (`f.call({}, 1)` into `function f(a, b)` answers `2`).
  Independent of the receiver; reads identically before and after this fix.
- [**#6417**](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6417-axios-residual-round-2) — the other 23 axios failures, re-measured after this fix:
  `buildURL` (6), the `instance mismatch` cluster (4, belongs to [#5347](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5347-redux-getprototypeof-compiled-class-instance)), the
  "validator returned 1" cluster (3), `isX` (3), `isNativeError` (2),
  `composeSignals` (1), and `platform` (2 — a `randomFillSync` host-shim gap
  that is not a compiler bug).

### DIFF
--- a/plan/issues/5341-axios-residual-buckets.md
+++ b/plan/issues/5341-axios-residual-buckets.md
@@ -1,10 +1,11 @@
 ---
 id: 5341
 title: "axios residual: 31 failures across nine files after the three landed blockers — prioritised by bucket"
-status: ready
+status: done
 sprint: current
 created: 2026-09-05
-updated: 2026-09-05
+updated: 2026-09-12
+completed: 2026-09-12
 priority: high
 horizon: m
 feasibility: medium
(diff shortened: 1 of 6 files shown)


ttraenkler:
Self-imposed `hold` by the authoring agent: the 17-suite dogfood A/B (base vs fix at one HEAD, per test file) is still running. Every required check is green and `mergeStateStatus` is CLEAN; the hold is only to keep the PR out of the queue until the A/B confirms no suite regressed. I will remove it myself with the results. This is NOT an auto-park-bot hold.

ttraenkler:
A/B complete — the hold is moot (this merged at 14:42Z, already in the queue when I labelled it), and the result is clean.

17 dogfood upstream suites, one HEAD (`upstream/main` cf82f78d6d), base = same tree with the one changed clause reverted, both variants measured back to back on the same box, per test file:

| suite | base | fix | delta |
|---|---|---|---|
| webpack | 16/16 | 16/16 | 0 |
| three | 17/18 | 17/18 | 0 |
| clsx | 32/32 | 32/32 | 0 |
| cookie | 63740/63740 | 63740/63740 | 0 |
| lodash | 59/62 | 59/62 | 0 |
| redux | 67/82 | 67/82 | 0 |
| **axios** | **202/231** | **208/231** | **+6** |
| stylelint | 108/108 | 108/108 | 0 |
| tailwindcss | 13/13 | 13/13 | 0 |
| jsdom | 6/6 | 6/6 | 0 |
| styled-components | 9/9 | 9/9 | 0 |
| uuid | 75/75 | 75/75 | 0 |
| marked | 16/30 | 16/30 | 0 |
| moment | 10/10 | 10/10 | 0 |
| prettier | 105/151 | 105/151 | 0 |
| jest | 335/356 | 335/356 | 0 |
| hono | 258/324 | 258/324 | 0 |

Every status change across all 17 suites, all `failed -> passed`: axios `core/transformData` (supports an array of transformers; passes headers through to transformers) and `transformResponse` (parses json; ignores XML; does not parse the empty string; does not parse undefined). Nothing else moved in either direction.

The table is recorded on the issue file via the open docs PR #5844. Removing the hold label.

PR #5844, opened 2026-09-12
docs(issues): file #6424 (worker uncaughtException still zeroes a file) and #6425 (hono TextEncoder)

ttraenkler:
Two issue files, both surfaced while fixing [#5369](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5369-unhandled-host-rejection-zeroes-test-module) (PR #5843). Docs-only — no `src/`, `tests/`, `scripts/`, `.github/` or `benchmarks/` change.

**[#6424](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6424-worker-uncaught-exception-still-zeroes-file)** — the dogfood Wasm worker still has no `uncaughtException` listener, so a host throw from a stray timer zeroes the whole test file exactly the way an unobserved rejection used to. #5369 fixed the rejection half; the issue records *why* the other half was left alone (an uncaught exception in the worker is usually a harness bug, and recording-and-continuing risks turning a fast readable failure into a 180 s worker timeout) and the in-loop / out-of-loop rule a fix needs to distinguish the two.

**[#6425](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6425-hono-crypto-textencoder-not-a-constructor)** — hono `src/utils/crypto.test.ts` reads 0/4 in the Wasm lane on `TextEncoder is not a constructor`, even though `getWebHostConstructors()` does forward the binding, so the defect is on the compiled side of the extern-constructor boundary. [#5338](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5338-hono-ipaddr-null-string-split) called it "adjacent, do not chase" and closed without filing it. Filed now because #5369 exposed its fourth casualty: that test was reading `passed` only because it compared two unawaited Promises while dropping two `TextEncoder` rejections on the floor.

### DIFF
--- a/plan/issues/6424-worker-uncaught-exception-still-zeroes-file.md
+++ b/plan/issues/6424-worker-uncaught-exception-still-zeroes-file.md
@@ -1,0 +1,72 @@
+---
+id: 6424
+title: "An uncaught exception in the dogfood Wasm worker still zeroes the whole test file — the other half of #5369"
+status: ready
+sprint: current
+created: 2026-09-12
+updated: 2026-09-12
+priority: medium
+horizon: s
+feasibility: medium
+reasoning_effort: medium
+task_type: infra
+area: testing
+goal: correctness
+---
+
+## Problem
+
+[#5369](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5369-unhandled-host-rejection-zeroes-test-module)
+removed the whole-file cliff for one failure mode — an unobserved host-promise
+**rejection** — by installing a sink in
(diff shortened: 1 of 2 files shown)


ttraenkler:
Added one docs-only commit to this branch per the project's single-open-docs-PR rule: `docs(#5341)` records the 17-suite dogfood A/B for the under-applied `.call` receiver fix that landed in #5836. It touches only `plan/issues/5341-axios-residual-buckets.md` — no code, no other issue file.

PR #5856, opened 2026-09-12
fix(#5365): a compiled closure keeps its arity across a call boundary

ttraenkler:
Slice 1 of [#5365](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5365-host-closure-bridge-loses-length-and-name)
— `Function.prototype.length` of a compiled closure that crossed a call boundary
as a value, in `gc`/JS-host mode.

## The plan's mechanism was wrong, and measuring it first is what fixed this

The issue (and its `String(f)` row) said the callee holds the **host bridge
wrapper** `_wrapWasmClosure` mints, so slice 1 should `defineProperty` a `length`
onto it at wrap time. Probed through the dogfood runner on `upstream/main`
d4108568d4:

```
wasm:   len=0  name=undefined  hasOwnName=false  tag=[object Object]  typeof=function
native: len=2  name=f          hasOwnName=true   tag=[object Function] typeof=function
```

`[object Object]` and `hasOwnName=false` are not a JS function — the callee holds
the **raw WasmGC closure carrier**. `String(f)` looks bridge-shaped only because
`compiledClosureNativeSource` puts the same native-code facade on a carrier. A
marker return at the head of `_wrapWasmClosure` never fired; one in the
`case "extern_get"` intent binding did. The read lowers to
`__extern_get(carrier, "length")` — there is no wrap on this path at all.

## Root cause — a wrong answer, not a missing one

`__extern_get`'s WasmGC arm walks sidecar → descriptor table → delete tombstone →
`masksField` → `__sget_<key>` field getter. A closure struct has no field-name
registry, so `_structOwnFieldStatus` says "unknown shape" and `wsh.readField`
probes the getter anyway. `__sget_length` exists in almost every real module —
minted for the vec shape whose struct genuinely has a `length` field — it
cast-succeeds on the closure and returns its **miss-default 0**. The #1629
anti-pattern, on the one receiver with no positive discriminator. (Without a vec
in the module there is no getter to probe and the read answered `undefined`
instead; both are wrong, the `0` is the silent one, and it is the one every
package hits.)

## Fix

`src/runtime/compiled-closure-length.ts` answers from `__closure_arity` — the
`$arity` header slot (#3673) every funcref-wrapper struct already carries, via an
export that already exists. No codegen, no new struct field, no module-size
delta.

Wired in **ahead of** the `__sget_` probe at three sites in `src/runtime.ts`,
gated on `ownFieldStatus !== true`: both `__extern_get` bindings (the by-name one
and its `case "extern_get"` intent twin — the live one for a compiled module, and
a twin that had to be found by marker, since the by-name edit alone changed
nothing) and `__extern_length` for the numeric lowering (`handler.length > 1`).
The sidecar, descriptor table and delete tombstone all run first, so
`defineProperty(f, "length", …)`, `delete f.length` and a struct that really owns
a `length` field all still win.

## Counts both ways

`tests/issue-5365-host-closure-length.test.ts`, untyped `.js` two-file fixtures:
**7 failed / 5 passed** on the parent → **12 passed** with the fix. The 5 that
pass both ways are the anti-vacuity controls — direct `f.length`, a live array
length, a string length, an object that really owns a `length` field, and an
explicit `defineProperty` override.

| declaration | parent | fix | spec |
| --- | --- | --- | --- |
| `(a, b) => a + b` | 0 | **2** | 2 ✓ |
| `function g(a,b,c)` | 0 | **3** | 3 ✓ |
| `() => 1` | 0 | 0 | 0 ✓ |
| `(a, ...rest) => a` | 0 | **1** | 1 ✓ |
| `(a, b = 1) => a` | 0 | **2** | 1 ✗ |

## A/B — 17 dogfood suites, base vs fix at one head (upstream/main d4108568d4)

| suite | base | fix | delta |
| --- | --- | --- | --- |
| webpack | 16/16 | 16/16 | +0 |
| three | 17/18 | 17/18 | +0 |
| clsx | 32/32 | 32/32 | +0 |
| cookie | 63740/63740 | 63740/63740 | +0 |
| lodash | 59/62 | 59/62 | +0 |
| redux | 67/82 | 67/82 | +0 |
| axios | 202/231 | 202/231 | +0 |
| stylelint | 108/108 | 108/108 | +0 |
| tailwindcss | 13/13 | 13/13 | +0 |
| jsdom | 6/6 | 6/6 | +0 |
| styled-components | 9/9 | 9/9 | +0 |
| uuid | 75/75 | 75/75 | +0 |
| marked | 16/30 | 16/30 | +0 |
| moment | 10/10 | 10/10 | +0 |
| prettier | 105/151 | 105/151 | +0 |
| jest | 335/356 | 335/356 | +0 |
| **hono** | **258/324** | **261/324** | **+3** |

One per-file mover, no losses anywhere: hono `src/helper/dev/index.test.ts`
**1/8 → 4/8**.

Also green: `equivalence-gate` (1720 passing, 22 known-failures, no new
regressions), `check:dogfood-validation`, `check:oracle-ratchet`,
`check:dead-exports`, `check:coercion-sites`, LOC/function budgets (allowances in
#5365's frontmatter with dated rationale), `check:host-import-policy`
(`runtimeTsLines` ceiling raised to the merged 19735, the usual re-merge drift).

## Residuals, both filed

- [#6427](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6427-host-mode-closure-name-carrier)
  — `name` is untouched, and a defaulted parameter still reports the declared
  count (`(a, b = 1)` → 2, §15.1.5 says 1). `$arity` cannot be re-pointed:
  `closure-exports.ts` widens an under-applied dispatch to `max(n, $arity)` and
  would stop padding omitted arguments (#4436 R2). It also records why slice 2's
  carrier **cannot** be the lookup table the plan preferred: there is no wrap to
  hand an index to, and a `ref.test` ladder over struct types cannot separate
  declarations, because WasmGC canonicalizes structurally. The index must live on
  the carrier — which is what `$fnmeta` (#4437) already is.
- [#6429](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6429-namedevaluation-static-name-folds)
  — the static folds `o.h.name` → `"h"` and `arr[0].name` → `""` are a
  compile-time NamedEvaluation bug and never reach the runtime.
- `_wrapWasmClosure`'s wrapper still reports `length === 0` when a compiled
  closure is handed to a *real host* function that reflects on it. No dogfood
  suite measured it, and the `arity` that function is handed is the call-site
  expectation rather than the declaration's — so stamping it there would be wrong
  as often as right. Folded into #6427, which has the carrier the stamp needs.


ttraenkler:
Unholding — the park is the spurious kind its own footnote describes: an infra step, not a verdict step.

Run 34713008706, every non-success job:

- `equivalence-shard (4)` — failing step `Setup Node and pnpm (cached)`: `actions/setup-node` could not fetch Node 25 (`Not found in manifest. Falling back to download directly from Node` → `##[error]read ECONNRESET`). No equivalence test ran in that shard.
- `equivalence-gate` — failing step `Verify every equivalence shard passed its baseline gate`, which is a one-line relay of the matrix result (`SHARD_RESULT: failure`). It has no verdict of its own.

Shards 1,2,3,5,6,7,8, `quality` and the test262 jobs all succeeded in that same merge_group run, so no baseline gate reported a regression anywhere.

Locally on this branch at the merged state: `scripts/equivalence-gate.mjs` → 1720 passing, 22 failing, 22 known-failures in baseline, "No new equivalence regressions". The change is runtime-only and its 17-suite dogfood A/B is 16 suites byte-identical + hono 258→261 with zero per-file losses.

Removing `hold` so auto-enqueue can re-admit it.

PR #5868, opened 2026-09-12
fix(#6415): a host predicate read as a value keeps its boolean crossing back to a typed caller

ttraenkler:
Closes [#6415](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6415-isview-first-class-value-host-lane).

## What was wrong

In an untyped `.js` module, a host builtin read as a **first-class value** and
returned to a **typed** caller always answered `false`:

```js
// mod.js (untyped)
export function viaIsView(value) {
  const f = ArrayBuffer.isView;   // value read, not a direct call
  return f(value);
}
```

```ts
// entry.ts (typed)
const viaIsView_ = viaIsView as unknown as (value: unknown) => boolean;
viaIsView_(new Uint8Array(3));   // false — Node says true
```

The direct spelling `ArrayBuffer.isView(value)` in the same module answers
`true` (that is what [#5370](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5370-typed-array-carrier-host-boundary-fidelity)
fixed), so the two spellings of one predicate disagreed.

## Mechanism

Neither arm the issue proposed is responsible, and the defect is **not
`isView`-specific**. Measured on `e8a778638f`:

* **The untyped callee is correct.** The same `f(value) ? "view" : "not-view"`
  computed *inside* `mod.js` answers `"view"` on the parent. The value read is
  `__get_builtin("ArrayBuffer")` → `__extern_get("isView")` — a real host
  function — and `__call_function`'s runtime already `_wrapForHost`s every
  WasmGC argument.
* **The value is lost at the typed caller.** The untyped callee's Wasm
  signature returns `externref` (it boxes via `__box_boolean`); the call site
  expects the boolean-branded `i32` this lane lowers `boolean` to. The
  funcref-ladder return bridge (`scalarBridgePlan` in
  `src/codegen/expressions/call-identifier.ts`) had **no externref →
  boolean-i32 arm**, so the **live** arm fell into the dead-arm placeholder
  `drop; i32.const 0`. WAT confirms `call_ref N; drop; i32.const 0`.

`Array.isArray` and `Object.is` read as values fail identically — it is the
boolean return bridge, not the property read.

## The fix

One arm at the end of `scalarBridgePlan`:

```ts
if (isHostExtern(from) && to.kind === "i32" && to.boolean === true) {
  return helpers.isTruthyIdx === null ? null : [{ op: "call", funcIdx: helpers.isTruthyIdx }];
}
```

`__is_truthy` is the coercion engine's canonical host-lane ToBoolean primitive
— exact for a real boolean, spec-correct if the callee hands back a
non-boolean the caller's declared type says to read as one. Deliberately **not**
widened to plain `i32`: that carrier also spells native ints and symbol ids.
`scalarAbiTypesMatch` untouched.

`isTruthyIdx` is plumbed through the `helpers` type, `dispatchBridgePlan`
(`ctx.funcMap.get("__is_truthy") ?? null`) and `theoreticalHelpers`
(`isTruthyIdx: 0`). That last field is load-bearing: it is what makes the
`needsScalarBridge` pre-pass see a non-null plan and run `addUnionImports` +
`flushLateImportShifts` **before** the ladder is baked, so no late import can
shift an already-emitted `ref.func` (the #2174 desync).

**Standalone / WASI byte-identical**: `scalarBridgePlan` returns `null` for
`ctx.standalone || ctx.wasi` two lines in, above every arm.

## Regression test

`tests/issue-6415-host-value-predicate-boolean-return.test.ts` — untyped
`mod.js` + typed `entry.ts`, results as strings, `target: "gc"`,
`platform: "web"`, `deferTopLevelInit`.

**Parent: 4 failed / 5 passed. With the fix: 9 / 9.**

| case | parent | fix |
| --- | --- | --- |
| `isView` of compiled `Uint8Array(3)` | `not-view` ❌ | `view` ✅ |
| `isView` of host `TextEncoder` bytes | `not-view` ❌ | `view` ✅ |
| `Array.isArray([1,2,3])` | `no` ❌ | `yes` ✅ |
| `Object.is(1, 1)` | `no` ❌ | `yes` ✅ |
| **anti-vacuity** `isView([1,2,3])` | `not-view` ✅ | `not-view` ✅ |
| **anti-vacuity** `Array.isArray(new Uint8Array(3))` | `no` ✅ | `no` ✅ |
| **anti-vacuity** `Object.is(1, 2)` | `no` ✅ | `no` ✅ |
| **control** `f(v) ? true : false` (i32 return, no bridge) | `view` ✅ | `view` ✅ |
| `DataView` — two spellings agree | ✅ | ✅ |

## Dogfood A/B

Base vs fix at one HEAD (`e8a778638f`), all 17 suites, one at a time.
**No suite moved**, as the plan predicted — the shape needs a typed caller
casting an untyped module's function to `=> boolean`, and the suites are
untyped-to-untyped.

## Gates

`check:oracle-ratchet` · `check:dead-exports` · `check:dogfood-validation` ·
`check:host-import-policy` all green. `tsc --noEmit -p tsconfig.ts7.json`
clean.

`check:compiler-boundaries` fails identically on the parent and with this
change (same output but for the content fingerprint) — pre-existing
`graphComplete: false` state on `upstream/main`, unrelated to this change,
which adds no module.

Growth granted in `plan/issues/6415-…md` frontmatter with dated rationale:
`loc-budget-allow` and `func-budget-allow` for +24 lines in
`call-identifier.ts` / `compileIdentifierCall` (the arm must live beside the
other `scalarBridgePlan` rows — they are one ordered ladder whose pre-pass
reserves imports from their combined answer), and `coercion-sites-allow` for
`__is_truthy` 1 → 2 in that file (a single `call` to the coercion engine's own
ToBoolean primitive, which is what the gate asks for).

## Out of scope

`ArrayBuffer.isView(new DataView(new ArrayBuffer(8)))` answers `false` in the
**direct** spelling too on this HEAD — pre-existing and independent. Filed as
[#6433](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6433-arraybuffer-isview-dataview-host-lane)
(issue file included here). This change makes the two spellings agree; it does
not change what they agree on, and the test asserts only the agreement.

`tests/issue-5150-es2015-buffers.test.ts` has the same 8 failed / 10 passed
(identical failure set) on the parent and with this change — pre-existing on
`upstream/main`, standalone lane untouched.


ttraenkler:
## Dogfood A/B (post-merge record)

Measured at one HEAD in one worktree, one suite at a time; the only difference
between the two columns is this PR's edit to
`src/codegen/expressions/call-identifier.ts`. The box was shared with other
agents throughout (load average 35-223), which is why the base column is
partial — see the note below.

| suite | base (this HEAD) | fix (this HEAD) | main anchor 2026-09-12 | delta |
| prettier | **107/151** | **108/151** | 107/151 | **+1 (this change)** |
| jest | 335/356 | 335/356 | 335/356 | 0 |
| hono | 261/324 | 261/324 | 259/324 | 0 (anchor stale by +2) |
| marked | 16/30 | 16/30 | 16/30 | 0 |
| redux | 67/82 | 67/82 | 67/82 | 0 |
| three | 17/18 | 17/18 | 17/18 | 0 |
| lodash | 59/62 | 59/62 | 59/62 | 0 |
| webpack | (not re-run) | 16/16 | 16/16 | 0 vs anchor |
| clsx | (not re-run) | 32/32 | 32/32 | 0 vs anchor |
| cookie | (not re-run) | 63740/63740 | 63740 | 0 vs anchor |
| axios | (not re-run) | 208/231 | 208/231 | 0 vs anchor |
| stylelint | (not re-run) | 108/108 | 108 | 0 vs anchor |
| tailwindcss | (not re-run) | 13/13 | 13 | 0 vs anchor |
| jsdom | (not re-run) | 6/6 | 6 | 0 vs anchor |
| styled-components | (not re-run) | 9/9 | 9 | 0 vs anchor |
| uuid | (not re-run) | 75/75 | 75 | 0 vs anchor |
| moment | (not re-run) | 10/10 | 10 | 0 vs anchor |

**All 17 suites ran on the fix side, exit 0, `admitted` headline present.
Nothing regressed anywhere.**

Two things worth recording:

* **prettier gains one test (107 → 108)** — measured base-vs-fix at the same
  HEAD, so it is attributable to this change, not drift. The plan predicted no
  dogfood movement because the defect's shape needs a typed caller casting an
  untyped module's function to `=> boolean`; prettier evidently has one.
* **hono's published anchor (259) is stale by +2.** Both my base and my fix
  measure 261/324, so the apparent "+2" is main-side drift since the anchor was
  taken, NOT this change. This is exactly the failure mode that comparing to a
  published anchor instead of your own base produces, and it is why the base
  column exists.

The remaining ten suites were not re-run on the base side: the shared box was
saturated (load up to 223) and each base run costs the same as a fix run. Their
fix-side numbers equal main's own published anchors exactly, so there is no
delta to attribute for any of them — but that is an anchor comparison, not a
base run I executed, and I am naming it as such rather than reporting 17/17
both ways.

Standalone lane is untouched by construction: `scalarBridgePlan` returns `null`
for `ctx.standalone || ctx.wasi` two lines in, above every arm.
`tests/issue-5150-es2015-buffers.test.ts` measures the same 8 failed / 10
passed (identical failure set) on the parent and with this change —
pre-existing on `upstream/main`.

PR #5869, opened 2026-09-12
fix(#6414): type an async-frame spill field from the same predicate the resume local uses

ttraenkler:
Fixes [#6414](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6414-hono-cloudflare-pages-async-resume-struct-set).

`hono/dist/adapter/cloudflare-pages/index.js` compiled successfully and then
produced a binary the engine rejected:

```
Compiling function #318:"__async_resume_fanon_467" failed:
  struct.set[1] expected type i32, found local.get of type externref
```

## Mechanism

The shape is hono's `handleMiddleware`: `let response = void 0`, assigned from an
`await`, then read after a **later** `await` — so the binding is spilled into the
async frame.

Two places typed that one slot, and they disagreed:

| | types it from | answer |
|---|---|---|
| the frame **field** (`resumeBindingValType`) | `checker.getTypeAtLocation(response)` → `resolveWasmType` | **i32** |
| the resume **local** (var-decl path) | `varBindingNeedsExternrefForUndefined` (#2806) | **externref** |

A `void 0` initializer pins the declared TS type to pure `undefined` — unlike a
bare `undefined` initializer, which TS gives evolving-`any` — and
`resolveWasmType(undefined)` is a numeric slot. `stack-balance.ts` patched the
*restore* side, but `storeSpills` wrote `local.get <externref>` straight into the
i32 field with no coercion, and the module failed validation.

`resumeBindingValType` now resolves the binding's declaration and consults the
same #2806 predicate before the checker query, so both halves agree by
construction. All three spill-layout builders, both spill-safety gates and the
delivery-local typing already route through this one function, so no other edit
was needed.

Deliberately **not** touched: `storeSpills`/`restoreSpills` (a coercion there
would change every existing frame) and `resolveWasmType`'s `undefined` → i32 arm
(#1112's delete / optional-property f64-sentinel machinery depends on it — which
is exactly why #2806's predicate is void-**expression**-only).

## Correction to the plan

The implementation plan expected the standalone lane to already validate. It does
not: `target: "wasi"` and `target: "standalone"` emitted the same invalid
`struct.set` on the parent (`__async_resume_fhandle`, @+97099 / @+133139) and both
validate with the fix. The native drive layer shares the layout, so the
one-function fix covers all three lanes; the regression test asserts gc and
standalone.

Not folded with
[#6412](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6412-hono-jwt-async-resume-extern-convert-any):
that is an `extern.convert_any` on an already-external value in
`importPublicKey` — a different instruction and a different defect. hono's
`dist/middleware/jwk|jwt/index.js` rows stay in `KNOWN_INVALID_MODULES`.

## Evidence

**Reduction** (`.tmp/6414/probe.mjs`, 6 cases at one HEAD):

| case | shape | parent | fix |
|---|---|---|---|
| e | `void 0` + try/catch + later await (hono) | INVALID | valid |
| g | `void 0`, no try, two awaits | INVALID | valid |
| j | same in a plain named `async function` | INVALID | valid |
| f | control — `let response;` | valid | valid |
| h | control — `= undefined` | valid | valid |
| i | control — `= null` | valid | valid |

**The real module.** `dogfood-surface-probe.mjs --package hono --modules
dist/adapter/cloudflare-pages/index.js` goes `"verdict":"invalid"` →
`"verdict":"valid"` (251,181 bytes).

**Regression test** `tests/issue-6414-void-init-async-resume-spill.test.ts`
(untyped two-file `.js` project via `compileProject`): 3 of 4 tests fail on the
parent with the exact `struct.set[1]` message, all 4 pass with the fix. The 4th
is the `let response;` anti-vacuity control — identical body, green both ways.
One test instantiates and asserts the awaited `42` survives the second await, so
the field is not merely well-typed.

**Gates.** loc (+14, granted in the issue frontmatter), func, coercion-sites,
oracle-ratchet (checker usage +0), dead-exports, host-import-policy,
dogfood-validation (6/6 packages, 20/20 subpath modules), ts7 typecheck — all
green. `check:compiler-boundaries` exits 1 identically on the parent
(pre-existing). `tests/equivalence/delete-sentinel.test.ts` and
`tests/async-census.test.ts` fail identically on the parent (pre-existing).

**Dogfood A/B** over all 17 upstream suites, base vs fix at one HEAD: see the
comment below.


ttraenkler:
## Dogfood A/B — all 17 upstream suites, base vs fix at one HEAD

Base = `upstream/main`'s `src/codegen/async-frame.ts` (byte-identical to
`e8a778638f`), fix = this PR's, swapped in place by file copy so nothing else in
the tree moves. Suites run one at a time.

| package | base | fix | Δ |
|---|---|---|---|
| webpack | 16/16 | 16/16 | = |
| three | 17/18 | 17/18 | = |
| clsx | 32/32 | 32/32 | = |
| cookie | 63740/63740 | 63740/63740 | = |
| lodash | 59/62 | 59/62 | = |
| redux | 67/82 | 67/82 | = |
| axios | 208/231 | 208/231 | = |
| stylelint | 108/108 | 108/108 | = |
| tailwindcss | 13/13 | 13/13 | = |
| jsdom | 6/6 | 6/6 | = |
| styled-components | 9/9 | 9/9 | = |
| uuid | 75/75 | 75/75 | = |
| marked | 16/30 | 16/30 | = |
| moment | 10/10 | 10/10 | = |
| prettier | 107/151 | 107/151 | = |
| jest | 335/356 | 335/356 | = |
| hono | 261/324 | 261/324 | = |

**No movers in either direction.** Every headline carried an `admitted` line and
`exit=0`. This is the expected shape: the affected module
(`dist/adapter/cloudflare-pages/index.js`) is not exercised by hono's upstream
suite, so the gain here is validation-only.

Two notes on the numbers themselves:

- **hono reads 261/324, not the 259/324 anchor** — but it reads 261 on the
  **base** lane too, so that is main-side drift since the anchor was taken, not
  this change.
- **axios's first base run was killed (exit 137, no headline)** under contention
  from a concurrent suite; re-run alone it gives 208/231, matching the fix lane
  and the anchor. The 137 run is not counted.

PR #5880, opened 2026-09-13
docs(#6437): file the async-census module-init cycle found while validating #6414

ttraenkler:
Files [#6437](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6437-async-census-collection-kind-module-init-cycle) — a pre-existing module-init cycle found in passing while validating [#6414](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6414-hono-cloudflare-pages-async-resume-struct-set).

Docs only: one new issue file, no code.

`tests/async-census.test.ts` does not fail an assertion — it fails to **collect**,
so all of its tests are silently lost:

```
TypeError: Cannot read properties of undefined (reading 'MAP')
 ❯ src/codegen/collections-brand.ts:100:24   Map: COLLECTION_KIND.MAP,
 ❯ src/codegen/expressions/calls.ts:36:1
```

Same family as
[#3324](https://js2wasm.loopdive.com/dashboard/issue.html?slug=3324-boolToStringEmitter-module-init-cycle)
(`Cannot access 'boolToStringEmitter' before initialization`), which was fixed for
a different entry point. The trigger is importing a codegen module directly
instead of `src/index.js` first — several sibling tests carry that workaround as
a hand-written comment, which is the actual defect: the next test that forgets it
hits the same wall.

Reproduced on unmodified `upstream/main` at `e8a778638f` (2026-09-12) and again at
`7adc0a6e89` (2026-09-13), with and without an unrelated working-tree edit.

This file was written on #6414's branch but the merge queue landed that PR at an
earlier sha, so it is carried here rather than stranding the reserved id.

### DIFF
--- a/plan/issues/6437-async-census-collection-kind-module-init-cycle.md
+++ b/plan/issues/6437-async-census-collection-kind-module-init-cycle.md
@@ -1,0 +1,66 @@
+---
+id: 6437
+title: "tests/async-census.test.ts fails at collect time with a module-init cycle: `Cannot read properties of undefined (reading 'MAP')` in collections-brand.ts"
+status: ready
+sprint: current
+created: 2026-09-13
+updated: 2026-09-13
+priority: low
+horizon: s
+feasibility: medium
+task_type: bug
+area: codegen
+goal: correctness
+related: [3324]
+origin: "found in passing while validating #6414 (2026-09-12) — reproduces on unmodified upstream/main, unrelated to that PR"
+---
+
+## Problem
+
+`tests/async-census.test.ts` does not fail an assertion — it fails to **collect**,
+so all of its tests are lost:
+
+```
+TypeError: Cannot read properties of undefined (reading 'MAP')
+ ❯ src/codegen/collections-brand.ts:100:24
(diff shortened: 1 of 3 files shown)


ttraenkler:
Update while this PR was in flight: **#6437 is already fixed on `main`.**

[#6419](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6419-three-closure-area-tests-red-on-main)
(`53ba0b7b46`, PR #5876) extracted `COLLECTION_KIND` into
`src/codegen/collection-kind.ts`, which cuts the cycle edge this issue describes.
Re-measured: `Test Files 1 failed / Tests no tests` at `e8a778638f` and
`7adc0a6e89`, `Test Files 1 passed / Tests 13 passed` at `f1462c4d15`.

The file is kept (now `status: done` with a Resolution) rather than dropped, so the
reserved id is not a hole in the sequence and the failure mode stays searchable.
One acceptance criterion is genuinely unmet and noted as such: nobody wrote the
note naming which import edge is load-bearing.

PR #5759, opened 2026-09-08
refactor(ir): extract native delay and combinator runtime bodies

ttraenkler:
## Summary

Stacked on #5758. Move native Promise-delay and combinator instruction bodies into two canonical WasmGC runtime owners, retaining allocator/cache/registration ownership in compatibility adapters. Both new owners are mandatory clean-boundary roots.

Preserve the original 13 donor receipts and add semantic-field coverage plus a donor-derived ledger of 20 combinator and four delay declarations. High-effort independent review cleared production and the revised preservation evidence.

## Validation

- 159/159 focused ownership and preservation tests.
- 77/77 unchanged historical regression tests across seven suites.
- 121/121 boundary tests; inventory has no errors.
- Paired candidate and historical compiler execution: 11 artifacts, 19 executions per root, complete bytes/WAT/resources/results equal; both children exited 0.
- Canonical TypeScript 7 and scoped Prettier checks pass. Normal commit and push hooks run without bypasses.

## Scope and handoff

This checkpoint establishes runtime-body ownership and preservation, not full prepared native physical acceptance. Complete native resource materialization, full async-family execution through preparation, IR-only cutover, strict closure, and direct-codegen retirement remain open. Four unknown dynamic-import sites remain.

Implementation and validation handoff is recorded in `plan/issues/3518-ir-only-default-and-direct-frontend-retirement.md`, including parallel recursive-type emitter progress. Hold retained; no auto-merge requested. Related to #3518, not a closure of it.

PR #5883, opened 2026-09-13
fix(promise): preserve observable intrinsic combinator protocol

ttraenkler:
## Description

- add the bounded direct-VEC observable `Promise.all` / `Promise.race` pipeline: one `Get(Promise, "resolve")`, per-element `Call`, and captured `then` `Invoke`
- retain only direct, unshadowed module-scope intrinsic `Promise.resolve` writes so the original module-shaped row reaches that pipeline
- preserve #5759's receipt-owned legacy `promise-combinators.ts` adapter byte-for-byte by routing through a dedicated module
- document the remaining boundaries: mutable vector length, generic iterator closing, and custom constructors are follow-up work

## Validation

- `pnpm run typecheck` — passed (23.0s)
- focused #5197 observable protocol suite — 13/13 passed (39.94s)
- isolated standalone Test262 positive control and unchanged `built-ins/Promise/all/invoke-resolve.js` — 2/2 `ROW pass`
- mandatory merge hook — changed-root 12 files passed; LOC/function/oracle gates passed
- normal pre-push hook — typecheck, lint, format, oracle/coercion ratchets, numeric-local parity 18/18, and issue integrity passed

## CLA


ttraenkler:
Independent review confirms a semantic landing blocker beyond the missing inventory record. On preserved clean df94fdac (relevant observable/admission/retention/test files verified byte-identical to published647343cc), a captured Promise.resolve callback shrinking [1,2,3] to length1 is called3 times instead of Node1; pushing4 calls3 instead of4, for both all and race. Two unchanged-length controls match; four mutation cases diverge. Follow-up12rows:4 controls match,8 diverge, including initial holes passed as NaN and stale values after shrink/regrow. All compiled successfully, zero host imports, no guard filtering. This is not a full published-head suite claim. Exact sources/results preserved in local review logs. High reviewed a bounded observable-only dynamic-state/live-vector-read repair; implementation is isolated, original literal/static/legacy donor code and fixtures remain preserved. Keep semantic HOLD; an inventory-only repair is insufficient.

PR #5885, opened 2026-09-13
fix(standalone): a table-free `Intl.DateTimeFormat` for UTC and `Etc/GMT±N` in the Temporal provider shim — `unknown time zone UTC` 69 → 0, linked lane 85 → 122/360 (#5383 S9 / #6442)

ttraenkler:
Land order: after PR #5875 (S8) — this branch is stacked on it.

Slice S9 of [#5383](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5383-standalone-temporal-provider), filed as [#6442](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6442-standalone-intl-datetimeformat-fixed-offset).

## What this is

After S8 the linked standalone Temporal lane scores 85/360 and the largest
single bucket is **`RangeError: unknown time zone UTC` — 62 of the 116
`Temporal/ZonedDateTime/prototype/**` failures**. This slice removes it by
giving the provider a real, host-free `Intl.DateTimeFormat` for the time zones
that need **no CLDR/tzdata tables**: `UTC`, its aliases, and the fixed-offset
`Etc/GMT±N` family. Everything else — every tzdata zone, every other locale or
option shape, `formatRange*` — still refuses with the same catchable
`RangeError` as before, so [#5355](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5355-intl-datetimeformat-shell)'s
bound ("no ICU in pure Wasm") is narrowed, not broken.

## Two census findings that changed the route

Both are in the new module's header, because a reader without them rewrites
this in the wrong place:

1. **It cannot be a `src/codegen/` arm.** The polyfill never sees the compiler's
   `Intl`. The S2c shim declares a module-scoped `const Intl = {…}` in the
   provider's own compilation unit, so all 24 `Intl.` reads in the bundle
   resolve to that **lexical** binding — an arm in `new-intl-host-bridge.ts`
   would be shadowed and could not move one row. Standalone *user* code is a
   separate, still-open gap ([#5206](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5206-intl-global-missing):
   the `Intl` global is absent entirely) and is untouched here.
2. **The load-bearing surface is `format()`, not `formatToParts()`.** The offset
   path is `Fn → lr → ur → br → ht(tz).format(epochMs)`, and `br` parses the
   result with `split(/[^\w]+/)` into exactly seven word tokens — month, day,
   year, era, hour, minute, second — inverting the era itself (`o = 1 - o`).
   The contract is the en-US string `M/D/Y AD, HH:MM:SS`.

## Correctness evidence

Every accepted spelling, canonical id, offset sign and format string was
measured against the real ICU implementation, not derived from the spec — the
three traps being that `Etc/GMT+1` is UTC **minus** one hour (POSIX), that
`gmt+1` without the `Etc/` prefix is rejected while `gmt+0` is accepted, and
that a leading zero (`Etc/GMT+01`) is rejected.

**Differential oracle** (`.tmp/s9/oracle.mts` — the generated source evaluated
in Node, no compiler in the loop, against the runner's own ICU): 44 zone
spellings × 4,022 epoch values = **176,968 pairs, 0 mismatches**, including both
±8.64e15 extremes, the out-of-range values either side, `NaN`, the BC boundary,
and 4,000 deterministic pseudo-random epochs. `resolvedOptions()` and
`formatToParts(0)` are byte-identical to ICU's for both admitted option shapes.

The civil date comes from Hinnant's `civil_from_days` rather than `Date`, with
the two divisions that straddle zero done by exact remainder: at 8.64e15 the
rounding error of `ms / 86400000` exceeds the distance to the next integer, so
`Math.floor` of it is not reliable, while `%` on exactly-representable doubles
is exact.

## Measurements

Three families, 120 rows each, `--target standalone`, provider linked
(`cacheHit=true`, `js2wasm:npm:@js-temporal/polyfill:dc43b7a43e0bd370`,
3,312,078 B, **0 `__temporal_*` leaks**), run sequentially:

| family | S8 pass | **S9 pass** | fail | CE | pass→fail |
| `Temporal/PlainDate/**` | 51 | **55** | 64 | 1 | **0** |
| `Temporal/Duration/**` | 32 | **35** | 82 | 3 | **0** |
| `Temporal/ZonedDateTime/prototype/**` | 2 | **32** | 86 | 2 | **0** |
| **total** | **85** | **122** | 232 | 6 | **0** |

37 rows flipped `fail→pass`, 30 of them in ZonedDateTime/prototype.
`RangeError: unknown time zone UTC` is **69 → 0**. No pass→fail and no
pass→compile_error row; all six S9 `compile_error` rows carry
`compilation timeout` in their detail, as do all eight on the S8 side.

**Reduction** (linked, host-free standalone; base by file-copy revert; verdict
`1` = the exact value Node's polyfill answers, `0` = threw "unknown time zone"):
`ZonedDateTime(0n,"UTC")`'s `.offset` / `.hour` / `.timeZoneId` and
`Temporal.Now.timeZoneId()` all go to `1`; the two `from(…)` rows stop being
"unknown time zone" and join the **pre-existing** `from()` bucket, which the
`PlainDate` control row — no time zone anywhere — shows answering the same on
**both** trees.

**Must-not-move samples** (base by file-copy revert on the same post-merge
tree): `intl402/DateTimeFormat/**` 2 pass / 115 fail / 3 CE on both sides;
`built-ins/Date/prototype/**` 120 pass on both — **0 flips** in either.

**Byte A/B**: all 22 (module, target) artifacts in the fixed corpus identical on
`gc` and `standalone`. The `gc` Temporal provider key is unchanged
(`372a41be…`); standalone (`53f91978…` → `ee23a4e9…`) and wasi
(`4ff8fb73…` → `359c6e28…`) re-key by design, which is what makes serving a
stale standalone artifact impossible.

**Gates**: typecheck · loc-budget · func-budget (both also with
`LOC_GATE_BASE=origin/main`) · coercion-sites · speculative-rollback-sites ·
oracle-ratchet · dead-exports · compiler-boundaries · host-import-policy ·
biome lint · update-issues --check · equivalence-gate (22 failing / 1720
passing, no new regressions) · the new #6442 suite (7/7, five of which fail on
the base tree) · #5383 (81 pass / 3 todo) · #5355 · #5404 · #6432.

Full tables, flip lists and the post-change bucket census are in the
`## Resolution` section of the issue file and in the `### S9 findings` section
of #5383.

## One deliberate contract change

`#5383 S2c`'s assertion "a method on the refusal class throws the same way" is
updated: `DateTimeFormat` is no longer a pure refusal, so a call on the bare
**prototype** is now a brand-check **TypeError** (what the spec says) rather
than the refusal-era `RangeError`. The property that assertion exists for is
unchanged and still asserted — touching `ai.prototype` can never trap and the
throw is catchable — and the `"formatToParts" in ai.prototype` probe the
polyfill actually performs is asserted in the new suite.

Re-verified by the coordinator on the pushed head (f1678ac461, S8 c577aecfbe merged in): typecheck, compiler-boundaries inventory, speculative-rollback gate, #6442 + #5383 suites (88 passed, 3 todo).

---


ttraenkler:
Park diagnosis (run 34736979338): the verdict never ran. All 102 shards — both `js-host` and `standalone` — exited ~15 s after `vitest` started, each with an **empty** results JSONL (the uploaded shard artifact is a 0-byte file), i.e. no test body executed before `Test262 shard did not reach afterAll` fired. The setup steps in the same job (checkout, bundles, `temporal-provider` and `quickjs` artifacts, `--require-cache` verification) all passed.

Reproduction attempt on this branch merged with current `main`: the exact chunk (`TEST262_CHUNK_INDEX=4 TEST262_CHUNK_TOTAL=50`, standalone, quickjs, prewarmed Temporal cache) collects 970 tests and is still running normally past 150 s; a Boolean-only slice completes to `afterAll` on both targets. This PR's diff touches only `src/`, `tests/`, and `scripts/compiler-boundaries.json`; nothing in the runner. Treating this as the "setup/infra step rather than a verdict step" case named in the park notice, and releasing the hold once. A second park will be treated as real.

---

ttraenkler:
Root cause found; the second park is real but not this PR's code. `main` (and this branch, inherited from the same commit) carries `test262` as a symlink instead of the submodule gitlink, so on a runner the corpus path is dangling and every shard dies at collection — #5889 and #5890 parked identically. Fix is #5892 (restores the `160000` gitlink). Keeping `hold` here until #5892 lands, then merging `main` in and releasing once. My earlier "infra flake" read was wrong about the cause, right about the verdict never running.

---

ttraenkler:
#5892 merged; `main` merged into this branch (head 787e2ddc6f), `git ls-tree HEAD test262` is now `160000 commit b363f29d3c…`. Releasing the hold once on that basis.

---

PR #5889, opened 2026-09-13
fix(#6421): a spread into a static builtin expands every element

ttraenkler:
Closes [#6421](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6421-spread-into-static-builtin-drops-arguments).

`String.fromCharCode(...bytes)` answered a single `"\0"`. hono's
`btoa(String.fromCharCode(...new Uint8Array(signature)))` therefore signed
every cookie as `AA==` — base64 of one zero byte — instead of the real digest.

## Mechanism

A call's argument list is a list of AST NODES, but its value list is a runtime
one: a spread contributes its element count, not one slot. Three lowerings
still assumed one slot per node — the shape #5361 and #6411 already removed
elsewhere.

1. **`String.fromCharCode` / `String.fromCodePoint`** (new
   `src/codegen/from-char-code-spread.ts`). The family built one string PART
   per argument node and folded the parts at compile time, so a spread compiled
   its SOURCE array as one code unit: `NaN` → ToUint16 → `"\0"`.
   `fromCodePoint` failed harder — `NaN` is not an integral code point, so the
   #2601 guard threw `RangeError: Invalid code point NaN`. A spread-containing
   list now goes through the shared `buildSpreadArgList` and folds at RUNTIME
   into an accumulator local seeded with `""`. The per-code-unit tail (§7.1.8
   ToUint16 in the f64 domain, the range guard, the 1-char-string helper) moved
   into a shared `emitCodeUnitPart` that both lanes call, so the loop and the
   accumulator cannot drift. A list with no spread keeps the old loop.

2. **`Array.of`** (new `src/codegen/array-of-spread.ts`). The host path is now
   `tryEmitSpreadHostArgs` — the exact #6411 edit — with the unrolled loop kept
   as the no-spread branch. The standalone path's `noJsHost && !hasSpreadArg`
   gate is gone: a runtime-length list is sized from the builder's count and
   filled at a running write index. Worth knowing: that gate was worse than the
   issue predicted — the spread fell onto the host path and imported
   `__js_array_new` / `__array_of`, which the native-first adapter refuses, so
   a standalone module containing `Array.of(...xs)` would not instantiate at
   all.

3. **`emitStores` handed out SHARED instruction objects.** This is why the
   first two fixes were still wrong in standalone, and it is a latent hazard
   for every existing caller of the builder. `SpreadArgSink.pre`/`post` were
   spliced into the body once per slot, so the same `Instr` objects sat at
   several points of one array; the late-import shifter (`shiftFuncIndices`)
   dedups by ARRAY identity and never by instruction identity, so it added the
   shift delta to a repeated `call` index once per occurrence. On
   `String.fromCharCode(48, ...[65,66,67])` in standalone — four static value
   slots — the accumulating `__str_concat` was shifted four times and the call
   answered `"0"`. `emitStores` now clones the sink per use.

## Measured

| form | gc parent | gc fix |
| --- | --- | --- |
| `fromCharCode(...[65,66,67])` | `"\0"` | `"ABC"` |
| `fromCharCode(48, ...[65,66,67])` | `"0\0"` | `"0ABC"` |
| `...new Uint8Array([72,73,74])` | `"\0"` | `"HIJ"` |
| host `ArrayBuffer` handed in | `"\0"` | `"PQR"` |
| `...s.split(",").map(Number)` | `"\0"` | `"ABC"` |
| `fromCodePoint(...[0x1F600, 65])` | threw `RangeError` | 3 code units |
| `Array.of(...[1,2,3])` | length 1, joins `"NaN"` | length 3, `1\|2\|3` |
| hono shape, 4 zero bytes | `"AA=="` | `"AAAAAA=="` |

Standalone: every row above also correct; the module now instantiates.
Controls correct on both sides: `fromCharCode(65,66,67)`, `.apply`,
`Math.max(...xs)`, `Array.of(1,2,3)`, `fromCodePoint(0x1F600,65)`.

`tests/issue-6421-static-builtin-spread-args.test.ts` — **20 failed / 8 passed
of 28 on the parent, 28 passed with the fix**; the 8 `control…` rows pass on
both sides. Across the 31 spread / charCode / `Array.of` test files: base 47
failed / 238 passed → 27 failed / 258 passed, with the failing SET unchanged
apart from this file's own 20 rows.

## Dogfood A/B

16 of 17 suites flat at one HEAD (base `699df289e1`); one mover.

| suite | base | fix |
| --- | --- | --- |
| hono | 261/324 | **262/324** |

webpack 16/16 · three 17/18 · clsx 32/32 · cookie 63740/63740 · lodash 59/62 ·
redux 67/82 · axios 208/231 · stylelint 108/108 · tailwindcss 13/13 · jsdom 6/6
· styled-components 9/9 · uuid 75/75 · marked 16/30 · moment 10/10 · prettier
108/151 · jest 335/356 — all unchanged.

hono `src/utils/cookie.test.ts` 24/35 → 25/35, one row flipped and none
regressed: **"Should serialize signed cookie with all options"**, which on base
emitted `great_cookie=banana.AA%3D%3D` — the literal symptom this issue is
named for.

**The issue predicted 24/35 → 33/35. It did not happen, and its attribution of
the other eight rows was wrong.** Signing is now provably correct — "Should
serialize a signed cookie" emits the real 44-character digest
`macha.diubJPY8O7hI1pLa42QSfkPiyDWQ0I4DnlACH%2FN2HaA%3D` and fails only on the
spurious `Max-Age=0` that #6423 owns. The eight "Should parse signed cookies…"
rows still answer `false` while being handed a correct signature, so
verification has its own defect rather than being downstream of this one. Filed
as [#6449](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6449-hono-signed-cookie-verify-returns-false).
33/35 is still reachable — from three issues, not one.

## Follow-ups filed

- [#6443](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6443-spread-into-user-rest-function-illegal-cast)
  — a spread into a USER rest function traps with `illegal cast` in an untyped
  `.js` project, on the parent AND with this fix, for every source shape. #6421
  listed that form as already correct; it is not, so it could not serve as the
  anti-vacuity control it was meant to be.
- [#6444](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6444-static-builtin-spread-audit-remaining-arms)
  — sweep the remaining static-builtin arms that unroll `expr.arguments`
  (`Object.assign`, `Math.hypot`, …). This is the fourth time the same idiom
  has been found from a downstream symptom rather than from a sweep.
- [#6449](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6449-hono-signed-cookie-verify-returns-false)
  — hono `getSignedCookie` answers `false` for a signature that is now
  correct; the eight remaining `cookie.test.ts` parse rows.

## Budget allowances

`loc-budget-allow` / `func-budget-allow` for
`src/codegen/expressions/call-builtin-static.ts` (+31 lines, one dispatcher),
granted in #6421's frontmatter with a dated rationale. The runtime-length vec
build itself went into a new module rather than into the god-file, and
`calls.ts` shrank by ~60 lines in the same change-set.


ttraenkler:
## Park cleared — main-side infrastructure outage, not a regression in this PR

Read the cited run before touching the label, per protocol. Diagnosis:

**Cited run:** https://github.com/loopdive/js2/actions/runs/34739544544 (`merge_group`, base `dcdd1efa`, 2026-09-13 05:08Z) — **103 of 115 jobs failed**: every one of the 52 js-host shards, every one of the 50 standalone shards, plus `merge shard reports`.

**Gate lines** (shard 1/52, job 103676937051 — identical in every failing shard):

```
blob report written to /home/runner/work/js2/js2/vitest-blob/blob-js-host-mg1.json
##[error]Test262 shard did not reach afterAll; refusing partial JSONL evidence
##[error]Process completed with exit code 2.
```

The vitest invocation ran for **14 seconds** and emitted no test output at all — it collected zero tests. No `GATE FAIL`, no `=== Net`, no `pass → fail` bucket, no trap categories, no `runtimeTsLines` ceiling: **there is no conformance verdict anywhere in this run**. The failing step is the downstream evidence guard, not a verdict step (see the auto-park footer's own caveat, #3597).

### Root cause (already fixed on main)

`7871adcc79` — *"restore the `test262` submodule gitlink that a lane grounding commit replaced with a symlink"*. Since `dfecafa7e9` (landed on main inside #5875) the tree carried `test262` as a **symlink to `/home/user/js2/test262`**, a path that exists only on one dev box. On a runner `actions/checkout` registers no such submodule (the checkout log shows only `test262-fyi/data` and `vendor/Porffor`), `TEST262_ROOT` resolves to a dangling link, and `findTestFiles` throws `ENOENT` at collection — hence a 14-second silent no-op and the `afterAll` guard firing. Gitlink restored at `b363f29d3c`, mode `120000 → 160000`.

### Evidence it is not this PR

| PR | run | window | failed jobs |
| --- | --- | --- | --- |
| **#5889** | 34739544544 | 05:08Z | **103** / 115 |
| #5885 (unrelated) | 34739853934 | 05:15Z | 103 / 115 |
| #5890 (unrelated) | 34740128635 | 05:22Z | 103 / 115 |
| #5885 — *same PR, after the fix* | 34744425905 | 07:07Z | **1** / 115 |
| #5890 — *same PR, after the fix* | 34745598631 | 07:35Z | **1** / 115 |

Two unrelated PRs produced the byte-identical signature inside the same window, and both went back to a full green shard matrix once `7871adcc79` was on main — with no change to the class of content under test. #5885 and #5890 have since merged.

Note that #5891's `merge_group` passing at 05:45Z is **not** counterevidence of the outage: it ran only 14 jobs because it touches no test262-relevant paths, so its shard matrix was skipped entirely. That trap is written up as Finding 2 of [#6454](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6454-merge-queue-test262-shards-die-before-afterall).

### Provenance of this park

`hold` applied `2026-09-13T05:12:21Z` by `github-actions[bot]` — squarely inside the 03:28Z–05:55Z outage window. [#6454](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6454-merge-queue-test262-shards-die-before-afterall) lists this PR by number as one of three evidence-free parks stranded by it, cites **this exact run and job** as its reference failure, and makes "#5885 and #5889 are re-admitted" acceptance criterion 3. #5885 and #5890 are already cleared and merged; this is the last one.

**Action:** removing the `hold`. `mergeStateStatus` is `CLEAN`, so `auto-enqueue.yml` owns it from here — not enqueueing by hand.

One thing the next reader should know: the queue currently has a *separate*, unrelated red gate — recent `merge_group` runs fail `merge shard reports` with `STANDALONE host-free pass floor breached: 35577 < high-water 35742 − 50 = 35692`. That is main-side and affects every PR in the queue, not this one, but this PR will likely meet it on its next pass.

ttraenkler:
Closing in favour of https://github.com/loopdive/js2/pull/5957, which republishes this work onto a branch in the main repository (the fork head is not reachable from the session that finished it) with `origin/main` merged in.

One substantive change on the way across: **the `String.fromCharCode` / `fromCodePoint` half of this PR is superseded.** Main landed #6430 independently while this sat open, implementing the same expansion fix, so the new PR drops this branch's twin module and restores main's inline per-code-unit tail. `src/codegen/expressions/calls.ts` there now diffs zero bytes against main.

The two parts main does **not** cover are kept, each verified against current main rather than assumed:

- `Array.of(...xs)` on the standalone target falls to a host fallback importing `__js_array_new` / `__js_array_push` / `__array_of`, which cannot bind there — a zero-host-imports violation, not just a wrong length.
- The shared-instruction bug in `spread-arg-list.ts`: on pristine main, `String.fromCharCode(48, ...[65,66,67])` still answers `"0"` instead of `"0ABC"`, and swapping only that file in fixes it. #6430 fixes the expansion, not the index-shift corruption after it.

All 28 pins in `tests/issue-6421-static-builtin-spread-args.test.ts` pass unchanged over there — no row was edited or weakened. This thread keeps the review history.

---

PR #5890, opened 2026-09-13
fix(#6426): an object-rest source keeps its values across a linked package

ttraenkler:
Closes [#6426](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6426-object-assign-option-bag-across-linked-package).

## Mechanism

Four host helpers in `src/runtime.ts` read a WasmGC struct in two steps: field
**names** from `_getStructFieldNames`, then each **value** from
`exports["__sget_<key>"]`. Since
[#5225](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5225-consumer-minted-struct-opaque-in-linked-provider)
the name step routes through `_decoderExportsFor`, so it answers with the struct
**owner's** decoder — but the value step still used the **reader's** own exports.

A `__sget_<key>` getter is module-global by field *name* and does **not** trap on
a foreign struct type: it `ref.test`-misses and returns that getter's default. So
when a consumer mints an options bag and a separately-linked provider does

```js
const { strict, ...rest } = options;
Object.assign(this, rest);
```

every key survives and every value is lost — `null` for references, `0` for
numbers. `typeof null === "object"` is the whole reason the issue's copied-*function*
row looked like a different defect from its copied-*object* row; both are the same
lost value.

The fix is `_decoderExportsFor` on the value read too, at the four sites with that
names-then-values shape: `__extern_rest_object`, `__object_values`,
`__object_entries`, and the tuple arm of `__extern_slice`. `__object_keys` is
names-only and was already correct; no write path moved.

**No codegen change.** `src/codegen/statements/destructuring.ts` already routes
rest patterns to the externref path correctly, and the standalone twin
(`__extern_rest_object` in `src/codegen/destructuring-params.ts`) never runs
`runtime.ts` — so the standalone floor/net is untouched.

**4 changed lines, zero net lines.** `src/runtime.ts` sits exactly at the #4401
ceiling (19735), so the write-up lives here and in the test header rather than
inline, and the sites carry the repo's existing `// (#5225/#6426)` marker. No
budget allowance needed.

## Probe, base → fix

`.tmp/6426-repro.mts`, `linkPlan.mode === "separate"`. 6 of 9 rows wrong on base,
0 after; the 3 controls never moved.

| row | base | fix | node |
| --- | --- | --- | --- |
| `restObj` — object through `...rest` | `"null"` | `"R"` | `"R"` |
| `restFn` — function through `...rest` | `"object"` | `"function"` | `"function"` |
| `restNum` — number through `...rest` | `"0"` | `"7"` | `"7"` |
| `restKeys` — `Object.keys(rest)` + `rest.router` | `router\|no` | `router\|has` | `router\|has` |
| `Object.values` of a consumer struct, in the provider | `"undefined,undefined"` | `"number,string"` | `"number,string"` |
| `Object.entries` of the same | `"a:undefined,b:undefined"` | `"a:number,b:string"` | `"a:number,b:string"` |
| control — direct `Object.assign(this, options)` | `"R"` | `"R"` | `"R"` |
| control — `{ ...options }` spread source | `"R"` | `"R"` | `"R"` |
| control — `this.router = options.router` | `"R"` | `"R"` | `"R"` |

The number row coming back `0` rather than `undefined` is the sharpest
confirmation of the mechanism: that is a `ref.test`-miss default, not a missing key.

## Regression test

`tests/issue-6426-linked-object-rest-source.test.ts` — 12 rows × 3 lanes (single
`compile`, `compileMulti` one unit, `compileProject` separate), one shared
`EXPECTED` table.

| lane | parent wrong | fix wrong |
| --- | --- | --- |
| single module | 0 | 0 |
| two modules, one unit | 0 | 0 |
| separately-linked package | **6** | **0** |

The 6 are `restObj`, `restFn`, `restNum`, `restKeys`, `restValues`, `restEntries`.
The other 6 rows are anti-vacuity controls that already passed on the parent
(direct assign, spread source, plain field store, `Object.keys`, rest with an
excluded key, empty bag), and the linked lane additionally asserts that all 6
regressed rows are actually present in that module's exports. The fixture's
provider exports **classes only** on purpose: a provider exporting untyped
*functions* falls back to a `bundled` plan, which would silently run the
single-unit lane a third time — hence the `linkPlan.mode === "separate"` assertion.

## Acceptance 2

`LINKED_RESIDUALS` is deleted from `tests/issue-5366-nullish-join-carrier.test.ts`
and its separately-linked lane now asserts the shared `EXPECTED` table. All three
lanes of that file pass.

## Dogfood A/B — all 17 suites, flat

Base vs fix at one HEAD (`upstream/main` `3e92241ecc` ± the 4-line change), one suite at a time, per test file. **Every suite flat; zero per-file movers.**

| suite | base | fix | | suite | base | fix |
| webpack | 16/16 | 16/16 | | styled-components | 9/9 | 9/9 |
| three | 17/18 | 17/18 | | uuid | 75/75 | 75/75 |
| clsx | 32/32 | 32/32 | | marked | 16/30 | 16/30 |
| cookie | 63740/63740 | 63740/63740 | | moment | 10/10 | 10/10 |
| lodash | 59/62 | 59/62 | | prettier | 108/151 | 108/151 |
| redux | 67/82 | 67/82 | | jest | 335/356 | 335/356 |
| axios | 208/231 | 208/231 | | hono | 261/324 | 261/324 |
| stylelint | 108/108 | 108/108 | | | | |
| tailwindcss | 13/13 | 13/13 | | | | |
| jsdom | 6/6 | 6/6 | | | | |

All 34 runs exited 0. Beyond the headlines, **every per-file `N/M Wasm` fraction is identical between the two lanes across all 17 suites** — so this is flat at file granularity, not just in the totals.

That is the predicted result: these suites compile the package's **own** source as the root unit, so the consumer→provider rest seam this fixes never fires.

**Two suites read differently from the dispatch brief's anchors, and neither is this change:** prettier **108** vs the brief's 107, and hono **261** vs 259. Both read identically on *my own base lane*, so they are main-side drift between 2026-09-12 and today, measured rather than assumed.

**One measurement was discarded and re-run.** My first hono fix-lane run overlapped the lane swap — my waiter counted log files, which are created when a suite *starts*, so the base lane began while hono was still going and `src/runtime.ts` changed underneath it. That number was thrown away and hono re-run cleanly on the fix runtime; the 261/324 above is the clean run. The other 16 suites were unaffected (they had completed), and pass counts here are deterministic.

## Gates

`check-loc-budget` · `check-func-budget` · `check-coercion-sites` · `check:oracle-ratchet` · `check:dead-exports` · `check:dogfood-validation` · `check:host-import-policy` — all green, re-run after merging current `main`. ts7 `--noEmit` clean. Biome clean on the changed files.

`check:compiler-boundaries` reports `inventory-valid-architecture-incomplete` — **identically on `upstream/main` with this change reverted**, so it is pre-existing (`inventoryValid: true`, `errors: []`, and this PR adds no module).

## Residuals

- **`tests/issue-2131.test.ts` fails 6/7 on `upstream/main`**, identically with this change applied and reverted. Not caused or masked by it; filed as [#6451](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6451-jshost-struct-enumeration-returns-empty-in-2131-harness). The six JS-host rows return an *empty* enumeration rather than a mis-ordered one, and only the standalone row survives.
- **27 more host helpers share this exact defect shape** — bind the reader's exports, then read `__sget_<key>`. #5225, #5365 and now #6426 have each fixed a subset one bug report at a time. Filed as [#6452](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6452-mixed-decoder-audit-remaining-sget-read-sites) with the scan, a per-site classification requirement (deliberately not a blind sed), and a gate so there is no fourth round.
- **The merge-queue outage that parked this PR**: [#6454](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6454-merge-queue-test262-shards-die-before-afterall).

### DIFF
--- a/plan/issues/6426-object-assign-option-bag-across-linked-package.md
+++ b/plan/issues/6426-object-assign-option-bag-across-linked-package.md
@@ -1,10 +1,11 @@
 ---
 id: 6426
 title: "`Object.assign(this, options)` in a base class drops the value when the class lives in a separately-linked package"
-status: ready
+status: done
 sprint: current
 created: 2026-09-12
-updated: 2026-09-12
+updated: 2026-09-13
+completed: 2026-09-13
 priority: medium
 horizon: m
 feasibility: medium
@@ -97,3 +98,73 @@ Then delete `LINKED_RESIDUALS` in `tests/issue-5366-nullish-join-carrier.test.ts
 ## Dispatch
 
 **Model: opus.** The defect is located to four one-line sites in `src/runtime.ts` with a known template (#5225), but the work needs a correct three-lane linked-package fixture (the fallback-to-bundled trap above) and a per-file dogfood A/B read honestly — medium, not mechanical.
+
+## Resolution
+
+**Fixed 2026-09-13** in `src/runtime.ts` (host runtime only, no codegen change),
+PR on `issue-6426`. The plan's diagnosis was right and its four sites were the
+right four; the probe confirmed it before any edit and added one row it did not
+predict.
+
+**Mechanism.** Four host helpers read a WasmGC struct in two steps: field NAMES
+from `_getStructFieldNames`, then each VALUE from `exports["__sget_<key>"]`.
+Since #5225 the name step routes through `_decoderExportsFor`, so it answers
+with the struct OWNER's decoder — but the value step still used the READER's
+own exports. A `__sget_<key>` getter is module-global by field NAME and does not
+trap on a foreign struct type: it `ref.test`-misses and returns its miss default.
+So when a consumer mints an options bag and a linked provider does
+`const { strict, ...rest } = options; Object.assign(this, rest)`, every key
+survives and every value is lost — `null` for references, `0` for numbers.
+`typeof null === "object"` is the whole reason the issue's function row looked
+like a different defect from its object row; both are the same lost value.
+
+The fix is one line at each of the four sites — `_decoderExportsFor(obj,
+callbackState?.getExports())`, exactly the L12569 template — in
+`__extern_rest_object`, `__object_values`, `__object_entries` and the tuple arm
+of `__extern_slice`. `__object_keys` is names-only and was already correct;
+nothing on the write side moved. The diff is 4 changed lines and **zero net
+lines**: `src/runtime.ts` sits exactly at the #4401 ceiling (19735), so the
+mechanism is written up here and in the test header rather than inline, and the
+sites carry the repo's existing `// (#5225/#6426)` marker.
+
+**Probe (`.tmp/6426-repro.mts`, `linkPlan.mode === "separate"`), base → fix.**
+6 of 9 rows wrong on base, 0 after; the 3 controls never moved.
+
+| row | base | fix | node |
+| --- | --- | --- | --- |
+| `restObj` (object through `...rest`) | `"null"` | `"R"` | `"R"` |
+| `restFn` (function through `...rest`) | `"object"` | `"function"` | `"function"` |
+| `restNum` (number through `...rest`) | `"0"` | `"7"` | `"7"` |
+| `restKeys` (`Object.keys(rest)` + `rest.router`) | `"router\|no"` | `"router\|has"` | `"router\|has"` |
+| `Object.values` of a consumer struct in the provider | `"undefined,undefined"` | `"number,string"` | `"number,string"` |
+| `Object.entries` of same | `"a:undefined,b:undefined"` | `"a:number,b:string"` | `"a:number,b:string"` |
+| control: direct `Object.assign(this, options)` | `"R"` | `"R"` | `"R"` |
+| control: `{ ...options }` spread source | `"R"` | `"R"` | `"R"` |
+| control: `this.router = options.router` | `"R"` | `"R"` | `"R"` |
+
+The number row (`0`, not `undefined`) is the sharpest confirmation of the
+mechanism: that is a `ref.test`-miss default, not a missing key.
+
+**Regression test** `tests/issue-6426-linked-object-rest-source.test.ts`, 12 rows
+x 3 lanes, one shared `EXPECTED` table:
+
+- parent: single 0 wrong / one-unit multi 0 / separately-linked **6** = 6
+- fix: 0 / 0 / 0 = 0
+
+The 6 are `restObj`, `restFn`, `restNum`, `restKeys`, `restValues`,
+`restEntries`. The other 6 rows are anti-vacuity controls that already passed on
+the parent (direct assign, spread source, plain field store, `Object.keys`,
+rest-with-an-excluded-key, empty bag), and the linked lane additionally asserts
+that all 6 regressed rows are actually present in that module's exports.
+
+**Acceptance 2** — `LINKED_RESIDUALS` is deleted from
+`tests/issue-5366-nullish-join-carrier.test.ts` and its linked lane now asserts
+the shared `EXPECTED` table; all three lanes of that file pass.
+
+**Dogfood A/B** — base vs fix at one HEAD over all 17 upstream suites, results
+in the PR body. Flat, as the plan predicted: these suites compile the package's
+own source as the root unit, so the consumer→provider rest seam never fires.
+
+**Residual.** `tests/issue-2131.test.ts` fails 6/7 on `upstream/main`
+`3e92241ecc` and fails identically with this change — pre-existing, untouched,
+not this issue's.
--- a/plan/issues/6451-jshost-struct-enumeration-returns-empty-in-2131-harness.md
+++ b/plan/issues/6451-jshost-struct-enumeration-returns-empty-in-2131-harness.md
@@ -1,0 +1,80 @@
+---
+id: 6451
+title: "JS-host `Object.keys`/`values`/`entries`/`for-in` return nothing in the #2131 harness — 6 of 7 rows regressed on main"
+status: ready
+sprint: current
+created: 2026-09-13
+updated: 2026-09-13
+priority: high
+horizon: m
+feasibility: medium
+reasoning_effort: high
+task_type: bug
+area: compiler
+goal: correctness
+---
+
+## Problem
+
+`tests/issue-2131.test.ts` passes **1 of 7** on `upstream/main` `3e92241ecc`.
+The six JS-host rows do not merely mis-*order* keys — the enumeration comes back
+**empty**:
+
+| row | expected | actual on main |
+| --- | --- | --- |
+| `Object.keys` puts array-index keys first, ascending | `1,2,b,a` | `""` |
+| `for-in` visits keys in the same order | `1,2,b,a,` | `""` |
+| `Object.values` follows the spec key order | `4,2,1,3` | `""` |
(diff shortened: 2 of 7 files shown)


ttraenkler:
### The `hold` on this PR is main-side, not this change — diagnosed, label deliberately left on

I read the cited run ([34739996390](https://github.com/loopdive/js2/actions/runs/34739996390)) before touching anything, per the park-handling rules. **No test262 verdict was produced**, so this park carries no evidence about this PR.

**What the run actually says.** 90 of 100 jobs failed, every one with the same two lines:

```
##[error]Test262 shard did not reach afterAll; refusing partial JSONL evidence
##[error]Process completed with exit code 2.
```

The shape is infra, not conformance:

- the vitest invocation lasts **~10 seconds** (05:21:11.85 → 05:21:21.61 on `test262 standalone shard 1/50`) — a real shard is minutes;
- the blob report **is** written and vitest's exit code is accepted by the wrapper, so vitest ran and returned cleanly; it just never wrote a `…shard-*.complete.json` completion marker, which is what the guard checks;
- everything upstream is healthy in the log — compiler and runtime bundles build (18.0 MB / 17.8 MB, ~400 ms each) and the Temporal provider artifact downloads with a matching SHA256;
- **both** the js-host and the standalone matrices fail, which rules out anything target-specific.

**It is not this PR.** The identical signature appears on unrelated PRs in the same window — #5889 ([job 103676937051](https://github.com/loopdive/js2/actions/runs/34739544544)) and #5885. Every `merge_group` on main today is failing `Test262 Sharded` the same way, so the whole queue is wedged.

This is also precisely the case the park comment's own footer names: a setup step rather than a verdict step, so the verdict never ran and the park is spurious.

**Why I am leaving the `hold` on.** Removing it would let `auto-enqueue` take the PR straight back into a group that fails identically — queue churn with no new information, and the next park would look like a second, independent regression. The label should come off once the underlying outage is fixed, together with the other PRs parked by it.

Filed as [#6454](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6454-merge-queue-test262-shards-die-before-afterall), which carries the bisect range and the acceptance criteria.

**This PR's own evidence is unchanged:** all six required checks are green at PR level, `mergeStateStatus` was `CLEAN` before the park, and the change is 4 lines in the host runtime with zero net line growth and no codegen change.

ttraenkler:
### Update: the queue recovered on its own — the outage was a ~15-minute window

PR #5891's `merge_group` **passed at 05:45:22Z**, so `Test262 Sharded` is healthy again. The failure window was roughly **05:08Z–05:22Z**, which covers every group in my previous comment (#5885, #5889, this PR's two).

Two things follow.

**The `hold` on this PR is now clearable.** It is confirmed collateral, not a regression: the same signature on unrelated PRs inside the window, no test262 verdict ever produced, and the gate demonstrably green afterwards. I will merge current `main` in and clear the label once the dogfood A/B finishes (it is mid-run and swaps `src/runtime.ts` between lanes, so the tree is not clean to merge into yet).

**The bigger finding is the aftermath, and I have corrected [#6454](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6454-merge-queue-test262-shards-die-before-afterall) to say so.** The outage healing itself also points away from a bad commit and toward infrastructure, so that issue now says to rule that out before bisecting main. **18 open PRs currently carry a `hold`.** `auto-enqueue` skips a held PR, so each strands until someone clears it — and because the rules rightly forbid removing a bot park without diagnosing its cited run, each one costs a full diagnosis that will reach this same conclusion. That sweep is the urgent half of #6454; a park carrying no evidence devalues the parks that do.

ttraenkler:
### Correction: 3 PRs were parked by the outage, not 18

My previous comment said 18 open PRs carry a `hold` and implied they were all collateral of this window. The first half is true and the second is not, so I have corrected [#6454](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6454-merge-queue-test262-shards-die-before-afterall).

18 is what a label query returns. Checking **when and by whom** each label was applied tells a different story:

```
gh api repos/loopdive/js2/issues/<n>/events \
  --jq '[.[]|select(.event=="labeled" and .label.name=="hold")]|last'
```

Only three were added by `github-actions[bot]` inside the failure window:

| PR | `hold` applied | by |
| --- | --- | --- |
| #5889 | 05:12:21Z | `github-actions[bot]` |
| #5885 | 05:15:57Z | `github-actions[bot]` |
| #5890 | 05:22:56Z | `github-actions[bot]` |

The other 15 are manual holds from `ttraenkler` (mostly 2026-09-09) or bot parks from earlier days. Sweeping those as collateral would silently re-admit work someone deliberately paused — the opposite of what the park rules are for. The actionable set is three, and the label alone cannot tell you which three.

ttraenkler:
### Root cause found (by another lane), my "recovery" reasoning was wrong, clearing the `hold`

Merging current `main` brought in **`7871adcc79`**, which is the actual answer and makes my earlier diagnosis more precise:

> since `dfecafa7e9` (the S6 lane's grounding commit, landed on main inside #5875) the tree carried `test262` as a **symlink to `/home/user/js2/test262`** — a path that exists only on one dev box. On a runner `actions/checkout` registers no submodule, `TEST262_ROOT` resolves to a dangling link, and `findTestFiles` throws `ENOENT` at collection.

The real window was **03:28Z–05:55Z**, wider than the 05:08–05:22 I had measured from three PRs.

**And I need to correct myself.** I cited PR #5891's `merge_group` passing at 05:45:22Z as proof the queue had recovered. It was not proof of anything: #5891 touches only `plan/` and `tests/dogfood/`, so its shard matrix was **skipped** — a green no-op. The real recovery was `7871adcc79` at 05:55Z, ten minutes later. I reached the right conclusion (this failure is not this PR's) by reasoning that does not hold, and that is worth saying plainly rather than leaving in the thread.

That trap is now the point of [#6454](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6454-merge-queue-test262-shards-die-before-afterall), which I have rescoped from "the outage" (fixed, nothing owed) to the two things it exposed and which survive the fix:

1. **A shard that collects zero tests is indistinguishable from one that crashed.** Ninety job logs said only `did not reach afterAll` — the downstream evidence guard firing — and not one named the dangling path or the `ENOENT`. That is why this read as a conformance regression and cost each parked owner a full diagnosis.
2. **`main` looked green for two and a half hours** because docs-only PRs skip the shard matrix. Green `main` was not evidence the gate worked; it was evidence nothing asked it to.

**Clearing the `hold` on this PR now**, with all four preconditions met: the cited run was read before the label was touched; the cause is known, external to this PR, and fixed on `main`; current `main` (including that fix) is merged in here; and everything is re-verified on top of it — both test files green (6/6) and every ratchet gate green, with `src/runtime.ts` still at exactly 19735 lines.

The dogfood A/B is complete and in the PR body: **all 17 suites flat, every per-file `N/M Wasm` fraction identical between lanes.**

Not enqueuing — `auto-enqueue.yml` owns that; clearing the label is what lets it see the PR again. #5885 and #5889 still carry a park from the same window and are not mine to clear.

PR #5893, opened 2026-09-13
fix(#6423): an absent number property stringifies as "undefined", not "NaN"

ttraenkler:
Fixes [#6423](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6423-absent-number-property-stringifies-as-nan).

`String(o.maxAge)` on an object with no `maxAge` printed `"NaN"` where §7.1.17
says `"undefined"`. Same for `` `${o.maxAge}` ``, `o.maxAge + ""` and
`String.raw`.

## Mechanism

`__extern_get` narrows a dynamic read of a number-shaped slot to
`{ kind: "f64", undefSentinel: true }` and materialises an **absent** slot as
`UNDEF_F64_BITS` ([#5251](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5251-extern-get-undef-sentinel-brand)).
That brand is exactly why `typeof o.maxAge`, `o.maxAge === undefined` and
`"maxAge" in o` all answer correctly — the property **read** was never broken.
The f64 ToString arms then ignored the brand and called `number_toString` on the
raw f64, and the host import renders that bit pattern as `"NaN"`.

New brand-gated helper `emitNumberToStringSentinelAware` in
`src/codegen/coercion-engine.ts`. For an unbranded operand it emits the same
single `call number_toString` as before, byte for byte. For a branded f64 it
tees into a scratch local, tests the exact `UNDEF_F64_BITS` i64 pattern
(`emitIsUndefF64`) and selects the `"undefined"` constant. It always leaves an
**externref** — the same shape `number_toString` leaves, in native-strings mode
too, because `stringConstantExternrefInstrs` appends `extern.convert_any` — so
every caller's tail (`emitNativeStringRefFromExternref`,
`emitStringBuiltinNumberResult`, the host `concat`) is untouched, and it
registers no late import.

**The bit-pattern compare is the load-bearing part.** `UNDEF_F64_BITS` is a
*signaling*-NaN payload JS arithmetic cannot produce; a genuine `NaN` is the
quiet `0x7FF8000000000000`. An `f64.ne` self-compare would have mapped every NaN
to `"undefined"` — worse than the bug being fixed.

Sites routed: `emitToString`'s f64/i32/i64 arm (which also covers the js-host
template span, already delegating there), `String(x)`, `+` concat left and
right, and `String.raw`.

### JS-host lane only — and that narrowing is itself a finding

A first cut also routed `compileNativeConcatOperand` and the native template
span through the helper: same defect, other lane. The merge group then breached
the standalone host-free pass-count floor (#2097) —
`current pass=35567, mark=35686, delta=-119`.

Attribution was genuinely ambiguous. The mark was set at `e8a778638f`
(2026-09-12T21:36Z); every merge group between then and this one **skipped the
shard matrix**, so that run was the first to exercise the floor in ~8.5 hours,
and a lot of standalone-touching source had landed in the meantime. Rather than
guess, the helper now returns the plain call unless
`coercionMode(ctx) === "js-host"`, which makes standalone / WASI /
native-strings codegen **byte-identical to the parent by construction** —
verified by SHA-256 of the emitted binaries across three fixtures (for-of
numeric concat, the full `String`/template/`+`/`String.raw` set, generators) in
all three non-js-host configurations: nine hashes, all equal. A repeat breach is
provably not this change.

The js-host binary is unchanged by the narrowing (same fixture, same hash before
and after), so the A/B below — entirely js-host, `target: "gc"` — stands as
measured.

The native lanes are not simply "two more arms": `for-of` over a numeric vec and
native generator `IteratorResult` reads both produce the same brand there, so
the blast radius is numeric-vec iteration rather than one property read. Filed
as [#6458](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6458-native-lane-f64-sentinel-tostring),
which needs a standalone test262 measurement — the dogfood A/B is structurally
blind to that lane.

## Evidence

| lane | parent | fixed |
| ---- | ------ | ----- |
| probe, dogfood `compileAndRunUpstreamModule`, untyped two-file fixture | wasm **3/9**, native 9/9 | wasm **9/9**, native 9/9 |
| `tests/issue-6423-absent-number-stringify.test.ts` bitmask | **448** (`0b111000000`) | **511** |
| standalone (`target: "standalone"`, `hostBridge: "off"`) | **127** | **127** (already correct, and now byte-identical too) |

The three passing cases on the parent are the anti-vacuity controls and they are
the point: a present `maxAge: 0` still prints `"0"`, a present `NaN` still
prints `"NaN"`, a present `5` still prints `"5"`.

One note carried into the test's own comment: the calls must be
**unspecializable**. Written as straight-line calls with literal arguments
inside one exported function, the same nine cases answer 9/9 *on the parent* —
the compiler resolves the read statically and no sentinel is produced.
Registering each case as a closure (what the dogfood harness does with its
`test(name, fn)` callbacks) keeps the read dynamic.

## A/B — 17 dogfood suites, one HEAD, base vs fix

Every headline identical and **zero per-test movers** (63,740 + 1,900 tests
compared by name and status):

webpack 16/16 · three 17/18 · clsx 32/32 · cookie 63740/63740 · lodash 59/62 ·
redux 67/82 · axios 208/231 · stylelint 108/108 · tailwindcss 13/13 · jsdom 6/6 ·
styled-components 9/9 · uuid 75/75 · marked 16/30 · moment 10/10 ·
prettier 108/151 · jest 335/356 · hono 261/324

This base measured prettier 108 and hono 261 where the 2026-09-12 anchors read
107 and 259 — main advanced in between; both are unchanged base→fix.

hono's spurious `Max-Age=0` is untouched, as expected: that is the guard
misfiring, not ToString, and it stays unattributed per the issue's own note.

## Gates

`check:oracle-ratchet` · `check:dead-exports` · `check:dogfood-validation` ·
`check:host-import-policy` all green; `tsc --noEmit -p tsconfig.ts7.json` clean.
`check:compiler-boundaries` reports `inventory-valid-architecture-incomplete`
identically on the unmodified base — pre-existing, no new module added.

`loc-budget-allow` / `func-budget-allow` granted in the issue frontmatter with a
dated rationale: the call-site edits are one-for-one substitutions and the added
lines are the comments explaining why the brand gate is a bit-pattern compare
rather than a NaN test; both files sit at their ceiling.


ttraenkler:
Diagnosed the cited run before touching the label, per the auto-park rules.

**The failure** (run 34742396970, job `merge shard reports`, step *Standalone pass-count high-water floor (#2097)*):

```
[standalone-highwater] current pass=35567, mark=35686 (floor=35636, tolerance=50, delta=-119).
```

**What I found.** Attribution was genuinely ambiguous, in both directions:

* The only standalone-touching part of this PR was two arms it routed through the new helper *beyond* its implementation plan — `compileNativeConcatOperand` and the native template span in `string-ops.ts`. That is in my area, so I treated it as real.
* But the high-water mark was set at `e8a778638f` (2026-09-12T21:36Z), and every merge group between then and this one **skipped the shard matrix** (non-test262-relevant changes). This run was the first to exercise the floor in ~8.5 hours, and a lot of standalone-touching source landed in that window (`standalone-link-boundary.ts`, `regexp-standalone.ts`, `link-boundary-tostring.ts`, `native-globalthis-outline.ts`, …). The group immediately after — pr-5894 stacked on this PR''s merge commit — reported the identical `pass=35567`.

**What I did instead of guessing.** Made the question answerable. `emitNumberToStringSentinelAware` now emits the plain `call number_toString` unless `coercionMode(ctx) === "js-host"`, and the two native arms are back to their original code. Standalone / WASI / native-strings codegen is therefore **byte-identical to the parent by construction** — verified by SHA-256 of the emitted binaries across three fixtures (for-of numeric concat, the full `String`/template/`+`/`String.raw` set, generators) in all three non-js-host configurations: nine hashes, all equal to the parent''s. If the floor breaches again on this PR, it is provably not this change.

The js-host binary is unchanged by the narrowing (same fixture, same hash before and after), so the 17-suite dogfood A/B in the PR body — entirely js-host, `target: "gc"` — stands as measured. Regression test still 511 (parent 448); standalone probe still 127.

The native lanes are not simply "two more arms": `for-of` over a numeric vec and native generator `IteratorResult` reads both produce the same `undefSentinel` brand there, so the blast radius is the numeric-vec iteration path rather than one property read, and the `$NativeString` ↔ `$AnyString` round trip in that arm was never measured. Filed as [#6458](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6458-native-lane-f64-sentinel-tostring) with a standalone test262 measurement as an acceptance criterion, because the dogfood A/B is structurally blind to that lane.

Removing `hold` on that basis.

ttraenkler:
Follow-up on the park diagnosis above — **the breach was not this PR**, and the record should say so.

PR #5897, which has no relation to this change, failed the same standalone host-free pass floor at 2026-09-13T08:27Z:

```
[standalone-highwater] current pass=35567, mark=35742 (floor=35692, tolerance=50, delta=-175).
```

Identical `current pass=35567`, different PR, against a *newer* mark (`35742`, set at `6aac84c0b6`, 06:43Z). Same current value on an unrelated change means the drop is **main-side**; this PR never caused it. My earlier comment said attribution was ambiguous — it is no longer, and it resolved against my own hypothesis.

**I am keeping the js-host-only narrowing anyway.** The native-lane extension is still covered by no measurement available in this lane — the 17-suite dogfood A/B is entirely js-host, `target: "gc"` — and it carries hazards the js-host arm does not (`for-of` over a numeric vec produces the same `undefSentinel` brand, and the `$NativeString` ↔ `$AnyString` round trip in that arm was never measured). Re-adding an unmeasured codegen change late, into a green PR, on the strength of a *refuted* alarm would be the same mistake in the opposite direction. It stays in [#6458](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6458-native-lane-f64-sentinel-tostring), behind a standalone measurement.

Separately, and not mine to fix here: **main''s standalone host-free pass count is sitting ~175 below its high-water mark** and has now taken two PRs out of the merge queue. Noted in #6458; it likely wants its own bisect.

ttraenkler:
Second park, same gate, and it is still not this PR — **leaving the `hold` on and escalating instead of re-enqueueing**.

Run 34749845086, job `merge shard reports`, step *Standalone pass-count high-water floor (#2097)*:

```
[standalone-highwater] current pass=35577, mark=35742 (floor=35692, tolerance=50, delta=-165).
```

Three data points now:

| time (UTC) | PR in the group | reported | mark | delta |
| ---------- | --------------- | -------- | ---- | ----- |
| 06:39 | #5893 | `pass=35567` | 35686 | −119 |
| 08:27 | **#5897** (unrelated) | `pass=35567` | 35742 | −175 |
| 10:07 | #5893 | `pass=35577` | 35742 | −165 |

The *current* value is ~35,567–35,577 across two different PRs and two different marks. This PR''s standalone / WASI / native-strings binaries are byte-identical to the parent — nine SHA-256 hashes across three fixtures × three non-js-host configurations, all equal — so it cannot move that number by construction. The mark was *raised* to 35742 at `6aac84c0b6` (06:43Z) while the measured value was already ~35,570, which is what turned a regression into a queue-wide block.

I am **not** removing `hold` a second time. The failure is deterministic against the current base, so re-admitting would just re-park — the re-enqueue loop the auto-park rules exist to prevent. This PR is otherwise complete and green (26/26 required checks, `CLEAN` before enqueue).

Filed as [#6461](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6461-standalone-floor-below-mark-parks-queue) with two things to chase: bisect `e8a778638f..0bba1b92bf` (the floor went unexercised for ~8.5 hours because every merge group in between skipped the shard matrix), and work out how a run that skipped the shard matrix was able to bank a *higher* mark.

**This PR is ready to merge the moment main''s standalone floor and its mark agree again** — either the regression is fixed, or the mark is re-seeded on a known-good main run. Happy to re-enqueue on a word from whoever picks up #6461.

ttraenkler:
Third diagnosis of the same park, and this time the root cause is nailed down: **the breached floor is a gate-infrastructure defect on main, not a conformance regression, and not this PR.** Removing `hold`.

## The gate, on six unrelated PRs

Every `merge_group` that actually ran the shard matrix since 05:22Z failed the same step — *Standalone pass-count high-water floor (#2097)* — with the same measured value:

| time (UTC) | group | gate line |
| --- | --- | --- |
| 07:35 | pr-5890 | `current pass=35567, mark=35742 (floor=35692, tolerance=50, delta=-175)` |
| 07:55 | pr-5897 | `current pass=35567, mark=35742 (floor=35692, tolerance=50, delta=-175)` |
| 08:47 | pr-5894 | `current pass=35567, mark=35742 (floor=35692, tolerance=50, delta=-175)` |
| 09:08 | pr-5896 | `current pass=35577, mark=35742 (floor=35692, tolerance=50, delta=-165)` |
| 09:37 | **pr-5893** | `current pass=35577, mark=35742 (floor=35692, tolerance=50, delta=-165)` |
| 10:08 | pr-5902 | `current pass=35577, mark=35742 (floor=35692, tolerance=50, delta=-165)` |

Six different PRs, two distinct measured values, one mark. Also independently reported in #5903. The only groups that went green in that window (pr-5898, pr-5901) did so with `SHARDS_RAN: false` — the gate was a no-op there, which is how #5894/#5896/#5897/#5899 merged *while failing this gate*, riding behind the shard-skipping `ci/npm-compat-refresh` group.

## Root cause: two producers write the same field and disagree by ~165

`benchmarks/results/test262-standalone-highwater.json` history:

| commit | when | producer | mark |
| --- | --- | --- | --- |
| `07c19331dc` | 09-07 19:45 | `refresh sharded baseline` | 35214 |
| `8ce47a4730` | 09-12 21:36 | **scheduled baseline summary sync (#1951)** | 35686 |
| `17ca7e8107` | 09-13 06:43 | **scheduled baseline summary sync (#1951)** | 35742 |

`baseline-summary-sync.yml` ratchets the mark from `test262-standalone-report.json` **fetched out of the `js2wasm-baselines` repo** (workflow L235-L240), while the gate asserts against the `full_summary.host_free_pass` computed by the merge group's *own* shard matrix. Nominally the same quantity; empirically 165-175 apart.

While the sharded producer owned the mark (35214) the floor was never near breaching. The scheduled sync then ratcheted it to 35686 and 35742 — both already **above** what the shard matrix measures — and the first group to run shards after that (06:17Z) breached immediately. That also answers the open question from my 10:19 comment: the higher mark was not banked by a shard-skipping merge group, it was banked by a scheduled workflow reading a different artifact.

## Why it cannot be this PR

This PR's standalone / WASI / native-strings codegen is byte-identical to its parent by construction — the `emitNumberToStringSentinelAware` change is gated on `coercionMode(ctx) === "js-host"` — verified by SHA-256 over the emitted binaries for three fixtures x three non-js-host configurations: nine hashes, all equal to the parent's. It cannot move a host-free standalone pass count.

## State

Branch is 8 ahead / 0 behind `main` (`6f68025` on base `54c36a9`), no conflicts, 26/26 required checks green, no failing check on the PR.

`hold` removed so the PR is eligible again; **I am not enqueueing** — `auto-enqueue` owns that. Caveat stated plainly: until main's mark and the shard matrix agree, a group that runs shards will re-park this PR (and every other one). The fix belongs in [#6461](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6461-standalone-floor-below-mark-parks-queue) — either re-seed the mark from the sharded producer, or reconcile the two producers' `host_free_pass`. Not something to paper over from a feature branch by editing the committed mark.

PR #5894, opened 2026-09-13
fix(#6422): Array.from over a buffer-backed typed-array view no longer traps

ttraenkler:
Closes [#6422](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6422-array-from-buffer-backed-view-illegal-cast).

`Array.from(new Uint8Array(buf))` **trapped** — `RuntimeError: illegal cast` — taking the whole module down. Node answers 32.

## Mechanism

The `Array.from` array-copy fast path picked its carrier from the **checker** type: a declared or inferred `Uint8Array` resolves to a `$Vec`, so the arm compiled the argument and `local.set` it into a `ref null $Vec` local without ever looking at what the argument actually lowered to. Two real carriers behind that same static type are not that vec:

| lane | `new Uint8Array(buf)` builds | parent |
| --- | --- | --- |
| JS host, untyped `buf` | a REAL host `Uint8Array` (externref) via `__construct_closure` | illegal cast |
| both lanes, compiled `ArrayBuffer` | the shared-backing `$__ta_view` struct (#3054) | illegal cast |

`repairStructTypeMismatches` silently papered the resulting validation mismatch over with `any.convert_extern; ref.cast_null $Vec` — which traps, because neither carrier is a WasmGC `$Vec`.

## Fix

One site, transactional, **no widened cast**. The argument is probe-compiled under `snapshotSpeculative`; the new `src/codegen/array-from-vec-carrier.ts` decides whether the value belongs in that `$Vec` local:

- already that vec → commit the `array.copy` lowering unchanged;
- a registered `$__ta_view` → **de-view** it first with #3054 B1's `emitTaViewToVec` (the same materialization the TypedArray prototype methods take), then commit — keeping the fast path *and* the element values;
- anything else → `rollbackSpeculative` into the existing native/host `Array.from` fallback, which reads a host typed array correctly.

The de-view arm is load-bearing, not a nicety: routing a view to the fallback instead answers the right **length** with `NaN` **elements**, because `__extern_get_idx` has no `$__ta_view` arm. That gap is dodged here, not fixed — filed as [#6453](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6453-extern-get-idx-ta-view-arm), which this PR adds.

## Contradicting the plan on one point

The issue's implementation plan recorded the standalone lane as *"measured 32, no trap, no change needed"*. **It traps there too**, for the `$__ta_view` reason the issue text originally guessed. Measured on the parent (`7adc0a6e89`), standalone: `Array.from(new Uint8Array(buf)).length` → `RuntimeError: illegal cast`.

## Evidence

**Probes, parent → fix.** Host lane: length `illegal cast` → 32, byte sum `illegal cast` → 96, bound form (`const v = new Uint8Array(buf); Array.from(v)`) `illegal cast` → 32; controls flat at 3 / 4 / 2. Standalone: length `illegal cast` → 32, sum `illegal cast` → 96; controls flat at 3 / 10.

**Regression test** (`tests/issue-6422-array-from-host-typed-array-view.test.ts`) — 12 cases, both lanes, untyped two-file fixture for the host half. On the parent **5 fail** (3 host + 2 standalone, all as `illegal cast`) and **7 pass**: the anti-vacuity controls, including a WAT assertion that the compiled `Uint8Array` carrier still emits `array.copy`, so a fix that merely routed everything to the fallback would not pass. With the fix 12/12.

**A/B over all 17 dogfood suites at one HEAD (`7adc0a6e89`) — every suite flat:**

| suite | base → fix | | suite | base → fix |
| webpack | 16/16 → 16/16 | | uuid | 75/75 → 75/75 |
| three | 17/18 → 17/18 | | marked | 16/30 → 16/30 |
| clsx | 32/32 → 32/32 | | moment | 10/10 → 10/10 |
| cookie | 63740 → 63740 | | prettier | 108/151 → 108/151 |
| lodash | 59/62 → 59/62 | | jest | 335/356 → 335/356 |
| redux | 67/82 → 67/82 | | hono | 261/324 → 261/324 |
| axios | 208/231 → 208/231 | | stylelint | 108/108 → 108/108 |
| tailwindcss | 13/13 → 13/13 | | jsdom | 6/6 → 6/6 |
| styled-components | 9/9 → 9/9 | | | |

Two anchors in the plan were stale against this HEAD and moved in **both** lanes, so neither is this change's doing: prettier 107 → 108, hono 259 → 261.

**Gates.** loc / func / coercion / oracle-ratchet / dead-exports / dogfood-validation / host-import-policy green; compiler-boundaries inventory green after classifying the new module in `scripts/compiler-boundaries.json`. `tsc --noEmit` clean. The loc/func growth (+15 / +14 — the probe and rollback pair that cannot leave the call site) is granted in the issue's frontmatter with a dated rationale.

## Residuals

- `__extern_get_idx` still has no `$__ta_view` arm → [#6453](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6453-extern-get-idx-ta-view-arm).
- Only `Array.from` was audited. `Array.of`, the spread arms and the other `resolveArrayInfo` consumers pick their carrier from the checker type the same way; no systematic sweep was done.


ttraenkler:
Diagnosed the cited run (34743292750, this PR's own merge group, base `a7265340b1`). The park is **real and mine**, not drift:

```
[standalone-highwater] current pass=35567, mark=35686 (floor=35636, tolerance=50, delta=-119)
STANDALONE host-free pass floor breached (#2097)
```

**Cause.** The first cut of this fix admitted only an exact
`typeIdx === vecTypeIdx` match into the `Array.from` array-copy fast path and
rolled every other WasmGC ref back into the fallback. A GC struct whose index
differs from the checker-derived vec is routinely cast-compatible with it — the
repair's `ref.cast` succeeds there and the `array.copy` lowering was correct all
along. Diverting those changed a working lowering for no reason, and cost 119
standalone host-free passes. Nothing at PR level could see it: all 17 dogfood
suites were flat and every required check was green.

**Fix, pushed as `65ed1cc56f`.** `admitArrayFromVecCarrier` now diverts or
materializes only the two carriers that provably TRAP — a non-GC value (host
externref, scalar) and a `$__ta_view` — and leaves every other ref
byte-identical to the parent. That is strictly *less* divergence from base than
the variant the dogfood A/B measured flat, so the A/B is not invalidated by the
relaxation.

Probes unchanged after the relaxation (host 32/96/32/3/4/2, standalone
32/96/3/96/3/10/4); the regression test is still 12/12 with the fix and
5-fail/7-pass on the parent.

Removing the `hold` to re-admit once, per the auto-park rules.

PR #5897, opened 2026-09-13
fix(#6425): new TextEncoder() constructs on the DOM-free node lane

ttraenkler:
Fixes [#6425](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6425-hono-crypto-textencoder-not-a-constructor).

## Mechanism

`--platform node` type-checks against the DOM-free composite lib
(`src/checker/index.ts`, `DOM_FREE_LIB_NAME`), which declares no `TextEncoder`
or `TextDecoder`. So `ctx.oracle.isUnresolvableIdentifier(TextEncoder)` was true
and the #4246 unresolvable arm of `tryNonConstructableNewTarget` turned every
`new TextEncoder()` into a **static** `TypeError("TextEncoder is not a
constructor")` — reached at `new-super.ts:6583`, before the recovery arm at
`new-super.ts:6905` (`!className && externClasses.has(name) &&
resolvesToAmbientGlobal`) that the synthetic `TextEncoder`/`TextDecoder`
registration in `extern-declarations.ts` exists to feed.

The web lane has lib.dom, resolved the same name, and emitted `TextEncoder_new`
all along — which is why only hono's two node-lane files
(`src/utils/crypto.test.ts`, `src/utils/buffer.test.ts`) ever reported it. The
binding was never missing: the node worker already supplies it
(`upstream-suite-compile-worker.mjs` assigns `getWebHostConstructors()`).

## Fix

One guard on that arm: decline when the unresolvable identifier names a
registered host extern class (`!ctx.externClasses.has(callee.text)`). Such a
name is an ambient host global the *lib* failed to declare, not an undeclared
name. `+17` lines, all but one of them the comment.

Order preservation: the arm stays where it is, only its unresolvable branch
narrows. `new undeclaredName()` still throws — asserted as a control.
Standalone/WASI are untouched: the synthetic registration is gated
`!nativeStrings && !strictNoHostImports`, so `externClasses` carries no
`TextEncoder` there.

## Evidence

**Probe** (`.tmp/6425/probe.mjs`, a two-test `new TextEncoder()` module):
node lane **0/2 → 2/2**; web lane 2/2 → 2/2 (unchanged control).

**Regression test** `tests/issue-6425-node-lane-textencoder-construct.test.ts`
with untyped JS fixtures at `tests/fixtures/issue-6425/` — 4 tests, **2 fail on
the parent** (`expected […] to include 'TextEncoder_new'`, plus the round-trip),
4 pass with the fix. Both anti-vacuity controls (the same fixture on the web
lane; `new NoSuchCtor()` still pooling `NoSuchCtor is not a constructor`) pass
on **parent and child**, which is what proves the other two assertions measure
the node lane rather than the guard's mere existence.

**A/B, 17 upstream suites at `3e92241ecc`, one arm at a time:**

| suite | base | fix |
| --- | --- | --- |
| hono | **261/324** | **265/324** |
| webpack | 16/16 | 16/16 |
| three | 17/18 | 17/18 |
| clsx | 32/32 | 32/32 |
| lodash | 59/62 | 59/62 |
| redux | 67/82 | 67/82 |
| axios | 208/231 | 208/231 |
| jsdom | 6/6 | 6/6 |
| uuid | 75/75 | 75/75 |
| marked | 16/30 | 16/30 |
| moment | 10/10 | 10/10 |
| prettier | 108/151 | 108/151 |
| jest | 335/356 | 335/356 |

The 16 non-hono suites are **byte-identical per file**, not merely equal in
headline. hono's +4 is exactly two files: `src/utils/crypto.test.ts` 0/4 → 3/4
and `src/utils/buffer.test.ts` 4/10 → 5/10.

**Guards green:** `tests/issue-1752.test.ts` (standalone/WASI TextEncoder, 5/5),
`tests/issue-1588-str-to-utf8.test.ts` (10/10), `tests/issue-1528.test.ts`
(11/11), `tests/issue-5096-global-shadow-class.test.ts` (31/31).

## Residual, filed not bundled

hono's fourth crypto test (`Should create hash for Buffer`) still reads `update
is not a function`. Reduced independently of hono: on the node lane
`import { createHash } from "crypto"` compiles to a **null** binding (native
2/2, Wasm 0/2, compile clean). Filed as
[#6450](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6450-node-crypto-createhash-null-provider),
together with the possible same-name local-export shadowing that makes hono's
message differ from the bare reduction's.

Pre-existing and unrelated: `pnpm run -s check:compiler-boundaries` exits 1 on
this branch **and identically on the parent** (`graphComplete: false`,
`external-unbound` entries for `src/codegen/prepared-async-frame-adapter.ts`).
This PR adds no module and no import, so it cannot move that inventory.


ttraenkler:
Diagnosed the cited `merge_group` run before touching the label, per the auto-park rules.

**Verdict: main-side drift on the `#2097` standalone high-water mark, not a regression from this PR. Removing `hold` once.**

The failing step is `merge shard reports` → *Standalone pass-count high-water floor (#2097)* in run [34746428311](https://github.com/loopdive/js2/actions/runs/34746428311):

```
[standalone-highwater] current pass=35567, mark=35742 (floor=35692, tolerance=50, delta=-175).
```

The same number appears, **to the digit**, in the merge groups of two unrelated PRs that do not contain this commit — each group is `main` + that one PR:

| merge group | run | reading |
| --- | --- | --- |
| `pr-5885-a172bc74` | [34744425905](https://github.com/loopdive/js2/actions/runs/34744425905) | `pass=35567, mark=35742, delta=-175` |
| `pr-5890-fd1448d6` | [34745598631](https://github.com/loopdive/js2/actions/runs/34745598631) | `pass=35567, mark=35742, delta=-175` |
| `pr-5897-88ee0c27` (this PR) | [34746428311](https://github.com/loopdive/js2/actions/runs/34746428311) | `pass=35567, mark=35742, delta=-175` |

`pr-5885`'s group ran at 07:34, before this PR's group existed. A change present in only one of the three cannot produce an identical delta in all three. The mark was re-seeded at `6aac84c0b6` (2026-09-13T06:43:24Z) to 35742, and every merge group since reads 35567 — the mark is ahead of what current main reproduces, which is what `#2097` calls a mark needing a `--update` re-seed on a known-good main run.

**Why this PR cannot move the standalone lane at all.** It adds one guard to the `#4246` unresolvable arm of `tryNonConstructableNewTarget`: decline when the identifier names a **registered host extern class**. The synthetic `TextEncoder`/`TextDecoder` registration that populates that set is gated `!ctx.nativeStrings && !ctx.strictNoHostImports` — i.e. host-import targets only — so under standalone/WASI `ctx.externClasses` carries no such entry and the guard never fires. Local guards confirm it: `tests/issue-1752.test.ts` (standalone/WASI TextEncoder) 5/5 and `tests/issue-1588-str-to-utf8.test.ts` 10/10, unchanged from the parent.

Every other required check on this PR is green, and the PR's own A/B over the 17 upstream dogfood suites moved hono 261→265 with the other 16 **byte-identical per file**.

Re-enqueue is left to `auto-enqueue.yml`; this comment only removes the park.

PR #5899, opened 2026-09-13
fix(#6428): unwrap a settled $Promise at the standalone async value sink

ttraenkler:
Fixes the standalone/WASI half of the async return-value contract that
[#5371](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5371-await-compiled-async-returning-host-promise)
deliberately left open:
[#6428](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6428-standalone-async-return-thenable-value-sink).

## Mechanism

On the native-`$Promise` carrier lane, `return <thenable>` from a
never-suspending async function reached the #1727 raw-value sink
(`f() as unknown as number`) as a `$Promise` externref, which the consumer
coerced straight to f64 -> `NaN`. #5371 fixed the *result-carrier* half; on the
host lane the call site's adopting `Promise.resolve` settles with the inner
value, but host-free there is no adopt.

New module `src/codegen/async-value-sink-unwrap.ts`:

- `emitStandaloneAwaitUnwrap` moved out of `expressions.ts` unchanged (the #3102
  god-file), imported back for the `await` site;
- `emitAsyncValueSinkUnwrap` emits that same guarded one-level unwrap
  (`ref.test $Promise` -> `struct.get $Promise.value`) for a `value` consumer
  when the carrier is active and the result is an externref, and otherwise hands
  `callResult` back untouched.

`asyncResultConsumedAsValue` became `asyncConsumerKind` (returns the
`AsyncConsumerKind` instead of collapsing it to a boolean) so the single call
site passes the kind through -- no new `ctx.checker` query, which keeps the
oracle ratchet flat. `await` consumers are untouched; off the carrier (gc/host)
the gate is false and the host ABI is byte-identical.

## Measured

Probe: `compile()` + `WebAssembly.instantiate(bin, {})`, no host imports.
Identical on `--target wasi` and `--target standalone`.

| case | before | after |
| --- | --- | --- |
| `return Promise.resolve(7)` | NaN | 7 |
| `return mk()` (plain `Promise`-returning callee) | NaN | 8 |
| `const p = Promise.resolve(9); return p` | NaN | 9 |
| `const v = await f(); return v` | NaN | 7 |
| `return await Promise.resolve(7)` (control) | 7 | 7 |
| `return 7` (control) | 7 | 7 |

No microtask drain is needed: the drive-lowered frame settles inside
`__async_resume_ff` before the sink reads it. That was the open question in the
issue's implementation plan; measuring it first is why step 3 of that plan was
not written.

## Tests

`tests/issue-6428-standalone-async-value-sink.test.ts` -- 6 shapes x 2 targets =
**12 assertions**: **8 red / 4 green on the parent, 12 green here**. The 4
already-green are the anti-vacuity controls (the raw-f64 shapes an "unwrap
everything" fix would have moved). `tests/async-await`, `async-census`,
`issue-2865-standalone-async-await-unwrap`,
`issue-3469-standalone-async-completion-sink` and
`tests/equivalence/async-function` also run green (41 tests).

**A/B over all 17 dogfood suites at one HEAD: byte-identical, base vs fix** --
webpack 16/16, three 17/18, clsx 32/32, cookie 63740/63740, lodash 59/62,
redux 67/82, axios 208/231, stylelint 108/108, tailwindcss 13/13, jsdom 6/6,
styled-components 9/9, uuid 75/75, marked 16/30, moment 10/10, prettier 108/151,
jest 335/356, hono 261/324 -- every one unchanged, exit 0.

Gates green with **no growth allowance**: extracting the unwrap shrank
`expressions.ts` by 38 lines and left `compileExpressionInner` at its baseline
size. The new module is classified in `scripts/compiler-boundaries.json`
(`errors: []`, `inventoryValid: true`).

## Also files a new issue

[#6456](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6456-untyped-js-module-in-project-graph-breaks-imported-call)
-- an untyped `.js` module anywhere in a `compileProject` graph breaks every call
into it: a plain `export function c2() { return 7 }` **traps** on
wasi/standalone and **silently returns `0`** on gc, with `success: true` and no
diagnostic. Found while building this fixture (it is why the fixture's producer
half is `.ts` rather than the planned `.js`); pre-existing and unchanged by this
PR.


ttraenkler:
Diagnosed the cited merge_group failure before touching the label, per the park-handling rules.

**Run:** https://github.com/loopdive/js2/actions/runs/34747735684 — the only failing step is
`Standalone pass-count high-water floor (#2097)`:

```
[standalone-highwater] current pass=35567, mark=35742 (floor=35692, tolerance=50, delta=-175)
```

**This is main-side drift, not a regression from this PR.** Three other merge_group runs report the
byte-identical figure, and two of them ran *before this PR existed* (it was created 07:35Z and
enqueued 08:27Z):

| run | PR | started | high-water line |
| --- | --- | --- | --- |
| 34744425905 | #5885 | 07:07Z | `current pass=35567, mark=35742 … delta=-175` |
| 34745598631 | #5890 | 07:35Z | `current pass=35567, mark=35742 … delta=-175` |
| 34746428311 | #5897 | 07:55Z | `current pass=35567, mark=35742 … delta=-175` |
| 34747735684 | #5899 (this PR) | 08:27Z | `current pass=35567, mark=35742 … delta=-175` |

An identical delta across four unrelated merged states is the signature of a high-water mark that
main itself no longer meets — it was seeded at 06:43:24Z from 6aac84c0b6 and has not been reproduced
since. #5898's green run is not counter-evidence: it was a `SHARDS_RAN: false` skip.

**The floor fires before the per-PR guards, so no PR-attributable evidence exists in this run at
all** — `Catastrophic regression guard (#1668)`, `Standalone regression guard (#1897)` and the
per-edition ratchet are all `skipped` behind the failed step.

**Independent reason this PR cannot move test262 by 175:** the new arm is reachable only when
`classifyAsyncConsumer` returns `"value"`, which requires an explicit `as` / `!` / `<T>`
cast-or-assertion node in the source. test262 files are plain JS and contain none, so the arm is
structurally unreachable there. The change is additionally gated on `isStandalonePromiseActive`, so
it is inert on the js-host lane — and the js-host shards in this very run are all green.

Local evidence on this branch: the #6428 regression test is 12/12 green (8 red on the parent),
`tests/async-await` + `async-census` + `issue-2865-standalone-async-await-unwrap` +
`issue-3469-standalone-async-completion-sink` + `tests/equivalence/async-function` are 41/41 green,
and an A/B of all 17 dogfood suites at one HEAD is byte-identical base vs fix.

Removing `hold` on that basis. The standalone high-water mark needs a re-seed from a known-good main
run — that is a queue-health item for the shepherd, affecting every PR in the queue, not this one.

PR #5902, opened 2026-09-13
fix(#6430): expand spread args in String.fromCharCode; a field-less struct is a known shape

ttraenkler:
Fixes [#6430](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6430-hono-signed-cookie-hmac-zero-signature) — hono's signed cookies came back with an all-zero HMAC (`macha.AA%3D%3D`) and every `serialize` appended a spurious `Max-Age=0`. Two independent compiler defects, both reduced to standalone probes first.

## (a) `String.fromCharCode(...codes)` folded the whole spread source into one code unit

`compileFromCharCodeFamily` walks `expr.arguments` and compiles each **node** as one f64 code unit. A `SpreadElement` has no lowering there, so the generic passthrough unwrapped `...codes` to `codes` — an array value — which coerced to `NaN`, went through ToUint16 to `0`, and produced a single NUL character. hono signs with `btoa(String.fromCharCode(...new Uint8Array(signature)))`, so every signature was the base64 of one zero byte, and every signed-cookie *read* then correctly failed verification.

The fix routes a spread-carrying argument list through the shared `buildSpreadArgList` builder (#5361) — the same substrate `Math.max(...xs)` uses (#2054), which reads a source by its actual representation (tuple struct / native vec / host iterable). It lives in a new `src/codegen/expressions/from-char-code-spread.ts`; the family's own ToUint16 and §22.1.2.2 RangeError guard are applied per expanded element, and the parts concatenate into a string accumulator instead of the per-node fold. One hook at the family's entry serves all four lanes plus the reflective `.call`/`.apply` synthetics.

Measured before → after, **gc and standalone**:

| expression | parent | fixed |
| --- | --- | --- |
| `String.fromCharCode(...codes())` | `"\0"` | `"hi"` |
| `String.fromCharCode(72, ...codes())` | `"H\0"` | `"Hhi"` |
| `String.fromCharCode(...new Uint8Array([104,105]))` | `"\0"` | `"hi"` |
| `String.fromCodePoint(...codes())` | `RangeError: Invalid code point NaN` | `"hi"` |

## (b) An absent field on a field-less struct read back `0`

`emitStructFieldNamesExport` emitted an arm only for shapes with at least one field, so `__struct_field_names` answered `ref.null` for the `(struct)` that `function f(opt = {})` lowers `{}` to. The host runtime reads `null` as "unknown legacy shape, keep probing" — not "known shape, no fields" — and the probe it keeps is `__sget_<key>`, a getter shared by every structurally compatible shape (one exists as soon as any caller passes `{ maxAge: … }`), whose f64 miss-default is `f64.const 0`. So `typeof opt.maxAge` answered `"number"` for an absent field, and hono's `typeof opt.maxAge === 'number' && opt.maxAge >= 0` appended `Max-Age=0`.

Every non-synthetic shape now gets an arm, field-less ones answering the empty CSV, and `_csvSplitCache` is seeded with `"" → []` so `_structFieldNamesRaw` returns a positive empty answer. `_structOwnFieldStatus` then answers `false`, `wsh.readField` yields `NO_GENERATED_FIELD`, and the read resolves to `undefined`. **`src/runtime.ts` changes at zero net lines** (it sits at the #4401 ceiling) — two lines replaced one-for-one.

**One guard is load-bearing and was found by measurement, not by reading.** `class A {}` / `class B extends A { x }` gives `A` a field-less struct that `B` declares as its supertype, and `ref.test $A` matches a `B` instance. Without the guard, an empty-CSV arm for `A` ahead of `B`'s stripped every field off `B`: measured `Object.keys(new Derived(1))` → `""` and `JSON.stringify(new Derived(1))` → `{}`. Field-less types that something extends therefore get no arm. (A supertype that *has* fields already answers for its subclasses — pre-existing, not widened here.)

The plan's second layer — an `UNDEF_F64_BITS` miss-default for `__sget_*` — turned out **not** to be needed and is not included: with the shape known, the miss-default is never reached.

## What this does NOT fix — the plan predicted 35/35, and that was wrong

`src/utils/cookie.test.ts` goes **24/35 → 27/35**, hono **261/324 → 264/324** (measured base, not the stale 259 anchor). The remaining 8 rows are a **third, independent mechanism**, now filed as [#6459](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6459-host-object-through-compiled-promise-loses-brand): a host object resolved through a **compiled function's promise** arrives as a mirror that passes property reads (`Object.prototype.toString.call(key)` is `"[object CryptoKey]"`, `key.type` is `"secret"` — which is exactly why #5371 read as disproved) but fails the WebIDL brand check, so `crypto.subtle.verify` answers `2nd argument is not of type CryptoKey` and hono's `catch { return false }` turns it into the observed `false`. Awaiting `crypto.subtle.importKey` in the frame that created it verifies correctly.

## Tests

`tests/issue-6430-fromcharcode-spread.test.ts` (gc **and** standalone lanes) and `tests/issue-6430-absent-field-empty-struct.test.ts`, both untyped `.js` two-file projects. **Parent: 3 failed / 3. With the fix: 3 passed / 3.** Anti-vacuity controls are pinned in both: the non-spread `fromCharCode` rows take the unchanged per-node path, and a *present* field must keep reading its real value.

## A/B — all 17 dogfood suites, one HEAD

Sixteen flat; `src/utils/cookie.test.ts` is the only file that moves anywhere.

| suite | base → fix | | suite | base → fix |
| webpack | 16/16 | | styled-components | 9/9 |
| three | 17/18 | | uuid | 75/75 |
| clsx | 32/32 | | marked | 16/30 |
| cookie | 63740/63740 | | moment | 10/10 |
| lodash | 59/62 | | prettier | 108/151 |
| redux | 67/82 | | jest | 335/358 |
| axios | 208/231 | | **hono** | **261/324 → 264/324 (+3)** |
| stylelint | 108/108 | | | |
| tailwindcss | 13/13 | | | |
| jsdom | 6/6 | | | |

The plan flagged cookie/jest/prettier as the ones to watch for (b) — it touches every host read of a field-less struct — and all three are byte-identical.

## Gates

LOC (+8 in `calls.ts`, granted in the issue frontmatter for the entry hook), func, coercion-sites, oracle-ratchet, dead-exports, dogfood-validation, host-import-policy, ir-fallbacks, ts7 typecheck: all green. `check:compiler-boundaries` exits 1 identically on the parent (`inventory-valid-architecture-incomplete`, pre-existing); the new module is classified in `scripts/compiler-boundaries.json` and `untracked` stays 0.


ttraenkler:
## Park diagnosed: main-side drift, not this PR — unholding

The cited run ([34751074112](https://github.com/loopdive/js2/actions/runs/34751074112)) failed exactly one step, `Standalone pass-count high-water floor (#2097)`:

```
[standalone-highwater] current pass=35577, mark=35742 (floor=35692, tolerance=50, delta=-165).
##[error]STANDALONE host-free pass floor breached: 35577 < high-water 35742 − 50 = 35692.
High-water set at commit 6aac84c0b6ef418bbfa6a97cceca25960db7a3f6 (2026-09-13T06:43:24Z).
```

No other gate fired — the catastrophic guard (#1668), the moving standalone regression guard (#1897) and the per-edition ratchet all show `skipped`, i.e. the step-level fail-fast stopped the job before them; the separate `check for test262 regressions` job concluded **success**.

### Why this is not #5902's regression

**Main itself breaches the same floor with the identical number.** The post-merge `Test262 Sharded` run on `main` at `54c36a9fe3` (run [34749841544](https://github.com/loopdive/js2/actions/runs/34749841544), 09:37Z) reports:

```
[standalone-highwater] current pass=35577, mark=35742 (floor=35692, tolerance=50, delta=-165).
```

Same pass count, same delta, same failing step — with this PR's changes absent. Its merged state therefore adds **0** to the standalone host-free count.

**Every other PR whose shards actually ran fails identically**, across unrelated areas:

| merge_group run | PR | current pass | delta |
| --- | --- | --- | --- |
| 34751074112 | #5902 (this PR) | 35577 | −165 |
| 34749845086 | #5893 | 35577 | −165 |
| 34748976257 | #5896 | 35577 | −165 |
| 34748482771 | #5894 | 35567 | −175 |
| 34747735684 | #5899 | 35567 | −175 |
| 34746428311 | #5897 | 35567 | −175 |

(The merge_group runs that did pass in this window — #5901, #5898, #5895 — are the ones whose shard matrix *skipped* entirely: their `merge shard reports` steps are all `skipped`, so the floor was never evaluated. They are not counter-evidence.)

The drop landed on `main` between `6aac84c0b6` (03:28Z, the commit the mark was seeded from, green) and `88ee0c2782` (07:55Z, first main run at 35567). It is a pre-existing main-side regression of ~165–175 host-free standalone passes that needs its own issue; it is not attributable to, and cannot be fixed on, this branch.

### Action

Removing the `hold` per the main-side-drift path. Not enqueuing — the server-side `auto-enqueue.yml` owns that, and this PR will keep failing the queue until the main-side drop is fixed or the high-water mark is deliberately re-seeded on a known-good main run (`--update`, #2097 / #2879).

PR #5909, opened 2026-09-13
fix(standalone): a spread from a `const` array binding gets a static arity — `new Temporal.Duration(...args)` stops evaluating to `null` (#5383 S12 / #6460)

ttraenkler:
Slice S12 of the standalone Temporal provider work, [#5383](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5383-standalone-temporal-provider). Standalone lane only; the JS-host lane and `src/codegen/temporal-native.ts` are untouched, and the `gc` lane is byte-identical. S11 (#5900) has already merged; this branch is on current `main`.

## What the census found

The brief attributed the largest residual to property-bag FIELD EXTRACTION inside the provider (`PrepareCalendarFields` / `ToIntegerWithTruncation`). **That does not reproduce.** Every constituent of that path answers correctly in a single standalone module with no link at all: computed `bag[k]` where `k` comes from a sorted, concatenated key array; `Object.create(null)` computed writes read back through a dot; and the whole `tn` shape with untyped parameters. The full table is in [#6460](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6460-standalone-property-bag-field-extraction).

What is actually broken, ranked by attributed rows:

1. **~40 rows** — every `X.from(…)` hands back an object whose accessor and method reads all answer `null` (`PlainDate.from("1976-11-18").day` → `null`, control `new Temporal.PlainDate(1976,11,18).day` → `18`). The polyfill builds those results with `Object.create(intrinsic.prototype)` + WeakMap slots, not with `new`. This is cross-module — the single-module control answers correctly — so it is a slice of its own and is written down with reductions rather than half-started.
2. **9 rows** — the one fixed here.
3. `typeof <provider-owned instance>` answers `"function"` (known S11 residual, unchanged).

## What this PR fixes

A spread whose source is a `const` array binding has no *static* arity, and two lowerings were asking the same too-narrow question (`flattenCallArgs`, which only recognises an inline array literal):

- the native construct drivers are minted **one per call-site arity**, so an unresolvable arity made the site decline and the `new` evaluate to `null`. That is the Temporal row: every `built-ins/Temporal/Duration/*-undefined.js` writes `const args = [1, 1, 1]; new Temporal.Duration(...args)`, and the harness then reports `Cannot access property on null or undefined` from `duration.years`;
- the ordinary-function `new` path pushes one operand per declared parameter, so the unexpanded spread degraded to its own inner expression and the whole ARRAY landed in the first parameter.

New module `src/codegen/static-spread-arity.ts` resolves such a spread only when the binding is `const`, initialised with a hole-free array literal of re-evaluable literals, declared in the same file, and **only ever spread** — no aliasing, no element write, no argument position. That last rule is the load-bearing one: a `const` array is mutable through its own reference, so proving the declaration is `const` proves nothing about `length`. The test asserts the REFUSAL as well as the successes, because getting that rule wrong produces a wrong answer rather than a missing one, and no byte A/B would show it.

Both call sites gate on `ctx.standalone || ctx.wasi`; the JS-host lane already repairs these shapes at runtime through `emitDynamicSpreadCall`.

## Measured

Base is a file-copy revert **on this tree**, not another worktree's numbers — the S11 Duration figure was 43 and this tree's base sample scored 42, a one-row compile-budget artifact that the solo re-runs resolve to 43. Comparing across worktrees would have manufactured a +1.

| three linked families, 120 rows each | base | S12 |
| --- | --- | --- |
| `built-ins/Temporal/PlainDate/**` | 62 | 62 |
| `built-ins/Temporal/Duration/**` | 43 | 43 |
| `built-ins/Temporal/ZonedDateTime/prototype/**` | 65 | 65 |
| **total** | **170** | **170**, 0 pass→fail |

**The score does not move, and that is the honest result.** What moves is the bucket: Duration's `Cannot access property on null or undefined at 164:22` goes 9 → 4, and five rows advance to `explicit: years result: Expected SameValue(«undefined», «1»)` — the constructor now returns a real Duration and the row dies one step later, on residual 1 above. Of the four rows left, two spread 8 elements plus a trailing `undefined` (arity 9, over the pre-existing `MAX_NATIVE_CONSTRUCT_ARITY` of 8) and two use no spread at all.

Every CE↔fail flip in the sample was a compile-budget artifact, established by re-running the same six rows solo at 60 s on **both** trees; all six agree. Had only the branch been re-run, one 29-second row would have been reported as a +1.

Byte A/B, base captured at the first edit and re-measured after merging `origin/main`: 80/80 artifacts identical on a 40-module corpus (which contains no spread-into-`new`, so it is a no-collateral control); on a targeted 11-shape corpus all 11 `gc` artifacts are identical and exactly the three repaired shapes move on standalone, with every refusal control byte-identical.

Must-not-move standalone samples, both flat with 0 flips: `language/expressions/new/**` (56/3) and `built-ins/Object/**` (106/14).

`equivalence-gate`: 22 failing / 1720 passing, matching baseline.

## Known residual, pinned

A spread into a *dynamic call* (`const g = f; g(...args)`) is still dropped — same root cause, a third lowering that reads `expr.arguments` positionally in about a dozen places. The last `it` in the new test PINS the current wrong answer so the day it is fixed the stale expectation fails loudly instead of rotting.

Issues: [#5383](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5383-standalone-temporal-provider) · [#6460](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6460-standalone-property-bag-field-extraction) · [#5408](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5408-standalone-temporal-plaindate-from)

Re-verified by the coordinator on the pushed head (7a917fad0e, current main merged in, `test262` gitlink 160000): typecheck, compiler-boundaries inventory, speculative-rollback gate, #6460 suite (5 passed).

---


ttraenkler:
Park diagnosis (run 34756760226, standalone host-free floor −157): **not this PR's change.** The merged-report diff against the standalone baseline shows 181 pass→non-pass rows: 174 `language/statements/for-await-of/*` (compile error "internal: async function `fn` activates a state machine") and 7 spread rows (`expressions/super/call-spread-*`, `arguments-object/*trailing-comma-spread-operator`, `harness/compare-array-arguments`). All of them fail on current `main` without this branch.

Bisected locally on `--target standalone`, forced checkouts with HEAD verified:

| commit | super spread | trailing-comma spread | for-await-of |
| --- | --- | --- | --- |
| `607c3c799f` (#5875, S8 merge) | pass | pass | pass |
| `c12c178e55` (#5882, issue-6416 `__extras_argv`) | **fail** | **fail** | pass |
| `82ca09dd09` (#5871, issue-6412 async Promise carrier) | fail | fail | **CE** |

Both #5882 and #5871 merged at 02:45–02:48 UTC, inside the window where `test262` was a dangling symlink on the runners (#5892), so no shard ever validated them; the high-water mark was seeded at `6aac84c0` before either landed, and the drop only crossed the 50-row tolerance now. Revert PR for both merges is coming; this PR stays held until `main`'s floor is restored and then gets one release.

---

ttraenkler:
Releasing this park. The condition set in the diagnosis above — "this PR stays held until `main`'s floor is restored" — is met, verified against the promoted standalone baseline rather than inferred.

The park cited a −157 standalone high-water breach whose 181 pass→non-pass rows were 174 `language/statements/for-await-of/*` compile errors plus 7 spread rows, attributed to #5871 and #5882 landing inside the broken-submodule-symlink window (#5892). In the standalone baseline promoted on 2026-09-16 at 10:46 UTC:

| cluster the park named | current baseline |
| --- | --- |
| `language/statements/for-await-of/*` | 971 pass, 263 non-pass |
| the 7 named spread rows' families | 60 pass, 16 non-pass |

The 174-row compile-error cluster is not there. A whole-corpus set-diff of that baseline against the one from 2026-09-07 finds 555 rows gained and exactly 1 lost (`language/statements/generators/default-proto.js`, unrelated to this branch and to the cited failure), so main's floor is restored rather than merely re-marked.

`origin/main` is merged into this branch as of 02b8bacb and the source-ratchet gates, issue-spec-coverage and issue-ids all pass on the result. Removing `hold` so `auto-enqueue` can take it; per the park comment's instruction the batched group members go one at a time, and this is the one being released.

---

PR #5913, opened 2026-09-13
fix(#6438): wrapExports refuses a raw exports record it cannot decode

ttraenkler:
Closes [#6438](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6438-wrapexports-raw-exports-marshals-struct-to-empty).

## The defect

`wrapExports` documents two inputs — the genuine `WebAssembly.Instance`
("preferred") and the raw exports record ("retains the historical API"). They
did not agree. Measured on `upstream/main` 69ccb3494f, one module, one call:

```
wrapExports(instance,         { signatures }).parse("x")  →  { type: "Program" }
wrapExports(instance.exports, { signatures }).parse("x")  →  {}
```

No error, no warning, and `{}` is indistinguishable from a genuinely field-less
object — so every embedder on the documented historical API read a wrong answer
with nothing to look at. A vec of such structs came back as `[{}]`.

## Mechanism

`_brandedInstanceExports` answers `undefined` for a bare record, so
`wrapExports` passes `mayEstablishDataStructAuthority: false` and the raw
overload can only CONSUME a [#3520](https://js2wasm.loopdive.com/dashboard/issue.html?slug=3520-data-struct-host-bridge-abi-ownership)
data-struct authority. With none established, `_hostBridgeExportView` masks the
compiler's own `__struct_field_names` to `undefined`, `_structFieldNamesRaw`
answers `null`, `looksMarshalable` still answers `true` through its `hasVecLen`
tail, and `_wasmToPlain` takes the [#3637](https://js2wasm.loopdive.com/dashboard/issue.html?slug=3637-vec-len-discriminator-vacuity)
"neither named struct nor vec" arm.

## The change

Contract: **refuse loudly, never decode from a raw record** — "decode anyway" is
ruled out by #3520's design (a mutable binding table cannot authenticate
callable identity).

- `wrapExports` computes `dataStructDecoderMasked` once per wrap; one arm in
  `invoke` — skipped under `marshal: false`, which still hands back the raw
  handle — asks the new `src/runtime/raw-exports-struct-authority.ts` whether
  the result is a struct no decoder can name, and throws a `TypeError` naming
  `__setInstance` if so. Vecs are scanned element-wise (depth-capped).
- `_wasmToPlain`, `_hostBridgeExportView` and the lifecycle adapter are
  untouched, so compiled-program `JSON.stringify` under legacy `__setExports`
  keeps its semantics.
- Consuming an authority established by `__setInstance(instance)` or by an
  earlier Instance wrap of the same module is unchanged.

**One widening past the plan, by measurement.** The plan gated the arm on
`marshalable`. The `__struct_field_names`-collision fixture showed the
NOT-marshalable arm is just as wrong: with the decoder masked AND no `__vec_len`
export, `looksMarshalable` falls through and a plain struct reaches JS as a
CALLABLE (the #1308 fallback). Same missing authority, different wrong answer,
so the guard runs whenever the decoder is masked; `__is_closure` still exempts
real closures. Second refinement: undecodability is tested as
`_structFieldNamesRaw(value, view) == null`, not "masked", so a #5225
cross-module decoder that CAN answer is still honoured.

## Fallout — measured, not assumed

All **70** test files that use the raw overload were run against the parent and
against the fix at one HEAD: **zero new failures**. Every site that now throws
was ALREADY red on main, reading `{}` where it asserted fields — the predicted
"suite-wide sweep" did not materialize because those sites were already broken.

Re-wiring six helpers with the documented one-liner (`__setInstance(instance)`
beside the existing `__setExports`) turns **21 previously-red tests green**:
`#2747` (5), `#2806` (3), `#2841` (4), `#2851` (4), `#3637` (5). Three failures
in those files survive and are unrelated (`#2747` multi-level `__proto__` chain,
`#2836` two vec round-trip cases).

`tests/issue-3520-data-struct-host-bridge-abi.test.ts` moves three `toEqual({})`
assertions to `toThrow(/data-struct authority/)` — same fail-closed outcome, no
forged field list, stated loudly instead of as an empty object.

## Tests

`tests/issue-6438-raw-exports-struct-decode.test.ts` (untyped `.js` fixtures):
Instance overload decodes · unauthenticated raw record refuses (struct AND vec)
· `__setInstance` then raw decodes · raw-after-Instance-wrap decodes ·
field-less struct still answers `{}` on an authenticated view (anti-vacuity) ·
`marshal: false` still hands back the raw handle · #3520 control with a
user-declared `__struct_field_names`. **2 of 7 fail on the parent, all 7 pass
with the fix, the other 5 pass both ways.**

## Gates

`check-loc-budget` (growth granted in the issue frontmatter: `src/runtime.ts`
+27, `src/index.ts` +6 — the 108-line decision module lives outside the
god-file), `check-func-budget`, `check-coercion-sites`, `check:oracle-ratchet`,
`check:dead-exports`, `check:dogfood-validation`, `check:host-import-policy`
(`maximumRuntimeTsLines` 19735 → 19762, exactly this PR's measured delta),
`check-compiler-boundaries --mode inventory`, `tsc --noEmit` — all green.

Dogfood anchors unchanged (17 suites measured on this branch).


ttraenkler:
Dogfood A/B, all 17 suites, one at a time, at this branch HEAD (the merge it carries touches no `src/`, `tests/` or `scripts/` file, so every row is one code HEAD):

webpack 16/16 · three 17/18 · clsx 32/32 · cookie 63740/63740 · lodash 59/62 · redux 67/82 · axios 208/231 · stylelint 108/108 · tailwindcss 13/13 · jsdom 6/6 · styled-components 9/9 · uuid 75/75 · marked 16/30 · moment 10/10 · prettier 108/151 · jest 335/356 · hono 268/324

Two rows differ from the dispatch anchors (prettier 107 → 108, hono ~261 → 268). Both were re-measured with `src/runtime.ts` reverted to its pre-change copy at this same HEAD and answered **identically** (prettier 108/151, hono 268/324), so they are main-side drift since the anchors were taken, not movement from this PR. Net dogfood delta: **zero**.

That is the expected result: `tests/dogfood/upstream-suite-compile-worker.mjs` calls `imports.__setInstance(instance)` and then wraps the **Instance**, so the masked-decoder arm this PR adds is unreachable from the dogfood path. The same holds for `scripts/generate-npm-compat-report.mjs` (every raw-record wrap there is preceded by `__setInstance`) and for the CLI-emitted `instantiateBytes`, which wires `setInstance` during instantiate.

PR #5944, opened 2026-09-15
fix(test262): #6477 — run the linked body after the consumer is registered; parity 4/47 → 25/47

ttraenkler:
## Summary

[#6477](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6477-linked-harness-descriptor-reads) — the linked-harness shadow lane's descriptor-read class (slice 3 of [#3451](https://js2wasm.loopdive.com/dashboard/issue.html?slug=3451-linked-harness-wasm-separate-compilation)).

**Root cause (measured):** a test262 body runs in the wasm `start` section, i.e. during `WebAssembly.instantiate`, but the consumer was registered in the #5225 cross-module decoder registry only afterwards (`wireCompiledInstance`). While the body ran, the registry held one module (the provider), so every provider-side read of a consumer struct resolved with the provider's exports and answered the `ref.test`-miss default.

**Fix:**
- P1 — `compileHarnessLinkedBody` compiles the body with `deferTopLevelInit: true`; `instantiateTest262Module` (linked arm only) calls `instance.exports.__module_init()` after `wireCompiledInstance(…, true)`. Honest lane byte-identical (control test pins: honest keeps its `start` section and exports no `__module_init`; linked is the reverse).
- P2 — `_readOwnDescriptor` and `__getOwnPropertyNames` resolve their exports through `_decoderExportsFor(obj, …)`, so the descriptor path redirects like `_wasmStructHasOwn` already did. No-op for single-module programs.

**Measured** (worker protocol, `COMPILER_POOL_SIZE=1`, linked vs honest, 47 rows = the 7 named `defineProperty` rows + the `class/elements/multiple-*privatename-identifier*` family):

| metric | before | after |
| --- | --- | --- |
| honest/linked agreement | 4 / 47 | 25 / 47 |
| linked fail→pass | — | 21 |
| linked pass→fail | — | 0 |

The 4 rows of the issue's 13 that did not flip are two different mechanisms, filed separately: [#6482](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6482-linked-consumer-vec-index-read-in-wasm) (provider index-reads a consumer vec in-wasm, never reaches the host) and [#6483](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6483-linked-class-fields-proto-method-own-property) (`verifyProperty(C.prototype, "m")` when the class also has instance fields).

## Changes

- `src/test262-harness-provider.ts`, `scripts/test262-import-object.mjs`, `src/runtime.ts`
- `tests/issue-6477-linked-descriptor-reads.test.ts` (new, 5 tests); `tests/issue-3451-linked-harness-substrate.test.ts` — the "own property" case moved from a parity assertion to a one-sided one (linked now passes; honest keeps its pre-existing `allowJs` gap)
- Issue files: #6477 done, #6482/#6483 filed, #3451 measurement row

## Validation

- Equivalence gate: no new regressions (22 failing / 1720 passing = baseline).
- `issue-6477-*` 5/5, `issue-3451-*` 11/11, `issue-6475-*` 6/6, `issue-6476-*` 3/3, `issue-5225/5353/5364/5738` 42/42.
- Lint, typecheck, loc/func budgets (vs origin/main; `src/runtime.ts` +19 LOC / `resolveImport` +10 granted in the issue file), coercion, oracle ratchet, dead-exports, host-import-policy, compiler-boundaries inventory, issue gates green.

---


ttraenkler:
Real regression, diagnosed and fixed in 1c8b440a74.

**Cause:** the sharded worker already compiles every host-lane row with `deferTopLevelInit` (#3123) and calls `__module_init()` itself after `setInstance`. The #6477 change added an unconditional `__module_init()` call inside `instantiateTest262Module`'s linked arm, so every linked Temporal row ran module init **twice** — hence the −401 (`ZonedDateTime/prototype` 67, mostly `wasm exception during module init`) in run 34923837185.

**Fix:** the call is now opt-in (`runDeferredInit: true`) and only the in-process callers that never go through the worker pass it (linked-harness smoke, `issue-3451/6475/6476/6477` suites). The worker keeps its single call and its own throw classification. Linked suites 47/47 locally; lint/typecheck green.

Removing `hold` to re-enqueue.

---

PR #5963, opened 2026-09-17
fix(test262): linked-lane parity batch — #6491 argc bridge, #6482 shape brands, #6492 Iterator statics + sandbox instanceof (+~190 rows)

ttraenkler:
[#6491](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6491-linked-lane-negative-early-errors-compile-clean) — the "early error not detected" parity bucket after the #3451 slice-6 flip.

## Two real fixes

1. **Under-applied consumer calls never ran the body.** `__call_fn_N` matches only closures of declared arity N. The host bridge's free-function arm (`_wrapWasmClosureUnknownArity`, `src/runtime.ts`) dispatched at the *call site's* argument count, so a harness helper calling a consumer function with fewer arguments than declared selected `__call_fn_0`, matched nothing and returned `undefined` — no default-parameter initializer, no throw. The method arm has handled this since #2664 via `__closure_arity`; the free arm now does the same, widening to the declared arity only when under-applied and padding with real `undefined`. Documented residual: `arguments.length` reports the declared arity on such a call. This is not test262-only: any provider→consumer callback that under-applied was a silent no-op.
2. **§16.2.1.1 duplicate `ImportedBoundNames`** early error added in `src/compiler/early-errors/module-rules.ts` — nothing enforced it; the single-source path only appeared to because its import preprocessing rewrote unresolvable imports into declarations the duplicate-lexical check caught.

## Measured (real runner, both lanes)

| | before | after |
| --- | ---: | ---: |
| linked, the 50 bucket rows | 0 pass | 15 pass (all 14 `eval-code/direct` + `import/dup-bound-names.js`) |
| honest, the 50 rows | 50 pass | 50 pass |
| honest, 1,297-row callback slice (`Array.prototype.{forEach,map,reduce,sort,filter}`, `Promise.prototype.then`) | 913 / 334 / 46 / 4 | identical, 0 flips |

## The other 34 rows are honest-lane false passes

The early-error pass does run on the linked graph. For 34 of the 36 negative rows the honest whole-assembly compile reports only an `IR path failed for $DONOTEVALUATE … [IR-FALLBACK]` *warning*, which the worker's #2912 lenient arm scores as a SyntaxError detection. The linked body-only unit has no such warning, so its `fail` is the correct verdict. Not "fixed" here on purpose; recorded in the issue for a compiler-side early-error implementation.

LOC allowance (+29 in `src/runtime.ts`) declared in the issue frontmatter. Gates, typecheck, and the #1931/#2664/#2623/#3451/#6474/#6492 suites green.

---


ttraenkler:
Park diagnosed (run 35256162280): real, and mine on both counts. `hold` stays until the fixes are pushed here.

The group carries the effect of #5962 (already on main) plus this PR: **138 improvements, 20 regressions, +118 net**, trap ratchet `illegal_cast` 30 → 31.

1. **19 rows** `built-ins/Iterator/{concat,from,zip,zipKeyed}/*` + `prototype/{chunks,windows}/next-method-returns-non-object.js` — from #5962's `%Iterator%` binding stratum. Both baselines in `js2wasm-baselines` show why: all 19 were honest **fail** (same messages) and linked **pass** before the stratum, i.e. the linked lane resolved the bare `Iterator` to the host engine's real global (CI's Node has `from/concat/zip/zipKeyed/chunks/windows`); the shim now shadows it and the rows revert to the honest verdict. Fix in flight: the shim keeps the host statics/prototype methods it does not provide itself, measured on all 654 `built-ins/Iterator/` rows in both lanes.
2. **`harness/verifyProperty-arguments.js`** — this PR's arity widening reports `arguments.length` as the declared arity, so `verifyProperty()`'s 0-argument guard no longer throws. Fix in flight: seed the real argc for the free-function bridge family the way the method family already does.
3. **`illegal_cast` +1** on `Array/prototype/sort/comparefn-resizable-buffer.js` (baseline: fail) — being A/B'd against the arity change on that row; fixed here if it is this PR's.

Both fixes will be pushed onto this branch; the PR is not re-enqueued until then.


---

ttraenkler:
Park fixed and the branch now carries the whole parity batch (head `45a1cec5b3`):

- **19 Iterator-statics rows** (`ee2898aae2`): the shim keeps the constructor role and `%IteratorPrototype%` but copies the intrinsic's own function-valued statics onto itself (enumerated, not hand-listed — CI's Node 25 has more than the container's Node 22), and `needsIteratorBinding` no longer counts `Iterator` mentions inside comments. All 654 `built-ins/Iterator/` rows both lanes: linked 341 → 352 (+114 / −2 vs the promoted baseline; the −2 are `prototype/{chunks,windows}` which this container's Node cannot exercise and are expected to pass on CI's — the one prediction here). Two honest-audit rows (`{from,zipKeyed}/non-constructible.js`) were accidental passes (js2's `new` does not honour a host function's non-constructibility, separate gap) and now fail honestly.
- **`verifyProperty-arguments.js` + `illegal_cast` +1** (`45a1cec5b3`): the free-function bridge gets an argc channel (`__host_argc`, new `__js2_call_fn_argc_N` wrappers; `PROVIDER_COMPILER_ABI_VERSION` → v2 because the provider cache key carries no compiler hash), and a widened call declines instead of trapping when a padded or mis-shaped argument cannot satisfy a non-nullable formal (`ref.test` mirrors the exact `ref.cast` predicate, so it only turns traps into declines). A/B on `sort/comparefn-resizable-buffer.js`: thrown → trap → thrown. 1,297-row honest callback slice: 0 flips.
- Also stacked: **#6482 round 1** (`51857d5a11`, +44 descriptor rows: shape-type branding across the link edge) and **#6492 round 5** (`3eb2b6118f`, +10: `__instanceof` resolved builtins in the worker realm while values were minted in the sandbox realm; honest 867-row control +1 / −0).

Expected in the merge group vs the promoted linked baseline: ≈ +190 improvements, 0 regressions on rows that were pass, trap counters flat or down. `hold` removed.


---

PR #5964, opened 2026-09-17
fix(linked-lane): parity batch 3 — #6492 r6 substrate bugs, #6482 r2 descriptor reads, #6495 coherent builtin realm (+79 rows, 6 traps gone)

ttraenkler:
Linked-lane parity batch 3 — [#6492](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6492-linked-lane-residual-parity-buckets) round 6/7, [#6482](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6482-linked-consumer-vec-index-read-in-wasm) round 2, and [#6495](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6495-host-realm-intrinsics) (new). None of the fixes is test262-specific.

## #6492 round 6 (+30 rows, 6 traps gone)
1. **Lib-globals gate missed the lib.es5 `declare function` globals** (`src/codegen/extern-declarations.ts`): a unit whose only lib-global was `eval`/`parseInt`/… skipped `collectDeclaredGlobals`, so `var s = eval; s("1+1")` **trapped**. 10 rows, 6 of them `null_deref`.
2. **UNDEF-SENTINEL f64 boxed to `null`** (`src/codegen/type-coercion.ts`): `ensureCanonicalUndefinedExtern` now runs before `__box_number`'s index is read. 8 rows.
3. **A host-thrown error lost its value crossing wasm→wasm** (`src/runtime.ts`): the last-host-exception latch was per import object; a linked graph has two. 12 rows.

Round 7 is diagnosis (shared-global mechanism, filed #6498); no code.

## #6482 round 2 (+49 rows) and #6495
4. **A vec receiver enumerates its own element indices** (`src/runtime.ts`): `__for_in_keys` returned `[]` for a vec, so propertyHelper's `for…in` pre-check short-circuited before any own-property predicate ran. +19 linked, **+7 honest**.
5. **#6495 — every test262 row gets a coherent builtin realm** (`tests/test262-runner.ts`): `builtin()` only resolved through the per-row `vm` realm when a source regex saw a literal `Object.defineProperty(Array.prototype, …)` in the body; harness-side mutation never matched, so a compiled program mutated the worker's own `Array.prototype`. Classifier deleted, every row marked coherent (≈ +40 % wall on a realm-heavy slice). Residual left open and documented: intrinsic *methods* are still the host's (`Array.prototype.forEach.name` canary).
[...]

PR #5967, opened 2026-09-18
fix(runtime): #6482 a sparse array's hole is not an own property (fixes 7 of batch-3's 12 regressions)

ttraenkler:
Forward fix for 7 of the 12 regressions PR #5964 introduced (merge-group run 35279861952): `Array/prototype/copyWithin/fill-holes.js`, `Object/defineProperties/15.2.3.7-6-a-{155,156,161,162}.js`, `Object/defineProperty/15.2.3.6-4-{159,160}.js` — all assert `hasOwnProperty("1") === false` for a hole.

## Mechanism
[#6482](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6482-linked-consumer-vec-index-read-in-wasm) round 2 read own-ness off `idx < __vec_len(obj)`. Sound for an `arguments` object (dense by §10.4.4), wrong for an ordinary array: `[0, , 2]` has an in-bounds hole that is not own, and the host cannot see sparseness (it lives in the in-wasm overlay; `__vec_gopd` is not an export). Rule now: an `arguments` vec → every in-bounds index; any other vec → only indices the host positively knows (a `_wasmPropDescs` entry or a sidecar value, both written by `defineProperty` or a host write, neither can be a hole). The arguments-only cut alone lost two further rows (`15.2.3.6-4-{201,203}`, a defineProperty-created index on an empty array); the second clause restores exactly those.

Round 2's honest "+7" (`Object/keys/*`, `getOwnPropertyNames/15.2.3.4-3-1`) was the same defect from its flattering side and is given back.

## Measured (real runner, both lanes)
| slice | rows | before → after |
| --- | ---: | --- |
| linked, wide (defineProperty/defineProperties/copyWithin/await/optional-chaining) | 1,862 | 1,451 → 1,458 (+7, 0 regressed) |
| honest, same | 1,862 | 1,415 → 1,422 (+7, 0 regressed) |
| linked, 114-row #6482 bucket | 114 | 98 → 98 |
| honest, 818-row control vs the pre-round-2 baseline | 818 | 692 → 692 |

Attribution by reverting each batch-3 commit's files one at a time: the 7 rows are this lane's vec-enumeration commit; the other 5 (async/await rows) are unaffected by any #6482/#6495 commit and fail identically in the honest lane — handed to the #6492 lane as a compiler regression from its round-6 commits.

Also carries #6492 round 10–12 notes and #6502's step-1 finding (docs only). Guard `tests/issue-6482-r3-sparse-vec-own-indices.test.ts` (linked-lane, fails 1/4 pre-fix). Gates, typecheck, equivalence gate, 23 guard tests green.

---


ttraenkler:
`equivalence-shard (8)` on `2b36bdd98d` (run 35291827651) failed in `actions/setup-node` with `read ECONNRESET` before checkout of the test body — an infra failure, not this PR's. Re-ran the failed jobs once; a second failure would be real.


---

ttraenkler:
Correction: the re-run API refused this session's token (403), so no re-run was spent. Pushed the pending #6492 round-13 notes commit (`85fced4673`, docs only) which retriggers the full check set on a fresh head.


---

ttraenkler:
Park diagnosed — real, not flake. Run 35293032556: 10 regressions, all array own-index rows (`Object/keys/15.2.3.14-*`, `getOwnPropertyNames/15.2.3.4-3-1`, `Array/prototype/{filter,map}` enumerable checks, `for-in/acquire-properties-from-array`). They are the flip side of `eb23ff171f`, which restricted the vec own-index rule to `arguments` objects to fix 7 hole rows: length alone cannot tell a dense literal from a hole, so dense arrays lost `hasOwnProperty(0)` / `Object.keys`.

Fix in progress on the branch: an exported per-index hole probe from the minting module so the host answers both the hole rows and the dense rows correctly. `hold` stays until both sets pass in the linked lane locally; the fix will be pushed to this PR. Tracked in [#6482](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6482-linked-consumer-vec-index-read-in-wasm).

---

ttraenkler:
Park fix pushed (head `ec46bb8370`), `hold` removed.

- The 10 dense-array rows and the 7 hole rows now both pass in the linked lane locally (fresh provider cache, both bundles rebuilt): a new `__vec_has_own_index` export answers own-ness from the raw element before boxing, and every `length` store marks the region it orphans (`0bed91b6e4`). 2,570-row vec-define matrix +6 / 0 regressed; 818-row for-in/keys slice unchanged.
- Also on this head since the park: #6491 round 3 (Script goal + 6 early-error rules, +13), #6492 round 4d (cross-module proxy traps + top-level destructuring assignment, +4 Iterator rows, +7 wider), #6492 round 5 (`Promise.allKeyed`/`allSettledKeyed` polyfill, +50 in `built-ins/Promise/`).
- The earlier local inversion (dense rows failing on pristine main) was a stale shared `$TMPDIR/js2wasm-test262-harness-cache`, not Node 25; recorded in [#6482](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6482-linked-consumer-vec-index-read-in-wasm).

Expected merge-group diff vs baseline: net positive, 0 regressions.

---

ttraenkler:
`quality` is red on `ec46bb8370`: `check:stack-balance` reports `default-value-lossy` 43 → 50. A/B against the pre-round-4 tree shows +1 in every playground example, so one emitter that runs once per module (the round-4 `__vec_has_own_index` / length-hole-fill ladder) has a branch that leaves too few values; the safety net fills it with a default. Fixing the emitter on the branch, not the baseline; will push the fix here.

---

ttraenkler:
Merge group on `887e87650a` (run 35313398232): `check for test262 regressions` passed, but `merge shard reports` failed the compile-time guard — aggregate compile time +345.6 % over the shared set (0 timeouts), a systemic compile slowdown, and the same cause made the previous group's shard matrix overrun the queue window (the 05:40 cancelled run). Bisecting the five code commits since `85fced4673` (which measured Δ +2.2 %) locally now; the fix lands on this PR before it is re-queued. Not requeueing until then.

---

ttraenkler:
Compile-time park fixed (head `51d78886d3`), `hold` removed. Root cause was not compiler cost: the round-5 `Promise.allKeyed` polyfill is installed onto the host `Promise` per instantiate, the worker's realm canary read that as drift, and in CI's `recycle` mode every test recycled its worker and re-loaded the harness provider (one shard: 922 drift lines, 868 provider loads vs 108 in the good run; shards 4× slower). The worker now primes that install before its canary snapshot. Local A/B with the canary in recycle mode: 327 → 7 drift lines, 8.1 s → 1.5 s mean compile per row. Expected merge-group compile-time Δ back near baseline.

---

PR #5968, opened 2026-09-18
fix(codegen): #6493 — first-class Function.prototype.call/apply and an [[ErrorData]] class tag in standalone

ttraenkler:
## Summary

Reading a builtin prototype method as a **value** and then calling it — the `Function.prototype.call.call(f, thisArg)` shape — threw `TypeError: <key> is not yet implemented in --target standalone`. The refusal is generic: any first-class builtin method value reaching that arm without a body degrades to a catchable TypeError, so a method that works perfectly at a static call site has no first-class value at all.

[#6493](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6493-es2015-standalone-first-class-builtin-method-values)

## Measured

Two trees — a merge-base worktree and this branch, same `.test262-cache` symlinked into both, `--isolate --standalone`.

| acceptance set (112 rows) | base | branch |
| --- | --- | --- |
| vs merge-base `747c0fee19` | 40 pass / 72 fail | **43 / 69** |
| vs `origin/main` `603ce3a4b2`, post catch-up merge | 56 / 56 | **59 / 53** |
| `Function.prototype.call is not yet implemented` in that slice | 27 | **0** |

**Zero rows lost** in both runs — the branch's non-pass set is a strict *subset* of the base's, checked per test rather than per count. The 805-row control set is identical chunk by chunk, non-pass line for line (677 pass / 124 fail / 4 CE on both).

## Changes

- `src/codegen/function-proto-call-apply.ts` — §20.2.3.3 / §20.2.3.1 bodies, wired into `makeGlue`'s `Function` arm ladder; `call` registered variadic, both routed through the existing `__apply_closure` bridge.
- `src/codegen/object-proto-tostring.ts` — §20.1.3.6 step 8 `[object Error]` arm.
- `src/codegen/error-stack-accessor.ts` — §1 of the `Error.prototype.stack` accessors rejects every primitive receiver, and the Proxy arm below.

## The last regression, and why it had to be closed rather than documented

Round 1 removed 27 refusals but scored +3/−3. Three rows had been passing **on the refusal**: each is `assert.throws(TypeError, …)` and the refusal was itself the TypeError, so the accessor never ran.

Two were genuine receiver-check rows, fixed by making `emitThisIsObjectCheck` reject every primitive rather than only null/undefined — measured, not assumed: on the first-class `get.call(1)` path the receiver arrives as the raw boxed primitive, not a wrapper object.

The third, `setter-proxy-trap-rejects.js`, needed a Proxy trap's `false` return to surface as a TypeError. It could not be left as a documented residual: the row is **ES2015**, ES2015 is `ratcheted: true`, it reads `"status":"pass"` in the standalone baseline CI's `--compare` arm reads, and `test262-edition-ratchet.ts` Check 2 fails on **any single** pass→not-pass with no waiver — inside the required `merge shard reports` check. Net +2 with it lost would have been ejected from the queue.

It is now closed via the Proxy dispatchers that already carry the trap's own result (`__proxy_define_dispatch`, `__proxy_set_receiver_dispatch`, `__create_descriptor`, `__is_truthy` — all `registerNative` standalone natives, so `result.imports` stays `[]`).

Three corrections to the plan this PR was built from, recorded because each was asserted the other way first:
- `__proxy_set_dispatch` **cannot** be used for the set half — its trap-absent arm pushes `ref.null.extern` as a placeholder the front guard *drops*, so checking it would throw on every trap-absent proxy.
- `__defineProperty_value` is **not** "declared with no result" — it returns `O`; the real blocker is that it carries no Proxy front guard, so the define trap ran **zero** times.
- Byte-identity for an ordinary receiver is not achievable as stated, because the arm is emitted in any module that mints the setter. Measured instead: suppressing the arm reproduces the pre-change sha exactly, so the whole delta is the arm.

## Validation

- `tests/issue-6493-first-class-builtin-method-values.test.ts` — **11/11**, red on exactly four lines against the round-2 tip, and pins that a throwing trap propagates unchanged and that the trap fires **exactly once** (a re-read probe would show 2 and add an observable `getOwnPropertyDescriptor` call).
- Gates all exit 0 run bare: loc/func/coercion budgets (also with `LOC_GATE_BASE=origin/main`), oracle-ratchet, dead-exports, compiler-boundaries inventory, typecheck, prettier, biome, `test:equivalence:gate` (1720 passing / 22 known failures).
- `result.imports` is `[]` on every standalone probe.

## Known, out of scope, named in the issue

A `$Proxy` assigned to a `var` whose inferred type is the target's object shape is **nulled** — `lib.d.ts` types `new Proxy<T>(t,h)` as `T` and the carrier fails that downcast. Verified independently. It deserves its own id; it is why half (b) of the trap-rejects row is pinned here over a `JSON.parse` target, so both halves exercise the real path.

Also still open: `apply` with >8 declared params hits the `__apply_closure` arity cap, and `.length` through a variable folds from the lib signature.

## CLA

---


ttraenkler:
Removed the `hold` — this park is **spurious**, per the bot's own footer ("If it is a setup/infra step rather than a verdict step, the verdict never ran").

Diagnosed against the cited run (35327264963) before touching the label:

`test262 standalone shard 19/50` died inside `actions/setup-node@v6`, before any test body ran:

```
Attempting to download 25...
##[error]read ECONNRESET
```

A network reset fetching the Node 25 toolchain. The shard produced no results artifact (`No files were found with the provided path: …mgchunk19.jsonl`), and the second failure — `merge shard reports` → *"Fail if required test262 shards did not succeed"* — is purely downstream of that missing shard. No verdict step ran, so no conformance judgement was made about this PR at all.

This is the "died before any test body ran (checkout, install, runner loss)" case, which is the one sanctioned re-admission. Nothing in the diff is implicated: this PR touches `Function.prototype.call`/`apply`, `Object.prototype.toString`'s `[object Error]` arm, and the `Error.prototype.stack` accessors — none of which can affect runner setup.

Worth distinguishing from the sibling park on #5969, which is a **real** merged-baseline regression (net −439) and is being fixed on its branch rather than re-admitted.

---

PR #5969, opened 2026-09-18
fix(standalone): #6494 — Proxy/Reflect steps that must throw now throw

ttraenkler:
## Summary

Six standalone paths whose entire spec step is "throw a TypeError" returned silently. Measured directly against `origin/main` @ `c698c755bb` with `result.imports === []` on every probe.

[#6494](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6494-es2015-standalone-proxy-reflect-must-throw)

| probe | base | branch |
| --- | --- | --- |
| `Reflect.get(null,"p")` / `Reflect.get(undefined,"p")` | no throw | **TypeError** |
| `Reflect.has(null,"p")` / `Reflect.has(undefined,"p")` | no throw | **TypeError** |
| revoked proxy via `Array.prototype.map` | no throw | **TypeError** |
| `Object.defineProperty(revoked.proxy, …)` | no throw | **TypeError** |
| `defineProperty` trap returns `false` | no throw | **TypeError** |
| strict `p.x = 1`, `set` trap returns `false` | no throw | **TypeError** |

**Controls, all already correct and all unchanged**: `deleteProperty` / `setPrototypeOf` / `preventExtensions` returning false, a present non-callable trap, `Reflect.construct(f,[],1)`, and a revoked proxy read directly. Plus three added negative controls — a *true*-returning trap does not throw, a trap-**absent** proxy still writes and defines through, and sloppy-mode assignment with a false trap correctly does **not** throw.

#5316 (r4, §10.5 descriptor-model invariants) is done and wired the success bit for exactly the three traps that work; `defineProperty` and `set` were left un-wired. The blocker comment in `object-runtime-proxy.ts` citing #797/#1460/#1462 is **stale** — all three are done.

## Measured row delta — both trees, same lists, same symlinked cache

| set | base | branch | delta |
| --- | --- | --- | --- |
| all of `built-ins/Proxy` + `built-ins/Reflect` (464 rows) | 367 pass / 91 fail / 6 CE | **370** / 88 / 6 | **+3, 0 lost** |
| 630 stride-sampled controls (`Object/{defineProperty,defineProperties,create,freeze,seal,getOwnPropertyDescriptor}`, `language/expressions/assignment`) | 611 / 14 / 5 | identical | **0 lost, 0 gained** |

1,094 rows per tree, compared per PATH so a one-in/one-out swap could not hide. The three gained are `Proxy/defineProperty/null-handler.js`, `Reflect/get/target-is-not-object-throws.js`, `Reflect/has/target-is-not-object-throws.js`.

**+3 against a planned "~21", and the shortfall is explained rather than padded**: of 24 unthrown-TypeError rows in that area, 15 are `-realm.js` (out of scope, #4444); of the 9 left, 3 need nested-proxy forwarding over exotic targets, 2 need CreateListFromArrayLike/IsConstructor, 1 needs `instanceof` routed through the `getPrototypeOf` trap. Three of the six fixes have **no dedicated ES2015 row at all** — real spec defects the corpus does not isolate.

## Changes

- `src/codegen/reflect-target-guard.ts` — a `staticallyNullish` option seeds the non-object guard's predicate; same emitted shape, no second throw site.
- `src/codegen/expressions/call-namespace-static.ts` — `Reflect.get`/`has` pass the target expression; `undefined` is decided by `ctx.oracle.typeFactOf`, not by spelling, so a shadowing `var undefined = {}` keeps its lowering. Oracle ratchet: `getTypeAtLocation +0, ctx.checker +0`.
- `src/codegen/object-ops.ts` — the proxy-receiver recogniser also accepts `Proxy.revocable(...).proxy` (requiring the *declaration* to be that, so it stays a proof); `emitDefinePropertyRejectionThrow` gains an **opt-in** `falsyIsRejection`, used only on the proxy arm.
- `src/codegen/object-runtime-proxy.ts` — a `$Proxy` front guard on `__extern_set_strict` (trap-present arm only) and the revoked bit on `__extern_length`. Both `ctx.standalone`-gated.

## Two plan corrections worth recording

- **`Reflect.get(1,"x")` already threw.** That was measured with `(Reflect as any).get(…)`, a *dynamic member call* that never reaches the namespace-static arm; test262 writes `Reflect.get(1,'p')`, which threw on base. The real residual was `null`/`undefined`. The dynamic spelling is still open — a different lowering path, out of this slice.
- **`__defineProperty_value` does not swallow every proxy define** — the trap fires for a `new Proxy(...)`-bound receiver; it ran zero times only for the `Proxy.revocable(...).proxy` spelling, because the reroute gate is syntactic. That made the fix a one-function recogniser widening instead of surgery on a very hot native.

## A near-miss, caught only by the row run

A front guard prepended to `__extern_get_idx` **silently dropped every typed-vec arm** — `fillExternGetIdxVecArms` locates its splice point by that function's 3-instruction preamble shape, so `body.unshift` breaks it. Two `Proxy/defineProperty` rows went **pass → fail** while all twelve probes stayed green. That guard was dropped; `__extern_length` alone still delivers the revoked-via-`map` throw, because LengthOfArrayLike is the first internal method `map` performs.

## Validation

- `tests/issue-6494-proxy-reflect-must-throw.test.ts` — **17/17**, covering the six defects, the six controls, and the widening-hazard shapes.
- gc byte-identical on both trees (`fd0aa18d8fad3947`, `94545f0dec39c32f`); wasi byte-identical except on a program literally writing `Reflect.get(null, …)`.
- `result.imports` `[]` on every standalone probe.
- Gates all exit 0 run bare: loc/func/coercion, oracle-ratchet, dead-exports, typecheck, `test:equivalence:gate` (1720 passing / 22 known, re-run against the final code). `check:godfiles` and compiler-boundaries `complete` mode fail **identically on base** — pre-existing, and neither names a file this PR touches.

## CLA

---


ttraenkler:
Park was **real and mine**. Fixed in `1d18bda24b`; `hold` removed only after the fixed head was actually on the branch.

## What the merge_group caught

Net **−439 pass** (35,829 → 35,390), 444 regressions against 5 improvements, 427 `assertion_fail`, plus the uncatchable-trap ratchet growing `oob` 7 → 9.

I ruled out the baseline-drift explanation the log itself suggests before doing anything else: #5968 and #5970 were both `clean` with no labels at the time, so the cluster was not shared.

## Root cause

`src/codegen/ta-dyn-mop.ts` (~L1045) **already** does `lenFn.body.unshift(...)` on `__extern_length`, prepending a TypedArray dyn-view arm that computes the live in-bounds element count — an arm its own comment requires to sit *ahead of* the `$__vec_base` arm, because `$__ta_dyn_view` subtypes it. This PR's revoked-bit guard was a second, independent `unshift` onto the same body. They collide, and `%TypedArray%.prototype.set` misreads its length (`RangeError: offset is out of bounds`).

**This is the second instance of the same hazard in this PR.** It had already been found for `__extern_get_idx` — `fillExternGetIdxVecArms` locates its splice point by that function's 3-instruction preamble shape, so a naive `unshift` drops every typed-vec arm — and that guard was dropped. The identical `unshift` on `__extern_length` was kept without checking whether the same dependency existed. It did.

## Measured, `--isolate --standalone`

| | with the guard | guard reverted |
| --- | --- | --- |
| `TypedArray/prototype/set/array-arg-set-values-in-order.js` | fail | **pass** |
| `TypedArray/prototype/set/array-arg-src-values-are-not-cached.js` | fail | **pass** |
| the three rows this PR gained | pass | **pass** |

5/5 on the combined list. Isolated by reverting *only* that loop, so the attribution is not a guess. The **+3 is intact**; the only behaviour given up is revoked-proxy-via-LengthOfArrayLike, which was never among the +3.

## The pin was not deleted or skipped

It is converted to a known-gap pin asserting the current (wrong) behaviour, per this repo's convention for named residuals, so it fails loudly the moment the guard lands correctly. Follow-up filed as [#6506](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6506-revoked-proxy-length-guard-unshift-collision) with the root cause and an explicit warning that any fix must be validated against `%TypedArray%.prototype.set` — a probe that merely reads a length stays green while the arm-spliced paths break, which is exactly how this shipped.

## Process note

I **dequeued this PR by closing and reopening it**, solely to release the merge-queue branch lock (`GH006: … queued for merging cannot be updated`) so the known-bad head could be replaced. That is the steward skill's sanctioned mechanism and is **not** a CI kick.

Also worth recording: every PR-level signal was green — 17/17 pins, all ratchet gates, typecheck, lint, equivalence. Only the full-corpus merge_group run could catch this, which is precisely why that gate exists.

---

PR #5970, opened 2026-09-18
fix(codegen): #6500 packed TypedArray reverse() compiles, #6501 the four helper-routed mutators validate a detached view

ttraenkler:
## Summary

Two stacked fixes in the same area. The second depends on the first: `reverse` cannot compile on a packed receiver without it.

- [#6500](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6500-packed-typedarray-reverse-binary-emit-crash) — a **compile-time kill**
- [#6501](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6501-ta-dyn-helper-methods-bypass-validate-guard) — the §23.2.4.4 guard four methods never reached

## #6500 — one line kills the whole module

```ts
export function probe(): number { const ta = new Uint8Array(8); ta.reverse(); return 1; }
// Binary emit error: encodeValType: packed storage type "i8" is not valid in a value position
```

Nothing in the module runs, so it takes out every other function in the same program. `compileArrayReverse` allocated its swap slot with the array's **storage** type; a packed type is legal only as a struct field or array element. The same function already knows this three lines later where it picks `array.get_u`/`array.get_s` — both push an **i32**. The slot just has to match the load.

Surface: a **statically-typed** receiver (`(ta as any).reverse()` routes elsewhere and compiles) and a **packed** element type — i8 (`Uint8Array`, `Int8Array`, `Uint8ClampedArray`) or i16 (`Uint16Array`, `Int16Array`). `sort`, `fill`, `copyWithin`, `slice`, `subarray`, `indexOf`, `join` already widen correctly. Detachment is irrelevant — the crash was first seen beside a detached-buffer probe during #1645, which made it look related.

Verified past "it compiles": reversal correct on all nine element types, and conversions match Node exactly (the check a widened slot could silently break) — Int8/Int16 sign-extend to 6995, Uint8 wraps −5→251 (7251), Uint8Clamped clamps −5→0 (7000). **Byte-identical where it used to compile**, both lanes: `a5faf7fc396fa2ca` (gc) / `60777d8347221172` (standalone), before and after.

## #6501 — four methods never reached the guard

#5961 put the ValidateTypedArray prologue on the generic `__call_m_*` dispatcher. `set`, `fill`, `copyWithin` and `reverse` are routed at the call site to dedicated `ensureTaDyn*Helper` natives and never consult it. `reverse` is even already listed in `TA_DYN_VALIDATE_METHOD_NAMES` — the list was right, the path never reached it.

| row | base | branch |
| --- | --- | --- |
| `reverse/detached-buffer.js` | fail — *no exception at all* | **pass** |
| `fill/detached-buffer.js` | fail — *TypeError expected, got Test262Error* | **pass** |
| `copyWithin/detached-buffer.js` | fail — *same* | **pass** |
| `sort/detached-buffer.js` | pass | pass — **control, unchanged** |
| `slice/detached-buffer.js` | pass | pass — **control, unchanged** |

The two passing controls are the point: they go through the generic dispatcher, so they prove the guard works and that the only variable is which path a method takes.

**Full `TypedArray/prototype/{reverse,fill,copyWithin,set}` — 249 rows, both trees:** 158 → **164** pass. Per-test diff: **6 fixed, 0 lost** (the three rows plus their `BigInt/` twins). Base for this measurement is the **#6500 tip**, not `origin/main`, or the `reverse` rows would not compile on base and would confound the diff.

`set` needed no separate splice — the emit block already coalesces `taSetIdx ?? taFillIdx`. Its 41 non-pass rows are the same 41 before and after.

## Why the dynamic shape is load-bearing

test262 never calls these on a statically-typed receiver — every row goes through `testWithTypedArrayConstructors`, so `TA` is a parameter and the receiver is a `$__ta_dyn_view`. A probe written `new Float64Array(8).fill(…)` takes the static path and looks perfectly healthy. That is exactly how this survived #5961, and there is a comment in the test file saying so.

## Two things that cost real time, documented in the code

- **Splice position is semantic.** `fill`/`copyWithin` failed with *"expected TypeError, got Test262Error"*, not "nothing thrown": their rows pass an object whose `valueOf` throws, from the **helper's** ToNumber. Splice before argument evaluation and you violate §13.3.6; splice inside the helper and ToNumber has already run. The head of `thenArm` is the only position satisfying both.
- **A local's NAME is load-bearing.** `deduplicateLocals` merges `__`-prefixed temps sharing a name *and* type. Suffixing the guard's buffer local "for consistency" silently grew the code section by 12 bytes on a four-call body.

## Validation

- `tests/issue-6500-packed-typedarray-reverse.test.ts` — **12/12**; **7 failed / 5 passed** against the unfixed source.
- `tests/issue-6501-ta-dyn-helper-validate.test.ts` — **9/9**; **4 failed / 5 passed** against the pre-#6501 source. It pins the regression this change could plausibly cause: on a **live** view, `reverse`/`fill`/`copyWithin` each mutate correctly *and* return the same object (checked by identity), and `set` writes at the offset — scored as a bitmask so a partial failure names itself.
- In both files the cases that pass either way are deliberate controls (unpacked receivers, live-buffer inertness, the dispatcher path); they exist to catch an over-reaching fix and would be worthless if they flipped.
- gc byte-identical on both trees; `result.imports` `[]` on every standalone probe; the gc lane cannot reach the #6501 code at all (`ensureTaDyn*Helper` returns undefined unless `noJsHost`).
- Gates all exit 0 run bare, including `LOC_GATE_BASE=f5506015f9` to simulate CI's base. Both budget grants carry dated rationales in their issue frontmatter; no `scripts/*-baseline.json` touched.

## Not fixed here

`copyWithin/coerced-values-{start,end}-detached.js` detach *inside* an argument's `valueOf`, after ValidateTypedArray has legitimately passed — a missing re-validation, a separate defect. `reverse` on a detached buffer is fixed by #6501; #6500 alone only stops the module failing to build.

## CLA

---


ttraenkler:
Removed the `hold` — this is **collateral from #5969's head**, not a regression in this PR. Diagnosed against the cited run (35328766509) before touching the label, per the auto-park rules.

## The two parks are the same failure

| | #5969's park (run 35325926529) | this park (run 35328766509) |
| --- | --- | --- |
| regressions | **444** | **444** |
| named new `oob` traps | `TypedArray/prototype/set/array-arg-set-values-in-order.js`, `…/array-arg-src-values-are-not-cached.js` | **the same two** |
| improvements | 5 | 20 |
| net | −439 | −424 |

Identical regression count and identical named trapping files. The improvement count differs only because this group carried more queued work.

## Why it cannot be this PR

Those two rows are exactly the ones I isolated as broken by #5969's revoked-bit guard, which did `body.unshift` on `__extern_length` and collided with the TypedArray dyn-view arm that `ta-dyn-mop.ts` already unshifts onto the same body. Reverting **only** that loop flips both rows back to pass — measured, not inferred.

This PR cannot produce them:

- its `%TypedArray%.prototype.set` change *adds* the ValidateTypedArray guard to the helper-routed path, and I measured `set`'s 41 non-pass rows as **the same 41** before and after — no gain, no loss;
- the packed-`reverse` fix (#6500) only widens a swap-slot type for i8/i16 element types, and is byte-identical wherever the module previously compiled, on both lanes;
- this PR's own measured delta is +6 / 0 lost across 249 rows, on both trees.

The merge group simply contained #5969's known-bad head alongside this one.

## Already fixed upstream of this

#5969's guard was reverted in `1d18bda24b`, pushed, and that PR is back in the queue with a clean head. The next merge group builds against the fixed version, so the 444 should not recur. If they do, that would be new information and this PR becomes the suspect — I'll treat a repeat as real rather than assume collateral twice.

---

PR #5998, opened 2026-09-20
fix(standalone): S67 — identity-keyed linked statics, runtime spread into inherited linked statics, spread super through a linked heritage (#5383/#6644)

ttraenkler:
## Description

Slice S67 of the standalone Temporal provider ([#5383](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5383-standalone-temporal-provider)), closing three of the residuals pinned by [#6644](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6644-link-static-inheritance-instanceof) (S66):

1. **Linked-static arms resolve by class IDENTITY, not name.** `resolveLinkedStaticClassName` in `standalone-linked-static-inheritance.ts` goes through `ctx.oracle.valueDeclarationOf`; S66's "enclosing-shape sensitivity" was a same-named twin class reading the other's `__linked_parent_<C>` global (the WRONG parent once the owner had run). A colliding twin with no synthetic identity minted is refused, not guessed.
2. **Runtime spread into an inherited linked static** (`S[k](...a)`): the call now ships through the boundary's `__apply_closure(__extern_get(P, key), P, argv)` terminal with the shared `tryEmitSpreadHostArgs`, gated on a spread being present. The fixed-arity inline dynamic call handed the source array over as a single argument.
3. **`super(...<runtime spread>)` through a linked heritage** builds `this` via S34's `__native_construct_argv` driver, now parameterised on the callee push (`new-super.ts`, `class-bodies.ts`); the arm used to decline and only evaluate the arguments for effect.

Rows: the two `from/subclassing-ignored` rows advance from `called value is not a function` to the pre-existing `SameValue(«null», «undefined»)` blocker shared with `PlainDate/from/argument-object-valid.js` and the #6633 `era` rows. Four-family sample holds 459/480 with 0 pass→fail. Findings appended to #5383, #6644 and the handover.

## Test plan

- [x] New witness `tests/issue-6644-link-computed-static-spread-super.test.ts` (18-case table); fails on the file-copy reverted base
- [x] Witness sweep `tests/issue-66*` + 6484 + 6493: 49 files / 275 tests green under Node 22 and Node 25.9 (also on the merged head)
- [x] Four families × 120 rows vs the S66 base: 0 pass→fail, 0 fail→pass (PlainDate 117, Duration 108, PlainDateTime 117, ZDT 117)
[...]

PR #6011, opened 2026-09-20
fix(standalone): S69 — a live-global-bound function declaration no longer answers null for an object result under a linked eval realm (#5383/#6647)

ttraenkler:
## Description

Slice S69 of the standalone Temporal provider ([#5383](https://js2wasm.loopdive.com/dashboard/issue.html?slug=5383-standalone-temporal-provider)), closing [#6647](https://js2wasm.loopdive.com/dashboard/issue.html?slug=6647-standalone-temporal-rangeerror-not-thrown).

**The fix (one file, +21 LOC).** When a module has `eval` reachable under a linked provider, every top-level declaration's call goes through the generic dynamic dispatcher (`ctx.runtimeEvalGlobalFunctionBindings`), which can only produce an `externref`. `ensureFuncClosureSingleton` kept the callee's CONCRETE struct result in the wrapper's funcref type, so `function g(){ return {a:1}; } g()` answered `null` while `.call`/`.apply`/`new` and primitive results were correct. The wrapper's result is now promoted to `externref` in exactly that case (`src/codegen/closures/method-trampolines.ts`), the same shape as the parked-async and native-generator bridges on that line. Byte-inert without `eval`: the standalone provider binary is identical before and after.

**The five briefed rows are attributed, not fixed**, and both hypotheses in the brief were falsified by measurement:
- Exceptions propagate correctly across the link and through `assert.throws` (probe `l3.js`).
- `Duration/compare/relativeto-propertybag-invalid-offset-string.js`, `relativeto-string-invalid.js`, `PlainDateTime/from/argument-string-invalid.js`: the vendored polyfill's offset grammar makes both separators independently optional, so `+00:0000` matches. Reproduces under plain Node importing the polyfill directly; the host lane fails them too. Only a polyfill upgrade moves them.
- `Duration/compare/throws-when-target-zoned-date-time-outside-valid-limits.js`, `ZonedDateTime/prototype/add/overflow-adding-months-to-max-year.js`: standalone BigInt is a branded i64, so `864n * 10n ** 19n` wraps at 2^64 and the limit check never fires. Needs arbitrary-precision BigInt (XL).

Also recorded: the runner's `assert.throws` line attribution names the first `assert.throws(` in a file, not the failing one. Four-family sample holds 463/480 with 0 pass→fail. Next target reduced: `PlainDate.prototype.add` fails for every input (~78 rows across `PlainDate`/`PlainYearMonth` add/subtract) because a spread-built object from a provider-local source breaks when it crosses a function return.

## Test plan

- [x] New witness `tests/issue-6647-live-global-binding-object-result.test.ts`: fails on a file-copy revert of the touched file to the branch base, passes on the fix (5 of 9 probes, all 4 controls pass both sides)
- [x] Witness sweep `tests/issue-66*` + 6484 + 6493 (54 files / 327 tests) under Node 22 and Node 25.9 on the merged head: 326 pass; the 1 failure is main's own `tests/issue-6648-regexp-capture-array-output.test.ts` residual, which fails on `origin/main` alone
- [x] Four families × 120 rows vs the S68 base: 0 pass→fail, 0 fail→pass (PlainDate 120, Duration 109, PlainDateTime 117, ZDT 117)
- [x] Must-not-move groups A/B/C/D/E-unlinked/E-linked/F-class/F-methoddef/F-objproto (3,204 rows): flat
- [x] Corpus byte A/B 42×{gc,standalone}: 0 status / 0 sha flips
- [x] Equivalence gate 22 / 1720 / 22, no new regressions
- [x] Gate chain (loc/func budget incl. `LOC_GATE_BASE=origin/main`, coercion-sites, oracle-ratchet, dead-exports, compiler-boundaries inventory, issue-ids), typecheck, lint

## CLA

- [x] I have read and agree to the project's Contributor License Agreement.

---


ttraenkler:
Park released once: the cited run (35516600403) failed in `equivalence-shard (6)` at the **Setup Node and pnpm (cached)** step with `##[error]read ECONNRESET` — an infra failure before any test body ran; `equivalence-gate` then failed only because that shard never reported. No verdict step ran against this PR's diff, so the park is spurious per the bot's own note. Second identical `setup-node` ECONNRESET park today (#5998 this morning was the first). Not re-enqueueing manually; leaving it to `auto-enqueue`.

---

